PR #69, opened 2013-06-27
Add support for a @FoldWith annotation

tburch:
I added a @FoldWith annotation that adds support for folding the ResultSet of the @SqlQuery annotation on sql object interfaces.


stevenschlansker:
I think this is now implemented more generally with Java 8 Collectors in #199, so I'm going to close this.  It'll be in Jdbi 3.  Let me know if you have renewed interest in picking this up, or if it is actually not the same feature.

PR #70, opened 2013-07-15
Introduce @BindMap, binds Map entries as statement parameters

stevenschlansker:
Feature request from [D. Reinert on the mailing list](https://groups.google.com/d/topic/jdbi/nuJFiFnpMgo/discussion)

Introduce a new binding method, where ```Map``` entries are bound as parameters.  Has a couple of tweaks on the annotation for its behavior.

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/Binding.java
+++ b/src/main/java/org/skife/jdbi/v2/Binding.java
@@ -63,7 +63,7 @@ public class Binding
      *
      * @param position starts at 0, not 1
      *
-     * @return arfument bound to that position
+     * @return argument bound to that position
      */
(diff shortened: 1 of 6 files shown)


stevenschlansker:
@christophercurrie fixed up and merged, sorry for the delay, mind testing it out a bit before the next release?

christophercurrie:
would like to, but probably won't have a chance this week, and you usually like to release more quickly than that. I'll take a look as soon as I'm able to context switch.

stevenschlansker:
No problem, I am actually out of town for the coming week anyway and don't plan a release.  So unless there's an urgent bug you have a bit of time.

PR #80, opened 2013-10-01
simple support for batches against a call

osi:
I needed to be able to do batches against a stored proc (that did inserts, so no out params).

this is a stupid-simple change that allows creation of a CallableStatement.


stevenschlansker:
@osi Any chance of that test case?

osi:
I'm not doing this any more, so I've stopped using my fork that had this. Closing the PR.

PR #105, opened 2014-06-14
Bind collection

jacksingleton:
The current @BindIn annotation is useful for when you want to use a collection of values in an IN statement, but does not allow us to further customize a complicated query.

This hit us when we tried to query for entities matching one or more of a set of free form tags (joined from an 'entity_tags' table)

This pr adds a @BindCollection annotation that makes these complicated queries possible.

You can use StringTemplate to iterate though collection items any way you want:

```
where
  value like <first(values)>
  <rest(values):{ v | OR value like <v> }>
```

translates roughly into:

``` sql
where
  value like :v1
  OR value like :v2
  OR value like :v3
```


stevenschlansker:
Sorry this branch didn't get much attention, does the new `@BindList` annotation cover this use case?  We can consider porting the code forward if not.  But it will have to be reopened against `jdbi2` so this PR is now outdated.

PR #107, opened 2014-07-12
Support for fluent map queries

arteam:
Maybe it's a quite a big change to add, but I will give a shot.

The problem this PR addresses is a complexity to perform map queries in JDBI. I think JDBI can do a lot better in this field.

Let's have a simple table in a database
```sql
insert into something(id, name) values (1, 'keith');
insert into something(id, name) values (2, 'brian');
```

Let's imagine we want to load data from this table to **Map**. *Canonical* way to make such queries is using *fold*. Something like that:
```java
h.createQuery("select id, name from something")
                .fold(new HashMap<String, Integer>(), new Folder2<Map<String, Integer>>() {
                    public Map<String, Integer> fold(Map<String, Integer> a, ResultSet rs, StatementContext context) throws SQLException {
                        a.put(rs.getString("name"), rs.getInt("id"));
                        return a;
                    }
                });
```
Quite cumbersome. A little better with *Guava* and *Java 8*:
```java
 h.createQuery("select id, name from something")
                .fold(Maps.newHashMap(), (a, rs, context) -> {
                    a.put(rs.getString("name"), rs.getInt("id"));
                    return a;
                });
```

Another problem is that we don't have a strict mapping between keys and values. So, in case if we already have some mappers for complex types it's quite not natural to use it the *fold* method.

This change proposes a fluent interface to work with map queries. Basic structure like that:
 ```java
h.createMapQuery("select id, name from something")
                .mapKey(new StringMapper("name"))
                .mapValue(new IntegerMapper("id"))
                .get();
```
I believe it's more concise and more focused on the real task. 

Registered mappers can be reused:
 ```java
h.registerMapper(new SomethingMapper());
h.registerMapper((index, r, ctx) -> 
        new City(r.getInt("city_id"), r.getString("city_name"), r.getString("state")));
h.createMapQuery("select u.id, u.name, u.city_id, c.name city_name, c.state " +
        "from users u inner join cities c on u.city_id = c.id")
        .mapKeyTo(Something.class)
        .mapValueTo(City.class)
        .get();
 ```

We can set the needed implementation of **Map** if the default **HashMap** is not sufficient.
 ```java
h.createMapQuery("select id, name from something")
                .mapKey(new StringMapper("name"))
                .mapValue(new IntegerMapper("id"))
                .accumulator(Maps.newHashMapWithExpectedSize(128))
                .get();
 ```

If we want some not standard structure with keys and values, we can use *fold* (I use more simple fold version without changing an accumulator, because it more handy for typical cases).
 ```java
h.createMapQuery("select u.id, u.name, u.city_id, c.name city_name, c.state " +
        "from users u inner join cities c on u.city_id = c.id")
        .mapKeyTo(Something.class)
        .mapValueTo(City.class)
        .fold(ArrayListMultimap.create(), (accumulator, key, value) -> 
                {accumulator.put(value, key);});
 ```  

I've tried to make everything backward compatible, but could make mistakes. Fell free to correct me.
Thanks in advance.

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/BaseQuery.java
+++ b/src/main/java/org/skife/jdbi/v2/BaseQuery.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright (C) 2004 - 2014 Brian McCallister
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.skife.jdbi.v2;
+
+
+import org.skife.jdbi.v2.tweak.SQLLog;
+import org.skife.jdbi.v2.tweak.StatementBuilder;
+import org.skife.jdbi.v2.tweak.StatementLocator;
+import org.skife.jdbi.v2.tweak.StatementRewriter;
+import org.skife.jdbi.v2.tweak.StatementCustomizer;
+import org.skife.jdbi.v2.tweak.ResultSetMapper;
+
+import java.sql.ResultSet;
+import java.util.Collection;
+
+/**
+ * Base class for statements providing convenience result handling for SQL queries
+ *
+ * @param <SelfType> concrete query type
+ */
+@SuppressWarnings("unchecked")
+public class BaseQuery<SelfType extends BaseQuery<SelfType>> extends SQLStatement<SelfType> {
+
+    protected final MappingRegistry mappingRegistry;
+
+    BaseQuery(Binding params,
+              StatementLocator locator,
+              StatementRewriter statementRewriter,
+              Handle handle,
+              StatementBuilder cache,
+              String sql,
+              ConcreteStatementContext ctx,
+              SQLLog log,
+              TimingCollector timingCollector,
+              Collection<StatementCustomizer> customizers,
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by stevenschlansker** [comment]
@arteam Sorry for the slow response, we (JDBI maintainers) have been busy and this is a much larger pull request than I'm used to reviewing :)
I'll take a look at it and get back to you soon.

**Comment by stevenschlansker** [comment]
Ok, again really sorry for the slow response.  I think my biggest concern here is that we are introducing a whole new type of query, when really we are only changing the results mapping phase.  Java 8 has the wonderful [Collectors.toMap](http://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toMap-java.util.function.Function-java.util.function.Function-) function, so you would then be able to write
```
h.createQuery("select id, name from something")
    .stream().collect(Collectors.toMap(r -> r.getString("name"), r -> r.getInt("id")));
```
Would you be interested in trying out this new way in the JDBI3 hacking branch I have?  (https://github.com/stevenschlansker/jdbi/tree/java-8-hacking)

I'm not saying that I won't consider merging this, I'm just worried that the complexity is very high (and it introduces an entirely new query type) and there is a much much nicer solution "just around the corner"...

arteam:
I close this PR in favor of Java 8 Stream API. It seems to be a more canonical way to solve the problem.

korri123:
Hey, createQuery doesn't have a stream() method, is this only for JDBI3?

qualidafial:
Correct.

JDBI v2 only requires Java 6, so we can't use any Java 8 types.

V3 requires Java 8, so we can take advantage of all the new types.

qualidafial:
If you're using JDBI v2 in a Java 8 environment, you could call
query.list().stream() if holding the entire result in memory isn't a
problem. Alternatively you can use
StreamSupport.stream(query.iterator().spliterator(), false).

-Matthew

qualidafial:
Oops, second example should have been:
StreamSupport.stream(query.spliterator(), false)

korri123:
Thanks! You've helped more than enough :)

PR #117, opened 2014-10-17
Supply name of column that generates key.

fiskeben:
Oracle needs the name of the column that generates a key in order for `@GetGeneratedKeys` to work.

This PR makes it possible to add a `columnName` to the `@GetGeneratedKeys` annotation which will be passed on to the `PreparedStatement`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -199,12 +199,7 @@
             <scope>test</scope>
         </dependency>
 
-        <dependency>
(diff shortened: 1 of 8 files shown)


stevenschlansker:
If you still care to fix the `FigureItOutResultSetMapper` feel free to send a followup PR.  Thanks!

PR #119, opened 2014-10-27
Enum support improvement

jeffjcrowe:
Allow enum null value in BeanMapper.
Fix BeanMapperFactory to not interfere with built-in argument types.
New EnumArgument argument type built-in which maps any Enum value as varchar


gsmith85:
I see that in 2.56 additional support was provided for `Enum` (see `EnumArgument`) and that on master there is now a `EnumColumnMapper` but a no `EnumMapper` type (that inherits from `TypedMapper`.  This seems to be an omission.

We have implemented this for some of our local projects and likely would be willing to share our implementation.  Before I execute a pull request I wanted to confirm that this:
# Isn't already in flight.
# There is some reason that `EnumMapper` not existing is undesirable / intentional.

Ping @qualidafial as you implemented the column mapper variant.

qualidafial:
This is supported using the column mapper. Column mappers are automatically
adapted into row mappers as needed.

qualidafial:
I can give a more complete answer now that I'm at my desk.

We now have two classes of mappers: ResultSetMapper (row mappers) and ResultColumnMapper (column mappers). Row mappers such as BeanMapper now have the ability to consult the column mapper registry, which means that if you register a column mapper for some new data type, the BeanMapper will now be able to map bean properties of that data type.

When you tell JDBI to map a query to a certain type, the row mapper registry is consulted first. If no row mapper is registered for that type, we search the column mapper registry. If a column mapper is found, we wrap it in a row mapper that maps only the first column by delegating to the column mapper.

PR #123, opened 2014-11-28
Restore Connection Auto-Commit on Exception

hawkan:
These changes addresses the problem when the underlying database driver throws error on commit in transactional mode. Previously the auto-commit state was not restored when throwing, thus creating inconsistent connections in the pool which then typically caused connections to hang when used non-transactional. This behaviour is especially sneaky in systems with large database pools where connection auto-commit state might be compromised over a longer period, creating hanging connections which might be hard to explain.

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/exceptions/UnableToRestoreAutoCommitStateException.java
+++ b/src/main/java/org/skife/jdbi/v2/exceptions/UnableToRestoreAutoCommitStateException.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C) 2004 - 2014 Brian McCallister
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.skife.jdbi.v2.exceptions;
 
 public class UnableToRestoreAutoCommitStateException  extends DBIException {
--- a/src/main/java/org/skife/jdbi/v2/exceptions/UnableToRestoreAutoCommitStateException.java
+++ b/src/main/java/org/skife/jdbi/v2/exceptions/UnableToRestoreAutoCommitStateException.java
@@ -1,0 +1,8 @@
+package org.skife.jdbi.v2.exceptions;
+
+public class UnableToRestoreAutoCommitStateException  extends DBIException {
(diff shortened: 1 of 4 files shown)


src/main/java/org/skife/jdbi/v2/exceptions/UnableToRestoreAutoCommitStateException.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.skife.jdbi.v2.exceptions;
+
+public class UnableToRestoreAutoCommitStateException  extends DBIException {
```
stevenschlansker: Unnecessary double space

src/main/java/org/skife/jdbi/v2/exceptions/UnableToRestoreAutoCommitStateException.java
```diff
@@ -0,0 +1,23 @@
...
+
```
stevenschlansker: Missing serialVersionUID

src/main/java/org/skife/jdbi/v2/tweak/transactions/LocalTransactionHandler.java
```diff
@@ -228,6 +213,24 @@ public void setRollbackOnly()
...
+            final LocalStuff stuff = localStuff.remove(handle);
+            if (stuff != null) {
+                handle.getConnection().setAutoCommit(stuff.getInitialAutocommit());
+                stuff.getCheckpoints().clear();
+            }
+        } catch (SQLException e) {
+            throw new UnableToRestoreAutoCommitStateException(e);
+        } finally {
+            // prevent memory leak if rollback throws an exception
+            if (localStuff.containsKey(handle)) {
```
stevenschlansker: Checking `containsKey` before calling `remove` is redundant and should be removed

src/test/java/org/skife/jdbi/v2/DBITestCase.java
```diff
@@ -89,22 +89,27 @@ protected StatementLocator getStatementLocator()
...
+    {
         BasicHandle h = new BasicHandle(getTransactionHandler(),
-                                        getStatementLocator(),
-                                        new CachingStatementBuilder(new DefaultStatementBuilder()),
-                                        new ColonPrefixNamedParamStatementRewriter(),
-                                        conn,
-                                        new HashMap<String, Object>(),
-                                        new NoOpLog(),
-                                        TimingCollector.NOP_TIMING_COLLECTOR,
-                                        new MappingRegistry(),
```
stevenschlansker: Unnecessary whitespace changes

stevenschlansker:
@hawkan Thanks, looks very close to good - couple of style nits.  Mind fixing those, squishing the commits, and I'll merge it?

PR #127, opened 2014-12-31
Remove findbugs-annotation dependency

### LINKED ISSUE #106: Upgrade to findbugs jsr305 2.0.3 + make jsr305 dependency optional
Replace `com.google.code.findbugs:annotations:2.0.1` with:

```
<dependency>
    <groupId>com.google.code.findbugs</groupId>
    <artifactId>jsr305</artifactId>
    <version>2.0.3</version>
    <optional>true</optional>
</dependency>
```

It should also be optional.

(On a related note: `jsr305`, not `annotations`, is also the dependency used by Guava 17.)

arteam:
Fix #106
Findbugs-annotations is used only in one place and could be replaced by a rule in the XML annotator.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,11 +136,6 @@
         </dependency>
 
         <dependency>
-            <groupId>com.google.code.findbugs</groupId>
-            <artifactId>annotations</artifactId>
-        </dependency>
-
-        <dependency>
             <groupId>junit</groupId>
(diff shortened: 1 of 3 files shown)


stevenschlansker:
Yes, optional dependency sounds good to me.  Sorry for merging before the feedback came through.

PR #128, opened 2015-01-02
Update to basepom-8

hgschmie:
This also requires the included cglib to be upgraded to 3.1 but as this
dependency is shaded into the jar it should not be a compatibility problem.


stevenschlansker:
Looks like the annotations jar is compiled against Java 7

```
[WARNING] /home/travis/build/jdbi/jdbi/src/main/java/org/skife/jdbi/v2/sqlobject/BatchHandler.java: /home/travis/.m2/repository/com/google/code/findbugs/annotations/3.0.0/annotations-3.0.0.jar(edu/umd/cs/findbugs/annotations/SuppressFBWarnings.class): major version 51 is newer than 50, the highest major version supported by this compiler.
```

hgschmie:
... stupid java 6 ... :-P

PR #140, opened 2015-02-05
Support for executing a script as a set of statements

arteam:
It's highly useful during development. In that case JDBI returns a debug statement context, where we can inspect information about a failed statement.

If the script is executed as a batch, we don't have such information and are forced to search an error manually.


arteam:
Thank you for merging this.

By the way, do you have any ETA on 2.60 release?  
It's not an issue, I am asking just to get information.

stevenschlansker:
I was hoping to let @christophercurrie review #70 before releasing, but if you get bored of waiting poke me again and I'll make one.

PR #141, opened 2015-02-08
Give a warning if SingleValueResult annotation is omitted

arteam:
If a user set a parametrized type as a return type of method and didn't set @SingleValueResult then generic information is erased and JDBI will try to map only generic type.

The problem that the user doesn't have a clue about why his query doesn't work. 
He will just get a mapper exception or null.

In this this case we could give a warning to him that he probably forgot to place an annotation to the method. This is highly useful during development, when probability of missing an annotation is very high.

Because JDBI doesn't have any logger dependency, I've chosen *java.util.logging* as a target. Because this is message is intended only for development environment, printing to the error console should be fine.

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/ResultReturnThing.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/ResultReturnThing.java
@@ -29,9 +29,13 @@ import org.skife.jdbi.v2.tweak.ResultSetMapper;
 
 import java.util.Iterator;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 abstract class ResultReturnThing
 {
+    private static final Logger LOGGER = Logger.getLogger(ResultReturnThing.class.getName());
+
     public Object map(ResolvedMethod method, Query q, HandleDing h)
     {
         if (method.getRawMember().isAnnotationPresent(Mapper.class)) {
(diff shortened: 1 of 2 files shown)


stevenschlansker:
Hm, I've long been pushing for a `slf4j` dependency but Brian never wanted it.  But `j.u.l` is arguably worse :/
I wonder what to do... is there any case where the warning could trigger on a correct configuration?  Should we make it throw an exception instead?

arteam:
I gave more thought to this and came to a conclusion that there are valid use cases for parametrized
types as a return value. Consider this case:

````java
@Test
public void testParametrizedResult() {
    OrderDao dao = dbi.onDemand(OrderDao.class);
    dbi.withHandle(new VoidHandleCallback() {
            @Override
            protected void execute(Handle handle) throws Exception {
            handle.execute("create table orders (" +
                    "  order_id bigint primary key," +
                    "  result_id bigint not null" +
                    "); " +
                    "create table billing_errors(" +
                    "  result_id bigint not null," +
                    "  code_id int" +
                    "); " +
                    "insert into billing_errors values (1, 214), (2, 215);" +
                    "insert into orders values (78, 1);" +
                    "insert into orders values (79, 2);");
        }
    });

    Order<BillingError> order = dao.getOrderFailedInBilling(78L);
    System.out.println(order.id);
    System.out.println(order.result.code);
}

public static class Order<T> {
    public final long id;
    public final T result;

    public Order(long id, T result) {
        this.id = id;
        this.result = result;
    }
}

public static class BillingError {
    public final int code;

    public BillingError(int code) {
        this.code = code;
    }
}

public static class BillingOrderMapper implements ResultSetMapper<Order<BillingError>> {
    @Override
    public Order<BillingError> map(int index, ResultSet r, StatementContext ctx) throws SQLException {
        return new Order<BillingError>(r.getLong("order_id"), new BillingError(r.getInt("code_id")));
    }
}

public static interface OrderDao {

    @SqlQuery("select o.order_id, be.code_id from orders o " +
            "inner join billing_errors be on be.result_id=o.result_id " +
            "where o.order_id = :id")
    @Mapper(BillingOrderMapper.class)
    Order<BillingError> getOrderFailedInBilling(@Bind("id") long id);
}
````

It's a legitimitate code, there shouldn't be any warnings, but we will give one.

So, the problem that we can't distinct regular types from container types. We need access to for registers of mappers and containers, but we don't have it.

I guess nothing we can do in this situation and it's better to leave things as they are, than
frustrate users on updates.

PR #142, opened 2015-02-08
Support for a positional binder in SQL Object queries

arteam:
By default users of the SQLObject API must map method parameters to query arguments by specifying `@Bind` annotation with a name of the argument.

This is a standard and most reliable way. But in most cases it seems as overkill because method parameter names are often present in the same order as query arguments. We are forced to write boilerplate code to glue them.

It would be great to have facility not to specify binding names, but rely on parameter positions and let the library handle them. Something like that:

````java
interface SomethingDao {

    @SqlQuery("select name from something where something_id=?")
    String findNameById(int i);

    @SqlQuery("select something_id from something where name=? and code=?")
    Integer getIdByNameAndCode(String name, int code);

    @SqlUpdate("insert into something(something_id, name, code) values (?, ?, ?)")
    void insertSomething(int id, String name, int code);
}
````

Moreover, the Fluent API has this facility and support for it in the SQLObject API will make migration from one API to another more smooth.

This is related to #133 and #46, but in contrast with former it doesn't use any annotations and in 
contrast with later doesn't use parameter names at all. It's not mutually exclusive with any of them, so
these features could also be implemented.

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/CustomizingStatementHandler.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/CustomizingStatementHandler.java
@@ -76,8 +76,8 @@ abstract class CustomizingStatementHandler implements Handler
 
         final Annotation[][] param_annotations = method.getRawMember().getParameterAnnotations();
         for (int param_idx = 0; param_idx < param_annotations.length; param_idx++) {
-            final Annotation[] annotations = param_annotations[param_idx];
-            for (final Annotation annotation : annotations) {
+            boolean thereBindingAnnotation = false;
+            for (final Annotation annotation : param_annotations[param_idx]) {
                 final Class<? extends Annotation> anno_class = annotation.annotationType();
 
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by stevenschlansker** [comment]
So my biggest worry here is that if you did not intend to use this feature, but accidentally omit `@Bind`, it seems that suddenly your parameters may be bound out of order.  This could cause some seriously confusing errors.

What's your opinion of a slightly different approach -- instead of binding it positionally, bind it with a name that is the parameter index.  So your example would instead look like
```
    @SqlUpdate("insert into something(something_id, name, code) values (:0, :1, :2)")
    void insertSomething(int id, String name, int code);
```
Now it is extremely unlikely to accidentally use this feature.
You also have the advantage that the function parameter order does not have to match the SQL parameter order, in case you later wish to refactor but do not want to break your Java API.

In the upcoming JDBI3, I would suggest that we instead use the parameter names directly, which will make this obsolete IMO.

arteam:
I agree with all your points. Thanks for feedback.
I've rebased this branch, please take a look.

Of course if we complied against Java 8 it would be tempting to use parameter names for resolving this problem, but unfortunately JDBI3 is a far project and Java 8 is not available in all production servers yet. This approach is a kludge, but not very ugly and should work.

stevenschlansker:
Released in 2.60

PR #145, opened 2015-02-24
Switch cglib-nodep to cglib, so we can manage ASM version to bring in bugfixes

### LINKED ISSUE #144: IllegalArgumentException when a DAO is initialized (Java 8 support?)
I've been using jdbi via [dropwizard](https://github.com/dropwizard/dropwizard/tree/master/dropwizard-jdbi) for a while now. On dropwizard 0.7.1 version, it was using jdbi 2.55 and I had no problems.

Then when I upgraded my dropwizard to 0.8-rc2, they had updated jdbi to 2.57. I started having the following exception when I run some of my tests via gradle. (Was not happening when run on Intellij, but I think it was because how Intellij handles external libraries with multiple module projects)

*(Well, the stack trace was different as the cglib and asm libraries were external libraries and didn't have jdbi package names. It's a bit difficult to get the exact stack trace now.)*
```stacktrace
java.lang.IllegalArgumentException
	at org.skife.jdbi.cglib.asm.ClassReader.<init>(Unknown Source)
	at org.skife.jdbi.cglib.asm.ClassReader.<init>(Unknown Source)
	at org.skife.jdbi.cglib.asm.ClassReader.<init>(Unknown Source)
	at org.skife.jdbi.cglib.proxy.BridgeMethodResolver.resolveAll(BridgeMethodResolver.java:61)
	at org.skife.jdbi.cglib.proxy.Enhancer.emitMethods(Enhancer.java:911)
	at org.skife.jdbi.cglib.proxy.Enhancer.generateClass(Enhancer.java:498)
	at org.skife.jdbi.cglib.core.DefaultGeneratorStrategy.generate(DefaultGeneratorStrategy.java:25)
	at org.skife.jdbi.cglib.core.AbstractClassGenerator.create(AbstractClassGenerator.java:216)
	at org.skife.jdbi.cglib.proxy.Enhancer.createHelper(Enhancer.java:377)
	at org.skife.jdbi.cglib.proxy.Enhancer.create(Enhancer.java:285)
	at org.skife.jdbi.v2.sqlobject.SqlObject.buildSqlObject(SqlObject.java:78)
	at org.skife.jdbi.v2.sqlobject.SqlObjectBuilder.onDemand(SqlObjectBuilder.java:66)
	at org.skife.jdbi.v2.DBI.onDemand(DBI.java:350)
    ...
```

Gradle failing but not Intellij usually comes down to libraries clashing somehow, so I started digging into that and came across to [this stackoverflow question](http://stackoverflow.com/questions/27092857/cglib-throws-an-illegalargumentexception-when-enhancing-the-java-util-date-class). Figured it's something with java 8 support of asm, I came up with a temporary solution after many trial and errors:

```gradle
    compile('io.dropwizard:dropwizard-jdbi:0.8-rc2'){
        exclude group:'cglib'
    }
    compile 'cglib:cglib:3.1'
    compile 'org.ow2.asm:asm:5.0.3'
```

Now, dropwizard has a new version, rc3 and they have updated jdbi to 2.59. My temporary solution is not working anymore since jdbi has moved the sourcecode of asm and cglib into it's own package (or must be something like that as the classes' package name start with `org.skife.jdbi`). So I digged into the ClassReader code and I can see the difference between jdbi's asm and asm 5.0.3:

##### jdbi:
```java
public ClassReader(byte[] var1, int var2, int var3) {
    this.b = var1;
    if(this.readShort(var2 + 6) > 51) {
      throw new IllegalArgumentException();
    } else {
      this.a = new int[this.readUnsignedShort(var2 + 8)];
```

##### asm:
```java
 public ClassReader(final byte[] b, final int off, final int len) {
        this.b = b;
        // checks the class version
        if (readShort(off + 6) > Opcodes.V1_8) {
            throw new IllegalArgumentException();
        }
        // parses the constant pool
        items = new int[readUnsignedShort(off + 8)];
```

I'm not sure what exactly this code does, and why my DAO triggers this (I use Instants and Optionals that are specific to java 8) but it looks like there is a specific Java 8 check on the latest asm code. Any way to work around or fix?

stevenschlansker:
Need to exclude cglib-nodep from easymock
Manage ASM to 5.0.2 which works on Java 8, fixes #144

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -3,6 +3,8 @@
   - Add support for /* */ style comments in statements
   - Add @BindMap annotation which allows parameters passed in a Map<String, Object>
   - Add support for running Script objects as individual statements rather than batch
+  - Add support for default bind name based on argument position number (thanks @arteam)
+  - Fix SqlObject connection leak through result iterator (thanks @pierre)
 
 2.59
   - Fixes #137, broken ClasspathStatementLocator cache (thanks @HiJon89).
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -5,6 +5,8 @@
   - Add support for running Script objects as individual statements rather than batch
   - Add support for default bind name based on argument position number (thanks @arteam)
   - Fix SqlObject connection leak through result iterator (thanks @pierre)
+  - Switch to using cglib instead of cglib-nodep so we can pull ASM 5.0.2 which is Java 8 compatible
+  - Classmate to 1.1.0
 
 2.59
   - Fixes #137, broken ClasspathStatementLocator cache (thanks @HiJon89).
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
     <parent>
         <groupId>org.basepom</groupId>
         <artifactId>basepom-standard-oss</artifactId>
-        <version>9</version>
+        <version>11</version>
     </parent>
 
     <modelVersion>4.0.0</modelVersion>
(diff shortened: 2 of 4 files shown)


stevenschlansker:
requires basepom#14

Issue #133, opened 2015-01-16
Positional arguments for SQL Object Queries


StanSvec:
Hi,

I am finally about to implement the support for the positional arguments as promised in [this topic] ( https://groups.google.com/forum/?fromgroups#!topic/jdbi/WWoky-h23wY). Are you OK with the suggested solution like this:
```java
void insertIntoSomething(@Pos(0) String param1, @Pos(1) int param2, @Pos(2) int param3)
```
Thanks.

stevenschlansker:
Closed via #142

PR #149, opened 2015-03-15
Positional arguments for SQL Object Queries using Pos annotation

StanSvec:
Related discussion:
https://groups.google.com/forum/?fromgroups#!topic/jdbi/WWoky-h23wY
Related issue:
https://github.com/jdbi/jdbi/issues/133


brianm:
Hmm, the <code>@Pos</code> annotation is hard to grok out of context. @Position would be better, but gets long.

What if we just defaulted to positional if no annotation is present? I think most sql dialects let you do things like "where age > ?2 & age < ?1" to do reordering in the sql from the bind order.

arteam:
Brian, please, take a look at PR #142. It's very similar to your proposition.

stevenschlansker:
Closed via #142

PR #158, opened 2015-07-21
Fix regression in withHandle

christophercurrie:
PR #156 introduced a regression due to Java 8 overload resolution. This change renames all `withHandle` methods that take a `HandleConsumer` to `useHandle`, and all `inTransaction` methods that take a `TransactionConsumer` to `useTransaction`, to eliminate the ambiguous overload.

cc @osi

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/BasicHandle.java
+++ b/src/main/java/org/skife/jdbi/v2/BasicHandle.java
@@ -328,12 +328,12 @@ class BasicHandle implements Handle
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


hgschmie:
Test cases to catch the regression and then fix it would be nice.

christophercurrie:
That would require being able to test in Java 8. Do you know of a way to allow that in Maven?

stevenschlansker:
I'm going to release this at 4pm today if nobody puts a hold on it beforehand.

stevenschlansker:
Released as 2.63.1

Issue #154, opened 2015-05-22
Null handling Mappers of for boxed types


nileema:
Mappers for boxed types (LongMapper, ShortMapper, DoubleMapper, etc) return `0` if the column value was `null`. As a consequence, the SqlQuery returning a boxed type may return a `0` or `null` based on whether the column value. There is no way to know if the column value was actually `0` or `null`. 
We can change the Mapper to do the following
```java
        long l = r.getLong(name);
        if (r.wasNull()) {
            return null;
        }
        return l;
```
However that is a backwards incompatible change and does not provide a way to differentiate between `NULL` was found and no rows were found.

electrum:
As another example, consider this SQL object query:
```java
@SqlQuery("select id from something where name = :name")
Long findIdByName(@Bind("name") String name);
```
The above can actually return three things: `null` if no record matches, `0` if the record matches but `id` is null, or the value of `id` (which might legitimately be `0`).

Due to backwards compatiblity, I'm not sure if `LongMapper` and related can be fixed (at least in v2), but I think the SQL object behavior should be fixed. Also, I think both should be fixed for v3.

stevenschlansker:
+1, this should definitely get fixed in v3

qualidafial:
This is something I've been looking at as part of #164, and I'm not sure if it's possible to address this without breaking existing code.

Certainly it would be trivial to enhance the primitive mappers to distinguish between mapping primitives and wrappers, e.g.:

    class BooleanMapper {
      static final BooleanMapper FIRST = new BooleanMapper(1);
      static final BooleanMapper PRIMITIVE = FIRST;
      static final BooleanMapper WRAPPER = new BooleanMapper(1) {
        @Override
        protected Boolean extractByName(ResultSet r, String name)  {
          boolean value = super.extractByName(r, name);
          return r.wasNull() ? null : value;
        }
        @Override
        protected Boolean extractByIndex(ResultSet r, int index) {
          boolean value = super.extractByIndex(r, i);
          return r.wasNull() ? null : value;
        }
      };
    }

Then we could modify the primitives mapper factory to use `BooleanMapper.PRIMITIVE` for `boolean.class`, and `BooleanMapper.WRAPPER` for `Boolean.class`.

This is arguably a more correct way to handle mapping, and has the benefit of allowing users to throw away any checks for zero.

[...]

PR #164, opened 2015-08-26
Enhance BeanMapper to take advantage of registered mappers

qualidafial:
This pull request:
* Introduces a column-centric mapper interface: ResultColumnMapper.
* Adds a separate column mapper registry, with methods to register column mappers in `DBI`, `Query`, and `Handle`.
* Adds `@RegisterColumnMapper`, `@RegisterColumnMapperFactory` customizing annotations.
* Exposes the column mapper registry for reuse by `ResultSetMapper`s: `StatementContext.columnMapperFor(Class, StatementContext)`.
* Modifies `BeanMapper`, `ReflectionBeanMapper` to delegate to the column mapper registry for each property.


qualidafial:
Idea: implement SingleColumnResultSetView using proxies instead of static code. This might remove the need for a workaround for methods that are or are not present in different JDBC versions.

src/main/java/org/skife/jdbi/v2/StatementContext.java
```diff
@@ -55,6 +57,14 @@
...
     Map<String, Object> getAttributes();
 
+    /**
+     * Obtain a column mapper for the given type in this context.
+     *
+     * @param type the target type to map to
+     * @return a ResultColumnMapper for the given type
+     * @throws org.skife.jdbi.v2.exceptions.DBIException if no mappers are registered for the given type.
+     */
+    ResultColumnMapper columnMapperFor(Class type);
```
hgschmie: I am pretty sure that this is not strong enough. This should be `Type` so that it is possible to pass in e.g. parameterized types. Interfaces that take `Class<?>` almost always needs retrofitting for later Java versions.

qualidafial: I see how this could be useful for automatic composition, e.g. chaining an `OptionalMapper` and a `LocalDateMapper` to read a `date` column into an `Optional<LocalDate>`.

I chose `Class` here to make it agree with existing API, in particular `ResultSetMapperFactory.mapperFor(Class,StatementContext)`.

Are you ok with making the `ResultSetMapper` and `ResultColumnMapper` interfaces diverge in v2?

hgschmie: Yes. In hindsight, the ResultSetMapperFactory should have probably also taken a Type instead of Class, but I guess @brianm was young and needed the money. :-)

This is a case of "will need retrofitting in a later version". But as we are stuck with our public API, we will have to wait for JDBIv3.

qualidafial: So I should leave it as `Class`, and the change to `Type` will go on the v3 wishlist?

stevenschlansker: Yeah, let's be consistent for now, and fix both in v3.

qualidafial:
See my commit message in dfbfa0a.

After seeing both options, I think it's better to make a clean break, and have column mappers be their own distinct thing, rather than trying to make all `TypedMapper` classes automatically inherit `ResultColumnMapper`.

Making the column mapper registry a first-class member of the API brings the query API on par with the update API. By that I just like `ArgumentFactory` lets you use arbitrary argument types with an SQL statement, `ResultColumnMapper` lets you receive arbitrary value types without having to write one-off mappers.

This will allow third-party libraries to distribute `ResultColumnMapperFactory` that allows any mapper that uses the column mapper registry to automatically recognize new types. e.g. a `Jsr310ColumnMapperFactory` might provide mappers for `LocalDate`, `OffsetDateTime`, etc.

The mapping registry will wrap a registered column mapper into a first-column `ResultSetMapper` if no row mapper is explicitly registered for a given type.

The column mapper registry is distinct and strictly opt-in. So no behavior should change until users adopt this interface.

A few items left to do:
* Add `@RegisterColumnMapper`, `@RegisterColumnMapperFactory` customizing annotations.
* Take inventory of all new behaviors and add as many tests as possible to cover them.
  * Test adapting column mapper to row mapper
* Deprecate TypedMapper in favor of ResultColumnMapper?
* Deprecate PrimitivesMapperFactory?
* Split `EnumColumnMapperFactory.BY_NAME` and `EnumColumnMapperFactory.BY_ORDINAL` into distinct classes, so they can be used with `@RegisterColumnMapperFactory`?
* Remove row-level primitive mappers, and allow fall-back on column mapper registry? This could break backward compatibility, as the primitive column mappers distinguish between primitive and wrapper mappers. So where row mappers previously returned zero or false for SQL NULL, they would now return Java Null for SQL NULL.

src/main/java/org/skife/jdbi/v2/ResultColumnMapperFactory.java
```diff
@@ -0,0 +1,36 @@
...
+     * Can this factory provide a column mapper which maps to the desired type
+     * @param type the target type to map to
+     * @return true if it can, false if it cannot
+     */
+    boolean accepts(Class type, StatementContext ctx);
+
+    /**
+     * Supplies a column mapper which will map result set columns to type
+     */
+    ResultColumnMapper columnMapperFor(Class type, StatementContext ctx);
```
qualidafial: I'm considering renaming this to `mapperFor`, to guarantee that no class implements both `ResultSetMapperFactory` and `ResultColumnMapperFactory`.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2004 - 2014 Brian McCallister
+ * Copyright (C) 2004 - 2015 Brian McCallister and others
```
stevenschlansker: I am not at all a fan of per-file license headers, nor a lawyer, but if we change this here we should change it globally :)

src/main/java/org/skife/jdbi/v2/BeanMapper.java
```diff
@@ -81,55 +78,18 @@ public T map(int row, ResultSet rs, StatementContext ctx)
...
-                    value = rs.getDate(i);
-                }
-                else if (type.isAssignableFrom(String.class)) {
-                    value = rs.getString(i);
+                ResultColumnMapper mapper = ctx.columnMapperFor(type);
+                if (mapper != null) {
+                    value = mapper.mapColumn(rs, i, ctx);
                 }
                 else if (type.isEnum()) {
                     String str = rs.getString(i);
```
stevenschlansker: Should Enum handling move to the `ResultColumnMapperFactory`?
Seems weird to have this edge case here now.

src/main/java/org/skife/jdbi/v2/InferredColumnMapperFactory.java
```diff
@@ -0,0 +1,53 @@
...
+import com.fasterxml.classmate.ResolvedType;
+import com.fasterxml.classmate.TypeResolver;
+import org.skife.jdbi.v2.tweak.ResultColumnMapper;
+
+import java.util.List;
+
+class InferredColumnMapperFactory implements ResultColumnMapperFactory
+{
+    private static final TypeResolver tr = new TypeResolver();
+    private final Class maps;
```
stevenschlansker: Raw types are evil, unless they are particularly onerous.  Can this be `Class<?>`?

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
```diff
@@ -49,32 +58,75 @@ public void add(ResultSetMapper mapper)
...
+        if (columnMapper != null) {
+            ResultSetMapper mapper = new SingleColumnMapper(columnMapper);
+            rowCache.put(type, mapper);
             return mapper;
         }
 
         throw new DBIException("No mapper registered for " + type.getName()) {};
     }
+
+    public void addColumn(ResultColumnMapper mapper)
```
stevenschlansker: I don't like the `addColumn` name, could this be `addColumnMapper` instead?

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
```diff
@@ -49,32 +58,75 @@ public void add(ResultSetMapper mapper)
...
+    {
+        this.addColumn(new InferredColumnMapperFactory(mapper));
+    }
+
+    public void addColumn(ResultColumnMapperFactory factory) {
+        columnFactories.add(factory);
+        columnCache.clear();
+    }
+
+    public ResultColumnMapper columnMapperFor(Class type, StatementContext ctx) {
```
stevenschlansker: Could this reasonably be `Class<?>`?

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
```diff
@@ -49,32 +58,75 @@ public void add(ResultSetMapper mapper)
...
+        if (columnCache.containsKey(type)) {
```
stevenschlansker: `containsKey` is redundant here, just call `get`?

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
```diff
@@ -49,32 +58,75 @@ public void add(ResultSetMapper mapper)
...
+
+        for (ResultColumnMapperFactory factory : columnFactories) {
+            if (factory.accepts(type, ctx)) {
+                ResultColumnMapper mapper =  factory.columnMapperFor(type, ctx);
+                columnCache.put(type, mapper);
+                return mapper;
+            }
+        }
+
+        if (BUILD_IN_COLUMN_MAPPERS.accepts(type, ctx)) {
```
stevenschlansker: Can we just unconditionally add this to the `columnFactories` and get rid of the special case here?

src/main/java/org/skife/jdbi/v2/PrimitivesColumnMapperFactory.java
```diff
@@ -0,0 +1,81 @@
...
+
+        mappers.put(BigDecimal.class, BigDecimalColumnMapper.INSTANCE);
+
+        mappers.put(String.class, StringColumnMapper.INSTANCE);
+
+        mappers.put(byte[].class, ByteArrayColumnMapper.INSTANCE);
+
+        mappers.put(Timestamp.class, TimestampColumnMapper.INSTANCE);
+
+        mappers.put(URL.class, URLColumnMapper.INSTANCE);
```
stevenschlansker: `URI` might be worthwhile here, if we support `URL`.

src/main/java/org/skife/jdbi/v2/PrimitivesColumnMapperFactory.java
```diff
@@ -0,0 +1,81 @@
...
+    private static final Map<Class, ResultColumnMapper> mappers = new HashMap<Class, ResultColumnMapper>();
+
+    static {
+        mappers.put(boolean.class, BooleanColumnMapper.PRIMITIVE);
+        mappers.put(byte.class, ByteColumnMapper.PRIMITIVE);
+        mappers.put(short.class, ShortColumnMapper.PRIMITIVE);
+        mappers.put(int.class, IntegerColumnMapper.PRIMITIVE);
+        mappers.put(long.class, LongColumnMapper.PRIMITIVE);
+        mappers.put(float.class, FloatColumnMapper.PRIMITIVE);
+        mappers.put(double.class, DoubleColumnMapper.PRIMITIVE);
```
stevenschlansker: `char` and `Character` are noticeably absent here, although I'm not sure how many people would actually use it.

src/main/java/org/skife/jdbi/v2/Query.java
```diff
@@ -420,4 +426,13 @@ public void registerMapper(ResultSetMapperFactory m)
     {
         this.mappingRegistry.add(m);
     }
+
+    public void registerColumnMapper(ResultColumnMapper m)
```
stevenschlansker: I know this is just mirroring existing design decisions, but I think adding methods on `Query` which modify state scoped more appropriately to the `Handle` or even `DBI` itself is bad.  Without understanding the internals, I might assume this register method only applies to the scope of the `Query`.  Maybe we break with tradition here and omit these convenience methods?

src/main/java/org/skife/jdbi/v2/ReflectionBeanMapper.java
```diff
@@ -80,54 +77,19 @@ public T map(int row, ResultSet rs, StatementContext ctx)
...
-                else if (type.isAssignableFrom(String.class)) {
-                    value = rs.getString(i);
+                ResultColumnMapper mapper = ctx.columnMapperFor(type);
+                if (mapper != null) {
+                    value = mapper.mapColumn(rs, i, ctx);
                 }
                 else if (type.isEnum()) {
-                    value = Enum.valueOf(type, rs.getString(i));
+
+                    String str = rs.getString(i);
```
stevenschlansker: Again, feels weird to have a special case for `Enum` types here.

src/main/java/org/skife/jdbi/v2/ResultColumnMapperFactory.java
stevenschlansker: I'm a bit hesitant to introduce arbitrary restrictions without a justification.  I could be convinced, I see why you are tempted, but maybe someone who knows what they are doing could implement both usefully...

src/main/java/org/skife/jdbi/v2/util/BooleanColumnMapper.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.skife.jdbi.v2.StatementContext;
+import org.skife.jdbi.v2.tweak.ResultColumnMapper;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+public enum BooleanColumnMapper implements ResultColumnMapper<Boolean> {
+    PRIMITIVE(false),
+    WRAPPER(true);
```
stevenschlansker: :+1: This is very nice

src/main/java/org/skife/jdbi/v2/util/EnumColumnMapperFactory.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+package org.skife.jdbi.v2.util;
+
+import org.skife.jdbi.v2.ResultColumnMapperFactory;
+import org.skife.jdbi.v2.StatementContext;
+import org.skife.jdbi.v2.tweak.ResultColumnMapper;
+
+public enum EnumColumnMapperFactory implements ResultColumnMapperFactory {
+    BY_NAME() {
```
stevenschlansker: These empty double parens confused me greatly for a moment, can they go?

src/test/java/org/skife/jdbi/v2/BeanMapperTest.java
```diff
@@ -0,0 +1,270 @@
...
+        expect(resultSet.wasNull()).andReturn(false);
+        replay(resultSet);
+
+        SampleBean sampleBean = mapper.map(0, resultSet, ctx);
+
+        assertSame(aLongVal, sampleBean.getLongField());
+    }
+
+    @Test
+    public void shouldHandleEmptyResult() throws Exception {
```
stevenschlansker: I'm a little confused about this behavior.  Seems to me if you attempt to map a Bean but do not include the appropriate columns in your `SELECT` it may be better to throw an error rather than silently return an empty bean.  Then again maybe selecting a subset of properties and defaulting the rest is useful.

src/test/java/org/skife/jdbi/v2/BeanMapperTest.java
```diff
@@ -0,0 +1,270 @@
...
+        replay(ctx);
+
+        expect(resultSet.getMetaData()).andReturn(resultSetMetaData);
+        Long aLongVal = 100l;
+        expect(resultSet.getLong(1)).andReturn(aLongVal);
+        expect(resultSet.wasNull()).andReturn(false);
+        replay(resultSet);
+        SampleBean sampleBean = mapper.map(0, resultSet, ctx);
+
+        assertSame(aLongVal, sampleBean.getLongField());
```
stevenschlansker: `assertSame` here unintentionally tests the `Long` cache, if I'm not mistaken... is `assertEquals` better?

src/test/java/org/skife/jdbi/v2/BeanMapperTest.java
```diff
@@ -0,0 +1,270 @@
...
+
+        expect(resultSet.getMetaData()).andReturn(resultSetMetaData);
+        long expected = 1L;
+        expect(resultSet.getLong(1)).andReturn(expected);
+        expect(resultSet.wasNull()).andReturn(false).anyTimes();
+        replay(resultSet);
+
+        SampleBean sampleBean = mapper.map(0, resultSet, ctx);
+
+        assertSame(expected, sampleBean.getLongField());
```
stevenschlansker: Ditto.

src/test/java/org/skife/jdbi/v2/SampleBean.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public void setLongField(Long longField) {
+        this.longField = longField;
+    }
+
+    public String getStringField() {
+        return stringField;
+    }
+
+    protected void setStringField(String stringField) {
```
stevenschlansker: Seems that for these the methods might reflect what it actually important, e.g. `setProtectedStringField`, just so it is more obvious that they should *not* be called.

stevenschlansker:
In general I am OK with the approach.  One worry is that almost all of the included test cases are extensively mocked, which is fine for testing in isolation, but it would make me feel better to have a couple of "end to end" tests as well.

src/test/java/org/skife/jdbi/v2/SampleBean.java
stevenschlansker: In fact, you could just make them outright `throw new AssertionError()` or something, so calling it obviously fails big.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
qualidafial: Only reason I looked at license headers was because the Maven build barked at me if I left it out of any new files. Then muscle memory from my time as an Eclipse committer took over...

I'm not sure from your comment if you want me to change anything here.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
qualidafial: I left it in to avoid breaking backward compatibility.

By default the mapping registry has no enum mappers, but this class (and `ReflectionBeanMapper`) had some code to read enums by name. I left this here to preserve that behavior so as to not break existing client code.

src/main/java/org/skife/jdbi/v2/InferredColumnMapperFactory.java
qualidafial: Fixed.

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
qualidafial: I was just trying to match the pattern of the existing `add(ResultSetMapper)` methods. I can rename them.

Do you want me to rename the existing `ResultSetMapper[Factory]` methods while I'm at it? This class internal so it shouldn't hurt compatibility.

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
qualidafial: Not in any meaningful way. The compiler can't make any guarantees, and we don't have enough information at runtime to make them statically.

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
qualidafial: A bit redundant, yes. Want me to remove the equivalent block in `mapperFor` as well?

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
qualidafial: This matches the existing design of the `ResultSetMapper` registry, and I think we should stick with that pattern.

Mappers and mapper factories are being consulted in the order they were registered. Built in mappers are consulted last. This allows the user to substitute their own mapper for built in types, if they choose.

If we were to eagerly add the built in mappers, they would be at the front of the line, before any user-registered mappers, which would take away the ability to provide substitute mappers.

src/main/java/org/skife/jdbi/v2/PrimitivesColumnMapperFactory.java
qualidafial: Maybe, but it's not directly supported by `ResultSet`, so I'd rather that be a separate issue. This PR already huge as it is.. :smile:

src/main/java/org/skife/jdbi/v2/PrimitivesColumnMapperFactory.java
qualidafial: I noticed that too, but as they are not directly supported by `ResultSet` I opted to leave them out for now.

src/main/java/org/skife/jdbi/v2/Query.java
qualidafial: I think this needs to exist so that the upcoming `@RegisterColumnMapper` and `@RegisterColumnMapperFactory` annotations on query methods can be correctly scoped.

src/main/java/org/skife/jdbi/v2/ReflectionBeanMapper.java
qualidafial: Agreed. I'm just trying to preserve existing behavior so I don't make existing users cranky.

src/main/java/org/skife/jdbi/v2/util/EnumColumnMapperFactory.java
qualidafial: I could have sworn they were required (like for anonymous inners) but the compiler says they're ok. So.. yes.

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
stevenschlansker: Yes please.

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
stevenschlansker: Even if it's not meaningful, parameterizing it removes an IDE warning.

src/test/java/org/skife/jdbi/v2/BeanMapperTest.java
qualidafial: This test was copied over from `ReflectionBeanMapper` before I started, to capture existing behavior.

I agree it's better to throw exception than to silently return empties, however this behavior is already present in `BeanMapper`. Can this be handled in a separate PR?

src/test/java/org/skife/jdbi/v2/BeanMapperTest.java
qualidafial: Will fix

src/test/java/org/skife/jdbi/v2/BeanMapperTest.java
qualidafial: ditto too

qualidafial:
I will add end to end tests.

src/main/java/org/skife/jdbi/v2/ResultColumnMapperFactory.java
qualidafial: My reasoning is that there's no need to implement both. If you have a column mapper, it will already be adapted into a first-column row mapper at need.

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
qualidafial: Copy that. Will fix.

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
qualidafial: Fixed.

src/test/java/org/skife/jdbi/v2/BeanMapperTest.java
stevenschlansker: OK.

src/main/java/org/skife/jdbi/v2/ReflectionBeanMapper.java
stevenschlansker: OK.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
stevenschlansker: I'm not sure either.  It's just something I noticed, I don't have any reason to insist a change, just feels weird to have it inconsistent

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
qualidafial: What about adding `@SuppressWarnings("raw")`?

src/test/java/org/skife/jdbi/v2/BeanMapperTest.java
qualidafial: Fixed.

src/test/java/org/skife/jdbi/v2/BeanMapperTest.java
qualidafial: Fixed

src/test/java/org/skife/jdbi/v2/SampleBean.java
qualidafial: Actually in the tests we are expecting an error from `BeanMapper`, so throwing an assertion error in the setter might trigger a false positive.

I will rename the properties to reflect intent.

qualidafial: Fixed

src/main/java/org/skife/jdbi/v2/BeanMapper.java
qualidafial: I take that to mean that every file touched in this PR should have updated license text?

brianm: I'd prefer we just got rid of the license text, personally.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
qualidafial: @stevenschlansker So I found something that may impact design on enums: `BuiltInArgumentFactory` implicitly encodes enum arguments by their name. Because `Foreman` eagerly registers the built-ins on construction, clients cannot override this with their own alternative arguments.

We have to choose between:
* Sticking with `name()`-only encoding. In that case we should drop `EnumColumnMapper.byOrdinal(Class)` and `EnumColumnMapperFactory.BY_ORDINAL` since they are moot.
* Make enum encoding configurable. This would require making `Foreman.waffle` use `BuiltInArgumentFactory` as a last-ditch fallback rather than eagerly adding it to the factory list on construction--similar to what the mapping registry does with built-in mappers.

I prefer the latter.

In either case, it now makes sense for `EnumColumnMapperFactory.BY_NAME` to be included in the built-in column mappers, for parity with the argument factory registry. If we pick the latter option above, then clients would be expected both an argument factory and a column mapper factory if they wish to override enum encoding.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
qualidafial: @brianm do you mean all license text project-wide, or just the changes in this PR?

stevenschlansker: Let's just leave this PR as-is and remove the header globally in a separate followup.  Easier review.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
stevenschlansker: If the only ill effect of merging as-is is the useless (for now) `byOrdinal` stuff, how about we change to allow different enum behavior in a followup PR?

src/main/java/org/skife/jdbi/v2/MappingRegistry.java
stevenschlansker: That could work too.  I'm not a huge fan of suppressing warnings unless there truly is no way to get around them, but this really is a nit given that the existing code already is largely raw so I will defer to what you want to do.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
qualidafial: Let's hold off on merge just a little bit. For one, we want to split out EnumColumnMapperFactory enum instances into distinct classes, so they can be used with `@RegisterColumnMapper`

stevenschlansker: OK in that case, I will go ahead with #171, there may be tiny merge conflicts but should be trivially fixed.

qualidafial:
Current status..

To Do - please don't merge before these are done:
* New end-to-end tests
  * BeanMapper and ReflectionBeanMapper using custom column mapper - I think this is already captured in existing e2e tests that use BeanMapper.
  * Registered column mapper is automatically adapted into first-column row mapper - this might be considered captured as well, in `TestEnums.testMapEnumValues`.
  * `@RegisterColumnMapper` and `@RegisterColumnMapperFactory` work
  * Any other scenarios we think of

Maybe - awaiting input from project members:
* Rename `ResultColumnMapperFactory.columnMapperFor` to `.mapperFor`?
* Generify `MappingRegistry.columnMapperFor`, or suppress raw type warnings
* BeanMapperFactory should consult the column mapper registry instead of `BuiltInArgumentFactory.canAccept`?
* Rename `MappingRegistry.add` to `.addMapper`?
* Deprecate TypedMapper and PrimitivesMapperFactory in favor of ResultColumnMapper?

Deferred until after this PR is merged:
* Make `Foreman.waffle` consult built in argument factories last, so default enum encoding by `name()` can be overridden.
* Add column mappers for char, Character, and URI
* Remove primitive type row mappers, allowing fall-through to column mappers - this could plausibly cause NullPointerExceptions in existing client code. Probably best to defer to v3.
* Add LICENSE.md at project root, and remove license headers in all other sources?

qualidafial:
@stevenschlansker @brianm @hgschmie I have a few minor questions I'd like to resolve, and then I think the PR is ready:

* Rename `ResultColumnMapperFactory.columnMapperFor` to `.mapperFor`? This would prevent implementing both `ResultSetMapperFactory` and `ResultColumnMapperFactory`. The idea is that since registered column mappers are automatically adapted into result set mappers, there is really no reason for someone to implement both factories types for the same value type.
* Should BeanMapperFactory consult the column mapper registry instead of `BuiltInArgumentFactory.canAccept`? My instinct is yes.
* Should TypedMapper and PrimitivesMapperFactory be deprecated, in favor of ResultColumnMapper?

src/main/java/org/skife/jdbi/v2/BeanMapper.java
qualidafial: submitted issue [175](https://github.com/jdbi/jdbi/issues/175)

src/main/java/org/skife/jdbi/v2/PrimitivesColumnMapperFactory.java
qualidafial: submitted issue [176](https://github.com/jdbi/jdbi/issues/176)

src/main/java/org/skife/jdbi/v2/PrimitivesColumnMapperFactory.java
qualidafial: submitted issue [176](https://github.com/jdbi/jdbi/issues/176)

qualidafial:
I believe I've addressed all comments, and this PR should be good to merge. :tada:

stevenschlansker:
Cool!  Thanks for your patience and persistence here, I will review shortly.

qualidafial:
Any new feedback?

qualidafial:
Follow up: I have a [branch](https://github.com/zanebenefits/jdbi/commits/deprecate-primitives-mapper-factory) with a single commit, which deprecates the primitives mapper factories in favor of the column mapper registry, and addresses issue #154

brcolow:
What is the best way for new users of jdbi to learn about new features introduced in this PR (assuming there are some for consumer of the jdbi API and it's not all internal stuff). I am up-to-date with 2.63 offerings via documentation and I understand 2.69 is in alpha, but wanted to ask just in case there was some way for me to learn about these changes. Thanks!

stevenschlansker:
@brcolow You bring up a good point.  Probably the upcoming 2.70 release is worthy of a blog post describing the new features.

qualidafial:
What about a PR against [brianm/jdbi-docs](https://github.com/brianm/jdbi-docs)?

stevenschlansker:
Yeah, that's what I meant.  But I've never built the site before so it might take some figuring out :P

qualidafial:
Ok, you figure out the site build process, and I'll start a branch to update the docs.

Side question: why isn't that repository under the jdbi org?

stevenschlansker:
Yeah I noticed that too.  @brianm want to transfer it?

brianm:
Transferred. Looks like it still deploys to an ancient VPS of mine, should
move it to S3 as well.

Will figure that out :-)

On Wed, Sep 30, 2015 at 6:48 PM, Steven Schlansker <notifications@github.com

brcolow:
Thanks much - looking forward to it.

qualidafial:
Opened pull request jdbi/jdbi-docs#15

PR #166, opened 2015-09-01
Throw appropriate exception in case of nested transactions

### LINKED ISSUE #165: Nested transactions
Hey there,

I have a DAO with those methods:

```java
public abstract class UserJdbiRepository{
{
    @SqlUpdate("insert into User ...")
    @GetGeneratedKeys
    public abstract long saveUser(@BindBean @BindUser final UserEntity user);

    @Transaction
    public void insertUser(final UserEntity user)
    {
            [...]
            saveUser(user);
            [...]
            saveUserRoleAssignments(user.getId(), user.getRoles().get());
            [...]
    }

    @Transaction
    public void saveUserRoleAssignments(final String userId, final Set<Role> roles)
    {
        [...]
    }
}
```

When I insert a user, I want to save the roles as part of it. But I also have the use case to separately save the roles, hence I added the `@Transaction` annotation to saveUserRoleAssignments.

When calling insertUser (and working with a MySQL database), I receive this error: 

"Failed to rollback transaction" - `Caused by: com.mysql.jdbc.exceptions.jdbc4.MySQLNonTransientConnectionException: Can't call rollback when autocommit=true`

But there is no exception that could have triggered the rollback.

When I, however, remove the `@Transaction` from saveUserRoleAssignments, then insertUser works fine without any problems.

Is it possible that you cannot make nested calls with functions that have the `@Transaction` annotation? I'd expect the second call to recognise that a transaction is already open and not do anything (but to open a transaction when called from outside, not via insertUser).

BernhardBln:
* When making a call through several methods of a DAO, and more than one
of those methods have a @Transaction annotation, an exception is thrown
by certain databases (e.g. MySQL). 

* This exception, however, is not really clear and does not help to
detect the error.

* Since nested transactions are not supported by JDBI at the moment,
throw a TransactionException with a more meaninful error message.

Fixes #165

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/PassThroughTransactionHandler.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/PassThroughTransactionHandler.java
@@ -20,6 +20,7 @@ import org.skife.jdbi.v2.Handle;
 import org.skife.jdbi.v2.TransactionCallback;
 import org.skife.jdbi.v2.TransactionIsolationLevel;
 import org.skife.jdbi.v2.TransactionStatus;
+import org.skife.jdbi.v2.exceptions.TransactionException;
 
 import java.lang.reflect.Method;
 
@@ -38,6 +39,11 @@ class PassThroughTransactionHandler implements Handler
         ding.retain("pass-through-transaction");
         try {
             Handle h = ding.getHandle();
+
+            if (h.isInTransaction()) {
+                throw new TransactionException("Nested @Transaction detected - this is currently not supported.");
+            }
+
             if (isolation == TransactionIsolationLevel.INVALID_LEVEL) {
                 return h.inTransaction(new TransactionCallback<Object>()
                 {
--- a/src/test/java/org/skife/jdbi/v2/sqlobject/TestClassBasedSqlObject.java
+++ b/src/test/java/org/skife/jdbi/v2/sqlobject/TestClassBasedSqlObject.java
@@ -21,6 +21,7 @@ import org.junit.Test;
 import org.skife.jdbi.v2.DBI;
 import org.skife.jdbi.v2.Handle;
 import org.skife.jdbi.v2.Something;
+import org.skife.jdbi.v2.exceptions.TransactionException;
 import org.skife.jdbi.v2.sqlobject.customizers.RegisterMapper;
 import org.skife.jdbi.v2.sqlobject.subpackage.SomethingDao;
 
@@ -85,6 +86,28 @@ public class TestClassBasedSqlObject
         dao.totallyBroken();
     }
 
+    @Test
+    public void testSimpleTransactionsSucceed() throws Exception
+    {
+        final SomethingDao dao = dbi.onDemand(SomethingDao.class);
+
+        dao.insertInSingleTransaction(10, "Linda");
+    }
+
+    /**
+     * Currently, nested transactions are not supported. Make sure an appropriate exception is
+     * thrown in that case.
+     * <p>
+     *
+     * Side note: H2 does not have a problem with nested transactions - but MySQL has.
+     */
+    @Test(expected = TransactionException.class)
+    public void testNestedTransactionsThrowException()
+    {
+        final SomethingDao dao = dbi.onDemand(SomethingDao.class);
+        dao.insertInNestedTransaction(11, "Angelina");
+    }
+
     @RegisterMapper(SomethingMapper.class)
     public static abstract class Dao
     {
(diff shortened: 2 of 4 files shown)


stevenschlansker:
It looks like this causes an existing test case to fail:
```
Tests run: 5, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.793 sec <<< FAILURE! - in org.skife.jdbi.v2.sqlobject.TestTransactionAnnotation

testNestedTransactions(org.skife.jdbi.v2.sqlobject.TestTransactionAnnotation) Time elapsed: 2.471 sec <<< ERROR!

org.skife.jdbi.v2.exceptions.TransactionException: Nested @Transaction detected - this is currently not supported.

at org.skife.jdbi.v2.sqlobject.PassThroughTransactionHandler.invoke(PassThroughTransactionHandler.java:44)

at org.skife.jdbi.v2.sqlobject.SqlObject.invoke(SqlObject.java:175)

at org.skife.jdbi.v2.sqlobject.SqlObject$1.intercept(SqlObject.java:75)

at org.skife.jdbi.v2.sqlobject.TestTransactionAnnotation$Dao$$EnhancerByCGLIB$$becd51b2.insertAndFetch(<generated>)

at org.skife.jdbi.v2.sqlobject.TestTransactionAnnotation$Dao.insertAndFetchWithNestedTransaction(TestTransactionAnnotation.java:202)

at org.skife.jdbi.v2.sqlobject.TestTransactionAnnotation$Dao$$EnhancerByCGLIB$$becd51b2.CGLIB$insertAndFetchWithNestedTransaction$4(<generated>)

at org.skife.jdbi.v2.sqlobject.TestTransactionAnnotation$Dao$$EnhancerByCGLIB$$becd51b2$$FastClassByCGLIB$$f4796ffb.invoke(<generated>)

at net.sf.cglib.proxy.MethodProxy.invokeSuper(MethodProxy.java:228)

at org.skife.jdbi.v2.sqlobject.PassThroughTransactionHandler$1.inTransaction(PassThroughTransactionHandler.java:54)

at org.skife.jdbi.v2.tweak.transactions.LocalTransactionHandler.inTransaction(LocalTransactionHandler.java:184)

at org.skife.jdbi.v2.BasicHandle.inTransaction(BasicHandle.java:327)

at org.skife.jdbi.v2.sqlobject.PassThroughTransactionHandler.invoke(PassThroughTransactionHandler.java:48)

at org.skife.jdbi.v2.sqlobject.SqlObject.invoke(SqlObject.java:175)

at org.skife.jdbi.v2.sqlobject.SqlObject$1.intercept(SqlObject.java:75)

at org.skife.jdbi.v2.sqlobject.TestTransactionAnnotation$Dao$$EnhancerByCGLIB$$becd51b2.insertAndFetchWithNestedTransaction(<generated>)

at org.skife.jdbi.v2.sqlobject.TestTransactionAnnotation.testNestedTransactions(TestTransactionAnnotation.java:91)
```

Also maybe @hgschmie or @brianm want to comment on whether this is a correct interpretation of how we should behave.

BernhardBln:
Hi, 

I don't think it is "supported" by H2, it just doesn't throw an error. It seems to be on the roadmap, though:

http://www.h2database.com/html/roadmap.html

So I *think* H2 just silently ignores the commit, which means - coming back to your original argument -that a function with `@Transactional?` marker cannot rely on its changes being committed. (Haven't tested, though, what happens with nested transactions and rollbacks in H2).

Sorry about the failing tests - in fact, I even had compile errors in some parts of the code when I checked it out from your master, and when I test with the classes that compile, I even get more failing tests when the travis build. Is there anything I need to do in addition to get this compiling, like running some byte-code generators?

The failing test is actually a test that checks if using nested transactions work - I suggest to remove it completely.

I don't think catching that unhelpful exception would be a good and clean solution, as you have the risk to catch this exception also in cases when it's actually the right one (I remember it said something about autocommit being enabled).

stevenschlansker:
OK, I agree with you -- removing the test is the correct move, assuming that this is the behavior we want.  If you update that it'll be good by me, but I want signoff from another committer on such a fundamental change before merging.  Thanks again.

BernhardBln:
Ok removed the tests.

stevenschlansker:
LGTM, just want signoff from another committer due to the sensitivity of changing transaction semantics.

BernhardBln:
Sure.

Especially it should be kept in mind that this will break code for folks that use H2 or other databases that are more relaxed, and have nested transactions. While it worked well for them in the past, the code change will make their code throw an exception.

But I think that's good, as it helps them to figure out if their code really does what they want, and they don't rely on wrong assumptions regarding transaction boundaries.

PR #167, opened 2015-09-03
Initial support for plugging in new exception handling strategies.

stockmaj:
We'd like to move off of Spring JdbcDaoSupport/JdbcTemplate, but one of the features that is lacking is the exception mapping capabilities.  This change looks like it accomplishes that goal and looks like, with a proper default ExceptionPolicy, it would be easliy backwards compatible


stevenschlansker:
@stockmaj, it looks like this needs to be rebased on top of current master.
@brianm, did you have any further thoughts on this change?  I know we discussed it a while back (almost 3 years ago, where did the time go?!); do you feel like it is good to merge?

stevenschlansker:
This PR needs rebase.  We can reopen the issue if anyone wishes to pursue it.

PR #168, opened 2015-09-05
Correct splitting SQL statements by a semicolon

arteam:
Currently semicolons in a script are treated as statements separators. Not all are. Semicolons in string literals are just symbols. The statement separator should be followed by a reserved
SQL word or the end of string.

The implementation is not guaranteed to work for all cases, but should cover the most common.

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
+++ b/src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
@@ -15,29 +15,25 @@
  */
 package org.skife.jdbi.v2;
 
+import org.antlr.runtime.ANTLRInputStream;
+import org.antlr.runtime.Token;
 import org.skife.jdbi.v2.exceptions.UnableToCreateStatementException;
 import org.skife.jdbi.v2.tweak.StatementLocator;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
-import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.nio.charset.Charset;
 import java.util.Collections;
 import java.util.Map;
 import java.util.WeakHashMap;
 import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 /**
  * looks for [name], then [name].sql on the classpath
  */
 public class ClasspathStatementLocator implements StatementLocator
 {
-    private static final Pattern MULTILINE_COMMENTS = Pattern.compile("/\\*.*?\\*/");
-
     private final Map<String, String> found = Collections.synchronizedMap(new WeakHashMap<String, String>());
 
     /**
@@ -97,9 +93,9 @@ public class ClasspathStatementLocator implements StatementLocator
             return name;
         }
         final ClassLoader loader = selectClassLoader();
-        BufferedReader reader = null;
+        InputStream in_stream = null;
         try {
-            InputStream in_stream = loader.getResourceAsStream(name);
+            in_stream = loader.getResourceAsStream(name);
             if (in_stream == null) {
                 in_stream = loader.getResourceAsStream(name + ".sql");
             }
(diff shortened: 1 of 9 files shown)


stevenschlansker:
@arteam Have you tested this with comments? e.g. 
```
SELECT 1; -- select something
SELECT 2;
```

arteam:
You are right. We don't currently handle such comments in `ClasspathStatementLocator`. We should probably switch to an ANTLR generated lexer for handling them. If it works, we can't generate one also for semicolons instead of regular expressions.

src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -118,24 +119,30 @@ public String locate(String name, StatementContext ctx)
...
-            String line;
+            StringBuilder sb = new StringBuilder();
             try {
-                while ((line = reader.readLine()) != null) {
-                    if (isComment(line)) {
-                        // comment
-                        continue;
+                SqlScriptLexer lexer = new SqlScriptLexer(new ANTLRInputStream(in_stream));
+                while (true) {
+                    Token t = lexer.nextToken();
```
stevenschlansker: This logic could be combined with the similar structure in `Script` below into a helper class, a `SqlStatementSplitter` perhaps?

src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -118,24 +119,30 @@ public String locate(String name, StatementContext ctx)
...
+                    int type = t.getType();
+                    if (type == SqlScriptLexer.EOF) {
```
stevenschlansker: This if construct is dodgy, if new token types are introduced they could be silently ignored.  Perhaps either a `switch` statement, or maybe move the big `else if` into a generic `else` block.

stevenschlansker:
Very cool, a couple of comments left inline, and there's a few FindBugs nits to fix too.

arteam:
Ouch, I didn't intended to push it for a review, it was an intermediate result. Anyway, thanks for the feedback!

arteam:
Okay, please take a look at this version. I believe that I covered most of your concerns.

I didn't created a helper class, because the parsing logic in `Script` and `ClasspathStatementLocator` is actually differ, despite that this classes use the same lexer.

src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -97,9 +94,9 @@ public String locate(String name, StatementContext ctx)
             return name;
         }
         final ClassLoader loader = selectClassLoader();
-        BufferedReader reader = null;
+        InputStream in_stream = null;
         try {
-            InputStream in_stream = loader.getResourceAsStream(name);
```
stevenschlansker: `inStream` would be canonical Java style.

src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -118,31 +115,42 @@ public String locate(String name, StatementContext ctx)
...
+                        case SqlScriptLexer.COMMENT:
+                        case SqlScriptLexer.MULTI_LINE_COMMENT:
+                            break;
+                        case SqlScriptLexer.NEWLINES:
+                            if (sb.length() > 0) {
+                                sb.append(" ");
+                            }
+                            break;
+                        default:
+                            sb.append(t.getText());
```
stevenschlansker: Is this code reachable today?  It might be preferable to cover all possibilities as `case` statements and then have the `default` throw an IllegalStateException or something.

src/main/java/org/skife/jdbi/v2/Script.java
```diff
@@ -65,24 +62,55 @@
...
+                    endOfFile = true;
+                    break;
+                case SqlScriptLexer.SEMICOLON:
+                    addStatement(sb.toString(), statements);
+                    sb = new StringBuilder();
+                    break;
+                default:
+                    sb.append(t.getText());
+                    break;
+                // We don't handle comments, because they are already removed in StatementLocator.
```
stevenschlansker: I am worried that if this comment does not hold (say, someone changes `StatementLocator` and does not check this code) that non-SQL tokens may be appended to the buffer since they will hit the `default` case.  I would cover all cases explicitly here too and consider making the `default` case throw again.

stevenschlansker:
In general the patch looks good.  My remaining worry is that some existing obscure SQL script out there may break, if it does not match what the lexer expects, and throw where it previously succeeded.  But really people should test their SQL and notice it when they update the version, so I'm tempted to not worry too much.

Final change, how about an entry in the `CHANGELOG` that notes the change as well as a small warning along the lines of "existing SQL statements may break due to lexer changes, ensure you have test coverage", but written better :)

arteam:
The next round! Please, take a look at this version. 

I eventually extracted the common code to `SqlScriptParser`, which seems convienent. In the final
version the parsing logic is very similar between `Script` and `ClasspathStatementLocator`.

stevenschlansker:
Great, thank you much for this contribution!  I am fairly happy with it now.  I will give @brianm or @hgschmie a few days to review this, and if they do not object or cannot get to it by say end of next week, I will merge it.

src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -203,4 +188,11 @@ private static String left(String str, int len)
         }
         return str.substring(0, len);
     }
+
+    private static final SqlScriptParser SQL_SCRIPT_PARSER = new SqlScriptParser(new SqlScriptParser.TokenHandler() {
```
stevenschlansker: Ah, so, should this be a `ThreadLocal` maybe?  [Are Parser instances reentrant?](https://stackoverflow.com/questions/19325803/are-antlr-parsers-for-java-thread-safe)

arteam: We don't use generated `org.antlr.runtime.Parser` class. It's just a simple immutable class, which delegates events from a lexer to a token handler. A new lexer instance is created every time for a new character stream.

PR #169, opened 2015-09-15
Switch to standard maven toolchains for building

hgschmie:
Use toolchains.xml abstraction to select build JDK. This is supported
(but not well documented) in maven for a long time. This alleviates the
need for the various build profiles.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: java
 jdk:
-  - openjdk6
   - oraclejdk7
   - oraclejdk8
+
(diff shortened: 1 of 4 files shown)


hgschmie:
This is *NOT* ready for merging.

stevenschlansker:
That's nifty!

hgschmie:
It needs to work first. :-)

hgschmie:
ok, now it is good to go. @stevenschlansker take a look?

PR #181, opened 2015-09-29
BasicHandle.close() can fail to close connection if StatementBuilder.close() fails

### LINKED ISSUE #180: BasicHandle.close can fail to close
If the `statementBuilder.close(...)` call throws, the rest of the close logic is skipped.

stevenschlansker:
Fixes #180


stevenschlansker:
@electrum Is this a good fix for you?

PR #182, opened 2015-09-29
Deprecate primitives mapper factory

qualidafial:
With the introduction of the column mappers, the `PrimitivesMapperFactory` is obsolete and can be deprecated.

We can also address #154 by relying on the column mapper registry, which respects nullability of primitive wrapper types.

Lastly, the internal FigureItOutResultSetMapper can rely on the column mapper registry instead of embedding knowledge of type mappers.


qualidafial:
@stevenschlansker ping

PR #183, opened 2015-09-30
statement rewriters use a cache

rudydai:
avoids doing redundant work in the lexer, since parsing a sql statement should be deterministic

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/ColonPrefixNamedParamStatementRewriter.java
+++ b/src/main/java/org/skife/jdbi/v2/ColonPrefixNamedParamStatementRewriter.java
@@ -26,6 +26,8 @@ import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 import static org.skife.jdbi.rewriter.colon.ColonStatementLexer.DOUBLE_QUOTED_TEXT;
 import static org.skife.jdbi.rewriter.colon.ColonStatementLexer.ESCAPED_TEXT;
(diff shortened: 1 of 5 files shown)


src/main/java/org/skife/jdbi/v2/ColonPrefixNamedParamStatementRewriter.java
```diff
@@ -41,6 +43,8 @@
  */
 public class ColonPrefixNamedParamStatementRewriter implements StatementRewriter
 {
+    private final Map<String, ParsedStatement> cache = new HashMap();
```
stevenschlansker: This needs to be a `ConcurrentMap` I think?

src/main/java/org/skife/jdbi/v2/HashPrefixStatementRewriter.java
```diff
@@ -39,6 +41,8 @@
  */
 public class HashPrefixStatementRewriter implements StatementRewriter
 {
+    private final Map<String, ParsedStatement> cache = new HashMap();
```
stevenschlansker: `ConcurrentMap`?

stevenschlansker:
Looks like there's a compilation error, and I think the current `Map` usage is not threadsafe.  Other than that, looks OK.

rudydai:
changed the commit to use a concurrenthashmap instead, can't reproduce the compilation error locally though - and the method signature seems to match other invocations in the same test class. is there something I'm missing here?

stevenschlansker:
The compile error is a merge conflict with #164 that Git didn't catch.

rudydai:
thanks for the fixups :)

stevenschlansker:
BTW, I switched it to a `Collections.synchronizedMap(new WeakHashMap<String, ParsedStatement>());` for the cache -- otherwise we may cause previously working applications that generate lots of SQL statements to crash.

PR #186, opened 2015-10-10
Fix returning generated keys

arteam:
It looks like that returning generated keys by the  `@GetGeneratedKeys` annotation was broken by #117.

The problem is that the method `columnName` in this annotation by default returns an empty string. But the new code checks only for a null reference. As the result, we pass to a JDBC driver an array
with one empty string as column names in the corresponding method  `prepareStatement`.

Database vendors handle this behavior differently. MySQL and H2 permit it,  but PostgreSQL, Oracle, and HSQLDB don't. Therefore we should not pass an empty string as a column name, because it doesn't work in major databases.

We have a test on generating keys for PostgreSQL, but it's ignored. This change adds a test for HSQDB, which shows broken behaviour.


stevenschlansker:
Sorry I missed this during code review, I thought we had tests covering it :/
Thanks for the fix.

PR #187, opened 2015-10-11
Add support for running tests against PostgreSQL

arteam:
Travis CI kindly provides a PostgreSQL instance on their virtual machines, so we can actually add tests against PostgreSQL to the test suite.

### DIFF
--- a/src/test/java/org/skife/jdbi/v2/sqlobject/TestGetGeneratedKeysPostgres.java
+++ b/src/test/java/org/skife/jdbi/v2/sqlobject/TestGetGeneratedKeysPostgres.java
@@ -13,10 +13,9 @@
  */
 package org.skife.jdbi.v2.sqlobject;
 
-import org.h2.jdbcx.JdbcConnectionPool;
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.skife.jdbi.v2.DBI;
 import org.skife.jdbi.v2.Handle;
@@ -25,15 +24,20 @@ import org.skife.jdbi.v2.tweak.HandleCallback;
 
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assume.assumeTrue;
 
 public class TestGetGeneratedKeysPostgres
 {
-    private JdbcConnectionPool ds;
     private DBI                dbi;
+
+    @BeforeClass
+    public static void isPostgresInstalled() {
+        assumeTrue(Boolean.parseBoolean(System.getenv("TRAVIS")));
+    }
 
     @Before
     public void setUp() throws Exception {
-        dbi = new DBI("jdbc:postgresql:test", "postgres", "postgres");
+        dbi = new DBI("jdbc:postgresql:jdbi_test", "postgres", "");
         dbi.withHandle(new HandleCallback<Object>() {
             @Override
             public Object withHandle(Handle handle) throws Exception
@@ -67,7 +71,6 @@ public class TestGetGeneratedKeysPostgres
         public String findNameById(@Bind long id);
     }
 
-    @Ignore
     @Test
     public void testFoo() throws Exception {
         DAO dao = dbi.open(DAO.class);
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,3 +11,6 @@ before_install:
 
 install: mvn -DskipTests=true -Dbasepom.check.skip-all=true -B install
 script: mvn -B verify
+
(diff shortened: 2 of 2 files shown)


src/test/java/org/skife/jdbi/v2/sqlobject/TestGetGeneratedKeysPostgres.java
```diff
@@ -25,19 +23,23 @@
 
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assume.assumeTrue;
 
-public class TestGetGeneratedKeysPostgres
-{
-    private JdbcConnectionPool ds;
-    private DBI                dbi;
+public class TestGetGeneratedKeysPostgres {
```
qualidafial: It might seem petty, but can we revert back to the original brace style? This line change doesn't contribute to the fix, and can cause merge conflicts.

stevenschlansker:
Nice!

src/test/java/org/skife/jdbi/v2/sqlobject/TestGetGeneratedKeysPostgres.java
```diff
@@ -67,9 +72,10 @@ public Object withHandle(Handle handle) throws Exception
         public String findNameById(@Bind long id);
     }
 
-    @Ignore
     @Test
     public void testFoo() throws Exception {
+        assumeTrue(postgresIsInstalled);
```
stevenschlansker: Would this be better in the `@Before` ?  Gets rid of the if blocks entirely, and should work just the same: https://github.com/junit-team/junit/wiki/Assumptions-with-assume

qualidafial:
:+1:

src/test/java/org/skife/jdbi/v2/sqlobject/TestGetGeneratedKeysPostgres.java
arteam: Makes sense, I also proponent of more cleaner diffs.

src/test/java/org/skife/jdbi/v2/sqlobject/TestGetGeneratedKeysPostgres.java
arteam: I believe, unfortunately, JUnit doesn't handle assumptions in the `@Before` or `@After` auxiliary methods. 

See [this](http://stackoverflow.com/questions/30112317/how-does-assumetrue-work-in-junit) (unfairly downvoted) SO question for the clarification.

stevenschlansker: Hm, the JUnit Wiki that I linked above says
> A failing assumption in a @Before or @BeforeClass method will have the same effect as a failing assumption in each @Test method of the class.
which I think means it should work?

(I voted the question up, wow that does not deserve a -3!)

arteam: I tested again and it looks like assumptions actually work in `@Before`. The problem is that the test is failed if I place an assumption is the `@After` method, because JUnit converts `AssumptionViolatedException` to `MultipleFailureException`. So we can't use `Assume.assumeTrue` twice in the same test.

qualidafial: Maybe put the assumption in a `@BeforeClass`? Then neither your `@Before` or `@After` methods will ever be called.

arteam: It looks like a great idea and I've pushed a new version. Thanks for your suggestion!

qualidafial:
@stevenschlansker This looks ready. Any remaining concerns before I merge?

stevenschlansker:
Go for it!

src/test/java/org/skife/jdbi/v2/sqlobject/TestGetGeneratedKeysPostgres.java
```diff
@@ -25,15 +24,20 @@
...
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assume.assumeTrue;
 
 public class TestGetGeneratedKeysPostgres
 {
-    private JdbcConnectionPool ds;
     private DBI                dbi;
 
+    @BeforeClass
```
stevenschlansker: :+1:

qualidafial:
@arteam Thanks for the contribution!

PR #189, opened 2015-10-15
WIP: jdbi 3

stevenschlansker:
tag @qualidafial now the branch is on the jdbi org, not my personal repo.


qualidafial:
I suspect merge the latest from master into this branch would be simpler if we temporarily renamed the packages back to `v2`. Will try it locally when I get a minute.

stevenschlansker:
I almost have it up to date at this point!  Hopefully by end of day today.

qualidafial:
Should get rid of TypedMapper and its subclasses entirely.

stevenschlansker:
so @qualidafial I'm not sure how this isn't broken in master -- but the way that the MappingRegistry is copied from the Statement into the StatementContext breaks registering of additional mappers through the Query.  Basically it gets registered with the Query's registry *after* the statement context copies it out, and then it isn't searched during execution.

qualidafial:
I ran into that on my branch. Basically the solution I found was to only clone on the calling side. i.e. Handle.createQuery should copy the registry. It should then pass the same registry to the statement context and the query.

qualidafial:
Shouldn't the `CloseMe` mixin extend `java.lang.AutoCloseable`?

stevenschlansker:
What do you think of https://github.com/jdbi/jdbi/commit/79631037c1a83038e704fdb027e0cfea3d4d4a62 ?

stevenschlansker:
I think `CloseMe` should just be gone in favor of Closeable.

qualidafial:
Only thing I dislike about AutoCloseable is it throws Exception, so you can't just extend it. You have to redeclare the close method without the exception--annoying.

stevenschlansker:
Yeah, I'm all for avoiding the mysterious changes at a distance -- all of the Handle calls already copy the MappingRegistry, so hopefully this only is shared with the Context and not the Handle.  But that's a legit TODO so I'll add it.

stevenschlansker:
I'm going to reinstate CloseMe for now.

qualidafial:
A few thoughts while exploring the code in its current state:

* [ ] Does `ResultSetMapper.map` need the row index parameter? Mappers can already query the current row through `ResultSet.getRow()`.
* [ ] Commit c4f32369 seems to have inadvertently replaced the word "get" with "getAttribute" in several places where "get" was used as a word and not as a method name. Sometimes IDE refactoring support is just a little *too* helpful..
* [ ] Is BaseResultSetMapper providing any value, to make it worth keeping?
* [ ] We should support Optional<T> as a return type from SQL object methods.
* [ ] Need to add core/src/doc/v3, and update core/src/doc/index.html current version.
* [ ] Codehaus links are dead--these should be redirected to live sites.
* [ ] We discussed switching from `Class<?>` to `Type` in mapper and argument factories.
* [ ] BeanMapperFactory.accepts could/should verify that the type is actually a bean.
* [ ] VoidHandleCallback is redundant now that we have HandleConsumer
* [ ] Now that the typed mappers are gone, I've been considering renaming all the column mappers without the word "column" for brevity at the usage site.
* [ ] Do we want built-in argument types and factories to live in their own package? Maybe in `util` with the built-in mappers?

Bookmark: I'm down to PreparedBatch

qualidafial:
Idea: Since `SqlObjectBuilder.open` opens a Handle, we should change the method to accept only a `Class<T extends AutoCloseable>`, which would make the result usable in try-with-resource blocks.

miere:
Hi guys,

I quite excited about the JDBI 3 and I'm wondering when it is planned to be released...
Do you guys have any idea when this new version will be released?

Regards

qualidafial:
Hi Miere,

I'm also excited about where JDBI 3 is heading.

We're not targeting a specific date for release at the moment. Personally I'm working toward early January, but it all depends on what happens between then and now. There's still a healthy list of todos left to deal with. If we discover any major flaws before release, we're going to hold off until we feel we've got the design right.

With that said, I wouldn't be opposed to releasing a few alphas along the way, to let folks kick the tires and provide feedback from the field.

qualidafial:
@stevenschlansker Need to use `-P oracle` on release builds. I will fix

qualidafial:
OMG look at the line count on this PR

qualidafial:
:tada:

qualidafial:
@jdbi/contributors Since v3 has been our main development focus for quite a while now, what would you think about spinning off a v2 maintenance branch, merging this PR to master, and developing off of master going forward?

stevenschlansker:
I was considering the same thing myself last night.  I think I'm in favor of it.

arteam:
I'm also for that.

qualidafial:
What about merging changes from v2 going forward? My concern is that the code bases have diverged so dramatically, that we're basically doing merges just to keep Git happy--not for any actual value to the project.

This has me wondering: would it be valuable to fork a repository for v2 so we can maintain it independently?

stevenschlansker:
I do not think we port changes from v2 by default anymore.  It is now in maintenance mode.

PR #190, opened 2015-10-19
SqlObject now supports default methods

stevenschlansker:
tag @qualidafial so that approach you linked doesn't seem to work:
```
java.lang.RuntimeException: java.lang.IllegalAccessException: no private access for invokespecial: interface org.jdbi.v3.sqlobject.TestJava8Features$Spiffy, from org.jdbi.v3.sqlobject.TestJava8Features$Spiffy/package
	at com.google.common.base.Throwables.propagate(Throwables.java:160)
	at org.jdbi.v3.sqlobject.PassThroughHandler.invokeDefault(PassThroughHandler.java:71)
	at org.jdbi.v3.sqlobject.PassThroughHandler.invoke(PassThroughHandler.java:41)
	at org.jdbi.v3.sqlobject.SqlObject.invoke(SqlObject.java:173)
	at org.jdbi.v3.sqlobject.SqlObject$1.intercept(SqlObject.java:73)
	at org.jdbi.v3.sqlobject.CloseInternalDoNotUseThisClass$$EnhancerByCGLIB$$bc09080d.insertAndReturn(<generated>)
	at org.jdbi.v3.sqlobject.TestJava8Features.testDefaultMethod(TestJava8Features.java:34)
Caused by: java.lang.IllegalAccessException: no private access for invokespecial: interface org.jdbi.v3.sqlobject.TestJava8Features$Spiffy, from org.jdbi.v3.sqlobject.TestJava8Features$Spiffy/package
	at java.lang.invoke.MemberName.makeAccessException(MemberName.java:855)
	at java.lang.invoke.MethodHandles$Lookup.checkSpecialCaller(MethodHandles.java:1565)
	at java.lang.invoke.MethodHandles$Lookup.unreflectSpecial(MethodHandles.java:1224)
	at org.jdbi.v3.sqlobject.PassThroughHandler.invokeDefault(PassThroughHandler.java:67)
	... 30 more
```

but I think I got it working!

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObject.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObject.java
@@ -27,10 +27,12 @@ import com.fasterxml.classmate.ResolvedTypeWithMembers;
 import com.fasterxml.classmate.TypeResolver;
 import com.fasterxml.classmate.members.ResolvedMethod;
 
+import net.sf.cglib.proxy.Callback;
 import net.sf.cglib.proxy.Enhancer;
 import net.sf.cglib.proxy.Factory;
 import net.sf.cglib.proxy.MethodInterceptor;
 import net.sf.cglib.proxy.MethodProxy;
+import net.sf.cglib.proxy.NoOp;
 
 class SqlObject
 {
@@ -65,14 +67,12 @@ class SqlObject
             }
             e.setInterfaces(interfaces.toArray(new Class[interfaces.size()]));
             final SqlObject so = new SqlObject(buildHandlersFor(sqlObjectType), handle);
-            e.setCallback(new MethodInterceptor()
-            {
-                @Override
-                public Object intercept(Object o, Method method, Object[] objects, MethodProxy methodProxy) throws Throwable
-                {
-                    return so.invoke(o, method, objects, methodProxy);
-                }
+            e.setCallbackFilter(m -> m.isDefault() ? 1 : 0);
+            e.setCallbacks(new Callback[] {
+                (MethodInterceptor) so::invoke,
+                NoOp.INSTANCE
             });
+
             T t = (T) e.create();
             T actual = (T) factories.putIfAbsent(sqlObjectType, (Factory) t);
             if (actual == null) {
@@ -81,14 +81,11 @@ class SqlObject
             f = (Factory) actual;
         }
 
+        // TODO 3: this is duplicated from the above setCallbacks, can we clean that up?
         final SqlObject so = new SqlObject(buildHandlersFor(sqlObjectType), handle);
-        return (T) f.newInstance(new MethodInterceptor()
-        {
-            @Override
-            public Object intercept(Object o, Method method, Object[] objects, MethodProxy methodProxy) throws Throwable
-            {
-                return so.invoke(o, method, objects, methodProxy);
-            }
+        return (T) f.newInstance(new Callback[] {
+            (MethodInterceptor) so::invoke,
+            NoOp.INSTANCE
         });
     }
 
--- a/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestDefaultMethods.java
+++ b/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestDefaultMethods.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.sqlobject;
+
+import static org.junit.Assert.assertEquals;
+
+import org.jdbi.v3.H2DatabaseRule;
+import org.jdbi.v3.Something;
+import org.jdbi.v3.sqlobject.customizers.Mapper;
+import org.jdbi.v3.sqlobject.mixins.CloseMe;
+import org.junit.Rule;
+import org.junit.Test;
+
+@SuppressWarnings("resource")
+public class TestDefaultMethods
+{
+    @Rule
+    public H2DatabaseRule db = new H2DatabaseRule();
+
+    @Test
+    public void testDefaultMethod() throws Exception {
+        Spiffy dao = SqlObjectBuilder.onDemand(db.getDbi(), Spiffy.class);
+        Something test = dao.insertAndReturn(3, "test");
+        assertEquals(3, test.getId());
+        assertEquals("test", test.getName());
+    }
+
+    @Test
+    public void testOverride() throws Exception {
+        SpiffyOverride dao = SqlObjectBuilder.onDemand(db.getDbi(), SpiffyOverride.class);
+        assertEquals(null, dao.insertAndReturn(123, "fake"));
+    }
+
+    @Test
+    public void testOverrideWithDefault() throws Exception {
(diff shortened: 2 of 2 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObject.java
```diff
@@ -82,13 +94,15 @@ public Object intercept(Object o, Method method, Object[] objects, MethodProxy m
...
-        return (T) f.newInstance(new MethodInterceptor()
-        {
-            @Override
-            public Object intercept(Object o, Method method, Object[] objects, MethodProxy methodProxy) throws Throwable
+        return (T) f.newInstance(new Callback[] {new MethodInterceptor()
             {
-                return so.invoke(o, method, objects, methodProxy);
-            }
+                @Override
+                public Object intercept(Object o, Method method, Object[] objects, MethodProxy methodProxy) throws Throwable
```
stevenschlansker: Yeah, duplicating this sucks.  I chose to keep the cleanup of this to a separate refactoring PR rather than mixing it in with the feature request.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObject.java
```diff
@@ -65,14 +68,23 @@
...
                 }
             });
+            e.setCallbacks(new Callback[] {new MethodInterceptor()
+                {
+                    @Override
+                    public Object intercept(Object o, Method method, Object[] objects, MethodProxy methodProxy) throws Throwable
+                    {
+                        return so.invoke(o, method, objects, methodProxy);
+                    }
+                },
```
qualidafial: This anonymous inner class could be replaced with `so::invoke`.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObject.java
qualidafial: Ditto `so::invoke`

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestDefaultMethods.java
```diff
@@ -0,0 +1,66 @@
...
+        Something test = dao.insertAndReturn(3, "test");
+        assertEquals(3, test.getId());
+        assertEquals("test", test.getName());
+    }
+
+    @Test
+    public void testOverride() throws Exception {
+        SpiffyOverride dao = SqlObjectBuilder.onDemand(db.getDbi(), SpiffyOverride.class);
+        assertEquals(null, dao.insertAndReturn(123, "fake"));
+    }
```
qualidafial: Another scenario: what if the default method has e.g. `@SqlQuery` annotation on it? Does the annotation or the default method win?

My feeling is the annotation should win. e.g. somebody declares a base SqlObject interface, with a default method. The default method would have a generic but possibly suboptimal implementation. The subclass might have a specialized implementation that can be achieved with annotations.

qualidafial:
Looks awesome, can't wait. A few comments, then :shipit:

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestDefaultMethods.java
qualidafial: Let me revise my position: whatever the subclass does should take precedence. So if superclass has `@SqlQuery` but subclass wants to do it with a default method impl, use the default method. Conversely, if superclass has default method but subclass is abstract with `@SqlQuery` annotation, follow the annotation.

stevenschlansker: Isn't that exactly the case of the `SpiffyOverride` below?  Or am I misunderstanding?
The default method should always lose in case there is any other reasonable action to take.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObject.java
```diff
@@ -81,14 +87,11 @@ public Object intercept(Object o, Method method, Object[] objects, MethodProxy m
             f = (Factory) actual;
         }
 
+        // TODO 3: this is duplicated from the above setCallbacks, can we clean that up?
```
qualidafial: Fixed now? Or did you want to extract the common array?

stevenschlansker: It's not actually common, the value of `so` changes :/

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestDefaultMethods.java
stevenschlansker: Added a test as such.

qualidafial: :+1:

qualidafial:
I believe this fails to handle `@Transactional` on default methods.

qualidafial:
Fixed (mostly?) in jdbi3-extensions branch in revision b697c0cd34f92092f9c037ec9facf0c2463cd6f8

qualidafial:
Essence of the fix, for future reference:

    if (method.isDefault()) {
        // TERRIBLE, HORRIBLE, NO GOOD, VERY BAD HACK
        // Courtesy of:
        // https://rmannibucau.wordpress.com/2014/03/27/java-8-default-interface-methods-and-jdk-dynamic-proxies/

        // CGLIB does not yet support calling interface default methods through mp.invokeSuper().

        // We can use MethodHandles to look up and invoke the super method, but since this class is not an
        // implementation of method.getDeclaringClass(), MethodHandles.Lookup will throw an exception since
        // this class doesn't have access to the super method, according to Java's access rules. This horrible,
        // awful workaround allows us to directly invoke MethodHandles.Lookup's private constructor, bypassing
        // the usual access checks.

        // We should get rid of this workaround as soon as a viable alternative exists.

        final Constructor<MethodHandles.Lookup> constructor = MethodHandles.Lookup.class.getDeclaredConstructor(Class.class, int.class);
        if (!constructor.isAccessible()) {
            constructor.setAccessible(true);
        }

        Class<?> declaringClass = method.getDeclaringClass();
        return constructor.newInstance(declaringClass, MethodHandles.Lookup.PRIVATE) // *shudder*
                .unreflectSpecial(method, declaringClass)
                .bindTo(target)
                .invokeWithArguments(args);
    }

    return mp.invokeSuper(target, args);

Nectarius:
Hello,
We're using IBM sdk Java. And we can't upgrade to JDBI 3.
We can't invoke method after newInstance(declaringClass, MethodHandles.Lookup.PRIVATE), it's like it is considered private
java.lang.RuntimeException: java.lang.IllegalAccessException: $Dao' no access to: $Dao.method:(Dao,Levels,List,List,List,String,Instant)void/invokeSpecial'

It should be MethodHandles.Lookup.PRIVATE | MethodHandles.Lookup.PUBLIC since interface default method is public.
constructor.newInstance(type, MethodHandles.Lookup.PRIVATE | MethodHandles.Lookup.PUBLIC);

arteam:
Does your JVM have a security manager which forbids reflection? Maybe we can wrap the invocation to `PrivilegedAction`, so the code can invoke the private constructor on `MethodHandles.Lookup`.

Nectarius:
No, IBM Jvm allows us to call private construcor, 
but lookup (inside org.jdbi.v3.sqlobject.DefaultMethodHandler) 

MethodHandles.Lookup lookup = lookupFor(declaringClass)
has access mode PRIVATE

and it not enough for checkSpecialAccess in this JVM

it works if we pass MethodHandles.Lookup.PRIVATE | MethodHandles.Lookup.PUBLIC to the private constructor
return constructor.newInstance(type, MethodHandles.Lookup.PRIVATE | MethodHandles.Lookup.PUBLIC );  

or method.setAccessible(true); in DefaultMethodHandler constructor to workaround checkSpecialAccess internal method

https://developer.ibm.com/javasdk/downloads/sdk8/
 it requires registration to upload unfortunately

qualidafial:
@Nectarius Would you open a new issue for this, so it doesn't get lost?

Nectarius:
@qualidafial sure

PR #191, opened 2015-10-19
SqlObject implicit @Bind by parameter name

stevenschlansker:
We added support for `@Bind` recognizing parameter names, now let's go even further and ditch the annotation if you don't actually need it.

Open question: should we just get rid of the default "name by parameter index" entirely?  This wouldn't prevent people from using positional parameters in fluent queries, but it would mean
```
@SqlUpdate("insert into table values(:0)")
void doSomething(int param);
```
would need to be rewritten to
```
@SqlUpdate("insert into table values(:param)")
void doSomething(int param);
```

The only reason I can see this being used would be if your code does not have parameter names enabled (which is a javac flag IIRC) but then you might as well just `@Bind("param")` instead... Thoughts?


sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultObjectBinder.java
```diff
@@ -19,20 +19,36 @@
...
+                value = Integer.toString(paramIndex);
+            } else {
                 throw new UnsupportedOperationException("A parameter was annotated with @Bind "
                         + "but no name was specified, and parameter name data is not present "
                         + "in the class file.  " + param.getDeclaringExecutable() + " :: " + param);
             }
-            value = param.getName();
         }
 
+        q.bind(paramIndex, arg);
```
qualidafial: Should this be called if this `DefaultObjectBinder` is created with the no-arg constructor?

qualidafial:
Do we really want to drop positional parameter support? We're sure there aren't enough users that use it to justify keeping it?

I'm also not sure I like falling back to positional parameters behavior depending on the user's compiler flags. IMO it should only work (or not) depending on compiler flags, not change behavior.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultObjectBinder.java
qualidafial: The more I look at this, the more I think we don't even want this class doing anything positional.

stevenschlansker:
To clarify, I wasn't suggesting removing positional parameter support.  There was a bizarre mode where `@Bind` would create a *named* parameter with the stringified position index as a name.  So it's a named positional parameter.  Confusing!

Now that's gone.  I simplified the rules a lot:

It's always bound as a positional parameter, which will only be used if the name doesn't match.
If you omit the name, you *must* have the parameter name available, otherwise it's an error.

Think that's reasonable?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/Bind.java
```diff
@@ -19,7 +19,11 @@
 import java.lang.annotation.Target;
 
 /**
- * A binding annotation
+ * A binding annotation.  Will bind both as the {@code value}
+ * of the annotation as well as the parameter index.  If the
+ * {@code value} is not specified, it will default to using the parameter
+ * name.  It is an error to not specify the name when there is no
+ * parameter naming information in your class files.
```
qualidafial: :+1:

sqlobject/src/main/java/org/jdbi/v3/sqlobject/BinderFactory.java
```diff
@@ -19,12 +19,12 @@
  * Factor for building {@link Binder} instances. This interface is used by
  * the {@link BindingAnnotation}
  */
-public interface BinderFactory
+public interface BinderFactory<T extends Annotation, O>
```
qualidafial: Should add documentation for `T` and `O` generic parameters.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultObjectBinder.java
```diff
@@ -19,20 +19,35 @@
 
 class DefaultObjectBinder implements Binder<Bind, Object>
 {
+    private final int paramIndex;
+
+    DefaultObjectBinder()
+    {
+        this(-1);
+    }
```
qualidafial: Is this no-arg constructor necessary?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/CustomizingStatementHandler.java
```diff
@@ -85,8 +85,8 @@
                     // we have a binder
                     BindingAnnotation ba = annotation.annotationType().getAnnotation(BindingAnnotation.class);
                     try {
-                        BinderFactory fact = ba.value().newInstance();
-                        binders.add(new Bindifier(this.method, annotation, param_idx, (Binder<Annotation, Object>)fact.build(annotation)));
+                        BinderFactory<Annotation, Object> fact = (BinderFactory<Annotation, Object>) ba.value().newInstance();
+                        binders.add(new Bindifier<Annotation>(this.method, annotation, param_idx, fact.build(annotation)));
```
qualidafial: The presence of a generic cast here suggests that `Bindifier` should expect a `BinderFactory<AnnoType extends Annotation, ?>`. Bindifier isn't using the second parameter.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/BinderFactory.java
qualidafial: I looks like the `O` parameter is unused.

qualidafial:
Consider comments, then :shipit: when ready.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultObjectBinder.java
stevenschlansker: yeah, if you have `@Bind(binder=DefaultObjectBinder.class)`, it calls `DefaultObjectBinder.class.newInstance()`

sqlobject/src/main/java/org/jdbi/v3/sqlobject/CustomizingStatementHandler.java
stevenschlansker: added a TODO to clean up the generics

stevenschlansker:
I think we should rethink how binding happens entirely.  I've dropped a note to that effect on the CHANGES_FOR_3.0 document and we'll revisit it later.  I'm going to merge this partially as a WIP.

PR #192, opened 2015-10-20
Arbitrarily Paramaterized Sql Locators

brianm:
Change sql "naming" to be (Object,Object...) based instead of a String.
This allows for much more sophisticated naming strategies, such as using
the SQL Object class and the method name, or doing classpath lookups
relative to the passed in class, etc.

In order to keep flexibility, it is done as just objects. Ideally it
would be type parameterized on the sql locator, but that would require
the sql locator setter become a builder, plus it would make for *really*
verbose type names, which would suck. This seems like a pragmatic
compromise, and fits with other dynamic aspects of jdbi.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/BasicHandle.java
+++ b/core/src/main/java/org/jdbi/v3/BasicHandle.java
@@ -13,13 +13,6 @@
  */
 package org.jdbi.v3;
 
-import java.sql.Connection;
-import java.sql.SQLException;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
 import org.jdbi.v3.exceptions.UnableToCloseResourceException;
 import org.jdbi.v3.exceptions.UnableToManipulateTransactionIsolationLevelException;
 import org.jdbi.v3.tweak.ArgumentFactory;
@@ -32,6 +25,13 @@ import org.jdbi.v3.tweak.StatementRewriter;
 import org.jdbi.v3.tweak.TransactionHandler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 class BasicHandle implements Handle
 {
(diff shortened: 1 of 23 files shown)


brianm:
The travis build failure is from preserving a weird string check which I don't grok so didn't change.

qualidafial:
I'm not sure I understand the use case. What problem is this trying to solve?

brianm:
This experimental, but is designed to address a common problem I have seen where folks want to use externalized SQL but have to cram information into stringly typed things, like:

```
handle.createQuery("com.example.breakfast:pancake-insert")
```

Or similar messes. The new signature is designed to allow pretty arbitrary statement locators, with possible examples being things like:

```
handle.createQuery(CurrentClass.class, "insertPancake")
```

which might do a classpath relative lookup against CurrentClass, or some such clever thing.

I don't *love* the interface, but I like it more than forcing things into a String. The biggest drawback is that it stands to create confusion about the `extras` (ie, "insertPancake") and positionally bound params. I am confident new users will stumble on this.

I'd love better ideas on how to solve this!

stevenschlansker:
What if we start to think of SQL location more as part of the DBI object and an inherently separate concept to query execution?

```java
class SqlFragment {
  String locatedSql;
  private ParsedStatement cacheSomeData;
}

class DBI {
  static? SqlFragment sqlFromClasspath(String name); // implicit classloader context?
  static? SqlFragment sqlFromClasspath(Class<?> context, String name);
  static? SqlFragment sqlLiteral(String sql);
}

class Handle {
  createQuery(SqlFragment sql).bind(...).execute();
}
```

Lets you naturally help us reuse some data with idiomatic code:

```java
class Dao {
  static final SqlFragment FIND_BY_ID = DBI.sqlFromClasspath(Dao.class, "findById");
}
```

It's slightly more verbose, but having the builder-style even for inline code makes it more readable:

```java
handle.createQuery(sqlFromClasspath("findById")).bind(...).execute()
```

qualidafial:
@stevenschlansker I think you're onto something, but I want to take it a step further.

What if we removed StatementLocator from core, and made it an optional library?

This would mean all the core (fluent) APIs would only ever be passed SQL. Users of the fluent API would have to locate the SQL imperatively, e.g.:

    String sql = SqlLocators.findOnClasspath(Dao.class, "findById");
    return handle.createQuery(sql).bind("id", id).findOnly();

If SQLObject users wanted to use a SQL locator, they could configure the extension, or use annotations on a per-type or per-method basis to tell SQLObject where to find SQL for a given String:

    @UseClasspathStatementLocator
    class Dao {
      @SqlQuery Foo findById(@Bind long id);
    }

This would also be a good time to get rid of the "either-or" style of SQL resolution: if there's a statement locator, then the value in the annotation is some kind of statement ID (e.g. a filename on the classpath). If there's not a statement locator, then the annotation value is SQL--period. No more testing the string for "SQL-ness."

stevenschlansker:
I very much like this idea.

qualidafial:
One obstacle to this is that `StringTemplate3StatementLocator` is doing the interpolation of defines into the SQL e.g. `select id, name from <table>`. IMO this rightly belongs in the statement rewriter.

Opened #344 to handle that aspect.

qualidafial:
With #391 merged, can we close this ticket? I believe the original intent has been achieved, namely that different SQL locators can have different semantics for how SQL snippets are identified.

stevenschlansker:
Yes, @brianm can reopen it if he wants to look at it again :)

PR #198, opened 2015-10-23
Add mixin method GetHandle.useHandle(HandleConsumer)

qualidafial:


sqlobject/src/main/java/org/jdbi/v3/sqlobject/UseHandleHandler.java
```diff
@@ -0,0 +1,36 @@
...
+    public Object invoke(HandleDing h, Object target, Object[] args, MethodProxy mp)
+    {
+        final Handle handle = h.getHandle();
+        final HandleConsumer consumer = (HandleConsumer) args[0];
+        try {
+            consumer.useHandle(handle);
+            return null;
+        }
+        catch (Exception e) {
+            throw new CallbackFailedException(e);
```
stevenschlansker: Hm... is this the exception policy we want?  Means we could end up wrapping even our own exceptions, or things like `IllegalStateException`, etc.

qualidafial: I was just following the pattern in `WithHandleHandler`. I'm open to any alternative approaches.

It might also make sense to implement `withHandle` and `useHandle` with interface default methods instead.

PR #199, opened 2015-10-25
Resurrect container builders in JDBI 3

arteam:
Container builders are actually quite useful for collecting query results into types that are not shipped with the JDK.  For example, Google Guava's immutable collections and optional values.

Also, they are indispensable for SQLObject API, where we don't have Stream API for providing implementations of custom types.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/BasicHandle.java
+++ b/core/src/main/java/org/jdbi/v3/BasicHandle.java
@@ -30,6 +30,7 @@ import org.jdbi.v3.tweak.StatementCustomizer;
 import org.jdbi.v3.tweak.StatementLocator;
 import org.jdbi.v3.tweak.StatementRewriter;
 import org.jdbi.v3.tweak.TransactionHandler;
+import org.jdbi.v3.tweak.CollectorFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -46,6 +47,7 @@ class BasicHandle implements Handle
 
     private final Map<String, Object>      globalStatementAttributes;
     private final MappingRegistry          mappingRegistry;
+    private final CollectorFactoryRegistry collectorFactoryRegistry;
     private final Foreman                  foreman;
     private final TransactionHandler       transactions;
     private final Connection               connection;
@@ -59,7 +61,8 @@ class BasicHandle implements Handle
                 Map<String, Object> globalStatementAttributes,
                 TimingCollector timingCollector,
                 MappingRegistry mappingRegistry,
-                Foreman foreman)
+                Foreman foreman,
+                CollectorFactoryRegistry collectorFactoryRegistry)
     {
         this.statementBuilder = preparedStatementCache;
         this.statementRewriter = statementRewriter;
@@ -71,6 +74,7 @@ class BasicHandle implements Handle
         this.foreman = foreman;
         this.globalStatementAttributes = new HashMap<String, Object>();
         this.globalStatementAttributes.putAll(globalStatementAttributes);
+        this.collectorFactoryRegistry = collectorFactoryRegistry.createChild();
     }
 
     @Override
@@ -88,7 +92,8 @@ class BasicHandle implements Handle
                                               timingCollector,
                                               Collections.<StatementCustomizer>emptyList(),
                                               queryRegistry,
-                                              foreman.createChild());
+                                              foreman.createChild(),
+                                              collectorFactoryRegistry.createChild());
     }
 
     /**
@@ -244,7 +249,8 @@ class BasicHandle implements Handle
                           sql,
                           new ConcreteStatementContext(globalStatementAttributes, MappingRegistry.copyOf(mappingRegistry)),
                           timingCollector,
-                          foreman);
+                          foreman,
+                          collectorFactoryRegistry);
     }
 
     @Override
@@ -258,7 +264,8 @@ class BasicHandle implements Handle
                         new ConcreteStatementContext(globalStatementAttributes, MappingRegistry.copyOf(mappingRegistry)),
                         timingCollector,
                         Collections.<StatementCustomizer>emptyList(),
-                        foreman);
+                        foreman,
+                collectorFactoryRegistry);
     }
 
     @Override
@@ -289,7 +296,8 @@ class BasicHandle implements Handle
                                  new ConcreteStatementContext(globalStatementAttributes, MappingRegistry.copyOf(mappingRegistry)),
                                  timingCollector,
                                  Collections.<StatementCustomizer>emptyList(),
-                                 foreman);
+                                 foreman,
+                collectorFactoryRegistry);
     }
 
     @Override
@@ -447,5 +455,10 @@ class BasicHandle implements Handle
     public void registerArgumentFactory(ArgumentFactory<?> argumentFactory)
     {
         this.foreman.register(argumentFactory);
+    }
+
+    @Override
+    public void registerCollectorFactory(CollectorFactory<?, ?> factory) {
+        this.collectorFactoryRegistry.register(factory);
     }
 }
--- a/core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
+++ b/core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
@@ -1,0 +1,97 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3;
+
+import org.jdbi.v3.tweak.CollectorFactory;
+
+import java.util.List;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+
+/**
+ * Registry of collector factories.
+ * Contains a set of collector factories, registered by the application.
+ */
+class CollectorFactoryRegistry {
+
+    private final Set<CollectorFactory<?, ?>> factories = new CopyOnWriteArraySet<>();
+
+    CollectorFactoryRegistry() {
+        factories.add(new ListCollectorFactory<>());
+        factories.add(new SortedSetCollectorFactory<>());
+        factories.add(new SetCollectorFactory<>());
+    }
+
+    CollectorFactoryRegistry createChild() {
+        return copyOf(this);
+    }
+
+    void register(CollectorFactory<?, ?> factory) {
+        factories.add(factory);
+    }
+
+    @SuppressWarnings("unchecked")
+    <T, R> Collector<T, ?, R> createCollectorFor(Class<R> type) {
+        for (CollectorFactory<?, ?> factory : factories) {
+            if (factory.accepts(type)) {
+                return ((CollectorFactory<T, R>) factory).newCollector(type);
+            }
+        }
+
+        throw new IllegalStateException("No collector builder available for " + type.getName());
+    }
+
+    static CollectorFactoryRegistry copyOf(CollectorFactoryRegistry registry) {
+        CollectorFactoryRegistry newRegistry = new CollectorFactoryRegistry();
+        newRegistry.factories.addAll(registry.factories);
+        return newRegistry;
+    }
+
+    private static class SortedSetCollectorFactory<T> implements CollectorFactory<T, SortedSet<T>> {
+
+        public boolean accepts(Class<?> type) {
+            return type.equals(SortedSet.class);
+        }
+
+        public Collector<T, ?, SortedSet<T>> newCollector(Class<SortedSet<T>> type) {
+            return Collectors.toCollection(TreeSet::new);
+        }
+    }
+
+    private static class ListCollectorFactory<T> implements CollectorFactory<T, List<T>> {
+
+        public boolean accepts(Class<?> type) {
+            return type.equals(List.class);
+        }
+
+        public Collector<T, ?, List<T>> newCollector(Class<List<T>> type) {
+            return Collectors.toList();
+        }
+    }
+
+    private static class SetCollectorFactory<T> implements CollectorFactory<T, Set<T>> {
+
+        public boolean accepts(Class<?> type) {
+            return type.equals(Set.class);
+        }
+
+        public Collector<T, ?, Set<T>> newCollector(Class<Set<T>> type) {
+            return Collectors.toSet();
+        }
+    }
+}
--- a/core/src/main/java/org/jdbi/v3/GeneratedKeys.java
+++ b/core/src/main/java/org/jdbi/v3/GeneratedKeys.java
@@ -32,6 +32,7 @@ public class GeneratedKeys<Type> implements ResultBearing<Type>
     private final Statement                stmt;
     private final ResultSet                results;
     private final StatementContext         context;
+    private final CollectorFactoryRegistry collectorFactoryRegistry;
 
     /**
      * Creates a new wrapper object for generated keys as returned by the {@link Statement#getGeneratedKeys()}
@@ -45,7 +46,8 @@ public class GeneratedKeys<Type> implements ResultBearing<Type>
     GeneratedKeys(ResultSetMapper<Type> mapper,
                   SQLStatement<?> jdbiStatement,
                   Statement stmt,
-                  StatementContext context) throws SQLException
+                  StatementContext context,
+                  CollectorFactoryRegistry collectorFactoryRegistry) throws SQLException
     {
         this.mapper = mapper;
         this.jdbiStatement = jdbiStatement;
@@ -53,6 +55,7 @@ public class GeneratedKeys<Type> implements ResultBearing<Type>
         this.results = stmt.getGeneratedKeys();
         this.context = context;
         this.jdbiStatement.addCleanable(Cleanables.forResultSet(results));
+        this.collectorFactoryRegistry = collectorFactoryRegistry.createChild();
     }
 
     /**
@@ -73,5 +76,10 @@ public class GeneratedKeys<Type> implements ResultBearing<Type>
         catch (SQLException e) {
             throw new ResultSetException("Exception thrown while attempting to traverse the result set", e, context);
         }
+    }
+
+    @Override
+    public <ContainerType> ContainerType collectInto(Class<ContainerType> containerType) {
+        return collect(collectorFactoryRegistry.createCollectorFor(containerType));
     }
 }
--- a/core/src/main/java/org/jdbi/v3/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/Handle.java
@@ -25,6 +25,7 @@ import org.jdbi.v3.tweak.ResultSetMapper;
 import org.jdbi.v3.tweak.StatementBuilder;
 import org.jdbi.v3.tweak.StatementLocator;
 import org.jdbi.v3.tweak.StatementRewriter;
+import org.jdbi.v3.tweak.CollectorFactory;
 
 /**
  * This represents a connection to the database system. It usually is a wrapper around
@@ -287,4 +288,6 @@ public interface Handle extends Closeable
     TransactionIsolationLevel getTransactionIsolationLevel();
 
     void registerArgumentFactory(ArgumentFactory<?> argumentFactory);
+
+    void registerCollectorFactory(CollectorFactory<?, ?> factory);
 }
--- a/core/src/main/java/org/jdbi/v3/Query.java
+++ b/core/src/main/java/org/jdbi/v3/Query.java
@@ -50,9 +50,11 @@ public class Query<ResultType> extends SQLStatement<Query<ResultType>> implement
           TimingCollector timingCollector,
           Collection<StatementCustomizer> customizers,
           MappingRegistry mappingRegistry,
-          Foreman foreman)
+          Foreman foreman,
+          CollectorFactoryRegistry collectorFactoryRegistry)
     {
-        super(params, locator, statementRewriter, handle, cache, sql, ctx, timingCollector, customizers, foreman);
+        super(params, locator, statementRewriter, handle, cache, sql, ctx, timingCollector, customizers, foreman,
+                collectorFactoryRegistry);
         this.mapper = mapper;
         this.mappingRegistry = mappingRegistry;
     }
@@ -70,6 +72,11 @@ public class Query<ResultType> extends SQLStatement<Query<ResultType>> implement
                                                         stmt,
                                                         stmt.getResultSet(),
                                                         getContext()));
+    }
+
+    @Override
+    public <ContainerType> ContainerType collectInto(Class<ContainerType> containerType) {
+        return collect(getCollectorFactoryRegistry().createCollectorFor(containerType));
     }
 
     /**
@@ -115,7 +122,8 @@ public class Query<ResultType> extends SQLStatement<Query<ResultType>> implement
                             getTimingCollector(),
                             getStatementCustomizers(),
                             MappingRegistry.copyOf(mappingRegistry),
-                            getForeman().createChild());
+                            getForeman().createChild(),
+                            getCollectorFactoryRegistry().createChild());
     }
 
     /**
--- a/core/src/main/java/org/jdbi/v3/ResultBearing.java
+++ b/core/src/main/java/org/jdbi/v3/ResultBearing.java
@@ -15,6 +15,7 @@ package org.jdbi.v3;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collector;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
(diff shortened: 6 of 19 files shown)


core/src/main/java/org/jdbi/v3/ResultBearing.java
```diff
@@ -63,4 +72,43 @@ default ResultType findOnly() {
...
+     * A factory for building the container should be registered in the query's
+     * container factory registry.
+     * @param containerType the class that represents the container
+     * @param <ContainerType> the generic type of the container
+     * @return the container with the query result
+     */
+    default <ContainerType> ContainerType collect(Class<ContainerType> containerType)
+    {
+        ContainerBuilder<ContainerType> builder =  getContainerFactoryRegistry().createBuilderFor(containerType);
+        return stream().collect(new Collector<ResultType, ContainerBuilder<ContainerType>, ContainerType>() {
```
qualidafial: There's a Collector.of method that avoids need to use anonymous inner class.

core/src/main/java/org/jdbi/v3/ResultBearing.java
```diff
@@ -24,6 +31,8 @@
     @Override
     ResultIterator<ResultType> iterator();
 
+    ContainerFactoryRegistry getContainerFactoryRegistry();
```
qualidafial: I'm not sure I like exposing the registry in the ResultBearing interface. We try to keep these classes internal so they can evolve without breaking compatibility.

core/src/main/java/org/jdbi/v3/ContainerBuilder.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+package org.jdbi.v3;
+
+/**
+ * Accumulator for building containers.
+ *
+ * @param <ContainerType> the type of the container
+ */
+public interface ContainerBuilder<ContainerType> {
```
qualidafial: Can we use Collector from the streams API directly instead? So e.g. ContainerFactory would become CollectorFactory.

core/src/main/java/org/jdbi/v3/tweak/ContainerFactory.java
```diff
@@ -0,0 +1,40 @@
...
+     */
+    boolean accepts(Class<?> type);
+
+    /**
+     * Builds a new container builder for objects of the given type.
+     *
+     * @param type the object type
+     * @return the container builder
+     */
+    ContainerBuilder<T> newContainerBuilderFor(Class<?> type);
```
qualidafial: Huh. The wildcard argument coupled with the ContainerBuildet<T> parameterized result is a nice compromise on type safety.

core/src/main/java/org/jdbi/v3/ResultBearing.java
arteam: That's a great observation. Really reduces the amount of code.

core/src/main/java/org/jdbi/v3/ResultBearing.java
arteam: I did that because I wanted to reuse the code of collecting results to a container. Since we can use the `Collector.of` method, I think we could just implement the `collect` method  twice in `Query` and `GeneratedKeys` and didn't expose the internals.

core/src/main/java/org/jdbi/v3/ContainerBuilder.java
arteam: Makes sense. We don't care about backward compatability and `Collector` is a more advanced API.

core/src/main/java/org/jdbi/v3/tweak/ContainerFactory.java
arteam: Yes, that was kind of sloppy.

qualidafial: That was actually supposed to be a compliment. ;) I think you found the right way to generify factories.

core/src/main/java/org/jdbi/v3/GeneratedKeys.java
```diff
@@ -74,4 +77,9 @@
             throw new ResultSetException("Exception thrown while attempting to traverse the result set", e, context);
         }
     }
+
+    @Override
+    public <ContainerType> ContainerType collect(Class<ContainerType> containerType) {
+        return stream().collect(containerFactoryRegistry.createCollectorFor(containerType));
+    }
```
qualidafial: Let rename this to e.g. `collectInto`, and add another `collect(Collector)` method.

core/src/main/java/org/jdbi/v3/ContainerFactoryRegistry.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.function.Function;
+import java.util.stream.Collector;
+
+/**
+ * Registry of container factories.
+ * <p/>
+ * Contains a set of factories, registered by the application.
+ */
+class ContainerFactoryRegistry {
```
qualidafial: Rename this to CollectorFactoryRegistry?

core/src/main/java/org/jdbi/v3/GeneratedKeys.java
arteam: Makes sense. 

I also thought about using `java.lang.reflect.Type` as the parameter type and resolving the actual type on runtime. With that users can write type-safe code like 
```java
ImmutableList<String> rs = h.createQuery("select name from something order by id")
                .mapTo(String.class)
                .collect(new TypeToken<ImmutableList<String>>(){}.getType());
```

What do you think? Is it worth it?

core/src/main/java/org/jdbi/v3/BasicHandle.java
```diff
@@ -244,7 +249,8 @@ public Update createStatement(String sql)
                           sql,
                           new ConcreteStatementContext(globalStatementAttributes, MappingRegistry.copyOf(mappingRegistry)),
                           timingCollector,
-                          foreman);
+                          foreman,
+                          containerFactoryRegistry);
```
qualidafial: Can we revert this constructor argument change for classes that don't implement ResultBearing?

core/src/main/java/org/jdbi/v3/BasicHandle.java
```diff
@@ -448,4 +456,9 @@ public void registerArgumentFactory(ArgumentFactory<?> argumentFactory)
     {
         this.foreman.register(argumentFactory);
     }
+
+    @Override
+    public void registerContainerFactory(CollectorFactory<?, ?, ?> factory) {
```
qualidafial: Rename method

arteam:
Thanks @qualidafial for your very thoughtful feedback. I've pushed a new version, please take another look. I am going to go through another round of comments tomorrow.

core/src/main/java/org/jdbi/v3/GeneratedKeys.java
qualidafial: I would almost rather just pass in a Collector directly with the fluent API. E.g. `GuavaCollectors.toImmutableList`.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/RegisterContainerMapper.java
```diff
@@ -0,0 +1,83 @@
...
+            return new Customizer((RegisterContainerMapper) annotation);
+        }
+
+        public SqlStatementCustomizer createForType(Annotation annotation, Class sqlObjectType) {
+            return new Customizer((RegisterContainerMapper) annotation);
+        }
+
+        public SqlStatementCustomizer createForParameter(Annotation annotation, Class sqlObjectType, Method method,
+                                                         Object arg) {
+            throw new UnsupportedOperationException("Not Yet Implemented!");
```
qualidafial: Better for the error message to say that the annotation is not supported on parameters.

sqlobject/src/test/java/org/jdbi/v3/docs/TestInClauseExpansion.java
```diff
@@ -63,9 +65,24 @@ public void testInClauseExpansion() throws Exception
...
+    public static class ImmutableSetCollectorFactory<T>
+            implements CollectorFactory<T, ImmutableSet.Builder<T>, ImmutableSet<T>> {
+
+        public boolean accepts(Class<?> type) {
+            return ImmutableSet.class.isAssignableFrom(type);
+        }
+
+        @Override
+        public Collector<T, ImmutableSet.Builder<T>, ImmutableSet<T>> newCollector(Class<ImmutableSet<T>> type) {
+            return Collector.of(ImmutableSet.Builder::new, ImmutableSet.Builder::add, (first, second) -> first,
```
qualidafial: The `(first, second) -> first` combiner is making me uncomfortable. Better to throw an exception than to silently permit unsupported parallelism.

qualidafial: With that said, it is so cool that this works with just a few method references.

core/src/main/java/org/jdbi/v3/GeneratedKeys.java
qualidafial: @brianm @stevenschlansker @arteam This comment speaks to the broader discussion surrounding `Class<T>` vs `Type` parameters.

We don't get any compile-time guarantees when using `Type`. `Type` is just a marker interface, so there are no generic parameters or hints that the compiler could use to make type safety guarantees. It is a black box at compile time.

For this reason I don't think it makes sense to use `Type` in fluent APIs like e.g. Query, as in @arteam's suggestion above. The best we can offer is a method with a `Class<T>` parameter, alongside a fully generic `Whatever<T>` method e.g. `Query.map(ResultSetMapper<T>)` vs `Query.mapTo(Class<T>)`.

SQL Objects are a different story though. We have full type information available when inspecting the interface, so we can pass e.g. `getGenericReturnType()` to factories.

For this reason I propose that only factory classes (ArgumentFactory, ResultSetMapperFactory, ResultColumnMapperFactory, and CollectorFactory) be changed to use `Type` parameters.

core/src/main/java/org/jdbi/v3/ResultBearing.java
qualidafial: :+1:

core/src/main/java/org/jdbi/v3/Handle.java
```diff
@@ -287,4 +288,6 @@
     TransactionIsolationLevel getTransactionIsolationLevel();
 
     void registerArgumentFactory(ArgumentFactory<?> argumentFactory);
+
+    void registerContainerFactory(CollectorFactory<?, ?, ?> factory);
```
qualidafial: Another one to rename

core/src/main/java/org/jdbi/v3/ResultBearing.java
```diff
@@ -63,4 +63,15 @@ default ResultType findOnly() {
...
+    /**
+     * Collect the query result into a container of the type specified by the given class.
+     * A factory for building the container should be registered in the query's
+     * container factory registry.
+     *
+     * @param containerType   the class that represents the container
+     * @param <ContainerType> the generic type of the container
+     * @return the container with the query result
+     */
+    <ContainerType> ContainerType collect(Class<ContainerType> containerType);
```
qualidafial: Let's rename this to `collectInto(Class<ContainerType>)` and an additional `collect(Collector<T,A,R>)` method alongside it.

core/src/main/java/org/jdbi/v3/SQLStatement.java
```diff
@@ -88,12 +91,23 @@
...
         this.params = params;
         this.locator = locator;
+        this.containerFactoryRegistry = containerFactoryRegistry.createChild();
 
         ctx.setConnection(handle.getConnection());
         ctx.setRawSql(sql);
         ctx.setBinding(params);
     }
 
+    public ContainerFactoryRegistry getContainerFactoryRegistry()
```
qualidafial: Can this be made package private?

qualidafial:
Nice work! Just a few more comments to address, and this should be good to merge.

core/src/main/java/org/jdbi/v3/BasicHandle.java
arteam: We need the registry here to pass it to `Query` in the `createQuery` method.

arteam:
The next round. Please review.

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private static class SortedSetCollectorFactory<T> implements CollectorFactory<T, SortedSet<T>, SortedSet<T>> {
+
+        public boolean accepts(Class<?> type) {
+            return type.equals(SortedSet.class);
+        }
+
+        public Collector<T, SortedSet<T>, SortedSet<T>> newCollector(Class<SortedSet<T>> type) {
+            return Collector.of(TreeSet::new, SortedSet<T>::add, (first, second) -> {
+                throw new UnsupportedOperationException("Parallel collecting is not supported");
```
qualidafial: :+1:

core/src/main/java/org/jdbi/v3/ResultBearing.java
```diff
@@ -59,8 +60,32 @@ default ResultType findOnly() {
     }
 
     default List<ResultType> list() {
+        return collect(Collectors.toList());
```
qualidafial: :+1:

core/src/main/java/org/jdbi/v3/SQLStatement.java
```diff
@@ -88,12 +91,23 @@
...
         ctx.setBinding(params);
     }
 
+    CollectorFactoryRegistry getCollectorFactoryRegistry()
+    {
+        return collectorFactoryRegistry;
+    }
+
+    @SuppressWarnings("unchecked")
+    public SelfType registerContainerFactory(CollectorFactory<?,?, ?> collectorFactory) {
```
qualidafial: This one still needs to be renamed.

sqlobject/src/test/java/org/jdbi/v3/docs/TestInClauseExpansion.java
```diff
@@ -63,9 +65,25 @@ public void testInClauseExpansion() throws Exception
...
+        public boolean accepts(Class<?> type) {
+            return ImmutableSet.class.isAssignableFrom(type);
+        }
+
+        @Override
+        public Collector<T, ImmutableSet.Builder<T>, ImmutableSet<T>> newCollector(Class<ImmutableSet<T>> type) {
+            return Collector.of(ImmutableSet.Builder::new, ImmutableSet.Builder::add, (first, second) -> {
+                throw new UnsupportedOperationException("Parallel collecting is not supported");
+            }, ImmutableSet.Builder::build, Collector.Characteristics.UNORDERED);
+        }
```
qualidafial: I wouldn't be opposed to a `jdbi-guava` artifact for these to live in (probably after the jdbi3 release though).

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestCollectorFactory.java
```diff
@@ -0,0 +1,290 @@
...
+    @Test
+    public void testExists() throws Exception {
+        h.execute("insert into something (id, name) values (1, 'Coda')");
+        h.registerCollectorFactory(new MaybeCollectorFactory());
+
+        @SuppressWarnings("unchecked")
+        Maybe<String> rs = h.createQuery("select name from something where id = :id")
+                .bind("id", 1)
+                .mapTo(String.class)
+                .collectInto(Maybe.class);
```
qualidafial: Why does this need `@SuppressWarnings`?

qualidafial:
@arteam A few more minor comments.

core/src/main/java/org/jdbi/v3/BasicHandle.java
```diff
@@ -244,7 +249,8 @@ public Update createStatement(String sql)
                           sql,
                           new ConcreteStatementContext(globalStatementAttributes, MappingRegistry.copyOf(mappingRegistry)),
                           timingCollector,
-                          foreman);
+                          foreman,
+                collectorFactoryRegistry);
```
stevenschlansker: Total nitpick, these lines no longer line up.

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+
+        throw new IllegalStateException("No collector builder available for " + type.getName());
+    }
+
+    private static class SortedSetCollectorFactory<T> implements CollectorFactory<T, SortedSet<T>, SortedSet<T>> {
+
+        public boolean accepts(Class<?> type) {
+            return type.equals(SortedSet.class);
```
stevenschlansker: Why is `SortedSet` special here?  Should we also have implementations for `Set` or `List`?

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        public Collector<T, SortedSet<T>, SortedSet<T>> newCollector(Class<SortedSet<T>> type) {
```
stevenschlansker: Should `T` be something like `T extends Comparable<T>`?

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
stevenschlansker: Nice!

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
```diff
@@ -0,0 +1,72 @@
...
+    CollectorFactoryRegistry createChild() {
+        return new CollectorFactoryRegistry(this);
+    }
+
+    void register(CollectorFactory<?, ?, ?> factory) {
+        factories.add(factory);
+    }
+
+    @SuppressWarnings("unchecked")
+    <T, R> Collector<T, ?, R> createCollectorFor(Class<R> type) {
```
stevenschlansker: I know you are only mirroring how the existing registries work, but I wonder if it would make sense for these to either return a nullable type or an `Optional<Collector<...>>` so that lookups can be done without necessarily throwing an exception.  If so, probably a change that should be done to the existing registries as well.

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+class CollectorFactoryRegistry {
+
+    private final Set<CollectorFactory<?, ?, ?>> factories = new CopyOnWriteArraySet<>();
+
+    CollectorFactoryRegistry() {
+        factories.add(new SortedSetCollectorFactory());
+    }
+
+    CollectorFactoryRegistry(CollectorFactoryRegistry parent) {
```
stevenschlansker: I somewhat like a static method `.copyOf` just so it is clear that it is a copy, and detached from the parent.  (as in, further changes to a `DBI` will not affect existing `Handles` etc)

core/src/main/java/org/jdbi/v3/Query.java
```diff
@@ -50,9 +50,11 @@
           TimingCollector timingCollector,
           Collection<StatementCustomizer> customizers,
           MappingRegistry mappingRegistry,
-          Foreman foreman)
+          Foreman foreman,
+          CollectorFactoryRegistry collectorFactoryRegistry)
```
stevenschlansker: I wonder if we should start aggregating the types like `MappingRegistry`, `Foreman`, and `CollectorFactoryRegistry` -- they seem to all be passed around to largely the same places, and often when we copy one as a child we end up copying them all.

sqlobject/src/test/java/org/jdbi/v3/docs/TestInClauseExpansion.java
stevenschlansker: I think that is a great idea, and it could be part of the initial release.  It is pretty easy to add another module.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestCollectorFactory.java
```diff
@@ -0,0 +1,290 @@
...
+import static org.junit.Assert.assertThat;
+
+public class TestCollectorFactory {
+
+    private DBI dbi;
+    private Handle h;
+
+    @Before
+    public void setUp() throws Exception {
+        this.dbi = new DBI("jdbc:h2:mem:" + UUID.randomUUID());
```
stevenschlansker: Check out the new `H2DatabaseRule` which makes this harnessing code obsolete!

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestCollectorFactory.java
```diff
@@ -0,0 +1,290 @@
...
+
+        @Override
+        public Collector<T, MaybeBuilder<T>, Maybe<T>> newCollector(Class<Maybe<T>> type) {
+            return Collector.of(MaybeBuilder::new, MaybeBuilder::set, (first, second) -> {
+                throw new UnsupportedOperationException("Parallel collecting is not supported");
+            }, MaybeBuilder::build);
+        }
+    }
+
+    public static abstract class Maybe<T> {
```
stevenschlansker: Any reason to declare our own `Maybe<T>` here rather than just reuse `Optional<T>`?

stevenschlansker:
Thanks for working on this!  this is looking really good :)

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
qualidafial: I'm planning to make mapper and argument factories to deal with Optional, once the `Type` vs `Class` stuff is in place.

core/src/main/java/org/jdbi/v3/Query.java
qualidafial: :+1:

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
stevenschlansker: Sounds good.

core/src/main/java/org/jdbi/v3/Query.java
stevenschlansker: Filed as #202 for later followup.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestCollectorFactory.java
arteam: Because `colectInto` returns plain `Maybe` without a generic parameter.
So, we need to suppress a warning from the compiler about an unchecked cast.

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
qualidafial: Oh, gotcha. I like that much better than the exception.

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
arteam: For Fluent API users can use Stream API and the `collect` method with `Collectors.toSet()`
and `Collectors.toList`.

For SqlObject API we already handle `Set` and `List` in `ResultReturnThing.IterableReturningThing`. See [here](https://github.com/jdbi/jdbi/blob/jdbi3/sqlobject/src/main/java/org/jdbi/v3/sqlobject/ResultReturnThing.java#L263]).

But I think we should refactor this method and use collector factories instead.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestCollectorFactory.java
qualidafial: This actually makes me question whether we want a `collectInto(Class<R>)` method. The `collect(Collector<T,A,R>)` method can make type safety guarantees, because the `T` parameter can be inferred from the query's type parameter.

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
arteam: During refactoring I've realized that we don't need the accumulator type for container factories, because it's actually a detail of a collector implementation. 

The JDK `Collectors.toList` method returns `Collector<T, ?, List<T>>`

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
arteam: What's wrong with throwing an exception? How can we handle absence of an appropriate `CollectorFactory`?

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestCollectorFactory.java
arteam: No reason, just followed an existed test in JDBI2. Will refactor.

arteam:
I pushed a new version, please review. It looks more polished for me.

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
qualidafial: An Optional lets the caller decide whether to throw an exception if no appropriate collector can be found.

stevenschlansker: That said, we can fix this up in a later refactor, so we can make similar changes to `MappingRegistry` and `Foreman` and friends.  Filed as #208

sqlobject/src/test/java/org/jdbi/v3/docs/TestInClauseExpansion.java
stevenschlansker: #203

core/src/main/java/org/jdbi/v3/tweak/CollectorFactory.java
```diff
@@ -0,0 +1,42 @@
...
+ * Factory for building containers of elements.
+ * The collector accepts only objects of the type of the container elements.
+ *
+ * @param <T> the type of the container elements
+ * @param <R> the type of the container
+ */
+public interface CollectorFactory<T, R> {
+
+    /**
+     * Whether the corresponding collector accepts objects of the given type.
```
stevenschlansker: Is this "accepts objects of the given type" or more accurate as "produces objects of the given type" since it will compare against e.g. `List` or `Set`

stevenschlansker:
OK one final doc nitpick, then I think this is good to merge.

core/src/main/java/org/jdbi/v3/tweak/CollectorFactory.java
```diff
@@ -0,0 +1,42 @@
...
+     */
+    boolean accepts(Class<?> type);
+
+    /**
+     * Builds a new {@link Collector}.
+     *
+     * @param type the actual type of the container
+     * @return the {@link Collector}
+     */
+    Collector<T, ?, R> newCollector(Class<R> type);
```
qualidafial: One small nit: could we remove the `T` parameter from the class, and put it only on this method? e.g.

    <T> Collector<T,?,R> newCollector(Class<R> type);

arteam: I don't think that we could, because we will lose type safety in implemetations. Basicly, we won't be able to specify `R` as `ContainerType<T>`.

arteam:
I've pushed a new version for review with a javadoc change.

stevenschlansker:
:tada:

arteam:
Hurray! Thanks, @stevenschlansker and @qualidafial!

PR #200, opened 2015-10-26
BeanMapper: support for camel-case conversion

JeanRev:
Hi,

This PR is about extending the Bean Mapper to allow automatic mappings from column names using the (very common) "underscore" syntax. Such functionality is also supported out-of-the-box in the BeanPropertyRowMapper (http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/jdbc/core/BeanPropertyRowMapper.html) from Spring.

In addition, it allows to further customize the mappings with two additional protected methods (one to get the mappings related to a given property name, and another to customize the locale used for case manipulation).

This could be considered a breaking change, since setters that were not mapped before may suddenly start to be invoked. However, the potential disruptive effect seems to be low.

Jean


JeanRev:
I just noticed that the availability of the JDBI3 branch... This is maybe a more suitable target branch than master.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
```diff
@@ -51,6 +67,58 @@ public BeanMapper(Class<T> type)
...
+
+    /**
+     * Gets the locale used to manipulate the Bean properties name. This locale is useful, for example, when doing
+     * case conversion. The locale is expected to be constant across method invocations. By default the {@code US}
+     * locale will be used.
+     *
+     * @return the target locale, never {@code null}
+     */
+    protected Locale getLocale() {
+        return Locale.US;
```
qualidafial: Do you have some examples to justify making US locale the default for everyone worldwide?

This looks a bit heavy-handed.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
```diff
@@ -51,6 +67,58 @@ public BeanMapper(Class<T> type)
...
+            String s = propertyName.substring(i, i + 1);
+            String slc = s.toLowerCase(locale);
+            if (!s.equals(slc)) {
+                // Different cases: tokenize.
+                propertyNameWithUnderscores.append("_").append(slc);
+            }
+            else {
+                propertyNameWithUnderscores.append(s);
+            }
+        }
```
qualidafial: Why not just `Pattern.compile("_").matches(propertyName).replaceAll("")`?

qualidafial:
You are correct about the `jdbi3` branch--you're welcome to port this and submit a new pull request against the branch if you'd like.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
JeanRev: I choose Locale.US for safety since this is what the Spring implementation is currently using as default.

Now, whatever the default is, it should actually not interfere with the case mappings for maximum compatibility.

JDK < 8 is advising Locale.ENGLISH for this purpose (see for example java.lang.String.toLowerCase()), and Locale.ROOT starting from JDK 8 (Locale.ROOT is available since Java 6).

If I had to decide, my personal order of preference for this value would actually be: Locale.ROOT, Locale.English, Locale.US.

src/main/java/org/skife/jdbi/v2/BeanMapper.java
JeanRev: Well, in this precise situation, we want to insert underscores, not to remove them. While we may think of some similar implementation based on regexp, I replicated the Spring implementation because we can expect it to be well-tested (in addition, the StringBuilder based code should be fairly efficient).

Or did you meant mapping the fetched column labels to bean properties instead?

JeanRev:
Weird, I did a rebase/force push against jdbi3 but this do not seems to play well on existing PR... I close this PR and create a new one on JDBI3 as suggested.

JeanRev:
Here: https://github.com/jdbi/jdbi/pull/201

qualidafial:
@JeanRev I think it's because this pull request was originally made the master branch.

kkrgwbj:
hi,@stevenschlansker ,i am test the camel case,but I found a bug, spring JDBC beanPropertyRowMapper is have a problem, for this like 'user_name_2' this variable has a problem. I found, from inside the sql2o sql2o replacement inside under camel method, named hope you can accept. thank you

`public class UnderscoreToCamelCase {
    public static String convert(String  underscore){
        if(underscore==null || underscore.isEmpty()) return underscore;
        return convert00(underscore);
    }

    private static String convert00(String  underscore){
        char[] chars = underscore.toCharArray();
        int write=-1,len=chars.length;
        boolean upper=false;
        for (int read = 0; read < len; ++read) {
            char c = chars[read];
            if('_'==c){
                upper = true;
                continue;
            }
            if(upper){
                upper = false;
                chars[++write]=Character.toUpperCase(c);
            } else {
                chars[++write]=Character.toLowerCase(c);
            }
        }
        return new String(chars, 0, ++write);
    }
}`

Issue #74, opened 2013-09-05
Support snake case/camel case together.


cemo:
As a Jdbi 3 request:

Please support configuring JDBI to utilize naming strategy for `snake_case` naming.

PR #201, opened 2015-10-26
BeanMapper property names also map to underscored column names

JeanRev:
Note: This is a rebase on the JDBI 3 branch from this PR: https://github.com/jdbi/jdbi/pull/200

Hi,

This PR is about extending the Bean Mapper to allow automatic mappings from column names using the (very common) "underscore" syntax. Such functionality is also supported out-of-the-box in the BeanPropertyRowMapper (http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/jdbc/core/BeanPropertyRowMapper.html) from Spring.

In addition, it allows to further customize the mappings with two additional protected methods (one to get the mappings related to a given property name, and another to customize the locale used for case manipulation).

This could be considered a breaking change, since setters that were not mapped before may suddenly start to be invoked. However, the potential disruptive effect seems to be low.

Jean

### DIFF
--- a/core/src/main/java/org/jdbi/v3/BeanMapper.java
+++ b/core/src/main/java/org/jdbi/v3/BeanMapper.java
@@ -21,15 +21,21 @@ import java.lang.reflect.InvocationTargetException;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 import org.jdbi.v3.tweak.ResultColumnMapper;
 import org.jdbi.v3.tweak.ResultSetMapper;
 
 /**
- * A result set mapper which maps the fields in a statement into a JavaBean. This uses
- * the JDK's built in bean mapping facilities, so it does not support nested properties.
+ * A result set mapper which maps the fields in a statement into a JavaBean. The default implementation will perform a
+ * case insensitive mapping between the bean property names and the column labels, also considering camel-case to
+ * underscores conversion. This uses the JDK's built in bean mapping facilities, so it does not support nested
+ * properties.
  */
 public class BeanMapper<T> implements ResultSetMapper<T>
 {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jdbi/v3/BeanMapper.java
```diff
@@ -51,6 +67,58 @@ public BeanMapper(Class<T> type)
...
+
+    /**
+     * Gets the locale used to manipulate the Bean properties name. This locale is useful, for example, when doing
+     * case conversion. The locale is expected to be constant across method invocations. By default the {@code US}
+     * locale will be used.
+     *
+     * @return the target locale, never {@code null}
+     */
+    protected Locale getLocale() {
+        return Locale.US;
```
stevenschlansker: This can / should probably be `Locale.ROOT` yeah?

core/src/main/java/org/jdbi/v3/BeanMapper.java
```diff
@@ -51,6 +67,58 @@ public BeanMapper(Class<T> type)
...
+     *
+     * @param propertyName the bean property name
+     *
+     * @return a collection holding the possible column names (mutable in the default implementation)
+     */
+    protected Collection<String> mapToColumnNames(String propertyName)
+    {
+        Locale locale = getLocale();
+
+        List<String> columnName = new ArrayList<>(5);
```
stevenschlansker: This would be better sized as 2 it seems, although that's obviously a nitpick.

core/src/test/java/org/jdbi/v3/BeanMapperTest.java
```diff
@@ -61,6 +61,23 @@ public void shouldSetValueOnPublicSetter() throws Exception {
     }
 
     @Test
+    public void shouldHandleColumNameWithUnderscores() throws Exception {
+        expect(resultSetMetaData.getColumnCount()).andReturn(1).anyTimes();
+        expect(resultSetMetaData.getColumnLabel(1)).andReturn("LONG_FIELD");
```
stevenschlansker: Would be nice to have tests for a couple more cases -- say
```
long_field
longField
Long_Field
LongField
```

stevenschlansker:
Couple of minor comments, thank you for looking at this!

stevenschlansker:
Fixes #74

core/src/main/java/org/jdbi/v3/BeanMapper.java
JeanRev: Agreed. Done in 3012ebb

core/src/main/java/org/jdbi/v3/BeanMapper.java
JeanRev: Actually, I prefer to be slightly oversized regarding what subclasses may want to do with that list...

core/src/test/java/org/jdbi/v3/BeanMapperTest.java
JeanRev: I added a test to check for both underscore syntax and case insensitivity in 3012ebb.

The two other proposed input (longField and LongField) are actually already covered in other tests of the same class.

JeanRev:
@stevenschlansker thank you for the feedback and linking to the original feature request.

core/src/main/java/org/jdbi/v3/BeanMapper.java
```diff
@@ -51,6 +67,58 @@ public BeanMapper(Class<T> type)
...
+            String slc = s.toLowerCase(locale);
+            if (!s.equals(slc)) {
+                // Different cases: tokenize.
+                propertyNameWithUnderscores.append("_").append(slc);
+            }
+            else {
+                propertyNameWithUnderscores.append(s);
+            }
+        }
+        columnNames.add(propertyNameWithUnderscores.toString());
```
qualidafial: Two last things: First, I would add some collision detection and log warnings when two properties map to the same column name. It may never come up, but I would rather catch it and say something, than let users puzzle over why their columns got mapped to a different property than they expected.

Second, there is a CaseFormat class in Guava that converts between different cases: `CaseFormat.LOWER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, propertyName)`. I'm planning to move the bean mapping and binding stuff out of core to a separate artifact `jdbi-beans`, but you can just promote Guava to a compile dependency for now.

stevenschlansker: I'm fine with logging a warning on collision.  Maybe the Guava change should be held off until we actually do such a split out?  I'm curious why you want to move the bean stuff -- I'm not necessarily opposed to it but each module we make creates additional dependencies our end users need to know about.

stevenschlansker: https://github.com/jdbi/jdbi/commit/123b1b6b94a9280eef4e3f1968744d5bf882834c
https://github.com/jdbi/jdbi/commit/31a5fa3a84c432dcda3c4931cec5968c43051831

stevenschlansker:
Rebased with a minor whitespace fixup, added two final touches, and merged it in :)

PR #212, opened 2015-10-28
Factor out property naming strategies

stevenschlansker:
Lets our two different bean mappers use the same underlying code to figure out SQL name <-> Java name mappings, and makes it pluggable without needing to subclass.

Probably best viewed with `?w=1`

### DIFF
--- a/core/src/main/java/org/jdbi/v3/BeanMapper.java
+++ b/core/src/main/java/org/jdbi/v3/BeanMapper.java
@@ -21,16 +21,19 @@ import java.lang.reflect.InvocationTargetException;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 import org.jdbi.v3.tweak.ResultColumnMapper;
 import org.jdbi.v3.tweak.ResultSetMapper;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.jdbi.v3.util.bean.ColumnNameMappingStrategy;
+import org.jdbi.v3.util.bean.CaseInsensitiveColumnNameStrategy;
+import org.jdbi.v3.util.bean.SnakeCaseColumnNameStrategy;
 
 /**
  * A result set mapper which maps the fields in a statement into a JavaBean. The default implementation will perform a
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/jdbi/v3/ReflectiveFieldMapper.java
```diff
@@ -34,22 +39,18 @@
 public class ReflectiveFieldMapper<T> implements ResultSetMapper<T>
```
qualidafial: Unrelated, but can we just call this `FieldMapper`?

core/src/main/java/org/jdbi/v3/BeanMapper.java
```diff
@@ -143,44 +88,58 @@ public T map(int row, ResultSet rs, StatementContext ctx)
...
-            String name = metadata.getColumnLabel(i).toLowerCase();
+            String name = metadata.getColumnLabel(i);
 
-            PropertyDescriptor descriptor = properties.get(name);
+            final Optional<PropertyDescriptor> maybeDescriptor =
+                    descriptorByColumnCache.computeIfAbsent(name, this::descriptorForColumn);
 
-            if (descriptor != null) {
-                Class type = descriptor.getPropertyType();
+            if (!maybeDescriptor.isPresent()) {
```
qualidafial: Instead of checking `isPresent()`, I suggest:

    maybeDescriptor.ifPresent(descriptor -> {
      // do stuff with descriptor
    }

stevenschlansker: Only reason I did it this way is because I rather like escaping the loop early, rather than wrapping all of the logic in an `if` block... fewer levels of useless indentation :)

core/src/main/java/org/jdbi/v3/ReflectiveFieldMapper.java
stevenschlansker: Works for me.

core/src/main/java/org/jdbi/v3/ReflectiveFieldMapper.java
```diff
@@ -71,33 +72,52 @@ public T map(int row, ResultSet rs, StatementContext ctx)
...
+            for (Field field : aClass.getDeclaredFields()) {
+                for (ColumnNameMappingStrategy strategy : nameMappingStrategies) {
+                    if (strategy.nameMatches(field.getName(), columnName)) {
+                        return Optional.of(field);
+                    }
+                }
+            }
+            aClass = aClass.getSuperclass();
+        }
+        return Optional.empty();
```
qualidafial: I feel like by the end of the first failed lookup, we've already iterated through all possible field/strategy combinations, and we'd get better performance by eagerly caching all the column to field mappings in the constructor.

core/src/main/java/org/jdbi/v3/util/bean/ColumnNameMappingStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Strategy for mapping Java bean property and field names
+ * to SQL column names.
+ */
+public interface ColumnNameMappingStrategy {
+    /**
+     * Given the Java name and SQL column name,
+     * return true if they are logically equivalent.
+     */
+    boolean nameMatches(String propertyName, String sqlColumnName);
```
qualidafial: As a follow up to my eager caching comment, this strategy could be replaced by a `Function<String,String>` which maps a field/property name to a column name.

core/src/test/java/org/jdbi/v3/ReflectiveFieldMapperTest.java
```diff
@@ -41,7 +41,7 @@
 
     TestingStatementContext ctx = new TestingStatementContext(Collections.emptyMap());
 
-    ReflectiveFieldMapper<SampleBean> mapper = new ReflectiveFieldMapper<SampleBean>(SampleBean.class);
+    FieldMapper<SampleBean> mapper = new FieldMapper<SampleBean>(SampleBean.class);
```
qualidafial: Meh: may as well rename this test class too

core/src/main/java/org/jdbi/v3/util/bean/ColumnNameMappingStrategy.java
qualidafial: Column name mapping is doing two things: mapping the name, and normalizing (lowercasing) the name for comparison. We should separate the normalize step out of the name mapping.

stevenschlansker:
So, the reason I changed the normalize and mapping into the same step and made it cached rather than eager:

Not all mappings are necessarily one to one, for example we have our snake case mapper which could in theory map a Java `someField` into `some_field` and `Some_Field` (and other varied casing)

So at the very least it would have to be a `Function<String, Set<String>>`.

But this is also limiting -- what if you want to base it on `String#equalsIgnoreCase`?  There is no fixed `Set` of possible results, all you can do is try it and see if it matches.

So adding the possibility to implement such strategies is why I moved the evaluation to mapping time, and made the interface a general "do these names match?"

Make sense?  I'm open to changing it back, but I think this is a useful generalization.
I think the cost will also be very low -- mapping the first row may be somewhat slower, but by the second use it should be fully warmed up and I believe the overhead here will be entirely acceptable.

qualidafial:
I'm convinced that there's a simpler design staring me in the face, but I can't see it yet. Let's make `ColumnNameMappingStategy` private for now, and give ourselves time to learn and possibly redesign before we commit to an API for v3.

Adding snake-case support by default should satisfy 99% of what people really want anyway.

stevenschlansker:
OK, unfortunately it cannot be made private, otherwise it has to move back into the main package.  But I slapped a TODO on it and am totally open to reconsidering before release.

PR #215, opened 2015-11-01
Add a bridge for SqlObject API

arteam:
In JDBI 2 there is the possibilty to create proxies for SqlObject interfaces via the `DBI` and `Handle` class instances. Actually, it's very convienent and even JDBI official docs mentions this as the standard way for using the SqlObject API.

This API was removed in JDBI 3, because the JDBI core and the SQLObject API are now in different modules.

This commit adds a bridge from the JDBI core to the SQLObject API. It uses method handles to dynamically call the methods of the `SqlObjectBuilder` class. It changes some tests to show that the bridge to `SqlObjectBuilder` in the JDBI core actually works.

Also it fixes a broken link in the `SqlObjectBuilder` javadoc.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/SqlObjectBuilderBridge.java
+++ b/core/src/main/java/org/jdbi/v3/SqlObjectBuilderBridge.java
@@ -1,0 +1,115 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3;
+
+import org.jdbi.v3.exceptions.DBIException;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+
+/**
+ * A bridge to the <b>SqlObjectBuilder</b> class in the <b>jdbi3-sqlobject</b> module.
+ */
+class SqlObjectBuilderBridge {
+
+    private static final String SQL_OBJECT_BUILDER_CLASS = "org.jdbi.v3.sqlobject.SqlObjectBuilder";
+
+    private static final MethodHandle ATTACH;
+    private static final MethodHandle OPEN;
+    private static final MethodHandle ON_DEMAND;
+    private static final MethodHandle CLOSE;
+
+    static {
+        Class<?> sqlObjectBuilder = null;
+        try {
+            sqlObjectBuilder = Class.forName(SQL_OBJECT_BUILDER_CLASS);
+        } catch (ClassNotFoundException ignore) {
+        }
+
+        if (sqlObjectBuilder != null) {
+            try {
+                MethodHandles.Lookup lookup = MethodHandles.lookup();
+                ATTACH = lookup.findStatic(sqlObjectBuilder, "attach",
+                        MethodType.methodType(Object.class, Handle.class, Class.class));
+                OPEN = lookup.findStatic(sqlObjectBuilder, "open",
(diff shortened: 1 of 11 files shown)


qualidafial:
@stevenschlansker What was the design intent in separating out the SQL object APIs? Without knowing the motivation for extracting SQL objects, my impression is that this change might contradict that motivation.

Aside: Could we rename `SqlObjectBuilder` to `SqlObject` or perhaps `SqlObjects`?

stevenschlansker:
Really the only reason for the separation is because it brings in dependencies related to code generation (`cglib`) that are otherwise unnecessary, and because excepting for these factories you identified it is totally standalone code.  So it seemed like a natural choice to factor out first.

Idea: we could avoid a lot of the `MethodHandle` reflective lookup if the code used an interface / implementation pair, and maybe a `ServiceLoader` to find it if available.  It's not that the current approach is wrong, I am just wondering if it might be cleaner / clearer code.

arteam:
That was my first idea. The problem is that `SqlObjectBuilder` has only static method, which makes impossible to create a service interface for accessing it.

stevenschlansker:
Totally open to changing that, I am happy to reconsider the entire API if we come up with something better :)

arteam:
I don't think that this is really worth of refactoring.

* `SqlObjectBuilder` with static method is used in 119 places in JDBI tests. If we refactor it, we should change all these tests;
* `SqlObjectBuilder`doesn't have any state. So there is no difference between a new object with methods and an utility class with static methods;
* We will need to keep a link to an instance of `SqlObjectBuilder` in `DBI` and `BasicHandle`.

I prefer to keep this change small with the aim to keep compatability with JDBI 2. I believe the approach with method handles, despite being not realy idiomatic, is quite reasonable and practical.

core/src/main/java/org/jdbi/v3/SqlObjectBuilderBridge.java
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    public static <T> T open(DBI dbi, Class<T> sqlObjectType) {
+        try {
+            check(OPEN);
+            return (T) OPEN.invoke(dbi, sqlObjectType);
+        } catch (Throwable t) {
+            throw new RuntimeException("Error during `SqlObjectBuilder.open` invocation", t);
```
stevenschlansker: this will wrap e.g. `UnableToOpenConnectionException` or other `DbiException` subclasses

stevenschlansker:
Fair enough.  We can merge this for now and maybe I'll submit a followup later if I come up with something nice :)

Only complaint is that depending on which method you use, you could end up getting different exceptions thrown -- `catch (Throwable t)` is a bad pattern IMO.

Can we take care to only catch checked exceptions, and propagate e.g. `Error` and `RuntimeException` subclasses without transformation?

arteam:
`MethodHandle` throws `Throwable`, so we are forced to catch it. But you are right about runtim exceptions, we should propogate them. I pushed a new version, please review.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestNewApiOnDbiAndHandle.java
```diff
@@ -90,6 +91,24 @@ public void testAttach() throws Exception
         assertEquals("Diego", spiffy.findNameById(2));
     }
 
+    @Test(expected = UnableToObtainConnectionException.class)
+    public void testCorrectExceptionIfUnableToConnectOnDemand(){
+        DBI.create("jdbc:mysql://localhost/test", "john", "scott")
```
stevenschlansker: This assumes `localhost` doesn't run mysql

core/src/main/java/org/jdbi/v3/SqlObjectBuilderBridge.java
```diff
@@ -0,0 +1,115 @@
...
+        } catch (Throwable t) {
+            throw new RuntimeException("Error during `SqlObjectBuilder.onDemand` invocation", t);
+        }
+    }
+
+    public static void close(Object sqlObject) {
+        try {
+            check(CLOSE);
+            CLOSE.invoke(sqlObject);
+        } catch (RuntimeException e) {
```
stevenschlansker: This will wrap subclasses of `Error`

stevenschlansker:
Merging with some minor fixups.

PR #217, opened 2015-11-02
Add Plugin System to JDBI

stevenschlansker:
hi @qualidafial @brianm @arteam I am thinking about how we might make it easy to extend jdbi with a plugin like system.  Already, we have some obvious use cases -- Guava support without linking directly against Guava, SQLObject hooking into core, etc.

I have made a test branch here which adds support for automatic discovery of `JdbiPlugin`s via the classpath by a `ServiceLoader`.  This is little more than an idea at this point, but there are two in-tree use cases that are already interesting --
* Guava plugin can register itself with all DBI instances just by being present on classpath
* `HandleTrackerDBI` used in a test case can be a composable plugin, rather than needing to subclass `DBI`

What do you guys think of going down this direction?  I think it would be entirely optional, and the automatic registration should be used only for plugins which are transparent if not used.  For example adding support for custom types (Guava, joda-time) is OK as an automatic plugin, but replacing default transaction semantics would not be.

The diff is cluttered a bit because I converted DBI constructors to factory methods.  I'll try to annotate the interesting bits.


core/src/main/java/org/jdbi/v3/DBI.java
```diff
@@ -62,28 +65,38 @@
...
-     * @param connectionFactory PrvidesJDBC connections to Handle instances
+     * @param connectionFactory Prvides JDBC connections to Handle instances
      */
-    public DBI(ConnectionFactory connectionFactory)
-    {
-        assert connectionFactory != null;
-        this.connectionFactory = connectionFactory;
+    public static DBI create(ConnectionFactory connectionFactory) {
+        DBI dbi = new DBI(connectionFactory);
+        dbi.installPlugins();
```
stevenschlansker: By default, install all classpath plugins.

core/src/main/java/org/jdbi/v3/DBI.java
```diff
@@ -114,9 +136,37 @@ public DBI(final String url, final Properties props)
...
+    public DBI installPlugin(JdbiPlugin plugin)
+    {
+        plugins.add(plugin);
+        return this;
+    }
+
+    public DBI resetPlugins()
     {
-        this(() -> DriverManager.getConnection(url, username, password));
+        plugins.clear();
```
stevenschlansker: but you can get rid of them if you want total control

core/src/main/java/org/jdbi/v3/DBI.java
```diff
@@ -198,6 +248,7 @@ public Handle open()
                                        MappingRegistry.copyOf(mappingRegistry),
                                        foreman.createChild(),
                                        collectorFactoryRegistry.createChild());
+            plugins.forEach(p -> p.customizeHandle(h));
```
stevenschlansker: handle customization

core/src/main/java/org/jdbi/v3/spi/JdbiPlugin.java
```diff
@@ -0,0 +1,20 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.spi;
+
+import org.jdbi.v3.Handle;
+
+public interface JdbiPlugin {
+    default void customizeHandle(Handle handle) { }
```
stevenschlansker: everything is a default method, so we can add functionality later without breaking

guava/src/main/java/org/jdbi/v3/guava/GuavaPlugin.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package org.jdbi.v3.guava;
+
+import org.jdbi.v3.Handle;
+import org.jdbi.v3.spi.JdbiPlugin;
+
+public class GuavaPlugin implements JdbiPlugin {
+    @Override
+    public void customizeHandle(Handle handle) {
+        handle.registerCollectorFactory(GuavaCollectors.factory());
```
stevenschlansker: simplest useful plugin

guava/src/main/resources/META-INF/services/org.jdbi.v3.spi.JdbiPlugin
```diff
@@ -0,0 +1 @@
+org.jdbi.v3.guava.GuavaPlugin
```
stevenschlansker: easy to register with the `ServiceLoader`

guava/src/test/java/org/jdbi/v3/guava/TestGuavaCollectors.java
```diff
@@ -70,8 +70,6 @@ public void addData() {
             db.getSharedHandle().execute("insert into something(name, intValue) values (?, ?)", Integer.toString(i), i);
             expected.add(i);
         }
-
-        db.getSharedHandle().registerCollectorFactory(GuavaCollectors.factory());
```
stevenschlansker: no need to register by hand now

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestOnDemandSqlObject.java
```diff
@@ -268,19 +256,13 @@ public Something map(int index, ResultSet r, StatementContext ctx) throws SQLExc
         }
     }
 
-    public static class HandleTrackerDBI extends DBI
+    static class HandleTracker implements JdbiPlugin
```
stevenschlansker: now this can be composed in, rather than relying on subclassing

core/src/main/java/org/jdbi/v3/DBI.java
qualidafial: Minor nit: could be `plugins.forEach(JdbiPlugin::customizeHandle)`

stevenschlansker: If only... You have to bind 'h' :)

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindAutomaticNames.java
```diff
@@ -13,52 +13,41 @@
...
-
-    @Override
-    public void tearDown() throws Exception
-    {
-        handle.execute("drop table something");
-        handle.close();
+        handle = db.getSharedHandle();
     }
 
+    @Test
```
qualidafial: wut

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestMapBinder.java
```diff
@@ -24,33 +26,33 @@
...
     private Handle handle;
 
-    @Override
+    @Before
     public void setUp() throws Exception
     {
         JdbcDataSource ds = new JdbcDataSource();
         ds.setURL("jdbc:h2:mem:test");
-        dbi = new DBI(ds);
+        dbi = DBI.create(ds);
```
qualidafial: Should use `H2DatabaseRule` here too.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestMixinInterfaces.java
```diff
@@ -42,7 +42,7 @@ public void setUp() throws Exception
     {
         JdbcDataSource ds = new JdbcDataSource();
         ds.setURL(String.format("jdbc:h2:mem:%s;MVCC=TRUE", UUID.randomUUID()));
-        dbi = new DBI(ds);
+        dbi = DBI.create(ds);
```
qualidafial: Should use H2DatabaseRule

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestNewApiOnDbiAndHandle.java
```diff
@@ -40,7 +40,7 @@ public void setUp() throws Exception
     {
         JdbcDataSource ds = new JdbcDataSource();
         ds.setURL("jdbc:h2:mem:" + UUID.randomUUID());
-        dbi = new DBI(ds);
+        dbi = DBI.create(ds);
```
qualidafial: Should use H2DatabaseRule

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestPositionalBinder.java
```diff
@@ -17,39 +17,33 @@
...
     @Before
     public void setUp() throws Exception {
-        JdbcDataSource ds = new JdbcDataSource();
-        ds.setURL("jdbc:h2:mem:" + UUID.randomUUID());
-        DBI dbi = new DBI(ds);
-        handle = dbi.open();
+        handle = db.getSharedHandle();
         somethingDao = SqlObjectBuilder.attach(handle, SomethingDao.class);
 
+        handle.execute("drop table something");
```
qualidafial: If this test is isolated, why should there be any tables to drop?

core/src/main/java/org/jdbi/v3/DBI.java
qualidafial: This is the only thing I object to. I'm not a fan of Spring-style magic--by which I mean stuff just happens because a JAR is on the classpath (or sometime next year, in the Java 9 module graph).

JDBI has thus far done a great job of keeping the magic "visible." I would rather force users to make a single call to `DBI.installPlugins()` so that the ServiceLoader stuff is truly opt-in.

qualidafial:
This is a fantastic addition! :+1:

stevenschlansker:
Fixed up most of the comments, everything else can be fixed in post ;)

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestPositionalBinder.java
stevenschlansker: the rule creates it automagically, and this test uses a conflicting table definition, so we drop and recreate it.  not the best, but should be fixed as a separate PR

arteam:
This looks like a great framework for handle customizing. It could useful for some specific database features or Java containers.

stevenschlansker:
Yeah!  My thoughts exactly.

brianm:
I realize I am late to the party, but I much prefer explicit registration rather than "behavior magically changes because something was added to the classpath."

core/src/main/java/org/jdbi/v3/spi/JdbiPlugin.java
```diff
@@ -0,0 +1,20 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.spi;
+
+import org.jdbi.v3.Handle;
+
+public interface JdbiPlugin {
+    default Handle customizeHandle(Handle handle) { return handle; }
```
brianm: Customize statement?

brianm: Customize DBI?

brianm: I realize customizing a statement can be done by providing a wrapper handle, but that is kind of inconvenient compared to making it first class on the plugin. It also implies separate methods for customizing queries, statements, batches, etc, so as to handle all the downcast cases from statement correctly.

guava/src/main/java/org/jdbi/v3/guava/GuavaPlugin.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package org.jdbi.v3.guava;
+
+import org.jdbi.v3.Handle;
+import org.jdbi.v3.spi.JdbiPlugin;
+
+public class GuavaPlugin implements JdbiPlugin {
+    @Override
+    public Handle customizeHandle(Handle handle) {
+        handle.registerCollectorFactory(GuavaCollectors.factory());
```
brianm: I would expect this to also register handlers for passing in Guava's Optional<?> types for binding (ie, if none then bind null, if some then bind value).

core/src/main/java/org/jdbi/v3/spi/JdbiPlugin.java
stevenschlansker: Yes, both of these things are reasonable to add.

stevenschlansker:
Yeah, the merged version removes the magic auto-registration unless explicitly asked.

guava/src/main/java/org/jdbi/v3/guava/GuavaPlugin.java
stevenschlansker: Hm, could add that, although at this point Guava's optional type should be discouraged for the Java 8 version :)

PR #219, opened 2015-11-02
Replace CloseInternal with CloseMe, since they do the same thing

stevenschlansker:
Note that this breaks SqlObjects that have a close() method that do not implement AutoCloseable
or subinterfaces, but I think that's okay.  There's no downside to pulling in the interfaces.


qualidafial:
What about just adding a requirement that SQL Object classes implement `AutoCloseable`? e.g.

    <T> T SqlObjectBuilder.attach(Class<T extends AutoCloseable>)

qualidafial:
:+1:

stevenschlansker:
Unfortunately I don't think that fixes the problem -- a subinterface narrowing the thrown exception types means that implementations from the supertype are not considered.  For example, a method
```
@Override
void close() throws Exception {
  throw new Exception();
}
```

is compatible to `AutoCloseable` but not `Closeable`.  So I think that doing so won't actually simplify the code?

I think requiring SqlObjects implement `Closeable` would be a reasonable requirement, or maybe we make a `SqlObject` interface to just be more explicit.  But let's handle that as a new feature separately.

PR #220, opened 2015-11-03
Add jdbi3-jpa artifact

qualidafial:
Supports `@Entity`, `@MappedSuperclass`, `@Column`, and `@Transient` annotations.

Copied [with permission](https://github.com/shakiba/jdbi-annotation/issues/3) from shakiba/jdbi-annotation, which is already licensed under Apache 2.0.


shakiba:
+1

qualidafial:
@stevenschlansker ping

jpa/src/main/java/org/jdbi/v3/jpa/AnnoClass.java
```diff
@@ -0,0 +1,86 @@
...
+    public static <C> AnnoClass<C> get(Class<C> clazz) {
+        return (AnnoClass<C>) cache.computeIfAbsent(clazz, AnnoClass::new);
+    }
+
+    private final List<AnnoMember> setters = new ArrayList<>();
+    private final List<AnnoMember> getters = new ArrayList<>();
+
+    private AnnoClass(Class<C> clazz) {
+        if (logger.isDebugEnabled()) {
+            logger.debug("init " + clazz);
```
stevenschlansker: It would be more idiomatic to write `logger.debug("init {}", class);` and toss the `if()` block.

jpa/src/main/java/org/jdbi/v3/jpa/AnnoClass.java
```diff
@@ -0,0 +1,86 @@
...
+            if (member.getParameterTypes().length == 1) {
+                setters.add(new AnnoMember(clazz, member));
+            } else if (member.getParameterTypes().length == 0) {
+                getters.add(new AnnoMember(clazz, member));
+            }
+        }
+    }
+
+    public List<AnnoMember> setters() {
+        return setters;
```
stevenschlansker: these public getters return unprotected mutable `List`s, which offends my sensibilities.

jpa/src/main/java/org/jdbi/v3/jpa/AnnoClass.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+        for (Method member : clazz.getDeclaredMethods()) {
+            if (member.getAnnotation(Column.class) == null) {
+                continue;
+            }
+            if (member.getParameterTypes().length == 1) {
+                setters.add(new AnnoMember(clazz, member));
+            } else if (member.getParameterTypes().length == 0) {
+                getters.add(new AnnoMember(clazz, member));
+            }
```
stevenschlansker: `else { throw new IllegalStateException("..."); }` or at least log a warning?

jpa/src/main/java/org/jdbi/v3/jpa/AnnoClass.java
```diff
@@ -0,0 +1,86 @@
...
+            if (member.getAnnotation(Column.class) != null) {
+                setters.add(new AnnoMember(clazz, member));
+                getters.add(new AnnoMember(clazz, member));
+            }
+        }
+        for (Method member : clazz.getDeclaredMethods()) {
+            if (member.getAnnotation(Column.class) == null) {
+                continue;
+            }
+            if (member.getParameterTypes().length == 1) {
```
stevenschlansker: what happens here if a method is overridden?  seems like you're going to get duplicate `AnnoMember`s in your setters / getters?

jpa/src/main/java/org/jdbi/v3/jpa/AnnoClass.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public List<AnnoMember> setters() {
+        return setters;
+    }
+
+    public List<AnnoMember> getters() {
+        return getters;
+    }
+
+    private static Logger logger = LoggerFactory.getLogger(AnnoClass.class);
```
stevenschlansker: should be final

jpa/src/main/java/org/jdbi/v3/jpa/AnnoMapper.java
```diff
@@ -0,0 +1,80 @@
...
+import java.sql.SQLException;
+
+public class AnnoMapper<C> implements ResultSetMapper<C> {
+
+    private final Class<C> clazz;
+    private final AnnoClass<C> annos;
+
+    public static boolean accept(Class<?> clazz) {
+        if (logger.isDebugEnabled()) {
+            logger.debug("accept " + clazz);
```
stevenschlansker: not sure what this debug is here for, if we want more info about the mapping process maybe we should improve the logging in core instead

jpa/src/main/java/org/jdbi/v3/jpa/AnnoMapper.java
```diff
@@ -0,0 +1,80 @@
...
+            Constructor<C> constructor = clazz.getDeclaredConstructor();
+            constructor.setAccessible(true);
+            obj = constructor.newInstance();
+            for (AnnoMember annoMember : annos.setters()) {
+                annoMember.write(obj, get(annoMember, rs, ctx));
+            }
+        } catch (RuntimeException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
stevenschlansker: should this be a subclass of `DBIException`?

jpa/src/main/java/org/jdbi/v3/jpa/AnnoMember.java
```diff
@@ -0,0 +1,105 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.persistence.Column;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+class AnnoMember {
+
+    private Column column;
```
stevenschlansker: all these should be `final`?

jpa/src/main/java/org/jdbi/v3/jpa/AnnoMember.java
```diff
@@ -0,0 +1,105 @@
...
+        return name;
+    }
+
+    public Class<?> getType() {
+        return type;
+    }
+
+    public Object read(Object obj) throws IllegalArgumentException,
+            IllegalAccessException, InvocationTargetException {
+        if (method != null) {
```
stevenschlansker: ick

jpa/src/main/java/org/jdbi/v3/jpa/AnnoMember.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    public void write(Object obj, Object value)
+            throws IllegalArgumentException, IllegalAccessException,
+            InvocationTargetException {
+        if (logger.isDebugEnabled()) {
+            logger.debug("write" + clazz + "/" + name + "/" + type + "/" + value);
+        }
+        if (method != null) {
+            method.setAccessible(true);
```
stevenschlansker: could do `setAccessible` once in constructor?

jpa/src/test/java/org/jdbi/v3/jpa/Something.java
```diff
@@ -0,0 +1,44 @@
...
+    private Something() {
+    }
+
+    public Something(int id, String name, Long value) {
+        super(id);
+        this.name = name;
+        this.value = value;
+    }
+
+    public String name() {
```
stevenschlansker: would be nice to test a method setter / getter pair and a subclass overriding the superclass

jpa/src/test/java/org/jdbi/v3/jpa/Something.java
```diff
@@ -0,0 +1,44 @@
...
+        return name;
+    }
+
+    public void name(String name) {
```
stevenschlansker: also would be nice to test that bean conventions work?  i.e. `setName`

stevenschlansker:
Looks reasonable, a couple of minor code quality issues to address.

jpa/src/main/java/org/jdbi/v3/jpa/AnnoMember.java
```diff
@@ -0,0 +1,105 @@
...
+            return method.invoke(obj);
+        }
+        if (field != null) {
+            field.setAccessible(true);
+            return field.get(obj);
+        }
+        // unreachable!
+        throw new RuntimeException("Reached unreachable!");
```
stevenschlansker: this 1) is not a `RuntimeException` 2) could be verified eagerly in the constructor

jpa/src/main/java/org/jdbi/v3/jpa/AnnoMember.java
```diff
@@ -0,0 +1,105 @@
...
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+class AnnoMember {
+
+    private Column column;
+    private String name;
+    private Field field;
+    private Method method;
```
stevenschlansker: these two could be generalized into a `Consumer<T>` and then set to `method::invoke` or `field::get` or something like that?

qualidafial: Probably not `Consumer` with all the exceptions they throw, but I'll look into alternatives.

jpa/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+
+    <parent>
+        <groupId>org.jdbi</groupId>
+        <artifactId>jdbi3-parent</artifactId>
+        <version>3.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jdbi3-jpa</artifactId>
+    <name>jDBI JPA Integration</name>
+
```
qualidafial: Should add a developers block here to acknowledge @shakiba's work on the source project this was all copied from.

jpa/src/main/java/org/jdbi/v3/jpa/AnnoMember.java
```diff
@@ -0,0 +1,103 @@
...
+            InvocationTargetException {
+        logger.debug("write {}/{}/{}/{}", clazz, name, type, value);
+        if (method != null) {
+            method.setAccessible(true);
+            method.invoke(obj, value);
+        }
+        if (field != null) {
+            field.setAccessible(true);
+            field.set(obj, value);
+        }
```
qualidafial: Found a problem. If the `@Column` annotation is on the getter method, then `method` will be the getter method. Which means the entity becomes read-only or write-only depending on which method you annotate.

qualidafial: Oops, nevermind. `AnnoClass` is taking care of this.

qualidafial: Although it does look like you have to annotate both the getter and setter for it to work two-way. I think we'd be better off using the bean descriptor to tie both the getter and setter together in the same AnnoMember.

stevenschlansker:
PS thanks a lot @shakiba for this contribution and allowing us to merge it into core!

shakiba:
@stevenschlansker Thank you and @qualidafial for reviewing, editing and merging my code!

jpa/src/main/java/org/jdbi/v3/jpa/AnnoMapper.java
qualidafial: Possibly. At the moment, BeanMapper is throwing IllegalArgumentException whenever a reflective call fails.

qualidafial:
@stevenschlansker Can you look at this again? I've done some fairly major refactoring, which I think should address all your comments.

jpa/src/main/java/org/jdbi/v3/jpa/BindJpaFactory.java
```diff
@@ -0,0 +1,41 @@
...
+                prefix = bind.value() + ".";
+            }
+
+            try {
+                JpaClass<?> jpaClass = JpaClass.get(arg.getClass());
+                for (JpaMember member : jpaClass.members()) {
+                    q.dynamicBind(member.getType(), prefix + member.getColumnName(), member.read(arg));
+                }
+            } catch (Exception e) {
+                throw new IllegalStateException(
```
stevenschlansker: Should this wrap subclasses of `RuntimeException`? Will end up wrapping e.g. `DbiException` and friends.

jpa/src/main/java/org/jdbi/v3/jpa/JpaClass.java
```diff
@@ -0,0 +1,167 @@
...
+    private static final Map<Class<?>, JpaClass<?>> cache = new ConcurrentHashMap<>();
+
+    public static <C> JpaClass<C> get(Class<C> clazz) {
+        return (JpaClass<C>) cache.computeIfAbsent(clazz, JpaClass::new);
+    }
+
+    private final List<JpaMember> members;
+
+    private JpaClass(Class<C> clazz) {
+        logger.debug("init {}", clazz);
```
stevenschlansker: this is somewhat redundant with the following log statement

jpa/src/main/java/org/jdbi/v3/jpa/JpaClass.java
```diff
@@ -0,0 +1,167 @@
...
+            }
+        }
+    }
+
+    private static void inspectFields(Class<?> clazz,
+                                      Map<String, JpaMember> members,
+                                      List<String> transients) {
+        for (Field member : clazz.getDeclaredFields()) {
+            if (member.getAnnotation(Transient.class) != null) {
+                transients.add(member.getName());
```
stevenschlansker: Maybe this is unlikely, but it looks like if a super class has a transient property, and is overridden by a subclass that has the property non-transient, we could lose it?

jpa/src/main/java/org/jdbi/v3/jpa/JpaMapper.java
```diff
@@ -0,0 +1,74 @@
...
+    public C map(int i, ResultSet rs, StatementContext ctx) throws SQLException {
+        C obj;
+        logger.debug("map {}", clazz);
+        try {
+            Constructor<C> constructor = clazz.getDeclaredConstructor();
+            constructor.setAccessible(true);
+            obj = constructor.newInstance();
+            for (int colIndex = rs.getMetaData().getColumnCount(); colIndex >= 1; colIndex--) {
+                String columnLabel = rs.getMetaData().getColumnLabel(colIndex);
+                JpaMember member = jpaClass.lookupMember(columnLabel);
```
stevenschlansker: This could use the property mapping strategy from #212

jpa/src/main/java/org/jdbi/v3/jpa/JpaMapper.java
```diff
@@ -0,0 +1,74 @@
...
+                    Class<?> memberType = member.getType();
+                    ResultColumnMapper<?> columnMapper = ctx.columnMapperFor(memberType);
+                    Object value = columnMapper.mapColumn(rs, columnLabel, ctx);
+                    member.write(obj, value);
+                }
+            }
+        } catch (RuntimeException e) {
+            throw e;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
stevenschlansker: should there be something like an `UnableToMapResultException extends DbiException`

stevenschlansker:
This is amazing!  couple of small considerations, then :shipit:

jpa/src/main/java/org/jdbi/v3/jpa/JpaClass.java
qualidafial: Here I'm trying to match the admittedly confusing rules in JPA, in an effort to reduce pain for folks migrating away from JPA. See http://stackoverflow.com/questions/7016380/excluding-fields-from-embedded-properties-on-case-by-case-basis-with-hibernate

I think it makes sense to remove `@Transient` support for now. The rules about which annotation wins are not intuitive, and I'd rather not support it until we know there's a need for it.

jpa/src/main/java/org/jdbi/v3/jpa/JpaMapper.java
qualidafial: In JPA one generally provides the column name in the annotation e.g. `@Column("thing_id") long thingId;`. If the name is omitted, it defaults to the field/property name.

jpa/src/main/java/org/jdbi/v3/jpa/JpaMapper.java
qualidafial: Yep.

It should also catch when there's no column mapper for the member type and throw an informative exception.

qualidafial:
Heads up in case you didn't see it in the git logs: I've added Google auto-service as an optional dependency. auto-service contains an annotation processor which generates `META-INF/services/<service-class>` files automatically based on the presence of an `@AutoService(ServiceInterface.class)` annotation.

If the team doesn't like it I can take it out. It just looked like a nice way to avoid having to maintain those files by hand.

qualidafial:
Thanks again @shakiba

stevenschlansker:
Nifty, I'll check that out for the plugins...

shakiba:
@qualidafial Thank you too! When it is published as an artifact please let me know to add a deprecation notice to my project.

Issue #91, opened 2014-02-13
Use readOnly connections when possible


diwakergupta:
Specifically, if a connection is obtained automatically via `onDemand` objects, and callers are using the `SqlObject` API, then for `SqlQuery` calls, JDBI could call `setReadOnly(true)` on the underlying connection, which can hint the DB to use some optimizations.

http://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html#setReadOnly(boolean)

j4m4l:
@diwakergupta/ or anybody else, has implemented this feature? Or has any pointers about how to get this done, specifically implementing feedback given by sugis?

stevenschlansker:
I don't believe anyone has started implementing this feature yet.  PRs welcome!

divath:
It is easy to create @ReadOnly annotated with @SqlStatementCustomizingAnnotation that will support @SqlQuery annotated methods. However it would be more useful to annotate transaction methods.

Is it possible to annotate @Transaction annotated method with connection/statement customizing annotation? So if that @Transaction annotated method is called, then underlying connection will be modified?

stevenschlansker:
It might be possible, although honestly the annotation handling on SqlObjects is funky and could use some work.
If refactoring is necessary we'd probably prefer to implement this on the `jdbi3` branch coincidentally with annotation handling cleanup.

PR #222, opened 2015-11-06
After merging #91, adding a bare @Bind breaks binding

stevenschlansker:


qualidafial:
Is the `-parameters` compiler flag set?

stevenschlansker:
Yeah, the problem is that the `Binder` API doesn't pass in the parameter index.  But we need it to do the "named positional" binds.  There's a special case for "no `@Bind` annotation at all" but it doesn't cover a bare `@Bind`

qualidafial:
This is fixed in #327.

PR #224, opened 2015-11-09
Remove BooleanIntegerArgument in favor of just binding an int

stevenschlansker:
@brianm do you remember why this was done this way?  Seems overly complicated


stevenschlansker:
On second thought, should we just remove these `bindAsInt` methods entirely?  If you want to bind an int, you should just do that, rather than have JDBI magically do it for you?

qualidafial:
I'd say go ahead and remove the `bindAsInt` methods while you're at it.

Otherwise looks good. :shipit:

brianm:
I think it is fine to remove it.

PR #225, opened 2015-11-10
Implement polymorphic return types for SqlObject methods

### LINKED ISSUE #72: Support Inheritance with SqlObject Api Return Types
I have different kind of Users and I would like to use a common select for all of them. It seems that API does not support this.

``` java
   @SingleValueResult
   @SqlQuery("SELECT * FROM <type> limit 1")
   public <T extends User> Optional<T> select(@Define("type") Class<T> type);
```

or 

``` java
   @MapResultAsBean
   @SqlQuery("SELECT * FROM <type> limit 1")
   public  <T extends User> T select(@Define("type") Class<T> type);
```

Is there any way to do it?

stevenschlansker:
Fixes #72 
@cemo does this do what you want?

EDIT: ok this one needs to go on hold, it needs to consider what happens if we don't want bean mapping.  And reading into the code a bit, I think it'll require some refactoring of Query.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/CustomizingStatementHandler.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/CustomizingStatementHandler.java
@@ -30,7 +30,7 @@ import org.jdbi.v3.sqlobject.customizer.Bind;
 import org.jdbi.v3.sqlobject.customizer.SqlStatementCustomizerFactory;
 import org.jdbi.v3.sqlobject.customizer.SqlStatementCustomizingAnnotation;
 
-abstract class CustomizingStatementHandler implements Handler
+abstract class CustomizingStatementHandler<StatementType extends SqlStatement<StatementType>> implements Handler
 {
     private final List<FactoryAnnotationPair>           typeBasedCustomizerFactories   = new ArrayList<>();
     private final List<FactoryAnnotationPair>           methodBasedCustomizerFactories = new ArrayList<>();
@@ -109,15 +109,15 @@ abstract class CustomizingStatementHandler implements Handler
     public Object invoke(Object target, Object[] args, HandleSupplier hs) throws Exception {
         final Handle h = hs.getHandle();
         final String locatedSql = locateSql(h);
-        final SqlStatement<?> stmt = createStatement(h, locatedSql);
+        final StatementType stmt = createStatement(h, locatedSql);
         final SqlObjectStatementConfiguration cfg = stmt.getConfig(SqlObjectStatementConfiguration.class);
         configureReturner(stmt, cfg);
         applyCustomizers(stmt, args);
         return cfg.getReturner().get();
     }
 
-    abstract void configureReturner(SqlStatement<?> stmt, SqlObjectStatementConfiguration cfg);
-    abstract SqlStatement<?> createStatement(Handle handle, String locatedSql);
+    abstract void configureReturner(StatementType stmt, SqlObjectStatementConfiguration cfg);
+    abstract StatementType createStatement(Handle handle, String locatedSql);
 
     String locateSql(final Handle h)
     {
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/CustomizingStatementHandler.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/CustomizingStatementHandler.java
@@ -35,7 +35,6 @@ abstract class CustomizingStatementHandler implements Handler
     private final Class<?> sqlObjectType;
     private final Method method;
 
-    @SuppressWarnings("unchecked")
     CustomizingStatementHandler(Class<?> sqlObjectType, Method method)
     {
         this.sqlObjectType = sqlObjectType;
@@ -105,35 +104,22 @@ abstract class CustomizingStatementHandler implements Handler
 
     protected void applyCustomizers(SqlStatement<?> stmt, Object[] args)
     {
-        for (FactoryAnnotationPair pair : typeBasedCustomizerFactories) {
-            try {
+        try {
+            for (FactoryAnnotationPair pair : typeBasedCustomizerFactories) {
                 pair.factory.createForType(pair.annotation, sqlObjectType).apply(stmt);
             }
-            catch (SQLException e) {
-                throw new UnableToCreateStatementException("unable to apply customizer", e, stmt.getContext());
-            }
-        }
 
-        for (FactoryAnnotationPair pair : methodBasedCustomizerFactories) {
-            try {
+            for (FactoryAnnotationPair pair : methodBasedCustomizerFactories) {
                 pair.factory.createForMethod(pair.annotation, sqlObjectType, method).apply(stmt);
             }
-            catch (SQLException e) {
-                throw new UnableToCreateStatementException("unable to apply customizer", e, stmt.getContext());
-            }
-        }
 
-        if (args != null) {
             for (FactoryAnnotationParameterIndex param : paramBasedCustomizerFactories) {
-                try {
-                    param.factory
-                        .createForParameter(param.annotation, sqlObjectType, method, param.parameter, param.index, args[param.index])
-                        .apply(stmt);
-                }
-                catch (SQLException e) {
-                    throw new UnableToCreateStatementException("unable to apply customizer", e, stmt.getContext());
-                }
+                param.factory
+                    .createForParameter(param.annotation, sqlObjectType, method, param.parameter, param.index, args[param.index])
+                    .apply(stmt);
             }
+        } catch (SQLException e) {
+            throw new UnableToCreateStatementException("unable to apply customizer", e, stmt.getContext());
         }
     }
 
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/CustomizingStatementHandler.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/CustomizingStatementHandler.java
@@ -20,9 +20,12 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.extension.HandleSupplier;
 import org.jdbi.v3.core.statement.SqlStatement;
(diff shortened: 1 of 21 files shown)


cemo:
Yes :) :+1:

qualidafial:
Overall I like this idea. But I agree, it should use a more generic annotation that uses the mapping registry instead of being specific only to beans. `<T extends Foo> get(@MapResultAs Class<T> returnType, int id`)

qualidafial:
Actually, could we name the annotation `@MapTo` to make it agree with e.g. `Query.mapTo`?

qualidafial:
Consider renaming annotation, but otherwise :+1:

stevenschlansker:
`@MapTo` should now support arbitrary Types.  Rebased on top of 1 year (!) worth of changes.  More cleanup around SqlObject result handling, and some documentation.

stevenschlansker:
It's been a while and I made a few significant changes.  Can I get another review?

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
```diff
@@ -76,6 +93,8 @@
     @Override
     ResultIterator<T> iterator();
 
+    StatementContext getContext();
```
qualidafial: Does this refactor not work without this method? On the `ResultSetIterable` refactor, I took pains to avoid needing this method, so that it could be a `@FunctionalInterface`.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/CustomizingStatementHandler.java
```diff
@@ -103,37 +105,43 @@
...
+        final SqlObjectStatementConfiguration cfg = stmt.getConfig(SqlObjectStatementConfiguration.class);
+        configureReturner(stmt, cfg);
+        applyCustomizers(stmt, args);
+        return cfg.getReturner().get();
+    }
+
+    abstract void configureReturner(StatementType stmt, SqlObjectStatementConfiguration cfg);
+    abstract StatementType createStatement(Handle handle, String locatedSql);
+
+    protected String locateSql(final Handle h)
```
qualidafial: Meh: these methods could be package private

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/MapTo.java
```diff
@@ -0,0 +1,51 @@
...
+    class Factory implements SqlStatementCustomizerFactory {
+        @Override
+        public SqlStatementCustomizer createForParameter(Annotation annotation, Class<?> sqlObjectType, Method method, Parameter param, int index, Object arg) {
+            if (! (arg instanceof Type)) {
+                throw new UnsupportedOperationException("@MapTo must take a Type, got a " + arg.getClass().getName());
+            }
+            return s -> {
+                ResultReturner returner = ResultReturner.forMethod(sqlObjectType, method);
+                s.getConfig(SqlObjectStatementConfiguration.class).setReturner(
+                        () -> returner.result(((ResultSetIterable) s).mapTo((Type) arg)));
```
qualidafial: Semantically, this wants to be a `ConfigurerFactory`, but configurers are class- and method-level only at the moment.

I think this use case justifies extending configurers to parameter-level too.

This also has me wondering whether we ought to explore unifying configurer and statement customizers (in a separate PR)

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/SqlObjectStatementConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    public SqlObjectStatementConfiguration createCopy() {
+        return new SqlObjectStatementConfiguration(this);
+    }
+
+    void setReturner(Supplier<Object> returner) {
+        this.returner = returner;
+    }
+
+    Supplier<Object> getReturner() {
```
qualidafial: package private :+1:

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/SqlQuery.java
```diff
@@ -45,29 +44,26 @@
...
-            return new QueryHandler(sqlObjectType, method, ResultReturner.forMethod(sqlObjectType, method));
+            return new QueryHandler(sqlObjectType, method);
         }
     }
 
-    class QueryHandler extends CustomizingStatementHandler {
-        private final Class<?> sqlObjectType;
-        private final Method method;
+    class QueryHandler extends CustomizingStatementHandler<Query> {
         private final ResultReturner magic;
```
qualidafial: lol at this field name

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/SqlUpdate.java
```diff
@@ -88,11 +85,13 @@ public Handler buildHandler(Class<?> sqlObjectType, Method method) {
...
-            Update update = handle.getHandle().createUpdate(sql);
-            applyCustomizers(update, args);
-            return this.returner.apply(update);
+        Update createStatement(Handle handle, String locatedSql) {
+            return handle.createUpdate(locatedSql);
+        }
+
+        @Override
+        void configureReturner(Update u, SqlObjectStatementConfiguration cfg) {
+            cfg.setReturner(() -> returner.apply(u));
```
qualidafial: Feels like the if/else chain in the constructor wants to live in this method.

Maybe `SqlObjectStatementConfiguration.returner` wants to be a `Function<SqlStatement,Object>` instead?

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestPolymorphicReturn.java
```diff
@@ -0,0 +1,88 @@
...
+    public void testPolymorphicReturnSuperclass() throws Exception {
+        Sheep normalSheep = dao.get(Sheep.class, "Fluffy");
+        assertEquals("Fluffy", normalSheep.getName());
+    }
+
+    @Test
+    public void testPolymorphicReturnSubclass() throws Exception {
+        FlyingSheep flyingSheep = dao.get(FlyingSheep.class, "Fluffy");
+        assertEquals("Fluffy", flyingSheep.getName());
+        assertEquals(5, flyingSheep.getNumWings());
```
qualidafial: Alternatively, with AssertJ goodness:

```java
assertThat(flyingSheep).extracting("name", "numWings").containsExactly("Fluffy", 5);
```

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/UseRowMapper.java
```diff
@@ -13,23 +13,49 @@
...
+@SqlStatementCustomizingAnnotation(UseRowMapper.Factory.class)
 public @interface UseRowMapper
 {
     /**
      * The class implementing {@link RowMapper}. It must have a no-arg constructor.
      * @return the class of row mapper to use.
      */
     Class<? extends RowMapper<?>> value();
+
+    class Factory implements SqlStatementCustomizerFactory
```
qualidafial: Two things:

First, I love making this a meta-annotation that takes care of itself instead of handlers having to explicitly check fro it.

Second, I think this wants to be a `ConfigurerFactory`. for the same reasons `@MapTo` does.

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
stevenschlansker: Nope, was vestigial from an intermediate state.  I'll take it out.  It is weird that you can get the context at points before and after but not here though.

stevenschlansker: I guess I can pass them in as arguments instead.  Feels weird that we now require you to pass `(iterable, ctx)` when you get it from just so we can make the interface functional, especially when it actually does hold the context (exposed through the returned `Iterator`).

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestPolymorphicReturn.java
stevenschlansker: Oh, yeah, these were written before we went to assertj (!)

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/MapTo.java
stevenschlansker: https://github.com/jdbi/jdbi/issues/647

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/SqlQuery.java
stevenschlansker: I still wonder if it can be cleaned up more.  There's too many levels of indirection and change of control in the results creation process.  But it's probably better to ship this feature and do cleanup separately.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/SqlUpdate.java
stevenschlansker: But if you capture it instead, you get it as an `Update` or similar.  Writing it as a Function that way introduces the statement into your scope twice as different static types.  So you either capture the correctly typed one anyway, or do an ugly cast in your lambda...

And why would we move the if/else chain into here from the constructor?  Not seeing the advantage

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/UseRowMapper.java
stevenschlansker: Can't do it as is, needs access to the `SqlStatement` to capture it.
I guess this is where making it a `Function` would help.  Sucks to lose the concrete subtype of the statement though.

stevenschlansker: Let's sort this out as part of #647

stevenschlansker:
Integrated code feedback.  Punting the question about customizers vs configurers to a followup issue.  🚢

PR #226, opened 2015-11-10
Refactor JDBI factory types to take Type rather than Class

### LINKED ISSUE #205: Use `Type` instead of `Class` in mapper, argument, and collector factories
Per discussion on #164, try replacing the `Class` parameters with `Type` in `ResultSetMapperFactory`, `ResultColumnMapperFactory`, `ArgumentFactory`, and possibly `CollectorFactory`.

If it works, this should allow us to support a broader range of generic types in SQL Object interfaces. For full effect, the argument factory registry (Foreman) would be exposed through the `StatementContext` similar to the column mapper registry.

This would allow arguments to use composition as needed. e.g. an argument with a type `Optional<LocalDate>` would resolve to an `OptionalArgument` wrapped around a `LocalDateArgument`.

stevenschlansker:
Improves generics.  Fixes #205
WIP

### DIFF
--- a/core/src/main/java/org/jdbi/v3/BasicHandle.java
+++ b/core/src/main/java/org/jdbi/v3/BasicHandle.java
@@ -20,9 +20,12 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import com.fasterxml.classmate.TypeResolver;
+
 import org.jdbi.v3.exceptions.UnableToCloseResourceException;
 import org.jdbi.v3.exceptions.UnableToManipulateTransactionIsolationLevelException;
 import org.jdbi.v3.tweak.ArgumentFactory;
+import org.jdbi.v3.tweak.CollectorFactory;
 import org.jdbi.v3.tweak.ResultColumnMapper;
 import org.jdbi.v3.tweak.ResultSetMapper;
 import org.jdbi.v3.tweak.StatementBuilder;
@@ -30,7 +33,6 @@ import org.jdbi.v3.tweak.StatementCustomizer;
 import org.jdbi.v3.tweak.StatementLocator;
 import org.jdbi.v3.tweak.StatementRewriter;
 import org.jdbi.v3.tweak.TransactionHandler;
-import org.jdbi.v3.tweak.CollectorFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -51,6 +53,7 @@ class BasicHandle implements Handle
     private final Foreman                  foreman;
     private final TransactionHandler       transactions;
     private final Connection               connection;
+    private final TypeResolver             typeResolver;
 
 
     BasicHandle(TransactionHandler transactions,
@@ -62,7 +65,8 @@ class BasicHandle implements Handle
                 TimingCollector timingCollector,
                 MappingRegistry mappingRegistry,
                 Foreman foreman,
-                CollectorFactoryRegistry collectorFactoryRegistry)
+                CollectorFactoryRegistry collectorFactoryRegistry,
+                TypeResolver typeResolver)
     {
         this.statementBuilder = preparedStatementCache;
         this.statementRewriter = statementRewriter;
@@ -72,6 +76,7 @@ class BasicHandle implements Handle
         this.timingCollector = timingCollector;
         this.mappingRegistry = mappingRegistry;
         this.foreman = foreman;
+        this.typeResolver = typeResolver;
         this.globalStatementAttributes = new HashMap<String, Object>();
         this.globalStatementAttributes.putAll(globalStatementAttributes);
         this.collectorFactoryRegistry = collectorFactoryRegistry.createChild();
@@ -464,7 +469,12 @@ class BasicHandle implements Handle
     }
 
     @Override
-    public void registerCollectorFactory(CollectorFactory<?, ?> factory) {
+    public void registerCollectorFactory(CollectorFactory factory) {
         this.collectorFactoryRegistry.register(factory);
+    }
+
+    @Override
+    public TypeResolver getTypeResolver() {
+        return typeResolver;
     }
 }
--- a/core/src/main/java/org/jdbi/v3/Query.java
+++ b/core/src/main/java/org/jdbi/v3/Query.java
@@ -60,7 +60,7 @@ public class Query<ResultType> extends SQLStatement<Query<ResultType>> implement
     }
 
     /**
-     * Obtain a forward-only result set iterator. Note that you must explicitely close
+     * Obtain a forward-only result set iterator. Note that you must explicitly close
      * the iterator to close the underlying resources.
      */
     @Override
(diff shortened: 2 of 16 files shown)


core/src/main/java/org/jdbi/v3/Query.java
```diff
@@ -76,7 +76,7 @@
 
     @Override
     public <ContainerType> ContainerType collectInto(Class<ContainerType> containerType) {
-        return collect(getCollectorFactoryRegistry().createCollectorFor(containerType));
+        return collect(getCollectorFactoryRegistry().createCollectorFor(getHandle().getTypeResolver().resolve(containerType)));
```
qualidafial: I'm starting to think this `collectInto(Class)` method isn't buying us much. You can already use `collect(toSet())` and, because of the generics signature of `Collector.toSet()` is set up, we're guarantee to get back a `Set<ResultType>`.

core/src/main/java/org/jdbi/v3/tweak/CollectorFactory.java
```diff
@@ -30,7 +29,7 @@
      * @param type the object type
      * @return {@code true}, if accepts, otherwise {@code false}
      */
-    boolean accepts(Class<?> type);
+    boolean accepts(ResolvedType type);
```
qualidafial: I'd really like to avoid exposing this third party class in our API.

qualidafial:
So, it looks like we duplicated effort on this one. I've also got my own branch, in which I was targeting `java.lang.reflect.Type` directly instead of using `TypeResolver`.

core/src/main/java/org/jdbi/v3/tweak/CollectorFactory.java
qualidafial: We could still use `TypeResolver` internally e.g. `TypeResolver.resolve(Type)` to get a resolved type.

At the moment though it looks like we're just getting the erased type from the resolved type and using that. :-/

stevenschlansker:
Whoops, sorry about that @qualidafial.  I can hold off further work here if you'd like to publish your branch soon.

core/src/main/java/org/jdbi/v3/Query.java
stevenschlansker: That's very appealing.  But it removes the possibility of separating the declarative "I want this type" from the CollectorFactory's "how do I make this type" knowledge.  It also does not play well with SqlObjects, where you can have Classes annotated but not instances of a Collector.

The fact that Java's type system hurts us so badly here sucks.

core/src/main/java/org/jdbi/v3/tweak/CollectorFactory.java
```diff
@@ -38,5 +37,5 @@
      * @param type the actual type of the container
      * @return the {@link Collector}
      */
-    Collector<T, ?, R> newCollector(Class<R> type);
+    Collector<?, ?, ?> newCollector(ResolvedType type);
```
qualidafial: It seems like a step backward to lose the guarantee that the collector actually collects `T` into `R`.

core/src/main/java/org/jdbi/v3/Query.java
qualidafial: It does. Which is why I'd rather pivot on the fluent interfaces and only provide APIs which can make type safety guarantees.

I'd rather that collector factories were only used behind the scenes with SQL objects.

core/src/main/java/org/jdbi/v3/tweak/CollectorFactory.java
stevenschlansker: Yeah, makes sense to try to avoid that.  However, consider what you can do with a raw `Type` -- all you can do is figure out what it actually is via `instanceof` and downcast and hope.  We could then only provide JDK `Type` subclasses e.g. `Class` and `ParameterizedType` and each plugin would have to figure out how to introspect it itself.

Whereas if we just pass the `ResolvedType` it's clear where you go from there.

Not sure which concern is greater here.  I don't see a particularly good solution either way :(

qualidafial:
Yeah I should have communicated earlier that I was looking into it--far enough that I started to question whether it was worth it.

Maybe this is just self-fulfilling because JDBI has been built this way for a while, but it seemed like everything just devolved into "get the erased type and decide based on that." If that continues to the be the only use case, then I'm not sure it's worth switching to `Type`.

qualidafial:
I'll update my `Type` branch with the latest changes in the next day or so, then leave a link here so we can compare notes.

stevenschlansker:
I'm running into similar problems to you.  It's not clear exactly what we can accomplish here :/
Probably the coolest bit would be
```java
myQuery.mapTo(new GenericType<ContainingClass<SimpleBean>>(){}).collect(Collectors.toList())
```
and have that be type-safe.  This mirrors what you can do with SqlObject today, but cannot really do with fluent queries.

But is it worth it?

qualidafial:
A related discussion on this topic: https://github.com/jdbi/jdbi/pull/199#discussion_r42950693

Here's my attempt at Class -> Type before I got stuck: https://github.com/jdbi/jdbi/compare/jdbi3...qualidafial:jdbi3-types-over-classes

qualidafial:
My current thinking is that only factories and `sqlobject` codegen should work with `Type`. Everything is reflective so compile-time type safety is out the window anyway in that case.

By contrast, I think fluent APIs like Query should support the 99% case with helper methods like `mapTo(Class<T>)` for raw types `T`. But if you have a type with nested generic type like `Foo<Bar>` then you have to pass in a generic mapper e.g. `map(ResultSetMapper<Foo<Bar>>)`.

This lack of true type safety is why I'm opposed to `collectInto(Class<Container>)`, because `Container` doesn't tell the whole story. The `collect(Collector<T,?,R>)` method works and gives us full generic type safety because methods like e.g. `Collectors.toSet()` produce a `Collector<T, ?, Set<T>>`.

qualidafial:
With all that said, I think it would be great if the factories used a type token instead of `Type` directly, such as `GenericType`, or Guava's `TypeToken`. This would allow our factories and registries to make a few type safety guarantees:

    <T> ResultSetMapper<T> mapperFor(GenericType<T> type)

I'd just like to do it in a way that didn't use a third party class like `GenericType` in our own APIs.

qualidafial:
I'm trying out redoing factories using `TypeToken<T>` instead of just raw Type. I'm not very far, but so far I like it.

Guava is licensed Apache 2.0 so we can lift the class and copy it to JDBI if we want to avoid introducing an external dependency.

Another thing: with Java 8 I think it makes sense to reduce the `accept` and `build` methods in our factories to just a single `provide` method that returns an optional. For now I'm just adding the provide method as a default that calls the existing methods, but as I go forward I think it will make sense to just collapse it.

qualidafial:
Work in progress PR for TypeToken: https://github.com/jdbi/jdbi/pull/235

core/src/main/java/org/jdbi/v3/BasicHandle.java
```diff
@@ -51,6 +53,7 @@
     private final Foreman                  foreman;
     private final TransactionHandler       transactions;
     private final Connection               connection;
+    private final TypeResolver             typeResolver;
```
qualidafial: Is there a significant performance benefit to reusing this resolver?

qualidafial:
Close?

PR #227, opened 2015-11-11
Nuke Query*Munger classes, and just return the statement directly.

stevenschlansker:
The innerExecute leaves the Statement open anyway, so using callback
style just adds verbosity, two interfaces, and no gain.


stevenschlansker:
(recommend `?w=1`)

qualidafial:
LGTM.

PR #231, opened 2015-11-17
Fix statement leak from exceptional @GetGeneratedKeys

### LINKED ISSUE #230: Failing statements with @GetGeneratedKeys leaks cursors on Oracle
An `SqlObject` method with `@GetGeneratedKeys` that fails during execution (resulting in an `UnableToExecuteStatementException`) leaks a cursor on Oracle. After enough failures, the connection runs out of cursors, rendering it useless. Adding a call to `close` in the `catch` clause that maps `SQLException` to `UnableToExecuteStatementException` in `SQLStatement.internalExecute` fixes it, but I'm unsure of the implications of doing that.

stevenschlansker:
Fixes #230 
the fix is not great... but can be fixed better for JDBI3!

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/SQLStatement.java
+++ b/src/main/java/org/skife/jdbi/v2/SQLStatement.java
@@ -1329,6 +1329,11 @@ public abstract class SQLStatement<SelfType extends SQLStatement<SelfType>> exte
             timingCollector.collect(elapsedTime, getContext());
         }
         catch (SQLException e) {
+            try {
+                stmt.close();
+            } catch (SQLException e1) {
+                e1.printStackTrace(); // TODO: we have nowhere to log this, and suppressed exceptions were added in Java 7...
+            }
             throw new UnableToExecuteStatementException(e, getContext());
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/skife/jdbi/v2/SQLStatement.java
```diff
@@ -1329,6 +1329,11 @@ private String wrapLookup(String sql)
             timingCollector.collect(elapsedTime, getContext());
         }
         catch (SQLException e) {
+            try {
+                stmt.close();
+            } catch (SQLException e1) {
+                e1.printStackTrace(); // TODO: we have nowhere to log this, and suppressed exceptions were added in Java 7...
```
qualidafial: Why not SLF4J?

qualidafial:
One comment, then :shipit:

src/main/java/org/skife/jdbi/v2/SQLStatement.java
stevenschlansker: slf4j is not a mandatory dependency, so cannot be used or you risk linking errors:

```
        <dependency>
            <groupId>org.slf4j</groupId>
            <artifactId>slf4j-api</artifactId>
            <optional>true</optional>
        </dependency>
```

:cry:

PR #233, opened 2015-11-18
Throw exception at SqlObject creation time when using invalid return …

magnusr:
…type for @SqlBatch

Pretty much a copy-paste job based on Matteo Barbieri's work on compile-time errors for bad SqlUpdate return types.


stevenschlansker:
Thanks!

PR #243, opened 2015-12-04
Allows binding bean of inaccessible type (Option 1)

### LINKED ISSUE #242: @BindBean is incompatible with objects of non-public types

References:

gsmith85:
BindBeanFactory will now check the access modifier on arg's method and
use #setAccessible to allow it to be accessible from BindBeanFactory
whether or not this would typically be allowed via Reflection/ObjectAccess
rules.

This will allow for binding a bean, that has a type that is inaccessible
outside of its package, via one of its supertypes.

Adds test case demonstrating issue.

Fixes #242

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/BindBeanFactory.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/BindBeanFactory.java
@@ -20,12 +20,12 @@ import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 
 class BindBeanFactory implements BinderFactory
 {
     @Override
-    public Binder build(Annotation annotation)
-    {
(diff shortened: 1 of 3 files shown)


gsmith85:
Consider this to be Option 1 of 2 options to address this problem.  I will follow up with what I consider to be the pros and cons of each approach shortly.

src/main/java/org/skife/jdbi/v2/sqlobject/BindBeanFactory.java
```diff
@@ -20,12 +20,12 @@
...
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 
 class BindBeanFactory implements BinderFactory
 {
     @Override
-    public Binder build(Annotation annotation)
-    {
+    public Binder build(Annotation annotation) {
```
qualidafial: Careful with auto-formatting, it increases the chance of merge conflicts with no functional benefit.

PR #244, opened 2015-12-04
Allows binding bean of inaccessible type (Option 2)

### LINKED ISSUE #242: @BindBean is incompatible with objects of non-public types

gsmith85:
Changes the BinderFactory interface so that the type token of the
annotated parameter is accepted as a parameter on #build.

Changes BindBeanFactory to get the value of its argument using the
Method of the annotated parameter's type instead of the argument's type.
This allows for BindBeanFactory to access methods on args that are of
types that are inaccessible to the SqlObject package.  This changes the
behavior of the BindBeanFactory to behave in a way that is much more
consistent with Java's polymorphic/access mechanism.

Using BindBean with an iterable of a raw type will result in using
Object.  This may be considered a limiting factor or a feature.

Fixes #242

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/BindBeanFactory.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/BindBeanFactory.java
@@ -24,8 +24,7 @@ import java.lang.reflect.Method;
 class BindBeanFactory implements BinderFactory
 {
     @Override
-    public Binder build(Annotation annotation)
-    {
+    public Binder build(Annotation annotation, final Class<?> parameterClass) {
         return new Binder<BindBean, Object>()
         {
             @Override
@@ -40,7 +39,7 @@ class BindBeanFactory implements BinderFactory
                 }
 
                 try {
-                    BeanInfo infos = Introspector.getBeanInfo(arg.getClass());
+                    BeanInfo infos = Introspector.getBeanInfo(parameterClass);
                     PropertyDescriptor[] props = infos.getPropertyDescriptors();
                     for (PropertyDescriptor prop : props) {
                         Method readMethod = prop.getReadMethod();
@@ -52,8 +51,6 @@ class BindBeanFactory implements BinderFactory
                 catch (Exception e) {
                     throw new IllegalStateException("unable to bind bean properties", e);
                 }
-
-
             }
         };
     }
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/BindFactory.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/BindFactory.java
@@ -18,7 +18,7 @@ import java.lang.annotation.Annotation;
 class BindFactory implements BinderFactory
 {
     @Override
-    public Binder build(Annotation annotation)
+    public Binder build(Annotation annotation, Class<?> parameterClass)
     {
         Bind bind = (Bind) annotation;
         try {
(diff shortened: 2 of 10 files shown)


gsmith85:
Consider this to be Option 1 of 2 (see #243) options to address the related issue.

qualidafial:
This option is a non-starter for v2, since it modifies the parameters of an existing interface, which breaks backward compatibility with existing 3rd party binders.

For v3 however (which is in flight in the jdbi3 branch), this will probably be the preferred option going forward.

qualidafial:
Let's wait until #243 is merge before deciding how this will look in v3.

gsmith85:
The other strategy that's similar to this approach but *is* backwards compatible (in the sense we're not touching existing interfaces) would be adding a `ParameterTypeDependentBinder`, `ParameterTypeDependentBindingAnnotation` and then in `CustomizingStatementHandler` checking methods for this annotation as well as `BindingAnnotation`.  In the case of `ParameterTypeDependentBindingAnnotation`, we'd pass in the parameter type.

The downside of this is a bifurcated code path for creating binders and the need to add a couple additional types.  The upside is obviously backwards compatibility and allowing for developers to create their own parameter-type dependent bindings if they desired (you could actually add support for raw types for example if you wanted to or allow for duck-typing).

Using the type of the annotated parameter seems like the "right" way to do this, so if we agree that's the case then perhaps then we don't care about adding a little bit of additional code.

Any thoughts here @qualidafial ?

qualidafial:
What about just adding a type to the `@BindBean` annotation?

    selectStuff(@BindBean(type=PublicInterface.class) PublicInterface param)

gsmith85:
So this was an option I considered.  We could arugment `BindBean` with a parameter type:

```
public @interface BindBean
{
// ...
`Class<?> parameterType();`
}
```

And have access to the `parameterType` in our `BindBeanFactory`.  Which is great.

However, by doing so, we've just forced everyone who upgrades their version of JDBI2 to the latest to  change their `@BindBean` annotation to include a type token.  Required changes of JDBI consumers aside, it also seems a bit odd to have the API require the specification of a class token in duplicate to the type provided for the parameter itself.  It may have the one upside that for raw type iterables, you could specify a type that you expect all objects to be members of.

Of course we could provide a default for `parameterType` but `null` is not an option per the JLS.  With `null` we'd be able to say "use the type of arg as a default" and otherwise use parameterType ... but without it we'd have ambiguity of whether the parameterType was specified by the user or not.

If we wanted to use two additional optional fields:

```
public @interface BindBean
{
// ...
    boolean useTypeOverride() default false; 
    Class<?> typeOverride() default Object.class;
}
```

We could do `@BindBean(useTypeOverride=true, typeOverride=PublicInterface.class)`.

But then we're back at: should we ever be using `arg`'s type's method be used as the basis of invokation as a default?

TLDR: If we're looking for an extremely lightweight approach and saying "we'll solve this the right way in 3" we could add two fields to `@BindBean`. But my take is the bifurcated approach suggested above handles things the right way for the consumer of JDBI transparently and keeps the API simpler (no additional configuration).

qualidafial:
We already have a proven approach in `@SingleValueResult` which makes the type parameter optional:

    public @interface SingleValueResult
    {
        Class<?> value() default Default.class;
    
        final class Default{}
    }

gsmith85:
Okay, I actually was unaware it was possible to declare an inner class in an `interface`.  So I just learned something new and that simplifies things considerably!

The only question then is whether we're okay using arg's type by default.  It sounds like this is something that will be reconsidered in JDBI3, so if that's the focus right now, then perhaps it makes sense not to address this (and further it's possibly risky in that it could change behavior of `@BindBean` for users who upgrade to the latest 2.X release).

Give me your thoughts and I can get this code wrapped up.

qualidafial:
We've been using arg's runtime type for some time, and this is the first report I've seen of it being a problem.

I think the safer approach is to add the `type()` method to the annotation. This way the new behavior is opt-in and existing users will be unaffected.

PR #245, opened 2015-12-07
Allows binding bean of inaccessible type

### LINKED ISSUE #242: @BindBean is incompatible with objects of non-public types

gsmith85:
Changes BinderFactory so that it allows the specification of the
Annotation type via a type token.

Changes the BindBean annotation such that a type can be provided as an
override.  This override will be used instead of the provided
argument's type for the purpose of binding.  This allows for
BindBeanFactory to access methods on args that are of types that are
inaccessible to the SqlObject package.

Fixes #242

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/BindBean.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/BindBean.java
@@ -25,4 +25,14 @@ public @interface BindBean
 {
     String BARE_BINDING = "___jdbi_bare___";
     String value() default BARE_BINDING;
+
+    /**
+     * A type to use for resolving a bean's bindings via reflection.  If omitted, <code>argument.getClass()</code> is used.
+     *
+     * @return a type to use for resolving bindings as an alternative to the argument's type, or <code>Default.class</code> if
+     *   type is omitted.
+     */
+    Class<?> type() default Default.class;
+
+    class Default {};
 }
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/BindBeanFactory.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/BindBeanFactory.java
@@ -40,7 +40,11 @@ class BindBeanFactory implements BinderFactory
                 }
 
                 try {
-                    BeanInfo infos = Introspector.getBeanInfo(arg.getClass());
+                    Class<?> beanType = bind.type().equals(BindBean.Default.class)
+                        ? arg.getClass()
+                        : bind.type();
+
+                    BeanInfo infos = Introspector.getBeanInfo(beanType);
                     PropertyDescriptor[] props = infos.getPropertyDescriptors();
                     for (PropertyDescriptor prop : props) {
                         Method readMethod = prop.getReadMethod();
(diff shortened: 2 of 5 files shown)


src/main/java/org/skife/jdbi/v2/sqlobject/BindBean.java
```diff
@@ -25,4 +25,13 @@
 {
     String BARE_BINDING = "___jdbi_bare___";
     String value() default BARE_BINDING;
+
+    // This override allows for explicitly specifying the type used to invoke the bound bean's methods.  By default the type is
+    // inferred from the argument that the annotated method is called with.  Providing an override may be required in certain
+    // situations, where the annotated parameter type is accessible from the org.skife.jdbi.v2.sqlobject package but a subtype
+    // is not.  The result of attempting to bind an inaccessible subtype being used without providing this override is a
+    // java.lang.IllegalAccessException.
+    Class<?> typeOverride() default Default.class;
```
gsmith85: Possibly `beanType` as a name instead of `typeOverride`?  I could go either way and am open to suggestions that would make the purpose of this parameter more explicit.

qualidafial: I'd rather just name this `type()` for brevity at the call site.

Also, can we tighten up this comment, and convert it into a javadoc? e.g.

    /**
     * The bean class of the argument. If omitted, <code>argument.getClass()</code> is used.
     */

src/main/java/org/skife/jdbi/v2/sqlobject/BinderFactory.java
```diff
@@ -19,12 +19,12 @@
  * Factor for building {@link Binder} instances. This interface is used by
  * the {@link BindingAnnotation}
  */
-public interface BinderFactory
+public interface BinderFactory<T extends Annotation>
```
qualidafial: Can we move this generic parameter into a separate PR?

src/main/java/org/skife/jdbi/v2/sqlobject/BindBeanFactory.java
```diff
@@ -40,7 +39,11 @@ public void bind(SQLStatement q, BindBean bind, Object arg)
                 }
 
                 try {
-                    BeanInfo infos = Introspector.getBeanInfo(arg.getClass());
+                    Class<?> beanType = bind.typeOverride().equals(BindBean.Default.class)
+                        ? arg.getClass()
+                        : bind.typeOverride();
+
+                    BeanInfo infos = Introspector.getBeanInfo(beanType);
```
qualidafial: Have you confirmed that the new tests fail without this code change?

qualidafial:
@gsmith85 A few comments to take care of, then this should be good to merge.

src/main/java/org/skife/jdbi/v2/sqlobject/BindBeanFactory.java
gsmith85: I have confirmed this, I wrote the test before the implementation.

gsmith85:
I believe your changes have been addressed.  Let me know if you have any additional feedback!

qualidafial:
Looks great. Thanks for the contribution!

qualidafial:
@gsmith85 Heads up, you'll want to configure your name and contact email for future open source contributions:

    git config --global user.name "Greg Smith"
    git config --global user.email "your.email@example.com"

gsmith85:
Yeah I just noticed I didn't update the author information on this commit too.  

Sorry about that.  Not much we can do now as updating the author would change the sha of the commit IIRC.

stevenschlansker:
Cool.  This is a much better fix, thanks guys! :)

PR #247, opened 2015-12-10
Squashes checkstyle warnings

gsmith85:


stevenschlansker:
Thanks for the fixup here.  I'll merge this one in, but be advised that in general we are trying to keep cosmetic / stylistic fixups to the master branch to a minimum -- they end up causing large merge conflicts with our ongoing work on the `jdbi3` branch.  We're in general huge fans of cleaning up these warnings though, the effort is appreciated!

gsmith85:
Is it worth considering having the build fail if the check style fails?  I've found that's a pretty effective way to avoid style cleanup commits like this one.

stevenschlansker:
Absolutely!  I believe we have it enabled on the `jdbi3` branch already, and if not, certainly should :)
I don't think that's a goal for the 2.x train.

PR #249, opened 2015-12-22
Prove double-binding of arguments works, per mailing list thread

stevenschlansker:


evnm:
👍

Thanks for taking the time to verify that this works!

PR #250, opened 2015-12-26
Simplify BuiltInArgumentFactory using lambdas

stevenschlansker:
removes a ton of public API surface, at the cost of slightly
more complicated tests (which are arguably better, since they actually
test the behavior, rather than just checking the class of an Argument subtype)


stevenschlansker:
I don't know why Travis is unhappy with this, it's green here :disappointed:

stevenschlansker:
build fixed by #251

qualidafial:
Reviewing now. I like the look of that line count..

core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
```diff
@@ -90,30 +123,55 @@ public Argument build(Class<?> expectedType, T value, StatementContext ctx)
 
         // Enums must be bound as VARCHAR.
         if (expectedType.isEnum()) {
-            return new StringArgument(value.toString());
+            return STR_BUILDER.build(value.toString());
         }
 
         // Fallback to generic ObjectArgument
-        return new ObjectArgument(value);
+        return OBJ_BUILDER.build(value);
```
qualidafial: This method has some redundant logic when `expectedType == Object.class`.

Have a look at this class in [#235](https://github.com/qualidafial/jdbi/blob/jdbi3-type-tokens/core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java) for a simpler flow.

core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
```diff
@@ -90,30 +123,55 @@ public Argument build(Class<?> expectedType, T value, StatementContext ctx)
...
     }
 
-    private static class P
-    {
-        private final Constructor<?> ctor;
+    @FunctionalInterface
+    interface StatementBinder<T> {
+        void bind(PreparedStatement p, int index, T value) throws SQLException;
+    }
```
qualidafial: :+1: Elegant solution, taking advantage of the uniform method signatures in ResultSet.

core/src/test/java/org/jdbi/v3/TestForeman.java
```diff
@@ -13,51 +13,79 @@
...
     }
 
     @Test
-    public void testExplicitWaffling()
+    public void testWaffleLong() throws Exception
     {
-        final Foreman foreman = new Foreman();
+        stmt.setLong(1, 3);
+        replay(stmt);
+        foreman.waffle(Object.class, new Long(3L), null).apply(1, stmt, null);
```
qualidafial: In Mockito, stubbing an expected method call is a single statement:

    import static org.mockito.Mockito.verify;

    ...

    @Rule MockitoRule mockitoRule = MockitoJUnit.rule();

    @Mock PreparedStatement stmt;

    @Test public void testWaffleLong() {
      foreman.waffle(Object.class, new Long(3L), null).apply(1, stmt, null);
      verify(stmt).setLong(1, 3);
    }

core/src/test/java/org/jdbi/v3/TestForeman.java
```diff
@@ -13,51 +13,79 @@
...
+        foreman.waffle(short.class, (short) 2000, null).apply(2, stmt, null);
+    }
 
-        final Argument stringArgument = foreman.waffle(String.class, "I am a String!", null);
-        assertSame(StringArgument.class, stringArgument.getClass());
+    @Test
+    public void testExplicitWaffleString() throws Exception {
+        stmt.setString(3, I_AM_A_STRING);
+        replay(stmt);
+        foreman.waffle(String.class, I_AM_A_STRING, null).apply(3, stmt, null);
```
qualidafial: :+1: This is a big improvement to the tests.

core/src/test/java/org/jdbi/v3/TestQueries.java
```diff
@@ -13,11 +13,7 @@
  */
 package org.jdbi.v3;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
```
qualidafial: Thought wildcard imports were against the checkstyle rules?

qualidafial:
Looks good. A few comments to consider, then :shipit:

stevenschlansker:
Yeah, the line count drop (and file drop too!) made me feel really good :smile:

core/src/test/java/org/jdbi/v3/TestForeman.java
stevenschlansker: Nice, that definitely is better.  I guess I'm convinced enough to try Mockito out, the EasyMockRule isn't very impressive.

core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
stevenschlansker: I was honestly somewhat surprised when I realized I could do this!

core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
stevenschlansker: In fact, this duplication was even hiding a bug in plain sight!  Usage of `Enum#toString` vs `Enum#name`...

core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
qualidafial: Yep. They really did an amazing job with Java 8.

PR #252, opened 2015-12-27
Cache maven repo to speed builds on Travis

stevenschlansker:
seems to cut time of the `install` phase on Travis in half (67 -> 33 seconds) although that's a point measurement, not scientific.  saves bandwidth either way.


qualidafial:
LGTM

Issue #253, opened 2016-01-02
Connection.prepareStatement(String,int) results in invalid SQL in Prepared Batch when using pgjdbc


apjoseph:
`PreparedBatch.executeAndGeneratedKeys()` results in double `RETURNING` statements when used with stock pgjdbc drivers due to the use of  `Connection.prepareStatement()String,int)` [here ](https://github.com/jdbi/jdbi/blob/61a50d927e08b6f1bf993b5cb86fe7721abd5cd7/src/main/java/org/skife/jdbi/v2/PreparedBatch.java#L143-L147). `DefaultStatementBuilder` [is a bit more sensitive but still doesn't seem to account for an explicit `RETURNING` clause within the SQL](https://github.com/jdbi/jdbi/blob/295453016b05a56570aa4a58b46d311a0c0b4317/src/main/java/org/skife/jdbi/v2/DefaultStatementBuilder.java#L41-L56) For non batch statements, it is possible to work around this by just using a `Query` instead of an `Update` but there is no such option with `PreparedBatch`. 

Ideally `PreparedBatch` would have an additional `executeAndMapResultSet(Class<?>)` method which would use `Connection.prepareStatement(String)` and process the `ResultSet` using the registered `ResultSetMapperFactory`.

Example of failure from [related StackOverflow issue] (http://stackoverflow.com/questions/34560647/pg-jdbc-appends-returning-on-prepared-batch-wtih-existing-returning-clause):

```
PreparedBatch b = getHandle().prepareBatch("INSERT INTO death_star(id,exhaust_port) VALUES ( :id, :exhaust_port) RETURNING id;");

for (RebelPilot p : rebelPilots) {
    b.add().bind("id",p.getId()).bind("exhaust_port",p.getProtonTorpedo());
}

ResultSetMapper<Long> mapper = new IdMapper()
GeneratedKeys<Long> gk = b.executeAndGenerateKeys(mapper);
return gk.list()
```

When executed, the statement will be rewritten as the following malformed abomination:

`INSERT INTO death_star(id,exhaust_port) VALUES ( ?, ?) RETURNING id RETURNING *;`

I have also filed an [issue with pgjdbc to create a JDBC level option to prevent it from appending the extra `RETURNING *`](https://github.com/pgjdbc/pgjdbc/issues/488)

[...]

PR #254, opened 2016-01-02
Providing a list of column names returned in a prepared batch

arteam:
References to #253.

Sometimes it's desirable have ability to explicitly provide the names of the columns, which should returned to the client after a query execution. For example, when a row contains many columns, and many rows are returned in a batch response.

JDBC provides such functionality during preparing a statement by using the form: `Connection.preparedStatement(String query, String[] columnNames)`.

If a JDBC driver supports it, it will return only specified columns in a response. For example, the official PostgreSQL driver adds a `RETURNING` clause with a list of the specified columns to a query.

See this [SO discusstion](http://stackoverflow.com/questions/34560647/pg-jdbc-appends-returning-on-prepared-batch-wtih-existing-returning-clause).

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/PreparedBatch.java
+++ b/src/main/java/org/skife/jdbi/v2/PreparedBatch.java
@@ -25,6 +25,7 @@ import org.skife.jdbi.v2.tweak.StatementLocator;
 import org.skife.jdbi.v2.tweak.StatementRewriter;
 import org.skife.jdbi.v2.util.SingleColumnMapper;
 
+import java.sql.Connection;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.sql.Statement;
@@ -96,7 +97,7 @@ public class PreparedBatch extends SQLStatement<PreparedBatch>
      * @return the number of rows modified or inserted per batch part.
      */
     public int[] execute() {
-        return (int[]) internalBatchExecute(null);
+        return (int[]) internalBatchExecute(null, null);
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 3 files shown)


src/main/java/org/skife/jdbi/v2/PreparedBatch.java
```diff
@@ -118,7 +134,12 @@ public PreparedBatch define(final Map<String, ? extends Object> values)
         return executeAndGenerateKeys(new SingleColumnMapper<GeneratedKeyType>(mapper));
     }
 
-    private <Result> Object internalBatchExecute(QueryResultMunger<Result> munger) {
+    public <GeneratedKeyType> GeneratedKeys<GeneratedKeyType> executeAndGenerateKeys(ResultColumnMapper<GeneratedKeyType> mapper,
```
stevenschlansker: Does this invocation make sense with any number of columns other than 1?  Should we take a single `String` instead of a `String[]`?

src/main/java/org/skife/jdbi/v2/PreparedBatch.java
```diff
@@ -109,7 +110,22 @@ public PreparedBatch define(final Map<String, ? extends Object> values)
...
                         getContainerMapperRegistry());
             }
-        });
+        }, null);
+
+    }
+
+    @SuppressWarnings("unchecked")
+    public <GeneratedKeyType> GeneratedKeys<GeneratedKeyType> executeAndGenerateKeys(final ResultSetMapper<GeneratedKeyType> mapper,
+                                                                                     String[] columnNames) {
```
stevenschlansker: This could be `String...` and then varargs invocation would allow you to elide the `new String[] { ... }` dance, although maybe being explicit is better here.

stevenschlansker:
Couple of minor thoughts for your consideration, but OK to merge even if you decide to not make changes.

stevenschlansker:
Should add a note to the `RELEASE_NOTES` as well.

src/main/java/org/skife/jdbi/v2/PreparedBatch.java
arteam: I think it might. For example, one could want to get a generated id of a record along with the timestamp it was inserted (the timestamp is generated by the database). The JDBI spec does not prohibit such scenario and I don't see reasons why we should do this.

src/main/java/org/skife/jdbi/v2/PreparedBatch.java
arteam: I think the varargs declaration is quite fine. It provides a more terse API and works the same way internally as the explicit array declaration.

arteam:
@stevenschlansker, please take another look.

src/main/java/org/skife/jdbi/v2/PreparedBatch.java
stevenschlansker: I agree for the `ResultSetMapper` invocation, but this takes a `ResultColumnMapper` -- how are you to map the multiple columns?

stevenschlansker: Another thought, should this use a `SingleColumnMapper#byLabel`?  Or maybe that doesn't matter...

arteam: This change adds an overrided method that accepts `ResultSetMapper` as well as `ResultColumnMapper`. Users have a choice which one to use.  I've added a unit test that shows getting several column names.

stevenschlansker:
OK, good to merge after @apjoseph has a chance to green-light it.

PR #255, opened 2016-01-11
Add SqlObject support for returning Stream<T>

stevenschlansker:


sqlobject/src/main/java/org/jdbi/v3/sqlobject/ResultReturnThing.java
```diff
@@ -55,9 +56,13 @@ static ResultReturnThing forType(ResolvedMethod method)
                     "Method %s#%s is annotated as if it should return a value, but the method is void.",
                     method.getDeclaringType().getErasedType().getName(),
                     method.getName()));
-        } else if (return_type.isInstanceOf(ResultBearing.class)) {
+        }
+        else if (return_type.isInstanceOf(ResultBearing.class)) {
             return new ResultBearingResultReturnThing(method);
         }
+        else if (return_type.isInstanceOf(Stream.class)) {
```
stevenschlansker: Still want to clean up this ResultReturnThing pile, but we'll do that separately later.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestStream.java
```diff
@@ -0,0 +1,60 @@
...
+        Something one = new Something(3, "foo");
+        Something two = new Something(4, "bar");
+        Something thr = new Something(5, "baz");
+
+        Spiffy dao = SqlObjectBuilder.onDemand(db.getDbi(), Spiffy.class);
+        dao.insert(one);
+        dao.insert(thr);
+        dao.insert(two);
+
+        List<Something> results = dao.stream().collect(Collectors.toList());
```
qualidafial: `collect` is not documented to close the stream. The `close` method suggests wrapping streams in try-with-resource blocks.

stevenschlansker: that's going to lead to a lot of end users leaking resources, I bet :/

qualidafial:
One thing I just realized: returning streams from a DAO (or from a Query, for that matter) is likely to be misused and result in resource leaks:

    dao.streamThings()
        .filter(predicate)
        .map(mapFunction)
        .collect(toList());

The documentation should stress that users must always explicitly close these streams, and recommend using them inside try-with-resource blocks.

stevenschlansker:
You are 100% correct.

stevenschlansker:
@qualidafial I put that in the CHANGES document as a todo item.

qualidafial:
:shipit:

Issue #208, opened 2015-10-27
MappingRegistry and friends should use Optional types


stevenschlansker:
`MappingRegistry` `CollectorFactoryRegistry` and `Foreman` should return `Optional` types to let the caller decide how to handle errors.

qualidafial:
I would add `ResultSetMapperFactory`, `ResultColumnMapperFactory`, `ArgumentFactory`, and `CollectorFactory` to that list.

qualidafial:
Resolved in #259

PR #259, opened 2016-01-24
Registries and factories should use optional types

qualidafial:
Work in progress to address #208 

* [x] Coalesce ArgumentFactory methods and return optional
* [x] Coalesce ResultSetMapperFactory methods and return optional
* [x] Coalesce ResultColumnMapperFactory methods and return optional
* [x] Coalesce CollectorFactory methods and return optional
* [x] Foreman.waffle returns optional
* [x] MappingRegistry.columnMapperFor returns optional
* [x] MappingRegistry.mapperFor returns optional
* [x] CollectorFactoryRegistry.createCollectorFor returns optional
* [x] StatementContext.argumentFor return optional
* [x] StatementContext.columnMapperFor return optional
* [x] StatementContext.collectorFor return optional
* [x] Rename Foreman to ArgumentRegistry
* [x] Methods that return optional should be named with `find` prefix for consistency.
* [x] NamedArgumentFinder.find return optional
* [x] Binding.forName/forPosition return optional, rename with `find` prefix
* [x] Add `@FunctionalInterface` annotation as appropriate
* [x] Make order of factory precedence consistent between all registries.


qualidafial:
@stevenschlansker time for a code review?

core/src/main/java/org/jdbi/v3/ArgumentRegistry.java
```diff
@@ -0,0 +1,64 @@
...
+                        .flatMap(JdbiStreams::toStream))
+                .findFirst();
+    }
+
+    void register(ArgumentFactory argumentFactory)
+    {
+        factories.add(0, argumentFactory);
+    }
+
+    ArgumentRegistry createChild()
```
stevenschlansker: To me, calling this a 'child' implies that changes to the parent will affect it.  Which is not true.  Would `createCopy` be more accurate?

core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
```diff
@@ -125,22 +106,26 @@ public Argument build(Type expectedType, Object value, StatementContext ctx)
...
-            return ctx.argumentFor(nestedType, nestedValue);
+            return ctx.findArgumentFor(nestedType, nestedValue);
+        }
+
+        if (expectedType == Object.class) {
+            return Optional.of(OBJ_BUILDER.build(value));
         }
 
-        // Fallback to generic ObjectArgument
-        return OBJ_BUILDER.build(value);
```
stevenschlansker: This seems to be a behavior change -- previously, if `expectedType != Object.class` and no primitive mapper is found, it would bind with `setObject`.  Now it will not make a binding.
This was the final fallback in case everything else did not match.  Is this change intentional?

core/src/main/java/org/jdbi/v3/ArgumentRegistry.java
qualidafial: I like the `copyOf` method name already in `MappingRegistry`.

stevenschlansker: Even better.

stevenschlansker:
My only remaining concern is that a couple of formerly lazy operations are now eager (due to use of streams / mapping).
I doubt there is any actual measurable performance impact, so won't hold up the train based on that.

Consider the remaining note then :shipit:

core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
qualidafial: This was intentional.

Compare the old `Foreman.waffle` and the new `ArgumentRegistry.findArgumentFor`.

Foreman iterated the list once, and searched for a result as well as a fallback candidate in the same loop. Thus, BuildInArgumentFactory was always consulted last.

The new method is using streams (iterating once by exact argument type, then again if necessary by Object.class) and will return the first argument that comes out of the stream.

I added the if clause around OBJ_BUILDER, to prevent BuiltInArgumentMapper from returning a result on the first iteration, to allow each registered mapper a second chance to return a result for Object.class. A few tests actually depend on this.

qualidafial:
Streams are actually super lazy. All your intermediate operations are stacked up until you add 1. an intermediate operation that requires an intermediate state--like sorted; or 2. you use a terminal operation like `collect`, `reduce`, or `findFirst`.

core/src/main/java/org/jdbi/v3/ArgumentRegistry.java
```diff
@@ -0,0 +1,60 @@
...
+                // Search first for factories accepting specific argument type
+                factories.stream().flatMap(factory -> toStream(factory.build(expectedType, it, ctx))),
+                // Fall back to any factory accepting Object if necessary
+                factories.stream().flatMap(factory -> toStream(factory.build(Object.class, it, ctx))))
+                .findFirst();
+    }
+
+    void register(ArgumentFactory argumentFactory)
+    {
+        factories.add(0, argumentFactory);
```
qualidafial: Something I just noticed:
* With arguments, last registered factory wins, and built-in factory is last.
* With mappers, first registered factory wins, and built-in factory is last.
* With collectors, first registered wins, but built in factories are first (and thus cannot be overridden).

For consistency, we should pick one strategy for all registries.

MappingRegistry seems like the most logical choice to me.

@stevenschlansker, @brianm, any strong opinions?

arteam: > With arguments, last registered factory wins, and built-in factory is last

My 2 cents. In Dropwizard we actually depends on this order for processing optional JodaTime arguments. See [here](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/DBIFactory.java#L132).

qualidafial: We already have implementations for Guava and java.util optionals which arbitrary arguments, so you'll be able to throw those away. :)

I was kind of afraid of that. However since v3 is backward incompatible, this really is the best time to select a strategy and use it consistently. Whatever we change it to, we should document it and put it in the release notes and/or migration guide.

qualidafial: I beg your pardon, we do _not_ have an argument factory for Guava optional yet--just a collector implementation.

qualidafial: @arteam I'd be happy to take those joda-time mappers off your hands, if you'd care to submit a PR to the jdbi3 branch. ;)

stevenschlansker: :+1: to aligning the strategies here.
I think the "builtin wins last" is probably not controversial.
I can see good arguments for either of "first wins" or "last wins" otherwise, but I lean towards:
* Register builtins first
* Last registered factory wins

stevenschlansker:
You are correct about streams themselves being lazy, but writing things as streams caused some formerly-lazy things to be eagerly evaluated...  Again I don't think there is any real-world consequence here, just something I noticed.

core/src/main/java/org/jdbi/v3/ColonPrefixNamedParamStatementRewriter.java
```diff
@@ -150,18 +153,18 @@ public void bind(Binding params, PreparedStatement statement) throws SQLExceptio
...
-                        String msg = String.format("Unable to execute, no named parameter matches " +
-                                                   "\"%s\" and no positional param for place %d (which is %d in " +
-                                                   "the JDBC 'start at 1' scheme) has been set.",
-                                                   named_param, i, i + 1);
-                        throw new UnableToExecuteStatementException(msg, context);
-                    }
+                    final int index = i;
+                    Argument a = Stream.concat(
+                            JdbiStreams.toStream(params.findForName(named_param)),
+                            JdbiStreams.toStream(params.findForPosition(i)))
```
stevenschlansker: For example, here.
This one should be trivially cheap, though.

core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
stevenschlansker: OK, sounds good.

core/src/main/java/org/jdbi/v3/ArgumentRegistry.java
qualidafial: Filed #268 which adds Guava Optional support

core/src/main/java/org/jdbi/v3/ColonPrefixNamedParamStatementRewriter.java
qualidafial: This one I blame the JDK for not having a `Optional.orElseGetOptional(Supplier<Optional<T>>)` method. ;)

core/src/main/java/org/jdbi/v3/ArgumentRegistry.java
qualidafial: Your preferred strategy does yield the simplest implementation: add the built in factory/factories in the constructor, and consult factories in the reverse order registered.

qualidafial: Actually there's one existing factory that would make "last wins" a very bad choice: BeanMapperFactory. BeanMapperFactory thinks it can map anything--how arrogant!--so if it was registered last, it would become the mapper for everything, which is probably not what folks expect.

Which is another way of saying *never* register BeanMapperFactory with the DBI instance--only on handles on a case by case basis.

stevenschlansker: Yuck, what a mess.  We could provide a way to register it first, so it only handles types that aren't already handled...  I wonder if there's another good option.

core/src/main/java/org/jdbi/v3/ColonPrefixNamedParamStatementRewriter.java
qualidafial: I found a lazily-evaluated solution for this on stackoverflow. It's off topic but I'll push it to this branch since it's not worth its own PR.

core/src/main/java/org/jdbi/v3/ArgumentRegistry.java
qualidafial: I thought of making BeanMapperFactory do a little bean introspection but since almost every data structure in java is a bean, it wouldn't buy us much.

I think we'd be better off introducing a `@RegisterBeanMapper(Class...)` annotation, and removing `BeanMapperFactory`.

qualidafial: That worked. The last change makes it so you have to specify which classes you want to register bean mappers for:

    dbi.registerMapper(new BeanMapperFactory(Foo.class, Bar.class, ...))

Or for SQL objects:

    @RegisterBeanMapper({Foo.class, Bar.class, ...})

PR #261, opened 2016-01-25
Make ResultColumnMapper lambda-friendly

qualidafial:


qualidafial:
@stevenschlansker could you code review this?

stevenschlansker:
Instead of removing the `String` version of the methods, what would you say to having a default implementation that does the column lookup and delegates?

stevenschlansker:
Another option, leave the interface as is, and have a subinterface that defaults the unused method.  Then we leave the current state as is and allow people to have a functional implementation of the subinterface?

qualidafial:
That's a good idea. If we make the String variant a default method, it can still be a functional interface.

qualidafial:
Hmm, while I'm at it, I think I'll use a similar trick to what you did in BuiltInArgumentFactory and get rid of the need for all the primitive column mapper classes.

qualidafial:
@stevenschlansker Can you take another look now?

stevenschlansker:
Whoa, net -500 lines.  I'm excited :)

core/src/main/java/org/jdbi/v3/PrimitivesMapperFactory.java
```diff
@@ -0,0 +1,123 @@
...
+import org.jdbi.v3.util.EnumMapper;
+
+/**
+ * Result column mapper factory which knows how to map standard JDBC-recognized types.
+ */
+public class PrimitivesMapperFactory implements ResultColumnMapperFactory {
+    private static final Map<Class<?>, ResultColumnMapper<?>> mappers = new HashMap<>();
+
+    static {
+        mappers.put(boolean.class, primitiveMapper(ResultSet::getBoolean));
```
stevenschlansker: This makes me very happy :smile:

PR #270, opened 2016-01-30
Upgrade the Spring integration to the latest version

arteam:
Spring 2.x is a rather outdated version of the framework and not supported anymore. This change upgrades the Spring dependency to the latest version 4.2.4.RELEASE.

This allows to take advantage from modern Spring features like annotation processing and the Spring JUnit test runner.


stevenschlansker:
Awesome, thanks!  Needs to update `commons-logging` to 1.2 before the build will pass.

qualidafial:
Thanks @arteam :+1:

PR #273, opened 2016-01-31
Add stream wrappers which close streams automatically after terminal operations

qualidafial:
Adds a wrapper around streams returned from ResultBearing.stream() which automatically closes a stream before returning from any call a terminal operation.

Also documents requirement to explicitly close query streams.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/ResultBearing.java
+++ b/core/src/main/java/org/jdbi/v3/ResultBearing.java
@@ -13,8 +13,11 @@
  */
 package org.jdbi.v3;
 
+import static java.util.Spliterators.spliteratorUnknownSize;
+
 import java.util.List;
 import java.util.Optional;
+import java.util.Spliterator;
 import java.util.stream.Collector;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -55,8 +58,26 @@ public interface ResultBearing<T> extends Iterable<T>
         }
     }
 
+    /**
+     * Executes the SQL statement and returns the stream of results.
+     *
+     * <p>
+     * Note: the returned stream owns database resources, and must be closed via a call to {@link Stream#close()}, or
+     * by using the stream in a try-with-resources block:
+     * </p>
+     *
+     * <pre>
+     * try (Stream&lt;T> stream = query.stream()) {
+     *   // do stuff with stream
+     * }
+     * </pre>
+     */
     default Stream<T> stream() {
-        return StreamSupport.stream(spliterator(), false);
+        ResultIterator<T> iterator = iterator();
+        Spliterator<T> spliterator = spliteratorUnknownSize(iterator, 0);
+        Stream<T> stream = StreamSupport.stream(spliterator, false)
+                .onClose(iterator::close);
+        return new SelfClosingStream<>(stream);
     }
 
     default List<T> list() {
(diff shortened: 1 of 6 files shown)


qualidafial:
The intention here is not to absolve users of the requirement to close their streams, but to help them out when they forget.

There are still a few ways that query streams can be used that will not auto-close:
* If an exception is thrown prior to the terminal operation method call
* If users call the `iterator()` or `spliterator()` methods on the stream.

And possibly some other cases I haven't contemplated.

qualidafial:
@stevenschlansker Any concerns before I merge this?

stevenschlansker:
Hm.  I'm a little worried.

This introduces significant complexity, and doesn't absolve the user of the need to close streams.
For example,
```
Stream s = dao.stream(); // returns self-closing stream
s = s.limit(computeSomeLimit()); // computeSomeLimit() throws!
s.collect(...); // Would clean up here, but we bombed!
```

I worry that any place this code is actually used, we have the potential to leak resources when things go wrong.
I know it's nice to be helpful and close streams in the good case, but I worry that it only helps out people that have fundamentally broken code :/

stevenschlansker:
I'm tempted to just say "Stream objects hold heavyweight database resources, and must be used in a `try-with-finally` or otherwise disposed of correctly".  But I'm also a fan of usable software, and I'm waffling...

qualidafial:
We don't have to force through the self-closing streams if you're uneasy about it, although there are other changes I do think are critical:

* Adding `Stream.onClose` handler that closes the ResultIterator
* Modifying the unit tests that call `Query.stream()` to use try-with-resources blocks.

qualidafial:
Opened PR #274

qualidafial:
This PR now only adds self-closing streams on terminal operations.

It's only a hunch, but I seem to be getting more reliable builds (i.e. no timeouts) with this change, FWIW.

qualidafial:
I wonder whether the statement will get cleaned up on exit from a handle or SQL object try-with-resources:

    try (Handle handle = dbi.open()) {
      return handle.createQuery(sql)
        .bind("name", arg)
        .stream()
        .limit(computeLimit()) // throws exception
        .collect(toList());
    }

    try (MyDao dao = dbi.open(MyDao.class)) {
      return dao.stream(arg)
        .limit(computeLimit()) // throws exception
        .collect(toList());
    }

If so, this might provide a palatable alternative to putting all query streams into try-with-resource blocks.

stevenschlansker:
I think by the spec if you close the `Connection` you don't have to close any e.g. `Statement` or `ResultSet` resources, they will be freed -- but I know of at least one connection pool that violates that because "it's too slow"

stevenschlansker:
tag @arteam @brianm @hgschmie anyone got any opinions here?

qualidafial:
What if we introduced a callback method similar to the `withHandle`?

    interface StreamCallback<T, R> {
      R withStream(Stream<T> stream) throws Exception;
    }
    class ResultBearing {
      <T, R> R withStream(StreamCallback<T, R> callback) {
        try (Stream<T> stream = stream()) {
          return callback.withStream(stream);
        }
      }
    }

stevenschlansker:
I like that.  We'd probably still want to have a way to just return the `Stream` though?  For more advanced users who want to control the lifecycle themselves.

Then people can opt-in to the resource management, we don't have wrapper classes, but we have a nicer way to write things safely.

qualidafial:
Already working on it. I'm adding both `withStream` and `useStream` variants just like with handles and transactions.

qualidafial:
Do we want to do anything with SQL objects returning string?

stevenschlansker:
We could allow them to take a `StreamConsumer` as an argument, which is like `withStream`?  If it returns the `Stream` it behaves like `useStream`?

qualidafial:
I guess it's time for me to figure out how SQL object works.

I'll close this PR and submit a new one with the new methods as discussed.

qualidafial:
Posted #275

PR #275, opened 2016-02-02
Add ResultBearing.useStream(), .withStream() methods

qualidafial:


core/src/main/java/org/jdbi/v3/tweak/StreamCallback.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package org.jdbi.v3.tweak;
+
+import java.util.stream.Stream;
+
+/**
+ * Callback for use with {@link org.jdbi.v3.ResultBearing#withStream(StreamCallback)}
+ */
+@FunctionalInterface
+public interface StreamCallback<T, R> {
```
qualidafial: @stevenschlansker Should these callbacks/consumers really be in the `org.jdbi.v3.tweak` package?

stevenschlansker: I'm all for moving them out.
I think the intention of the `tweak` package were for things that modified the behavior of statement preparation / execution (plugin-like things).
This certainly doesn't fit under that definition.

Maybe we can get rid of or give a better name to `tweak` entirely for v3.

stevenschlansker:
LGTM.  Figure out whether you want to move the packages, then :shipit:

Issue #277, opened 2016-02-05
Remove object argument from BuiltInArgumentFactory


qualidafial:
My experience has been that most of the time the built in argument factory returns an ObjectArgument (as it used to be called), it was always a symptom of an unregistered factory, or a factory that did not return an argument when you thought it would.

This has led to many wild goose chases, since by the time you get an exception, you no longer have a stack trace which points directly to the argument that caused the problem.

I propose we introduce an ObjectArgumentFactory class, and remove object argument support from BuiltInArgumentFactory.

If a user wants to use an argument which is supported through their JDBC driver's `ResultSet.setObject` (e.g. Postgres supports UUID) then they should register it explicitly:

    dbi.registerArgumentFactory(ObjectArgumentFactory.for(UUID.class));

    @RegisterObjectArgumentFactory(UUID.class)
    interface MyDao { ... }

stevenschlansker:
Works for me.

qualidafial:
Closing as dupe of #279

PR #279, opened 2016-02-10
Remove object argument from BuiltInArgumentFactory

qualidafial:
Addresses #277

### DIFF
--- a/core/src/main/java/org/jdbi/v3/ObjectArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/ObjectArgumentFactory.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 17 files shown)


qualidafial:
I'm also debating getting rid of the second `Object.class` pass when searching for arguments.

qualidafial:
@stevenschlansker would you mind reviewing this?

core/src/main/java/org/jdbi/v3/ObjectArgumentFactory.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+import org.jdbi.v3.tweak.Argument;
+import org.jdbi.v3.tweak.ArgumentFactory;
+
+public class ObjectArgumentFactory implements ArgumentFactory
+{
+    public static ArgumentFactory create(Class<?> type) {
+        return create(type, java.sql.Types.NULL);
```
stevenschlansker: Is this right?  Should this be some sort of unknown type, rather than claiming it is of type `NULL`?

stevenschlansker:
One comment, then :shipit:

core/src/main/java/org/jdbi/v3/ObjectArgumentFactory.java
qualidafial: That's what the previous behavior was, but on closer inspection it looks like `Types.OTHER` is a better fit. Will change.

qualidafial: After a little digging I found #132 which deliberately chose `Types.NULL` so things would work in Oracle. Probably best to keep using `Types.NULL`.

PR #285, opened 2016-02-23
Introduce BuildInMapperFactory.build(Class) which avoids casting in user code

stevenschlansker:


qualidafial:
What about adding a `<T> mapTo(Class<T>)` method instead, and let the choice of mapper be an implementation detail?

stevenschlansker:
I don't think that's better, really.  For example the result is passed in to `batch.executeAndGenerateKeys(mapper)`, for which you actually need the returned Mapper instance.  And it's an overload of the more general `build(Type, StatementContext)` method.

Or am I misunderstanding?

qualidafial:
I was looking at the wrong class when I made my earlier suggestion.

Could we use `PreparedBatch.executeAndGenerateKeys(Class)` instead of looking up the mapper by hand?

qualidafial:
I included a change in #323 that removes the need for this: https://github.com/jdbi/jdbi/pull/323/files#diff-e24595282a6941c4903008b26c7541fc

PR #291, opened 2016-02-29
Add a chain of parameter binder factories that can provide binders, defaulting to PositionalBinder when all else fails

apatrida:
When annotations do not provide binders, use a chain of parameter binding factories to find a binder, falling back to the default PositionalBinder when nothing else accepts or provides a non-null binder.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -287,8 +287,8 @@
                     <configuration>
                         <toolchains>
                             <jdk>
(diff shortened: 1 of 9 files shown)


apatrida:
This is the required element for adding Kotlin support for 2.x branches

apatrida:
Doing this again on a cleaner branch.

PR #292, opened 2016-02-29
Add a chain of parameter binder factories that can provide binders

apatrida:
Add a chain of parameter binder factories that can provide binders, defaulting to PositionalBinder when all else fails

When annotations do not provide binders, use a chain of parameter binding factories to find a binder, falling back to the default PositionalBinder when nothing else accepts or provides a non-null binder.

This is required for Kotlin support, since it needs to register a binder.

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/ParameterBinderFactory.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/ParameterBinderFactory.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.skife.jdbi.v2.sqlobject;
+
+import java.lang.reflect.Method;
+
+/**
+ * Factory interface used to produce result set mappers.
+ */
+public interface ParameterBinderFactory
+{
+    /**
+     * Can this factory provide a parameter binder for binding a method parameter to a sql bound parameter?
+     * @param type the class containing the method
+     * @param method the method that requires binders
+     * @param param_idx the current parameter index that will be bound
+     * @return true if it can, false if it cannot
+     */
+    boolean accepts(Class type, Method method, int param_idx);
+
+    /**
+     * Supplies a binder for the given parameter, or null to fallback to a default binder
+     */
+    Binder binderFor(Class type, Method method, int param_idx);
+}
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/ParameterBinderRegistry.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/ParameterBinderRegistry.java
@@ -1,0 +1,128 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.skife.jdbi.v2.sqlobject;
+
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+class ParameterBinderRegistry
+{
+    static final ParameterBinderFactory DEFAULT_BINDER_FACTORY = new ParameterBinderFactory() {
+        @Override
+        public boolean accepts(Class type, Method method, int param_idx) {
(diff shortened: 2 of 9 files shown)


apatrida:
Question, this is static due to how `SqlObject` creates things from a static factory, and since there really isn't another instance of the factory, is this the best way to hold the `ParameterBinderRegistry`?  Should it be on the `SqlObjectBuilder` instead?  

I think so, therefore moving it.

src/main/java/org/skife/jdbi/v2/sqlobject/ParameterBinderFactory.java
```diff
@@ -0,0 +1,36 @@
...
+     * @param method the method that requires binders
+     * @param param_idx the current parameter index that will be bound
+     * @return true if it can, false if it cannot
+     */
+    boolean accepts(Class type, Method method, int param_idx);
+
+    /**
+     * Supplies a binder for the given parameter, or null to fallback to a default binder
+     */
+    Binder binderFor(Class type, Method method, int param_idx);
```
qualidafial: When you start porting this PR to v3 (after this one merges), remember to coalesce these two methods into a single method that returns `Optional<Binder>` to stay consistent with the other factories.

src/main/java/org/skife/jdbi/v2/sqlobject/ParameterBinderRegistry.java
```diff
@@ -0,0 +1,128 @@
...
+    static final ParameterBinderFactory DEFAULT_BINDER_FACTORY = new ParameterBinderFactory() {
+        @Override
+        public boolean accepts(Class type, Method method, int param_idx) {
+            return true;
+        }
+
+        @Override
+        public Binder binderFor(Class type, Method method, int param_idx) {
+            return new PositionalBinder(param_idx);
+        }
```
qualidafial: In v3, all binds are assumed to be by name even in the absence of a `@Bind` annotation, unless the sql object class was compiled without the `-parameters` compiler flag.

So there will have to be a discussion around positional bindings as the default when we get to v3.

src/main/java/org/skife/jdbi/v2/sqlobject/ParameterBinderRegistry.java
```diff
@@ -0,0 +1,128 @@
...
+    private static class ClassMethodParameter {
+        final Class type;
+        final Method method;
+        final int param_idx;
+
+        ClassMethodParameter(Class type, Method method, int param_idx) {
+            this.type = type;
+            this.method = method;
+            this.param_idx = param_idx;
+        }
```
qualidafial: I'm wondering if `type` is redundant considering that we already have `method.getDeclaringClass()`. (I'm mostly musing about v3 here)

src/main/java/org/skife/jdbi/v2/sqlobject/SqlObjectBuilder.java
```diff
@@ -22,6 +22,7 @@
  */
 public class SqlObjectBuilder
 {
+    static final ParameterBinderRegistry  binderRegistry = new ParameterBinderRegistry();
```
qualidafial: This being static makes me uncomfortable.

I'd rather this registry live in the DBI, and passed (possibly cloned) to each SqlObject.

We'll have to do something different for v3 since sqlobject is now a separate layer on top of core.

qualidafial:
> If there is some context created for SqlObjectBuilder that is stored into an JDBI instance then this could change to not be static. We would need to add something into JDBI to allow some state being held for other plugins or modules. This probably fits everything a bit more.

:+1: let's discuss this in its own issue

qualidafial:
@stevenschlansker can you weigh in on this PR?

qualidafial:
Needs tests.

src/main/java/org/skife/jdbi/v2/sqlobject/ParameterBinderFactory.java
apatrida: yes.

src/main/java/org/skife/jdbi/v2/sqlobject/ParameterBinderRegistry.java
apatrida: Yes, that default breaks when `-parameters` are not present on the javac command line.  It might be better to put that into a factory that does not accept anything that doesn't have parameters present, and then have the positional as the last in the chain.  Positional can be the default that is never removed, and the java 8 parameters a single item in the normal chain since it can turn on only when appropriate.

src/main/java/org/skife/jdbi/v2/sqlobject/ParameterBinderRegistry.java
apatrida: Yes, likely will drop the class parameter and get it from the method.

src/main/java/org/skife/jdbi/v2/sqlobject/SqlObjectBuilder.java
apatrida: It's a problem that SqlObjectBuilder has no where to store data like its registries or default handlers or anything else that might come up as configurable.  DBI could hold the values for components it does not know about that they can retrieve as needed.  Assuming they know about DBI.  I'll look at the chain backwards to see if every path can lead to DBI and come up with something.  As you say elsewhere, "discuss in its own issue".  In v2 flavour I can put into DBI and pollute it a bit more, and then look at the other issue so it can move out of static for v3.

apatrida:
I'll do another round on this shortly for comments above, just wanted feedback on roughed in idea to make sure it is basically a decent direction.

stevenschlansker:
I like the direction this is going.  I wonder if you could mimic poor man's "DBI-level storage" via a `static Map<DBI, ParameterBinderRegistry>` if you're impatient for the other PR to land.

stevenschlansker:
hi @apatrida we are considering cutting a `jdbi3` alpha soon.  It might be useful to get any more feedback you have, especially concerning the new Extension facility we added.

stevenschlansker:
Since we merged 3 in, this PR needs to be reopened against the `jdbi2` branch.

PR #293, opened 2016-02-29
jdbi3 - Add a chain of parameter binder factories that can provide binders

apatrida:
The JDBI3 version of: https://github.com/jdbi/jdbi/pull/292

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/ParameterBinderFactory.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/ParameterBinderFactory.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.sqlobject;
+
+import java.lang.reflect.Method;
+
+/**
+ * Factory interface used to produce result set mappers.
+ */
+public interface ParameterBinderFactory
+{
+    /**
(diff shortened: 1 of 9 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/ParameterBinderFactory.java
```diff
@@ -0,0 +1,36 @@
...
+     * @param method the method that requires binders
+     * @param param_idx the current parameter index that will be bound
+     * @return true if it can, false if it cannot
+     */
+    boolean accepts(Class type, Method method, int param_idx);
+
+    /**
+     * Supplies a binder for the given parameter, or null to fallback to a default binder
+     */
+    Binder binderFor(Class type, Method method, int param_idx);
```
qualidafial: Can we combine these methods into a single one that returns `Optional<Binder>`?

qualidafial:
Needs tests

sqlobject/src/main/java/org/jdbi/v3/sqlobject/ParameterBinderFactory.java
apatrida: Yes, this can be changed for 3.x, waiting to solve issues for the master then will merge up again and then change this.

apatrida:
Test coverage that it doesn't break anything is in current tests.   For interception and order of the factories, I'll add.  Starting with the master version then merging up.

stevenschlansker:
ping @apatrida still got time to write some tests? :)

qualidafial:
Hi @apatrida,

If you've got bandwidth to work on this, the new `JdbiConfig` stuff that just landed in 3.0.0-alpha4 would be a great way to implement this.

We've been using pluralized names for config classes, so if you rename `ParameterBinderRegistry` to `ParameterBinders`, and change it to implement `JdbiConfig<ParameterBinders>`, this should have a couple benefits:

* No need to modify all those the `*Handler` classes to add another constructor parameter. You can access the configuration via `handleSupplier.getConfig(ParameterBinders.class)`.
* You can have default configuration at the `Jdbi` level, and override it at the `Handle` or SQL Object level.

qualidafial:
I'm moving this to the post-v3 release milestone--meaning we won't hold back v3 final release if this isn't merged. However if it's ready before release, I'd happily include it.

stevenschlansker:
Since we are merging `jdbi3` to `master`, this PR will need to be re-opened against `master`.  Sorry!

Issue #295, opened 2016-03-07
JDBI extensions 


qualidafial:
In v3, we separated out sqlobject from core, but there is still strong coupling between them in `SqlObjectBuilderBridge`.

Also--as discussed in #282--`sqlobject` would be better if we could configure its behavior as readily as we can configure arguments, mappers, and collectors in core.

I think we could accomplish this goal by introducing one further registry in core:

    interface ExtensionFactory {
      <T> Optional<T> attach(Class<T> extensionClass, Handle handle);
    }

    class DBI {
      void registerExtension(ExtensionFactory factory)

      // Extension factory may expose configuration methods for the extension
      // e.g. register parameter binders as in #293
      <E extends ExtensionFactory> E getExtension(Class<E> extensionClass)

      <T> open(Class<T> extensionClass)

      // should be part of DBI or specific to sqlobject?
      <T> onDemand(Class<T> extensionClass)
    }

    interface Handle {
      void registerExtension(ExtensionFactory factory)
      <E extends ExtensionFactory> E getExtension(Class<E> extensionClass)

      <T> void attach(Class<T> extensionClass);
    }

Thus `sqlobject` would implement its own `JdbiPlugin` (and publish a services file so it was available through `DBI.installPlugins()`). This plugin would register `SqlObjectExtensionFactory`. This extension would take the place of `SqlObjectBuilder`.

This is just a first draft, and there are other ways we could structure this to enable custom configuration of extensions.

For example, in the above scheme, the `ExtensionFactory` is the repository of configuration, but this means that the factory becomes a configuration singleton for the DBI and all handles. Some other options might be:

1. Make ExtensionFactory also a configuration factory, and pass that configuration back to the factory when it's time to build an extension:

        interface ExtensionFactory<C> {
          <T> boolean accepts(Class<T> extensionClass)

          // configuration could be per extension class, or per factory, at the factory writer's discretion
          <T> C createConfiguration(Class<T> extensionClass)

          <T> Optional<T> attach(Class<T> extensionClass, C configuration, Handle handle);
        }

2. Add a `createCopy()` method to `ExtensionFactory`, and copy each factory when when creating an ExtensionFactoryRegistry one-off for handles, such that DBI-level configuration is handed down to the Handle, but Handle-level configuration is isolated.

        interface ExtensionFactory<F extends ExtensionFactory<F>> {
          <T> Optional<T> attach(Class<T> extensionClass, Handle handle);
          F createCopy();
        }

Thoughts?

[...]

Issue #271, opened 2016-01-31
Never return Stream or raw Iterable from onDemand sql objects


qualidafial:
Originally: "Remove DBI.onDemand(Class)"

orangecoding:
Why?

stevenschlansker:
I'm torn on this one.  On one hand, it's really convenient -- but the behavior is surprising, and the code to implement it in the library is actually really touchy and hard to get right.  And in the worst case, we cannot prevent resource leaks if the user doesn't use it correctly.

It opens up many opportunities for incorrect resource handling.  But it's also really, really convenient... maybe there's a better paradigm entirely?

orangecoding:
As we've understand it, it collects a connection if needed and release it, if not needed anymore. We are using this quit heavily and find it really helpful.

apatrida:
Is this for removing it just from the Handle class or from DBI and other places as well?

qualidafial:
This is about removing it entirely until resource leaks can be prevented.

@stevenschlansker do you have an example of improper use that leads to leaks? All I have is a gut feeling from my own usage; I've never been able to definitively track it down.

stevenschlansker:
I'm not sure if this is the most compelling example, but one simple way would be:

```java
@SqlQuery("...")
Iterator<Thing> getThings(...);

Thing getFirstThing() {
    dao = dbi.onDemand(ThingDao.class);
    return dao.getThings().next();
}
```

The `Iterator` magically inherits the resource ownership from the `dao`, so if you don't close it, it leaks.  And it's just an `Iterator` so you can't even see that it is `Closeable` unless you know to use `ResultIterator` instead.

stevenschlansker:
One option: we could change `onDemand` so it loses the transferability of resource ownership.  You always lose your `Handle` when the method returns.  This means for simple cases it still works, but you are forbidden from letting resources escape.  It'd be like a built-in `try-with-resources`

qualidafial:
I think that's a good compromise.

stevenschlansker:
Anyone object to doing this?  I also think it's a good compromise.  I guess the biggest loss is it prevents you from returning a `Stream<Thing>` from onDemand DAO methods.

qualidafial:
While working on #304 I found it easier to complete my generic `onDemand` implementation by putting this rule into effect. So when #304 is merged this issue will be resolved.

qualidafial:
[...]

PR #304, opened 2016-04-02
JDBI 3 Extensions

qualidafial:
## API changes

* Added `ExtensionFactory`, `ExtensionConfig`, `ExtensionCallback`, `ExtensionConsumer` interfaces
* `DBI` method `open(Class)` replaced with `withExtension(Class, ExtensionCallback)` and `useExtension(Class, ExtensionConsumer)`
* Added `DBI` methods `registerCollectorFactory(CollectorFactory)`,  `registerExtensionFactory(ExtensionFactory)`, and `configureExtension(Class<C>, Consumer<C>)`.
* `DBI.onDemand` restrictions: only public interface types are supported for now--allows us to use proxies without bringing in third party dependencies. The handle from an on-demand extension method call is _always_ closed as the method returns, which means that SQL object methods that return iterables, streams, or other SQL objects (`@CreateSqlObject`) will throw an error when you try to access the returned value, as the handle is closed. (See #271)
* Added method `JdbiPlugin.customizeDbi(DBI)` method. This was necessary so that extensions get installed on the DBI, making those extensions available for use with `DBI.onDemand(Class)`.
* Added `Handle` methods `registerExtension(ExtensionFactory)` and `configureExtension(Class, Consumer)`.
* Interface `CloseMe` removed. Extensions (including SQL objects) are no longer `AutoCloseable`. Users manage the handle resource explicitly and attach extensions to them, or use callback methods like `withExtension` to have the handle managed for them.

(Comprises #295, #271)

To do:

* [x] Finish generic `onDemand` implementation in core
  * [x] Lazy init handle so `onDemand` extensions only create handle when they need to (`equals`, `hashCode`, `toString` being counterexamples).
  * [x] Redirect tests to `DBI.onDemand`
  * [x] Revise tests that assume `onDemand` extensions can provide streams or fluent iterators, to expect the opposite now.
* [x] Replace `HandleDing` with `Supplier<Handle>`, and stop doing handle retain/release in sqlobject.
* [x] Move static configuration in SqlObjectBuilder into config class
* ~~Get rid of CallbackFailedException in favor of exception transparency in callback interfaces, while I'm thinking about it~~ Oh hell no
* [x] Remove method handlers for `AutoCloseable` and friends.
* [x] Remove CloseHandler
* [x] GetHandle's `useHandle` and `withHandle` could be default methods, and could remove cglib handlers `UseHandleHandler` and `WithHandleHandler`


qualidafial:
@stevenschlansker @brianm This pull request is ready to look at.

qualidafial:
@stevenschlansker Any chance I could get a code review on this one?

stevenschlansker:
Yeah sorry, I've been slammed at work... will try to get to all the PRs this week.

core/src/main/java/org/jdbi/v3/DBI.java
```diff
@@ -370,41 +382,100 @@ public void useTransaction(final TransactionIsolationLevel isolation, final Tran
...
-        return SqlObjectBuilderBridge.onDemand(this, sqlObjectType);
+    public <E, X extends Exception> void useExtension(Class<E> extensionType, ExtensionConsumer<E, X> callback)
+            throws NoSuchExtensionException, X {
+        withExtension(extensionType, extension -> {
+            callback.useExtension(extension);
+            return null;
+        });
+    }
+
+    private class LazyHandle implements Supplier<Handle>, AutoCloseable {
```
stevenschlansker: Probably worth making a standalone class?

core/src/main/java/org/jdbi/v3/DBI.java
```diff
@@ -370,41 +382,100 @@ public void useTransaction(final TransactionIsolationLevel isolation, final Tran
...
-    {
-        SqlObjectBuilderBridge.close(sqlObject);
+    public <E> E onDemand(Class<E> extensionType) throws NoSuchExtensionException {
+        if (!extensionType.isInterface()) {
+            throw new IllegalArgumentException("On-demand extensions are only supported for interfaces.");
+        }
+        if (!Modifier.isPublic(extensionType.getModifiers())) {
+            throw new IllegalArgumentException("On-demand extensions types must be public.");
+        }
+        if (!extensionRegistry.hasExtensionFor(extensionType)) {
```
stevenschlansker: Are you intending to `Optional`-ify this "check then use"?

core/src/main/java/org/jdbi/v3/ExtensionRegistry.java
```diff
@@ -0,0 +1,90 @@
...
+        return factories.stream()
+                .anyMatch(entry -> entry.factory.accepts(extensionType));
+    }
+
+    <E> Optional<E> findExtensionFor(Class<E> extensionType, Supplier<Handle> handle) {
+        return factories.stream()
+                .filter(entry -> entry.factory.accepts(extensionType))
+                .map(entry -> entry.factory.attach(extensionType, entry.config, handle))
+                .map(extensionType::cast)
+                .findFirst();
```
stevenschlansker: Should this sanity check that there are no additional elements?  i.e. `findOnly`

core/src/main/java/org/jdbi/v3/ExtensionRegistry.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Optional;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+
+import org.jdbi.v3.extension.ExtensionConfig;
+import org.jdbi.v3.extension.ExtensionFactory;
+import org.jdbi.v3.extension.NoSuchExtensionException;
+
+class ExtensionRegistry {
```
stevenschlansker: I worry that this class isn't sufficiently threadsafe -- if you publish a config from one thread, there's no guarantees any other thread would observe that config?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObject.java
```diff
@@ -13,236 +13,11 @@
...
-
-    static String getSql(SqlBatch q, Method m)
-    {
-        if (SqlQuery.DEFAULT_VALUE.equals(q.value())) {
-            return m.getName();
-        }
-        else {
-            return q.value();
-        }
+public class SqlObject implements ExtensionConfig<SqlObject> {
```
stevenschlansker: Should there be a built-in `NoConfig` or the like for extensions which have no configuration?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -0,0 +1,174 @@
...
+                Transactional.class.isAssignableFrom(extensionType)) {
+            return true;
+        }
+
+        MemberResolver mr = new MemberResolver(typeResolver);
+        ResolvedType extension_type = typeResolver.resolve(extensionType);
+        ResolvedTypeWithMembers d = mr.resolve(extension_type, null, null);
+        for (ResolvedMethod method : d.getMemberMethods()) {
+            Method rawMethod = method.getRawMember();
+            if (rawMethod.isAnnotationPresent(SqlQuery.class)
```
stevenschlansker: This hardcoded list sucks.  Maybe a meta-annotation?
```java
@SqlObjectAnnotation
public @interface SqlQuery { ... }
```

core/src/main/java/org/jdbi/v3/DBI.java
qualidafial: We do have `ExtensionRegistry.findExtensionFor(Class, Supplier<Handle>)` that returns an `Optional`, but that would require allocating a handle now instead of on-demand.

In this case I'm just verifying that the extension will be available when future calls to the proxy require it.

core/src/main/java/org/jdbi/v3/ExtensionRegistry.java
qualidafial: This registry follows a last-registered factory wins rule, similar to the other registries. If users need to ensure that a certain factory wins, they must explicitly register the factory (or plugin) last.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObject.java
qualidafial: I thought about that, but decided against it for forward compatibility. Over time a user would expect the config class to stay in the same place, but accumulate more options.

I'm afraid if we provide a `NoConfig` we'd be encouraging extension authors into a situation where they have to break backward compatibility in order to make their extensions configurable.

I'm still debating whether having the configuration class separate from the extension factory is really necessary.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
qualidafial: I could get behind that.

core/src/main/java/org/jdbi/v3/DBI.java
qualidafial: Wouldn't hurt. Do you have any reuse cases in mind?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -0,0 +1,174 @@
...
+            final MemberResolver mr = new MemberResolver(typeResolver);
+            final ResolvedType sql_object_type = typeResolver.resolve(sqlObjectType);
+
+            final ResolvedTypeWithMembers d = mr.resolve(sql_object_type, null, null);
+
+            final Map<Method, Handler> handlers = new HashMap<>();
+            for (final ResolvedMethod method : d.getMemberMethods()) {
+                final Method raw_method = method.getRawMember();
+
+                if (raw_method.isAnnotationPresent(SqlQuery.class)) {
```
stevenschlansker: The meta-annotation could drive this dispatch as well

sqlobject/src/main/java/org/jdbi/v3/sqlobject/mixins/GetHandle.java
```diff
@@ -41,7 +41,14 @@
      * @throws CallbackFailedException Will be thrown if callback raises an exception. This exception will
      *                                 wrap the exception thrown by the callback.
      */
-    <ReturnType> ReturnType withHandle(HandleCallback<ReturnType> callback) throws CallbackFailedException;
+    default <ReturnType> ReturnType withHandle(HandleCallback<ReturnType> callback) throws CallbackFailedException {
+        try {
+            return callback.withHandle(getHandle());
+        }
+        catch (Exception e) {
+            throw new CallbackFailedException(e);
```
stevenschlansker: This should be exception-transparent like the other PR, yeah?

core/src/main/java/org/jdbi/v3/DBI.java
stevenschlansker: Nope.  Just struck me as not related enough to the enclosing class to be inner.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
qualidafial: We could go a step further and make `@SqlObjectAnnotation` take a factory class as a parameter, similar to `@SqlStatementCustomizingAnnotation`. Then users are free to introduce their own method handlers.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObject.java
stevenschlansker: Excellent point.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
stevenschlansker:
```java
@SqlObjectAnnotation(QueryHandler.class)
public @interface SqlQuery { ... }
```

stevenschlansker:
It looks like the Extension interface essentially requires you to use cglib.
I don't suppose there would be any way to tease these apart, so we handle cglib under the hood, and your Extension doesn't need to know / care that we use it?  I only ask because it ties us to cglib, and we couldn't switch to e.g. a MethodHandle or Proxy based approach without breaking API compatibility...

stevenschlansker:
Dude this PR rocks!  I'm excited.  SqlObject being tied to core so intimately was one of my pet peeves, with `HandleDing` being second on the list.  Sending buckets of internet karma your way :smile: 

Will let you do one more pass, poke me when you want a re-review or merge.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/mixins/GetHandle.java
qualidafial: This is already such a big pull request, and the cognitive load was getting too high already. That's why I opened the separate PR for the exception transparency stuff.

This change was just about migrating from `GetHandleHandler` to a default method.

qualidafial:
> It looks like the Extension interface essentially requires you to use cglib.

There's no cglib in core, just sqlobject. Am I missing something?

stevenschlansker:
> On Apr 8, 2016, at 4:13 PM, Matthew Hall <notifications@github.com> wrote:
> 
> It looks like the Extension interface essentially requires you to use cglib.
> 
> There's no cglib in core, just sqlobject. Am I missing something?


Sorry, it's late on Friday and I can't read!

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
qualidafial: Hmm, if we give the meta-annotation a factory class, then we expose the Handler interface and by extension cglib's MethodProxy type into the API. The MethodProxy argument is still used in `PassThroughHandler`, so I can't remove it from the interface method just yet.

I'm going to call it `@SqlMethodAnnotation` and just annotate the other annotation classes for now. I'll open an issue to add the rest later.

qualidafial: Another idea to explore: modify the mixins to use meta-annotations

    public interface GetHandle
    {
      @SqlMethod(GetHandleHandler.class)
      Handle getHandle();
    }

Then the mixins go through the same annotation lookup procedure as e.g. `@SqlQuery`, `@SqlBatch`, etc.

qualidafial:
I think I've addressed most of your concerns. I'd like to hold off on adding factories to the meta-annotation for now--I feel like that's somewhat orthogonal, and should live in a separate issue.

qualidafial:
Ready for another review

qualidafial:
@stevenschlansker ping

qualidafial:
Actually, now I'm working on making the annotations specify a `HandlerFactory` class right now. Should be done in ~30min

qualidafial:
@stevenschlansker Ready for review

stevenschlansker:
What a mouthful!

Issue #216, opened 2015-11-02
Annotation to throw exception if >1 row returned


lewisd32:
I'd like to have some way to annotate a method to indicate that it should only return a single row, and that if more rows are returned an exception should be thrown.  I expected that that's what `@SingleValueResult` would do, but after more reading, I see that it's only for providing type information in the case of `Optional` return types? (I can't find much documentation about it, so I'm not entirely sure that's true.)

Having an attribute of it that could be used like:

``` Java
@SingleValueResult(Foo.class, constrain=true)
```

It would disable adding the `MAX_ROW_ONE` customizer, and throw an exception if more than one row was returned.

Alternatively, if there were some way of adding constraint annotations:

``` Java
@SingleValueResult(Foo.class)
@MaximumRowCountConstraint(1)
```

That would serve the same purpose.

stevenschlansker:
In jdbi3, you can accomplish this by wrapping your return type: mapping to `Optional<Foo>` will throw `IllegalStateException` if there's more than one row.

PR #311, opened 2016-04-05
Added MaxRowCountConstraint

anev:
Added MaxRowCountConstraint annotation to control SqlQuery result max size. 
issue #216

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/ResultReturnThing.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/ResultReturnThing.java
@@ -16,12 +16,13 @@ package org.skife.jdbi.v2.sqlobject;
 import com.fasterxml.classmate.ResolvedType;
 import com.fasterxml.classmate.TypeBindings;
 import com.fasterxml.classmate.members.ResolvedMethod;
-
 import org.skife.jdbi.v2.Query;
 import org.skife.jdbi.v2.ResultBearing;
 import org.skife.jdbi.v2.ResultIterator;
 import org.skife.jdbi.v2.exceptions.UnableToCreateStatementException;
 import org.skife.jdbi.v2.sqlobject.customizers.Mapper;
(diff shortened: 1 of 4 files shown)


qualidafial:
Hi Andrei! Thanks for contributing.

What's your use case for this pull request? Have you considered using a limit clause in the SQL itself?

    @SqlQuery("select id, name from something where name = :name limit :maxRows")
    Something findByNameNew(@Bind("name") String name, @Bind("maxRowCount") long maxRows);

Or perhaps you're looking to verify that an expected number of rows match a query? In that case I suggest a `select count(*)` query instead:

    @SqlQuery("select count(*) from something where name = :name")
    long countByName(@Bind("name") String name);

src/main/java/org/skife/jdbi/v2/sqlobject/ResultReturnThing.java
```diff
@@ -102,11 +104,32 @@ else if (return_type.isInstanceOf(Iterator.class)) {
...
         }
 
         @Override
         protected Object result(ResultBearing q, HandleDing baton)
         {
+            if (maxRowCount == 0) {
+                return returnFirst(q);
+            }
+
+            if (q.list(maxRowCount + 1).size() > maxRowCount) {
```
qualidafial: This is very inefficient. You are executing the query and pulling down all data in all rows just to verify the row count. Better to use a `select count(*)` query instead and avoid all the network and disk i/o.

stevenschlansker: It also means the query will be executed twice, I think?  That could be quite dangerous if it is e.g. an `UPDATE ... RETURNING *` style of statement.

stevenschlansker:
Hi @anev we didn't hear back from you so I'm going to close this PR.  If you're still interested in pursuing this let us know the problem you are trying to solve -- I believe it is strictly better to use SQL `LIMIT` so the database knows what's up.  We certainly can reopen the issue if you have more to add.

Issue #290, opened 2016-02-25
Exception transparency in handler/callback interfaces


qualidafial:
interface HandleCallback<R, X extends Exception> {
      T withHandle(Handle handle) throws X;
    }

    public <R, X> R withHandle(HandleCallback<R, X> callback) throws X

This way you only need to catch in the caller to `withHandle` the exceptions that a lambda actually throws.

stevenschlansker:
:+1:

qualidafial:
This is addressed in pull request #312

PR #312, opened 2016-04-06
Exception transparency in handler/callback interfaces

qualidafial:
Addresses issue #290 

To do:

* [x] ~~Remove `Transaction` callback interface in favor of `TransactionCallback`~~ Rename it to `TransactionalCallback` and move it to sqlobject where it belongs
* [x] Add `TransactionalConsumer` and friend `Transactional.useTransaction(TransactionalConsumer)`
* ~~See if `TransactionFailedException` can be removed~~ nah


core/src/main/java/org/jdbi/v3/tweak/transactions/LocalTransactionHandler.java
```diff
@@ -165,36 +165,28 @@ public boolean isInTransaction(Handle handle)
...
         catch (Exception e) {
             try {
                 handle.rollback();
             } catch (Exception rollback) {
                 e.addSuppressed(rollback);
             }
-            throw new TransactionFailedException("Transaction failed do to exception being thrown " +
-                                                 "from within the callback. See cause " +
-                                                 "for the original exception.", e);
+            throw (X) e;
```
stevenschlansker: Ah, what a sneaky throw ;)

stevenschlansker:
Nice!  Is this ready to merge, or still pending the TODOs?  I think the current state is good, happy to merge as is and handle the rest as followups, or block it -- your call.  Thanks!

core/src/main/java/org/jdbi/v3/tweak/transactions/LocalTransactionHandler.java
qualidafial: My reasoning is that the try can only throw a `RuntimeException` or `X`. So the cast is safe from a runtime perspective. At compile time `X` is erased to just `Exception`.

Plus there's the whole part about how checked exceptions are a purely compile time construct--the runtime makes no distinction and any method can legally throw any Throwable they like.

qualidafial:
I want to hold this one and merge after the extensions, since the merge effort (and pain) should be less on this PR than on the other one.

core/src/main/java/org/jdbi/v3/tweak/transactions/LocalTransactionHandler.java
stevenschlansker: Yeah, got it.  I like that it removes one of the catch clauses!

qualidafial:
@stevenschlansker I made one more change: renamed `Transaction` interface to `TransactionalCallback`, moved it from core to sqlobject, fixed its generic signature. Also added `TransactionalConsumer` interface to go with it.

If you're good with these changes, I think this is ready to merge.

PR #317, opened 2016-04-10
Fix #265: non-numeric return types when combining @SqlUpdate with @GetGeneratedKeys

### LINKED ISSUE #265: Support (again!) ID Objects for generated keys in update statements
Unfortunately latest releases removed this feature by introducing the eager check in the UpdateHandler

private boolean returnTypeIsInvalid(Class<?> type) {
        return !Number.class.isAssignableFrom(type) &&
                !type.equals(Integer.TYPE) &&
                !type.equals(Long.TYPE) &&
                !type.equals(Void.TYPE);
    }

Before it was possible to use IdObject(Wrappers) like
 @GetGeneratedKeys(value=IdObjectMapper.class)

that could contain several keys like:
public class IdObject {
 private Long id;
 private UUID uuid;
...
}

which worked perfect with Postgres...

stevenschlansker:
heads up @jeetah

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -2,6 +2,7 @@
   - (finally!) parse Postgres CAST syntax 'value::type' properly in colon
     prefix statements
   - fix @SqlBatch hanging if you forget to include an Iterable-like param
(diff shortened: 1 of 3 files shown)


qualidafial:
LGTM

PR #318, opened 2016-04-10
Cleanup: SQLStatement has redundant getParams / getParameters

stevenschlansker:


src/main/java/org/skife/jdbi/v2/SQLStatement.java
```diff
@@ -245,11 +245,6 @@ protected String getSql()
         return sql;
     }
 
-    protected Binding getParameters()
-    {
-        return params;
-    }
-
```
qualidafial: Strictly speaking this is a backward-incompatible change.

From a practical standpoint though, `SQLStatement` isn't intended to be extended by clients.

stevenschlansker: Fair point.  We can block it if you like, it's only mildly silly, just close it if you feel that way.

qualidafial: Meh, I'm coming down on the pragmatic side on this one.

PR #320, opened 2016-04-11
Fix #314: @UseStringTemplate3Locator creates a new locator for every invocation

### LINKED ISSUE #314: UseStringTemplate3StatementLocator returns a new locator on every invocation
When `createForType` is invoked, a new locator is created every time.  This means that templates are not cached, which has a large CPU impact.

stevenschlansker:
tag @devrandom

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/stringtemplate/UseStringTemplate3StatementLocator.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/stringtemplate/UseStringTemplate3StatementLocator.java
@@ -19,6 +19,10 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.WeakHashMap;
+
+import javax.annotation.concurrent.GuardedBy;
 
 import org.antlr.stringtemplate.StringTemplateErrorListener;
 import org.antlr.stringtemplate.StringTemplateGroup;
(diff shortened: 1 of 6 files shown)


devrandom:
I tried something similar, and also this fix, but I'm getting:

```
org.skife.jdbi.v2.exceptions.UnableToCreateStatementException: Exception while binding '__ids_0' [statement:"SELECT * FROM xxx WHERE yyy in (<ids>)",
 located:"SELECT * FROM xxx WHERE yyy in (:__ids_0:__ids_0,:__ids_1)", rewritten:"SELECT * FROM xxx WHERE yyy in (??,?)",
 arguments:{ positional:{}, named:{__ids_1:'9c96da6c7af62b46cb5952b7a3c240bed67384bb395a07c22c62f3689276aa23@1',__ids_0:'9c96da6c7af62b46cb5952b7a3c240bed67384bb395a07c22c62f3689276aa23@4'}, finder:[]}]
```
so it looks like there's some template substitution state that is not being cleared.  Note the `:__ids_0:__ids_0,:__ids_1` where it should just be `:__ids_0,:__ids_1`.

devrandom:
BTW, thank you for the quick response to the issue report.

qualidafial:
@devrandom Where is `<ids>` getting defined? Are you using the fluent API or a SQL object? Can you provide some code to help troubleshoot this?

stevenschlansker:
Huh.  That's strange.  And the build fails, too... I'll try to look again tonight :disappointed:

devrandom:
@qualidafial - just anSQL object.  I don't have a narrow test case, will try to get to it tomorrow.

stevenschlansker:
Yuck, if you call `lookupTemplate`, you get a *single shared* instance.  We should call `getInstanceOf` instead... bug was introduced way back in 2012 :cry:

stevenschlansker:
tag @devrandom @qualidafial @bartfaitamas for review

src/main/java/org/skife/jdbi/v2/sqlobject/stringtemplate/UseStringTemplate3StatementLocator.java
```diff
@@ -37,13 +41,32 @@
...
+        private static final Map<Class<?>, WeakHashMap<Annotation, SqlStatementCustomizer>> CUSTOMIZER_CACHE =
+                new WeakHashMap<Class<?>, WeakHashMap<Annotation, SqlStatementCustomizer>>();
+
         @Override
         public SqlStatementCustomizer createForType(Annotation annotation, Class sqlObjectType)
         {
             final UseStringTemplate3StatementLocator a = (UseStringTemplate3StatementLocator) annotation;
+
+            if (a.cacheable()) {
+                synchronized (LocatorFactory.class) {
```
qualidafial: Should use double-check pattern to prevent unnecessary locking.

qualidafial: In fact I'd say don't lock here at all, and move the `CUSTOMIZER_CACHE.put` call down at the bottom of the method, where you can lock unconditionally. That way the top section that checks the cache is a read-only operation and therefore safe to access without synchronization.

qualidafial: Hmm, WeakHashMap is not synchronized, so synchronization _is_ required for any access.

Never mind these comments then.

qualidafial:
:shipit:

src/main/java/org/skife/jdbi/v2/sqlobject/stringtemplate/UseStringTemplate3StatementLocator.java
stevenschlansker: Yeah, double checked locking wouldn't be safe against an unsynchronized Map access.

qualidafial:
Before we merge, is there a practical way to test this fix?

stevenschlansker:
Added a test.

PR #325, opened 2016-04-14
Expose Foreman on StatementContext in JDBI 2.X

brianm:
Add docs, rename waffle(), demonstrate use in documentation test


stevenschlansker:
fixes #325

PR #326, opened 2016-04-15
Additional fix for #314 .

bartfaitamas:
Resets the reused string template before parameter substitution.


stevenschlansker:
Hi @bartfaitamas Thanks for taking a stab at this.  You're entirely right about the problem, but I think the fix is not quite right.  Just calling `reset` introduces a race between threads, where Thread-1 may be in the middle of setting up attributes when Thread-2.

I've put this other fix here: https://github.com/jdbi/jdbi/pull/320/commits/013744d721e731ca118e32d8b9bc25bc3882faa3 and will close this PR.

Let me know if you disagree, and we can reopen this PR.  Thanks again!

Issue #258, opened 2016-01-24
Migrate sqlobject from classmate API to to java.lang.reflect


qualidafial:
Affected methods:
  * Handle.attach
  * DBI.open
  * DBI.onDemand
  * SqlObjectBuilderBridge methods
  * StatementContext.sqlObjectType property
  * SqlStatementCustomer methods
  * CustomizingStatementHandler ctor parameter
  * CreateSqlObjectHandler
  * SqlObject methods
  * SqlObjectBuilder methods
  * SqlStatementCustomizingFactory methods
  * StringTemplate3StatementLocator?
  * Binder.bind needs a resolved java.lang.reflect.Type parameter
  * Binder.bind needs a parameterIndex parameter to support bare `@Bind`s.

qualidafial:
Taking a stab at this.

qualidafial:
@stevenschlansker @brianm Question: what would you think of replacing the classmate/cglib stuff in sqlobject with a `java.lang.reflect.Proxy` implementation?

This would imply restricting sqlobject types to interfaces only, like we did with on-demand extensions.

It would also allow us to remove cglib and classmate as dependencies.

Thoughts?

qualidafial:
Let's move discussion to #327

PR #327, opened 2016-04-18
Migrate sqlobject from Classpath to Types class (currently backed by Guava TypeToken)

qualidafial:
Resolves issues #222, #258 

To Do

* [x] Use the new `StatementContext.elementTypeFor` method to help integrate third-party collectors without need for `@SingleValueResult`. ~~This might even let us coalesce all the `ReturnResultThing` subclasses into a single implementation.~~
* [x] Give Binder.bind a `parameterIndex` parameter to support bare `@Bind` annotations
* [x] Bonus! Support returning `java.util.Optional` from SQL object methods
* [x] Should add tests for new `Types.resolveType` method
* [x] Throw exception on multiple elements for Optional builders (java.util and guava)
* [x] BatchHandler take another look at streams vs nested for loops


qualidafial:
Let's move any discussion solicited in my comment from #258 here:

> @stevenschlansker @brianm Question: what would you think of replacing the classmate/cglib stuff in sqlobject with a java.lang.reflect.Proxy implementation?

> This would imply restricting sqlobject types to interfaces only, like we did with on-demand extensions.

> It would also allow us to remove cglib and classmate as dependencies.

> Thoughts?

Since classmate is gone, the discussion scope is now limited to migrating away from cglib toward a `java.lang.reflect.Proxy` implementation.

qualidafial:
@stevenschlansker This is ready for review.

qualidafial:
@stevenschlansker ping?

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
```diff
@@ -50,40 +51,108 @@ void register(CollectorFactory factory) {
...
+                    OptionalBuilder::build);
+        }
+
+    }
+
+    private static class OptionalBuilder<T> {
+        private Optional<T> optional = Optional.empty();
+
+        public void set(T value) {
+            optional = Optional.of(value);
```
stevenschlansker: Should this do something along the lines of `if (optional.isPresent()) throw new IllegalStateException();` ?

core/src/main/java/org/jdbi/v3/Query.java
```diff
@@ -146,8 +146,8 @@
      * @see Handle#registerRowMapper(RowMapperFactory)
      * @see Handle#registerRowMapper(RowMapper)
      */
-    private <T> Query<T> mapTo(Type resultType) {
-        return this.map(new RegisteredRowMapper<>(resultType, mappingRegistry));
+    public Query mapTo(Type resultType) {
```
stevenschlansker: Hm.... raw type... is the `<T>` really so dangerous that a raw type is better?  I guess it is more truthful about what protections exist...

core/src/main/java/org/jdbi/v3/CollectorFactoryRegistry.java
qualidafial: I had that same thought but forgot to follow up and fix it. Thanks for catching

core/src/main/java/org/jdbi/v3/Query.java
qualidafial: The `<T>` was ok while the method was private, but since I had to make it part of the API to bring sqlobject "up to code" on enhanced generics, I felt it was better to underpromise in this case.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/BatchHandler.java
```diff
@@ -36,65 +32,61 @@
...
     {
-        Annotation[][] param_annos = raw_method.getParameterAnnotations();
-        for (int i = 0; i < param_annos.length; i++) {
-            Annotation[] annos = param_annos[i];
-            for (Annotation anno : annos) {
-                if (anno.annotationType().isAssignableFrom(BatchChunkSize.class)) {
-                    return i;
-                }
+        Annotation[][] parameterAnnotations = method.getParameterAnnotations();
+        for (int i = 0; i < parameterAnnotations.length; i++) {
```
stevenschlansker: If you're going to do streams, might as well go whole hog?  I'm not sure the single element per Stream approach is idiomatic...

```java
return IntStream.range(0, parameterAnnotations.length)
                          .filter(i -> parameterAnnotations[i].annotationType().isAssignableFrom(BatchChunkSize.class))
                          .findFirst().orElse(-1);
```

stevenschlansker:
Couple comments, then :shipit:

core/src/main/java/org/jdbi/v3/Query.java
stevenschlansker: 👍

PR #328, opened 2016-04-20
Add support for URI arguments to builtin argument factory

stevenschlansker:


core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
```diff
@@ -90,6 +91,7 @@
         register(map, Time.class, Types.TIME, PreparedStatement::setTime);
         register(map, Timestamp.class, Types.TIMESTAMP, PreparedStatement::setTimestamp);
         register(map, URL.class, Types.DATALINK, PreparedStatement::setURL);
+        register(map, URI.class, Types.VARCHAR, stringifyValue(PreparedStatement::setString));
```
stevenschlansker: `URL` is this funky `DATALINK` type -- for symmetry, we could consider that, but I have no idea what it does or why it's used and Postgres / H2 are both perfectly happy with Strings.

qualidafial:
Let's add a URI column mapper for symmetry, then :shipit:

qualidafial:
Oops, I see URI support is already present. Good to go then.

PR #329, opened 2016-04-25
Add built-in arguments and column mappers for JSR-310 types

qualidafial:


core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
```diff
@@ -92,6 +99,15 @@
         register(map, Timestamp.class, Types.TIMESTAMP, PreparedStatement::setTimestamp);
         register(map, URL.class, Types.DATALINK, PreparedStatement::setURL);
         register(map, URI.class, Types.VARCHAR, stringifyValue(PreparedStatement::setString));
+
+        Calendar c = Calendar.getInstance();
+        Supplier<Calendar> cal = () -> (Calendar) c.clone();
```
stevenschlansker: Why is this not just `Calendar::getInstance`?

qualidafial: `getInstance` is an expensive operation. Clone is cheap. So we pay the `getInstance` penalty once.

core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
```diff
@@ -92,6 +99,15 @@
...
+        register(map, Instant.class, Types.TIMESTAMP, (p, i, v) -> p.setTimestamp(i, Timestamp.from(v), cal.get()));
```
stevenschlansker: Is this `Calendar` instance even necessary?
> If no `Calendar` object is specified, the driver uses the default timezone, which is that of the virtual machine running the application.

core/src/test/java/org/jdbi/v3/TestJsr310.java
```diff
@@ -0,0 +1,91 @@
...
+public class TestJsr310 {
+    @Rule
+    public H2DatabaseRule db = new H2DatabaseRule();
+
+    Handle h;
+
+    @Before
+    public void setUp() {
+        h = db.getSharedHandle();
+        h.execute("create table stuff (ts timestamp, d date)");
```
stevenschlansker: Is it worth having some tests for `timestamp with time zone` types, so that we can ensure zones are stored and retrieved correctly?

stevenschlansker:
Couple minor comments, then :shipit:

core/src/main/java/org/jdbi/v3/BuiltInArgumentFactory.java
qualidafial: If the tests pass without it, that works for me. I was just following `dropwizard-jdbi`'s lead, not fully understanding why they used the calendar in certain cases.

core/src/test/java/org/jdbi/v3/TestJsr310.java
qualidafial: It's worth a shot, but from my read of the ResultSet and Timestamp classes, I don't think that time zones or zone offsets can be recovered.

stevenschlansker: Hm.  Maybe you're right.  That's a bummer if so :(

qualidafial: Per http://mail.openjdk.java.net/pipermail/jdbc-spec-discuss/2012-December/000010.html it looks like the intention was that `ResultSet.setObject` would support some java.time types.

However I'm concerned it may not be uniformly implemented across all drivers. I'll do some digging.

qualidafial: I looked at H2 and Postgres drivers. Neither of them support java.time types in PreparedStatement.setObject or ResultSet.getObject. Looks like we'll have to give up on retaining time zone / zone offset for now.

We could provide override arguments and mappers on a per-driver basis however. For example, in Postgres, `ResultSet.getString` on a date column returns a specific string format. See `org.postgres.jdbc2.TimestampUtils.toTimestamp(Calendar, String)` to see how the Postgres driver parses it out.

PR #332, opened 2016-04-28
Configure maven-surefire-plugin forkCount to 1

qualidafial:
The default maven-surefire-plugin `forkCount` configuration from basepom-oss is `1.0C` meaning 1 VM forked per CPU core. Build logs suggest Travis build environments provide 8 cores.

This is often a problem for our Postgres tests since they are actively creating / modifying the same table in the database. I suspect this also explains why our builds time out at the end of the test run so often.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -293,6 +293,13 @@
                         </toolchains>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-surefire-plugin</artifactId>
+                    <configuration>
+                        <forkCount>1</forkCount>
+                    </configuration>
+                </plugin>
             </plugins>
         </pluginManagement>
         <plugins>


stevenschlansker:
Should we consider instead changing the tests so that they are safe to run in parallel?
It could be as simple as changing the default `public` schema to a random one on a per test basis.

qualidafial:
I could see that helping under the 3.x branch, but in 2.x without being able to isolate it in the test rules it's kind of a pain.

qualidafial:
So I tried it out on master branch. It's not easy making each test work on a separate schema. You can create the schema once in the `@Before` method (or in a test rule) but then every time you open a handle you have to execute the command `set search_path to schema123`.

Even on v3 with the test rule, the DBI is exposed, so any test that talks to DBI directly instead of the shared handle will have to do the same thing.

Based on this experiment I don't think schema-per-test is worth pursuing at this point.

stevenschlansker:
I'm uninterested in fixing it for master, but `jdbi3` should support running tests in parallel :)

PR #342, opened 2016-05-03
jdbi3 Postgres support

### LINKED ISSUE #221: Create jdbi-pg project with Postgres features e.g. UUID and SQL arrays
E.g. supporting UUID and SQL arrays in Postgres.

stevenschlansker:
fixes #221


pg/src/main/java/org/jdbi/v3/pg/ArrayColumnMapper.java
```diff
@@ -0,0 +1,70 @@
...
+        return buildArray(r.getArray(columnLabel));
+    }
+
+    private Object[] buildArray(Array array) throws SQLException {
+        if (UNSUPPORTED_TYPES.contains(array.getBaseType())) {
+            return buildFromResultSet(array);
+        }
+        try {
+            // more efficient, doesn't have to copy an unknown size array
+            return (Object[]) array.getArray();
```
qualidafial: Should this be mapping the array elements?

pg/src/main/java/org/jdbi/v3/pg/ArrayColumnMapper.java
```diff
@@ -0,0 +1,70 @@
...
+            }
+        }
+
+        if (length == capacity) {
+            return accumulator;
+        }
+        Object[] result = (Object[]) java.lang.reflect.Array.newInstance(componentType, length);
+        System.arraycopy(accumulator, 0, result, 0, length);
+        return result;
+    }
```
qualidafial: Also there's a nice method `Arrays.copyOf(T[] original, int newLength)` that you can use instead of the last three lines.

Is three a specific reason you're handling arrays manually here? Seems like ArrayList would be simpler.

pg/src/main/java/org/jdbi/v3/pg/PostgresJdbiPlugin.java
```diff
@@ -0,0 +1,26 @@
...
+package org.jdbi.v3.pg;
+
+import org.jdbi.v3.Handle;
+import org.jdbi.v3.spi.JdbiPlugin;
+
+public class PostgresJdbiPlugin implements JdbiPlugin {
+    @Override
+    public Handle customizeHandle(Handle handle) {
+        handle.registerArgumentFactory(new SqlArrayArgumentFactory());
+        handle.registerColumnMapper(new SqlArrayMapperFactory());
```
qualidafial: Something I forgot to bring up in the extensions PR: now that JdbiPlugin has a `customizeDbi` method, we should decide whether core plugins should be customizing the handle or the DBI.

Because anything registered on the handle will *always* take precedence over anything registered on the DBI, my vote would be to always register things on the DBI. That way the order of plugin registration results in a predictable order of precedence for arguments, mappers, collectors, and extensions they register.

pg/src/main/java/org/jdbi/v3/pg/SqlArrayArgumentFactory.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public boolean accepts(Class<?> expectedType, Object value, StatementContext ctx) {
+        return expectedType.isArray();
+    }
+
+    @Override
+    public Argument build(Class<?> expectedType, Object[] value, StatementContext ctx) {
+        return new ArrayArgument(guessSqlType(value), value);
+    }
```
qualidafial: These are v2 methods..

stevenschlansker: Sorry, you jumped on this PR before I was ready -- I branched off many months ago and am now trying to rebase on latest `jdbi3` and it's a bit of a mess at the moment :)

pg/src/main/java/org/jdbi/v3/pg/PostgresJdbiPlugin.java
stevenschlansker: I would say always customize at the highest level if possible (i.e. DBI).  That method didn't exist when this code was written, so I've updated it now.

pom.xml
```diff
@@ -246,6 +246,13 @@
                 <artifactId>postgresql</artifactId>
                 <version>9.1-901-1.jdbc4</version>
             </dependency>
+
+            <dependency>
+                <groupId>com.opentable.components</groupId>
+                <artifactId>otj-pg-embedded</artifactId>
+                <version>0.4.4</version>
```
qualidafial: Test scope?

pg/src/main/java/org/jdbi/v3/pg/ArrayColumnMapper.java
stevenschlansker: It's intended that this only happens for "obvious" mappings that the driver supports, e.g. `int[]` or `String[]` and that it's likely more efficient to let the driver return its array directly rather than materializing it ourselves.  Indeed, this would break if the JDBC driver handles the mapping differently than jdbi does -- maybe this (unproven) performance boost is not worth that possibility.

Or maybe a type whitelist would be more appropriate?

pom.xml
stevenschlansker: This is in `pom.xml` `dependencyManagement` which doesn't need a scope -- the scope is defined in `pg/pom.xml`

core/src/main/java/org/jdbi/v3/BuiltInMapperFactory.java
```diff
@@ -129,6 +131,14 @@ private static URI getURI(ResultSet r, int i) throws SQLException {
         }
     }
 
+    private static UUID getUUID(ResultSet r, int i, StatementContext ctx) throws SQLException {
+        String s = r.getString(i);
+        if (s == null) {
+            return null;
+        }
+        return UUID.fromString(s);
```
qualidafial: It's funny, when I looked closely at the Postgres JDBC driver, this is exactly what it's doing when you call getObject() on a UUID column.

pg/src/main/java/org/jdbi/v3/pg/ArrayColumnMapper.java
qualidafial: Right, but we're not validating that the element type of the `array.getArray()` matches the java array component type.

My other concern is that by using this method we may be bypassing any custom column mappers provided by the user.

pg/src/main/java/org/jdbi/v3/pg/SqlArrayArgumentFactory.java
```diff
@@ -0,0 +1,46 @@
...
+        map.put(double.class, "double precision");
+        BEST_GUESS = Collections.unmodifiableMap(map);
+    }
+
+    /**
+     * Look at a Java array and attempt to determine an appropriate
+     * SQL type to pass to the driver.
+     */
+    static final String guessSqlType(Object[] array) {
+        String guess = BEST_GUESS.get(array.getClass().getComponentType());
```
qualidafial: `array` could be null.

pg/src/main/java/org/jdbi/v3/pg/SqlArrayMapperFactory.java
```diff
@@ -0,0 +1,19 @@
...
+import org.jdbi.v3.Types;
+import org.jdbi.v3.tweak.ColumnMapper;
+
+public class SqlArrayMapperFactory implements ColumnMapperFactory {
+
+    @Override
+    public Optional<ColumnMapper<?>> build(Type type, StatementContext ctx) {
+        final Class<?> clazz = Types.getErasedType(type);
+        return clazz.isArray() ?
+                Optional.of(new ArrayColumnMapper(clazz.getComponentType(), ctx)) : Optional.empty();
```
qualidafial: What if `ArrayColumnMapper` were just a decorator around `ctx.findColumnMapperFor(clazz.getComponentType())`?

pg/src/main/java/org/jdbi/v3/pg/TypedEnumArgumentFactory.java
```diff
@@ -0,0 +1,19 @@
...
+import org.jdbi.v3.tweak.Argument;
+import org.jdbi.v3.tweak.ArgumentFactory;
+
+public class TypedEnumArgumentFactory implements ArgumentFactory {
+    @Override
+    public Optional<Argument> build(Type type, Object value, StatementContext ctx) {
+        if (!Types.getErasedType(type).isEnum()) {
+            return Optional.empty();
+        }
+        return Optional.of((p, s, c) -> s.setObject(p, value, java.sql.Types.OTHER));
```
qualidafial: Neat, I didn't know Postgres supported that

pg/src/test/java/org/jdbi/v3/pg/TestUuid.java
```diff
@@ -0,0 +1,86 @@
...
+
+import org.jdbi.v3.Handle;
+import org.jdbi.v3.sqlobject.SqlQuery;
+import org.jdbi.v3.sqlobject.SqlUpdate;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class TestUuid {
```
qualidafial: Since UUID support is in core, should this test class be there too?

pg/src/main/java/org/jdbi/v3/pg/PostgresJdbiPlugin.java
qualidafial: Should we get rid of `customizeHandle` to remove ambiguity? I could see that becoming a papercut for users.

pom.xml
qualidafial: You're right, my mistake.

pg/src/main/java/org/jdbi/v3/pg/ArrayColumnMapper.java
```diff
@@ -0,0 +1,65 @@
...
+    private Object[] buildFromResultSet(Array array) throws SQLException {
+        final ColumnMapper<?> mapper = ctx.findColumnMapperFor(componentType)
+                .orElseThrow(() -> new IllegalArgumentException("Unable to find column mapper for " + componentType));
+
+        int capacity = 16;
+        int length = 0;
+        Object[] accumulator = (Object[]) java.lang.reflect.Array.newInstance(componentType, capacity);
+        try (ResultSet rs = array.getResultSet()) {
+            while (rs.next()) {
+                accumulator[length++] = mapper.map(rs, 2, ctx);
```
qualidafial: Why column 2?

stevenschlansker:
@qualidafial The Array support is turning out to be way trickier than I anticipated.  I'm getting tired of playing rebase catchup with this branch, and I'm unsure as to whether this can be finished before we cut `alpha0`.  So how about we merge this code, which has working support for UUID and typed enums but Arrays are still broken?  Filed #369 as followup.

pg/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.jdbi</groupId>
+        <artifactId>jdbi3-parent</artifactId>
+        <version>3.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jdbi3-pg</artifactId>
```
qualidafial: Could we rename this artifact to `jdbi3-postgres`?

pg/src/test/java/org/jdbi/v3/pg/TestSqlArrays.java
```diff
@@ -0,0 +1,114 @@
...
+
+import org.jdbi.v3.Handle;
+import org.jdbi.v3.sqlobject.SqlQuery;
+import org.jdbi.v3.sqlobject.SqlUpdate;
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Rule;
+import org.junit.Test;
+
+@Ignore // XXX: this doesn't work yet!
```
qualidafial: Should fix this in follow-up ticket.

qualidafial:
One more comment to consider, then :shipit: when ready

stevenschlansker:
renamed, rebased, merged

PR #346, opened 2016-05-04
SQLStatement: getParams and getParameters are redundant

stevenschlansker:


core/src/main/java/org/jdbi/v3/Query.java
```diff
@@ -114,7 +114,7 @@
      */
     public <T> Query<T> mapTo(Class<T> resultType)
     {
-        return this.mapTo((Type) resultType);
+        return this.mapTo(resultType);
```
qualidafial: StackOverflow coming from this change.

stevenschlansker: Crap!  Stupid "helpful" IDE!!!

qualidafial:
Any reason this PR is merging into `jdbi3-pg` branch instead of `jdbi3`?

stevenschlansker:
Crap, because I'm bad at Git :(

PR #349, opened 2016-05-05
Introduce a ConstructorMapper, a row mapper for "immutable bean" objects

stevenschlansker:
also includes a `@DbName` annotation for customizing parameter and accessor names.  There are some bindings in SQLObject already but none in core...

### DIFF
--- a/core/src/main/java/org/jdbi/v3/ConstructorMapper.java
+++ b/core/src/main/java/org/jdbi/v3/ConstructorMapper.java
@@ -28,7 +28,6 @@ import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
-import org.jdbi.v3.exceptions.ResultSetException;
 import org.jdbi.v3.tweak.ColumnMapper;
 import org.jdbi.v3.tweak.RowMapper;
 import org.jdbi.v3.util.bean.ColumnNameMappingStrategy;
@@ -69,12 +68,7 @@ public class ConstructorMapper<T> implements RowMapper<T>
     }
 
     private ObjectCreator<T> createCreator(List<String> columnNames, ResultSetMetaData metadata, StatementContext ctx) {
-        final int length;
-        try {
-            length = metadata.getColumnCount();
-        } catch (SQLException e) {
-            throw new ResultSetException("Could not get column count", e, ctx);
-        }
+        final int length = columnNames.size();
 
         if (length != constructor.getParameterCount()) {
             throw new IllegalStateException(length + " columns in result set, but constructor takes " + constructor.getParameterCount());
@@ -129,7 +123,7 @@ public class ConstructorMapper<T> implements RowMapper<T>
     }
 
     private static String paramName(Parameter parameter) {
-        DbName dbName = parameter.getAnnotation(DbName.class);
+        ColumnName dbName = parameter.getAnnotation(ColumnName.class);
         if (dbName != null) {
             return dbName.value();
         }
--- a/core/src/main/java/org/jdbi/v3/ConstructorMapper.java
+++ b/core/src/main/java/org/jdbi/v3/ConstructorMapper.java
@@ -15,6 +15,7 @@ package org.jdbi.v3;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Parameter;
 import java.lang.reflect.Type;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
@@ -118,13 +119,21 @@ public class ConstructorMapper<T> implements RowMapper<T>
     {
         for (int i = 0; i < constructor.getParameterCount(); i++) {
             for (ColumnNameMappingStrategy strategy : nameMappingStrategies) {
-                if (strategy.nameMatches(constructor.getParameters()[i].getName(), columnName)) {
+                if (strategy.nameMatches(paramName(constructor.getParameters()[i]), columnName)) {
                     return i;
                 }
             }
         }
         throw new IllegalArgumentException("Constructor " + constructor + " has no parameter for '" + columnName +
                 "'. Are parameter names compiled into your class?");
+    }
+
+    private static String paramName(Parameter parameter) {
+        DbName dbName = parameter.getAnnotation(DbName.class);
+        if (dbName != null) {
+            return dbName.value();
+        }
+        return parameter.getName();
     }
 
     interface ObjectCreator<T> {
--- a/core/src/main/java/org/jdbi/v3/ConstructorMapper.java
+++ b/core/src/main/java/org/jdbi/v3/ConstructorMapper.java
@@ -1,0 +1,165 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/jdbi/v3/ConstructorMapper.java
```diff
@@ -0,0 +1,174 @@
...
+
+        if (length != constructor.getParameterCount()) {
+            throw new IllegalStateException(length + " columns in result set, but constructor takes " + constructor.getParameterCount());
+        }
+
+        final int[] columnMap = new int[length];
+        final ColumnMapper<?>[] mappers = new ColumnMapper<?>[length];
+
+        for (int i = 0; i < length; i++) {
+            final Type type = constructor.getGenericParameterTypes()[i];
```
qualidafial: Need to reconcile the parameter type in the context of `T`.

core/src/main/java/org/jdbi/v3/ConstructorMapper.java
```diff
@@ -0,0 +1,174 @@
...
+    @Override
+    public T map(ResultSet rs, StatementContext ctx)
+        throws SQLException
+    {
+        final ResultSetMetaData metadata = rs.getMetaData();
+        final List<String> columnNames = new ArrayList<>(metadata.getColumnCount());
+
+        for (int i = 1; i <= metadata.getColumnCount(); ++i) {
+            columnNames.add(metadata.getColumnLabel(i));
+        }
```
qualidafial: Meh: Since `columnNames` is entirely derived from `metadata`, this could just be moved into `createCreator()`

core/src/main/java/org/jdbi/v3/ConstructorMapper.java
```diff
@@ -0,0 +1,174 @@
...
+        return creator.create(rs);
+    }
+
+    private ObjectCreator<T> createCreator(List<String> columnNames, ResultSetMetaData metadata, StatementContext ctx) {
+        final int length;
+        try {
+            length = metadata.getColumnCount();
+        } catch (SQLException e) {
+            throw new ResultSetException("Could not get column count", e, ctx);
+        }
```
qualidafial: Could just use `columnNames.size()` instead of this try/catch.

core/src/main/java/org/jdbi/v3/ConstructorMapper.java
```diff
@@ -0,0 +1,174 @@
...
+            }
+        };
+    }
+
+    private int parameterIndexForColumn(String columnName)
+    {
+        for (int i = 0; i < constructor.getParameterCount(); i++) {
+            for (ColumnNameMappingStrategy strategy : nameMappingStrategies) {
+                if (strategy.nameMatches(paramName(constructor.getParameters()[i]), columnName)) {
+                    return i;
```
qualidafial: I wonder if we could get away with expecting the columns to be in the same order as the constructor parameters? Would simplify this class a lot.

core/src/main/java/org/jdbi/v3/ConstructorMapper.java
```diff
@@ -0,0 +1,174 @@
...
+            final int paramIndex = parameterIndexForColumn(columnNames.get(i));
+
+            if (mappers[paramIndex] != null) {
+                throw new IllegalArgumentException("Column named " + columnNames.get(i) + " maps to multiple parameters");
+            }
+
+            mappers[paramIndex] = ctx.findColumnMapperFor(type).orElseThrow(() ->
+                new IllegalArgumentException("Could not find column mapper for type '" + type + "' of parameter for constructor " + constructor));
+            columnMap[i] = paramIndex;
+        }
```
qualidafial: I we go with the column name to parameter name mapping, it would probably make sense to have a sanity check after this loop, to verify each column is used to exactly one parameter.

core/src/main/java/org/jdbi/v3/DbName.java
```diff
@@ -0,0 +1,29 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Specify the binding or mapping name for a property or parameter explicitly.
+ */
+@Retention(RUNTIME)
+@Target(PARAMETER)
+public @interface DbName {
```
qualidafial: I'd rather call this `ColumnName`. `DbName` sounds like the name of the database itself.

core/src/test/java/org/jdbi/v3/ConstructorMapperTest.java
```diff
@@ -0,0 +1,92 @@
...
+        assertEquals("3", bean.s);
+        assertEquals(2, bean.i);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testMismatch() throws Exception {
+        ConstructorBean bean = db.getSharedHandle().createQuery("SELECT i, '7' FROM bean").mapTo(ConstructorBean.class).findOnly();
+
+        assertEquals("3", bean.s);
+        assertEquals(2, bean.i);
```
qualidafial: These asserts are unreachable

core/src/main/java/org/jdbi/v3/ConstructorMapper.java
stevenschlansker: It's used as the cache key.

core/src/main/java/org/jdbi/v3/ConstructorMapper.java
stevenschlansker: Sorry, I don't understand this comment?  The type `T` is implicitly held by the `constructor` member, so won't it know what type we are talking about here?

core/src/main/java/org/jdbi/v3/ConstructorMapper.java
stevenschlansker: I think that is dangerous.  Right now if you accidentally do `SELECT b, a` with `MyBean(String a, String b)` it will catch it -- if we just use ordering, we cannot.  Additionally if you do a `SELECT *` I'm not entirely clear that we can or should rely on the column order.

core/src/main/java/org/jdbi/v3/ConstructorMapper.java
stevenschlansker: This is guaranteed -- we know that the count of columns and parameters are equal, and we check that we never replace any ( `mappers[paramIndex] != null` ).  So we could make the check explicit, but I think it is effectively dead code.

core/src/test/java/org/jdbi/v3/ConstructorMapperTest.java
stevenschlansker: Not anymore, I improved the test.

core/src/main/java/org/jdbi/v3/ConstructorMapper.java
qualidafial: Fair enough.

core/src/main/java/org/jdbi/v3/ConstructorMapper.java
qualidafial: Ok. `SELECT *` is a convincing counterexample.

core/src/test/java/org/jdbi/v3/ConstructorMapperTest.java
qualidafial: Aren't we expecting an `IllegalArgumentException` from the first line of `testDuplicate` and `testMismatch`? If so then the `assertEquals` calls are still unreachable.

qualidafial:
One last comment to consider, then :shipit:.

I do want to register an orthogonal concern though, that the current approach with `ColumnNameMappingStrategy` still isn't sitting right with me. I don't know that I can put my finger on the reason yet.

core/src/test/java/org/jdbi/v3/ConstructorMapperTest.java
stevenschlansker: Whoops.  I was thinking of a different test.  My bad.

stevenschlansker:
@qualidafial let's revisit `ColumnNameMappingStrategy` once you have concrete concerns.  it's not new here so we can do a followup later.

PR #354, opened 2016-05-09
cglib 3.2.2, asm 5.1

### LINKED ISSUE #340: Problem with bridged methods
I ran into a problem with JDBI that I tracked down to some kind of mishandling of bridged methods in either cglib or asm (can't recall which). I boiled the problem down to the bare essentials and uploaded to this project: https://github.com/AlasdairBruce/jdbi-parameters-problem. 

To trigger the problem, you have to compile using language level 1.8 with -parameters and you have to have a bridged method involved in the Dao being attached to a JDBI handle. With those in place, some kind of version mismatch is triggered like so:

```
java.lang.RuntimeException
    at org.skife.jdbi.asm.MethodVisitor.visitParameter(Unknown Source)
    at org.skife.jdbi.asm.ClassReader.b(Unknown Source)
    at org.skife.jdbi.asm.ClassReader.accept(Unknown Source)
    at org.skife.jdbi.asm.ClassReader.accept(Unknown Source)
    at org.skife.jdbi.cglib.proxy.BridgeMethodResolver.resolveAll(BridgeMethodResolver.java:61)
```

```java
    public void visitParameter(String var1, int var2) {
        if(this.api < 327680) {
            throw new RuntimeException();
```

I tested a fix locally by rebuilding JDBI using upgraded versions of cglib and asm (3.2.2 and 5.0.4) and it fixed this specific problem, although I think one JDBI unit test failed in the process.

I chatted to @brianm about this briefly already.

stevenschlansker:
Fixes #340 hopefully, @AlasdairBruce can you confirm?

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,3 +1,6 @@
+2.74
+  - cglib 3.2.2, asm 5.1; fixes codegen for new Java 8 bridge methods
+
 2.73
   - Allow clearing of bindings in SQLStatement
   - (finally!) parse Postgres CAST syntax 'value::type' properly in colon
--- a/pom.xml
+++ b/pom.xml
@@ -95,7 +95,7 @@
         <dependency>
             <groupId>cglib</groupId>
             <artifactId>cglib</artifactId>
-            <version>3.1</version>
+            <version>3.2.2</version>
         </dependency>
 
         <dependency>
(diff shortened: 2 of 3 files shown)


qualidafial:
LGTM. :shipit: when we have confirmation.

AlasdairBruce:
Confirmed - this fixes my test case. Thanks!

PR #356, opened 2016-05-15
jdbi3 doc outline and example test

stevenschlansker:
feel free to claim a page or make your own, if you have time.


qualidafial:
Oops, I spaced this off. I think this is a great start.

Adding Java compiler with snippet inclusion, so that snippets are guaranteed to be valid was a nice touch.

Rather than keep this open forever let's just merge it and make future improvements piecemeal.

qualidafial:
Also can we agree that merging doc improvements without peer review is OK so long as the build passes?

stevenschlansker:
I don't have a problem with it, but I somewhat feel that review still has value just to have someone act as the editor.  If you're blocked or something though and just want to push the button I'm not going to be upset.

PR #362, opened 2016-05-25
First pass at result mapping documentation

stevenschlansker:


stevenschlansker:
I'm actually just going to self-merge these, but would still love your review / edit.

Issue #353, opened 2016-05-09
Standardize casing of jdbi and sql


stevenschlansker:
In particular, `jDBI` vs `jdbi` vs `JDBI` and `Sql` vs `SQL` are used inconsistently.  Might also consider the `DBI` type itself.

qualidafial:
My vote: jdbi/sql when referring to the project/language in prose. In class names, use init caps e.g. JdbiPlugin, SqlObject.

qualidafial:
I would favor renaming `DBI` to `Dbi`.

stevenschlansker:
Those were exactly my preferences as well.  Perfect!

qualidafial:
:shipit:

stevenschlansker:
See #364 #366 #367

PR #364, opened 2016-05-25
DBI -> Jdbi

stevenschlansker:
part of #353


qualidafial:
👍  to this rename.

This is a good time to reconsider the naming idioms in our code. If we're renaming `DBI` to `Jdbi`, should variables be renamed too? `jdbi` is an option but I was thinking just `db` would be nice.

stevenschlansker:
Let's do that as a followup?  I'm trying to burn down the backlog to get to `alpha0` and fixing the API is more important than the internal names :)

stevenschlansker:
followup is #370

PR #365, opened 2016-05-25
Fix five minute doc indentation

stevenschlansker:
(depends on #364 so as to avoid merge conflicts)


qualidafial:
Thanks for cleaning these up

PR #366, opened 2016-05-25
SQLStatement -> SqlStatement

stevenschlansker:
(depends on #364 and #365 to avoid merge conflicts)


core/src/main/java/org/jdbi/v3/SqlStatement.java
```diff
@@ -46,9 +46,9 @@
  * <code>Update</code>. It defines most of the argument binding functions
  * used by its subclasses.
  */
-public abstract class SQLStatement<SelfType extends SQLStatement<SelfType>> extends BaseStatement
+public abstract class SqlStatement<SelfType extends SqlStatement<SelfType>> extends BaseStatement
```
qualidafial: What would you think of standardizing all self-return generic parameters like this:

    class Whatever<This extends Whatever<This>> {
      public This doSomething() {
        ...
        return (This) this;
      }
    }

qualidafial: ..in a separate bug of course

docs/src/test/java/jdbi/doc/FiveMinuteTourTest.java
```diff
@@ -34,27 +34,27 @@
...
     @Before
     public void setUp() {
-// tag::createJdbi[]
-// H2 in-memory database
-Jdbi dbi = Jdbi.create("jdbc:h2:mem:test");
-// end::createJdbi[]
+        // tag::createJdbi[]
+        // H2 in-memory database
+        Jdbi dbi = Jdbi.create("jdbc:h2:mem:test");
+        // end::createJdbi[]
```
qualidafial: Thanks for cleaning these up

qualidafial:
:shipit:

PR #378, opened 2016-05-26
Latest basepom, version management

stevenschlansker:
Haven't been able to reproduce test hang yet with the latest plugin versions, maybe it's fixed?  *hopes*


qualidafial:
LGTM. :shipit: assuming build passes

PR #382, opened 2016-06-01
First pass at transaction documentation

stevenschlansker:


docs/src/adoc/transaction.adoc
```diff
@@ -0,0 +1,52 @@
...
+transaction.
+
+== Transactions in SqlObject API
+
+Similarly, you may declare transactions with SqlObject annotations:
+
+[source,java,indent=0]
+-------------------------------------------
+include::TransactionTest.java[tags=sqlObjectTransaction]
+-------------------------------------------
```
qualidafial: You cannot use `@Transaction` with a `@Sql[Whatever]` annotation. `SqlObjectFactory` (lines 105-109) only pays attention to the first `@SqlMethodAnnotation`-annotated annotation on the method----all other are ignored.

This has been a paper cut for users for a long time--I think it's worth fixing.

qualidafial: I do feel like an example where you do multiple operations imperatively in a `@Transaction` method (possibly calling other `@SqlUpdate` methods) would be more à propos for demonstrating this feature. But I'm fine with doing that work in a separate PR, as long as it's tracked.

stevenschlansker: OK I'll do that then.

qualidafial:
:shipit:

PR #383, opened 2016-06-01
Warning Patrol

### LINKED ISSUE #371: Fix all compiler warnings

stevenschlansker:
fixes #371

### DIFF
--- a/core/src/main/java/org/jdbi/v3/ExtensionRegistry.java
+++ b/core/src/main/java/org/jdbi/v3/ExtensionRegistry.java
@@ -31,34 +31,38 @@ class ExtensionRegistry {
         return new ExtensionRegistry(registry.factories);
     }
 
-    private static class Entry {
-        static Entry copyOf(Entry entry) {
-            return new Entry(entry.factory, entry.config.createCopy());
+    private static class Entry<C extends ExtensionConfig<C>> {
+        static <C extends ExtensionConfig<C>> Entry<C> copyOf(Entry<C> entry) {
+            return new Entry<>(entry.factory, entry.config.createCopy());
         }
 
-        final ExtensionFactory factory;
-        final ExtensionConfig config;
+        final ExtensionFactory<C> factory;
+        final C config;
 
-        Entry(ExtensionFactory factory, ExtensionConfig config) {
+        Entry(ExtensionFactory<C> factory, C config) {
             this.factory = factory;
             this.config = config;
         }
+
+        <E> E attach(Class<E> extensionType, Supplier<Handle> handle) {
+            return factory.attach(extensionType, config, handle);
+        }
     }
 
-    private final List<Entry> factories;
+    private final List<Entry<? extends ExtensionConfig<?>>> factories;
 
     ExtensionRegistry() {
         this.factories = new CopyOnWriteArrayList<>();
     }
 
-    ExtensionRegistry(List<Entry> factories) {
+    ExtensionRegistry(List<Entry<? extends ExtensionConfig<?>>> factories) {
         this.factories = factories.stream()
                 .map(Entry::copyOf)
                 .collect(toCollection(CopyOnWriteArrayList::new));
     }
 
-    void register(ExtensionFactory factory) {
-        factories.add(0, new Entry(factory, factory.createConfig()));
+    <C extends ExtensionConfig<C>> void register(ExtensionFactory<C> factory) {
+        factories.add(0, new Entry<>(factory, factory.createConfig()));
     }
 
     boolean hasExtensionFor(Class<?> extensionType) {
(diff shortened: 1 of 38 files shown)


core/src/main/java/org/jdbi/v3/Jdbi.java
```diff
@@ -551,7 +551,7 @@ public void registerCollectorFactory(CollectorFactory collectorFactory)
         config.collectorRegistry.register(collectorFactory);
     }
 
-    public void registerExtensionFactory(ExtensionFactory extensionFactory)
+    public void registerExtensionFactory(ExtensionFactory<?> extensionFactory)
```
qualidafial: Unrelated, but we should rename this to `registerExtension` to match `Handle`. Or vice-versa (but I prefer the shorter name).

core/src/main/java/org/jdbi/v3/Query.java
```diff
@@ -139,8 +139,8 @@
      * @see Handle#registerRowMapper(RowMapperFactory)
      * @see Handle#registerRowMapper(RowMapper)
      */
-    public Query mapTo(Type resultType) {
-        return this.map(new RegisteredRowMapper(resultType, config.mappingRegistry));
+    public <T> Query<T> mapTo(Type resultType) {
```
qualidafial: Better to return `Query<?>` than to let shenanigans like this get past the compiler:

    Type type = int.class;
    String result = handle.createQuery("...").mapTo(type).findOnly();

core/src/main/java/org/jdbi/v3/ExtensionRegistry.java
```diff
@@ -31,34 +31,38 @@ static ExtensionRegistry copyOf(ExtensionRegistry registry) {
...
             this.config = config;
         }
+
+        <E> E attach(Class<E> extensionType, Supplier<Handle> handle) {
+            return factory.attach(extensionType, config, handle);
+        }
     }
 
-    private final List<Entry> factories;
+    private final List<Entry<? extends ExtensionConfig<?>>> factories;
```
qualidafial: I have mixed feelings on this one.

I originally wrote this Entry class with a generic parameter, but ended up removing it because all it did was add a bunch of noise to keep the type checker happy.

Case in point: the `<C>` parameter only has value inside the entry--the only value it provides is ensuring that the factory and the config types are compatible. We don't really need the wildcard parameter on this list, but Java barks at us anyway for REASONS.

This is really just a rant about the limitations of Java's generics. *sigh*

I like the new `Entry.attach` method though.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/RegisterObjectArgumentFactory.java
```diff
@@ -42,7 +42,7 @@
      * The argument type(s) which are compatible with {@link java.sql.PreparedStatement#setObject(int, Object)}.
      * @return the argument types
      */
-    Class<? extends Class>[] value();
+    Class<? extends Class<?>>[] value();
```
qualidafial: Shouldn't this just be `Class<?>[]`?

qualidafial: This annotation is untested.

Since `Class` is final, the only possible value that can be given to this annotation property is `Class.class`.

qualidafial:
Just a couple comments. :shipit: when ready.

core/src/main/java/org/jdbi/v3/Jdbi.java
stevenschlansker: done

core/src/main/java/org/jdbi/v3/ExtensionRegistry.java
stevenschlansker: Thankfully this is all private so we can change it later if necessary.  I hate warnings more than I hate extra types just to satisfy the checker :)

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/RegisterObjectArgumentFactory.java
stevenschlansker: Yikes, it sure is...

stevenschlansker:
thanks for the review :)

core/src/main/java/org/jdbi/v3/ExtensionRegistry.java
qualidafial: Which IDE do you use? I think my compiler warnings are different from yours.

stevenschlansker: Eclipse

PR #385, opened 2016-06-02
Merge BaseHandle into Handle, reorganize Handle and Jdbi methods, add fluent returns

stevenschlansker:
We never really had any implementations of `Handle` other than `BasicHandle` so we can reduce our code by just removing it.


core/src/main/java/org/jdbi/v3/Handle.java
```diff
@@ -28,109 +32,231 @@
...
+    }
 
     /**
-     * Start a transaction
+     * Register a column mapper which will have its parameterized type inspected to determine what it maps to.
      *
-     * @return the same handle
+     * Column mappers may be reused by {@link RowMapper} to map individual columns.
+     *
+     * @param mapper the column mapper
```
qualidafial: We should add a `@throws` here for the case that the mapper isn't concretely parameterized.

core/src/main/java/org/jdbi/v3/Handle.java
```diff
@@ -28,109 +32,231 @@
...
+    }
 
     /**
-     * Rollback a transaction
+     * Register a row mapper which will have its parameterized type inspected to determine what it maps to.
      *
-     * @return the same handle
+     * Will be used with {@link Query#mapTo(Class)} for registered mappings.
+     *
+     * @param mapper the row mapper
```
qualidafial: Ditto for `@throws`

stevenschlansker:
added the `UnsupportedOperationException` docs.

core/src/main/java/org/jdbi/v3/Handle.java
```diff
@@ -28,109 +32,237 @@
...
+import org.jdbi.v3.tweak.TransactionHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * This represents a connection to the database system. It usually is a wrapper around
  * a JDBC Connection object.
  */
-public interface Handle extends Closeable
+public class Handle implements Closeable
```
qualidafial: Should we make this final for now?

stevenschlansker: I'm ambivalent, you can't subclass it anyway (no public or protected constructor).  So I'm not going to do it unless there's a compelling reason.

umcodemonkey: I would discourage it; making it final makes it harder to mock.

qualidafial: Let's leave it then.

core/src/main/java/org/jdbi/v3/Handle.java
```diff
@@ -140,225 +272,337 @@
...
+        finally {
+            try {
+                setTransactionIsolation(initial);
+            }
+            catch (RuntimeException e) {
+                if (! failed) {
+                    throw e;
+                }
+
+                // Ignore, there was already an exceptional condition and we don't want to clobber it.
```
qualidafial: `addSuppressed` instead, maybe?

core/src/main/java/org/jdbi/v3/Jdbi.java
```diff
@@ -163,16 +229,38 @@ public Jdbi installPlugin(JdbiPlugin plugin)
...
      * @param locator StatementLocator which will be used by all Handle instances
      *                created from this DBI
      */
-    public void setStatementLocator(StatementLocator locator)
+    public Jdbi setStatementLocator(StatementLocator locator)
     {
-        assert locator != null;
+        if (locator == null) {
+            throw new IllegalArgumentException("null statement locator");
+        }
```
qualidafial: Could alternatively use `Objects.requireNonNull(locator, "null statement locator")`

core/src/main/java/org/jdbi/v3/Jdbi.java
```diff
@@ -278,8 +365,9 @@ public void registerRowMapper(RowMapperFactory factory) {
      *
      * @param mapper the column mapper
      */
-    public void registerColumnMapper(ColumnMapper<?> mapper) {
+    public Jdbi registerColumnMapper(ColumnMapper<?> mapper) {
```
qualidafial: This method wants a `@throws` too, similar to the one in `Handle`

core/src/main/java/org/jdbi/v3/Handle.java
stevenschlansker: fixed with try-with-resources.

core/src/main/java/org/jdbi/v3/Jdbi.java
```diff
@@ -289,8 +377,33 @@ public void registerColumnMapper(ColumnMapper<?> mapper) {
...
-    public void registerColumnMapper(ColumnMapperFactory factory) {
+    public Jdbi registerColumnMapper(ColumnMapperFactory factory) {
         config.mappingRegistry.addColumnMapper(factory);
+        return this;
+    }
+
+    /**
+     * Register a row mapper which will have its parameterized type inspected to determine what it maps to
+     *
+     * Will be used with {@link Query#mapTo(Class)} for registered mappings.
```
qualidafial: Do we want to also mention generated keys?

qualidafial:
A few minor comments to consider, then :shipit:

core/src/main/java/org/jdbi/v3/Jdbi.java
stevenschlansker: done

core/src/main/java/org/jdbi/v3/Jdbi.java
stevenschlansker: maybe later, we should do a separate pass over the javadoc.  Trying to get the code cleaned up right now :)

PR #386, opened 2016-06-02
Simplify Cleanable by extending AutoCloseable, lambdafy

stevenschlansker:


core/src/main/java/org/jdbi/v3/Cleanables.java
```diff
@@ -35,120 +35,17 @@ private Cleanables()
 
     static Cleanable forResultSet(final ResultSet rs)
     {
-        return new ResultSetCleanable(rs);
+        return () -> rs.close();
```
qualidafial: Meh: you could use `rs::close` instead of a lambda. They're semantically equivalent--I just like the way method handles look.

Not critical for merge, obviously.. :)

qualidafial:
Looks good. :shipit:

stevenschlansker:
whoops, fixed the method handle, good catch.

PR #387, opened 2016-06-02
Move mapping classes into a separate package

stevenschlansker:
InferredMapperFactory -> InferredRowMapperFactory
Move RegisteredRowMapper to a lambda on SqlStatement

(requires #385)


core/src/main/java/org/jdbi/v3/mapper/InferredRowMapperFactory.java
```diff
@@ -11,21 +11,22 @@
...
 import static org.jdbi.v3.Types.findGenericParameter;
 
 import java.lang.reflect.Type;
 import java.util.Optional;
 
+import org.jdbi.v3.StatementContext;
 import org.jdbi.v3.tweak.RowMapper;
 
-class InferredMapperFactory implements RowMapperFactory
+public class InferredRowMapperFactory implements RowMapperFactory
```
qualidafial: If this class is going public, we probably want to add a short javadoc blurb explaining what it's for.

qualidafial:
One minor comment to consider, then :shipit:

qualidafial:
Move `MappingRegistry` too?

stevenschlansker:
I wasn't planning to move `MappingRegistry` or `ArgumentRegistry` since we would have to make them public, and they are fairly tied to `JdbiConfig`.

PR #390, opened 2016-06-02
SqlStatement: Remove a bunch of warning suppression by doing it once in the constructor

stevenschlansker:


core/src/main/java/org/jdbi/v3/SqlStatement.java
```diff
@@ -66,6 +67,7 @@
     private       RewrittenStatement rewritten;
     private       PreparedStatement  stmt;
 
+    @SuppressWarnings("unchecked")
```
qualidafial: Take that, type checker!

qualidafial:
Looks good. Merging.

PR #391, opened 2016-06-03
Move StatementLocators out of fluent API

qualidafial:
This PR extracts the statement locator concept out of the fluent API. Going forward the fluent API will accept only SQL.

`SqlLocator` and `ClasspathSqlLocator` will live in core, but only as a convenience class for looking up SQL imperatively:

    handle.createQuery(ClasspathSqlLocator.findSqlOnClasspath(MyDao.class, "findById"))
          .bind("id", id)
          .mapTo(Foo.class)
          .findOnly();

A new artifact, `jdbi3-string-template` will host `StringTemplateSqlLocator` and the `@UseStringTemplateSqlLocator` annotation, separating out the dependency on StringTemplate.

The `SqlObject` config class will get a new `SqlLocator` configuration property, where the default locator (`AnnotationSqlLocator`, which will also live in `sqlobject`) returns the SQL in the annotation itself:

    class MyDao {
      @SqlQuery("select * from foo where id = :id")
      Foo findById(long id);
    }

If users want to use a different locator, they can either configure it application-wide:

    dbi.configure(SqlObject.class, config -> {
      config.setSqlLocator(new ClasspathSqlLocator());
    });
    
    class MyDao {
      Foo findById(long id);
    }

..or on a case-by-case basis with class or method annotations:

    @UseClasspathSqlLocator
    class MyDao {
      Foo findById(long id);
    }

To do:

* [x] Introduce `locator` subpackage in core
  * [x] Add `ClasspathSqlLocator` class
* [x] sqlobject:
  * [x] `SqlLocator` interface will have two arguments: extension class, and extension method.
  * [x] Add `AnnotationSqlLocator` implementation in `sqlobject`, which returns the value in the method's `@Sql*` annotation.
  * [x] Add a `SqlLocator` property to `SqlObject` config class, with `AnnotationSqlLocator` as the default.
  * [x] Introduce `@SqlObjectConfiguringAnnotation` meta-annotation, where `value()` is a class that implements `SqlObjectConfigurerFactory`, which in turn creates `SqlObjectConfigurer` instances. These configurers will be applied before creating the statement (or resolving the SQL).
  * [x] Add `@UseClasspathSqlLocator` annotation and factory.
* [x] Introduce `jdbi3-stringtemplate` artifact
  * [x] `StringTemplateSqlLocator` implementation
  * [x] `@UseStringTemplateSqlLocator` implementation
* [x] sqlobject: Resolve the SQL through the configured `SqlLocator` instead of directly from the annotation.
* [x] Identify all existing `StatementLocator` tests, and migrate them to use the corresponding `SqlLocator` APIs.
* [x] Remove `StatementLocator` and all implementations, and amend all classes that use them.
* [x] Rename `SqlObject` to `SqlObjectConfig`? The shorter name seems nice but it's a bit of a misnomer.


qualidafial:
@stevenschlansker Any reason not to just cache all `StringTemplateGroup`s in a weak hash map, instead of having to ask for it in the annotation?

qualidafial:
@stevenschlansker Do you have time to review this? It's a bit of a doozy..

string-template/src/main/java/org/jdbi/v3/stringtemplate/StringTemplateSqlLocator.java
```diff
@@ -0,0 +1,70 @@
...
+    private StringTemplateSqlLocator() {
+    }
+
+    public static String findStringTemplateSql(Class<?> sqlObjectType, String name) {
+        StringTemplateGroup group = CACHE.computeIfAbsent(sqlObjectType, StringTemplateSqlLocator::readGroupResource);
+
+        if (!group.isDefined(name)) {
+            throw new IllegalStateException("No StringTemplate group " + name + " for class " + sqlObjectType);
+        }
+        return group.getInstanceOf(name).getTemplate();
```
qualidafial: This `getTemplate()` call is the only thing I have reservations about. I'm uncertain whether references between templates in a group will resolve:

    group Foo;
    select() ::= << select * from foo >>
    selectById() ::= << $select() where id = :id >>

qualidafial:
Should we update to StringTemplate 4 while we're at it?

qualidafial:
@stevenschlansker ping?

stevenschlansker:
Sorry, another crazy week... I'm 👍  for upgrading to ST 4 but let's do that as a followup.  Reading through this now.

core/src/main/java/org/jdbi/v3/locator/ClasspathSqlLocator.java
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * Locates SQL for the given extension type and name. Example: Given an extension type <code>com.foo.Bar</code> and
+     * a name of <code>baz</code>, looks for a resource named <code>com/foo/Bar/baz.sql</code> on the classpath and
+     * returns its contents as a String.
+     *
+     * @param extensionType the extension type
+     * @param name          the SQL statement name (usually a method name of the extension type).
+     * @return the located SQL.
+     */
+    public static String findSqlOnClasspath(Class<?> extensionType, String name) {
```
stevenschlansker: Is this really limited to extension types?  Or should we term it just as a class, same way `class.getResource()` works?

core/src/main/java/org/jdbi/v3/locator/ClasspathSqlLocator.java
```diff
@@ -0,0 +1,102 @@
...
+        String path = resourcePathFor(extensionType, name);
+        return findSqlOnClasspath(extensionType.getClassLoader(), path);
```
stevenschlansker: There's a bit of a difference here between selecting by `String` vs `Class` -- whether the context class loader is considered.  Should the context class loader override the `class.getClassLoader()`?  (I'm honestly not sure, I've not used the context class loader much at all)

core/src/test/java/org/jdbi/v3/locator/TestClasspathSqlLocator.java
```diff
@@ -0,0 +1,125 @@
...
+public class TestClasspathSqlLocator {
+    @Rule
+    public H2DatabaseRule db = new H2DatabaseRule();
+
+    @Rule
+    public ExpectedException exception = ExpectedException.none();
+
+    @Before
+    public void setUp() {
+    }
```
stevenschlansker: Dead code

core/src/test/java/org/jdbi/v3/locator/TestClasspathSqlLocator.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Test
+    public void testNonExistentResource() throws Exception {
+        exception.expect(IllegalArgumentException.class);
+        findSqlOnClasspath("this-does-not-exist");
+    }
+
+    @Test
+    public void testCachesResultAfterFirstLookup() throws Exception {
+        ClassLoader ctx_loader = Thread.currentThread().getContextClassLoader();
```
stevenschlansker: Non-standard naming conventions

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectConfigurer.java
```diff
@@ -13,11 +13,15 @@
...
-public class SqlObject implements ExtensionConfig<SqlObject> {
-    @Override
-    public SqlObject createCopy() {
-        return new SqlObject();
-    }
+/**
+ * Configures a {@link SqlObjectConfig}.
+ */
+@FunctionalInterface
+public interface SqlObjectConfigurer {
```
stevenschlansker: Is this useful as a standalone interface, or would `Consumer<SqlObjectConfig>` work as well?

stevenschlansker:
Couple of comments, probably most don't need any action.  Then :shipit: !

core/src/main/java/org/jdbi/v3/locator/ClasspathSqlLocator.java
qualidafial: This isn't restricted to extensions. I will rename.

core/src/test/java/org/jdbi/v3/locator/TestClasspathSqlLocator.java
qualidafial: Yeah, some of these tests are just copy pasted from the old StatementLocator stuff--I will clean this up

core/src/main/java/org/jdbi/v3/locator/ClasspathSqlLocator.java
qualidafial: This is actually just preserving some of the behavior from the old `ClasspathStatementLocator`. I wasn't sure why it was needed, but figured there could be a good reason for it.

If you like, I can change the `findSqlOnClasspath(String)` method to `findSqlOnClasspath(ClassLoader, String)` and leave it up to the caller to specify the correct class loader.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectConfigurer.java
qualidafial: I don't remember why I did that as a separate interface. Will use `Consumer` until I can think of a good reason not to.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -92,16 +97,24 @@ public boolean accepts(Class<?> extensionType) {
...
-    private Map<Method, Handler> buildHandlersFor(Class<?> sqlObjectType, SqlObject config) {
+    private Map<Method, Handler> buildHandlersFor(Class<?> sqlObjectType, SqlObjectConfig config) {
         return handlersCache.computeIfAbsent(sqlObjectType, type -> {
 
             final Map<Method, Handler> handlers = new HashMap<>();
             for (Method method : sqlObjectType.getMethods()) {
+                // FIXME will applying configurers here and caching config create problems?
+                SqlObjectConfig methodConfig = config.createCopy();
+                forEachConfigurerFactory(method, (factory, annotation) ->
+                        factory.createForMethod(annotation, sqlObjectType, method).accept(methodConfig));
```
qualidafial: We should probably fix this before merge.

In a nutshell, whatever the `SqlObjectConfig` looks like the first time an extension class is created is cached and reused here, effectively making the default config set in stone.

What I'm thinking is that `Handler` should not receive the `SqlObjectConfig` until the invocation. Right now most handlers receive it as a constructor parameter, so this will be a little work to address. But since it's a breaking change I think it's better to do now.

qualidafial:
@stevenschlansker I believe I've addressed all your comments save one: using different classloaders in `ClasspathSqlLocator.findSqlOnClasspath` depending on whether you use the `(Class, String)` or the `(String)` variant.

Could you review the latest changes?

Do you mind if we open a separate issue for the classloader thing?

qualidafial:
Merging now to clear the path for other work. Let me know if you find any problems with the latest changes.

PR #392, opened 2016-06-03
Remove Maven Toolchains

arteam:
Toolchains are useful when we compile against different JDKs or want to use a
specific old JDK. Since we target JDK8 in JDBI3, there is no big benefit in
using toolchains. In most development installations JDK8 is default and usually
already set in the PATH environment variable, so Maven can find it without
specific assistance. We only need to set-up source and target for the Java
compiler.

Also, avoiding using toolchains simplifies the process of contributing to JDBI,
because developers don't need to add specific steps to configure the development
environment and can build the project with `mvn clean install`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,5 +12,5 @@ cache:
 before_install:
     - cp src/build/travis-toolchains.xml ~/.m2/toolchains.xml
 
-install: mvn -DskipTests=true -Dbasepom.check.skip-all=true -B install
-script: mvn -B verify
+install: mvn -Ptoolchains -DskipTests=true -Dbasepom.check.skip-all=true -B install
+script: mvn -Ptoolchains -B verify
--- a/pom.xml
+++ b/pom.xml
@@ -91,6 +91,53 @@
         <basepom.test.timeout>240</basepom.test.timeout>
     </properties>
 
+    <profiles>
+      <profile>
+        <id>toolchains</id>
+        <build>
+         <pluginManagement>
+           <plugins>
+             <plugin>
+               <groupId>org.apache.maven.plugins</groupId>
+               <artifactId>maven-compiler-plugin</artifactId>
+               <configuration>
+                 <compilerArguments>
+                   <parameters/>
+                 </compilerArguments>
+               </configuration>
+             </plugin>
+             <plugin>
+               <groupId>org.apache.maven.plugins</groupId>
+               <artifactId>maven-toolchains-plugin</artifactId>
+               <version>1.1</version>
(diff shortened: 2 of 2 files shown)


qualidafial:
👍  from me. @stevenschlansker?

stevenschlansker:
My only worry is that once Java 9 is released, I'll want to re-add this the very next day.  I think that the toolchains are fundamentally going to be necessary, the `jdbi` version will only rarely target the very latest JDK.  However, I am definitely sympathetic to the ease of setup for a new contributor.

Setting source and target versions is great, but only affects the compiler.  It does not make sure you use a compatible class library.  So you can relatively easily introduce code that targets an old JVM but does not actually work at runtime.

I'm going to ponder this a bit more.  Are you guys convinced that the simplification outweighs the likely need for it to come back in < 1 year?

qualidafial:
Could we only use the toolchain in TravisCI builds, using maven profiles? That would ease the burden for contributors, but still help identify any version-specific behaviors between Java 8 and 9.

qualidafial:
@arteam If you wouldn't mind making a few changes to this PR:

* Add a `toolchains` profile to the parent POM, and move the `maven-toolchains-plugin` stuff into that profile.
* Reinstate the `travis-toolchains.xml` stuff in `travis.yml`.
* Invoke Maven with `-Ptoolchains` in `travis.yml`.

This way we could retain the benefits of toolchains when Java 9 is released, but still get it out of the way for would-be contributors.

.travis.yml
```diff
@@ -9,8 +9,5 @@ cache:
   directories:
     - '$HOME/.m2/repository'
 
-before_install:
-    - cp src/build/travis-toolchains.xml ~/.m2/toolchains.xml
```
qualidafial: Just need to reinstate this `before_install` directive and it should work

pom.xml
```diff
@@ -91,46 +91,69 @@
...
+                    <plugins>
+                        <plugin>
+                            <groupId>org.apache.maven.plugins</groupId>
+                            <artifactId>maven-compiler-plugin</artifactId>
+                            <configuration>
+                                <compilerArguments>
+                                    <parameters/>
+                                </compilerArguments>
+                            </configuration>
+                        </plugin>
```
qualidafial: Is this plugin a leftover from trial and error?

arteam:
I've moved `maven-toolchains-plugin` to a separate profile and fixed the Travis configuration.
@qualidafial, please review.

qualidafial:
Looks good.

PR #393, opened 2016-06-03
Add support for Postgres arrays

arteam:
This PR attempts to fix issues with converting arrays between an application and a PostgreSQL JDBC driver. 

* Arrays are covariant, so we can't convert `int[]` to `Object[]` by casting. We should do conversion manually using `java.reflect.Array` which allows to work with arrays of any type.
* We should support boxing arrays as well as the primitive ones.

Unfortunately, it seems not possible to convert between arrays and lists, because JDBI considers lists as containers. I remember, we could tweak such behavior in JDBI2 by using  the `@SingleValueResult`
annotations, but it's gone in JDBI3.

### DIFF
--- a/postgres/src/main/java/org/jdbi/v3/pg/ArrayArgument.java
+++ b/postgres/src/main/java/org/jdbi/v3/pg/ArrayArgument.java
@@ -13,7 +13,7 @@
  */
 package org.jdbi.v3.pg;
 
-import java.sql.Array;
+import java.lang.reflect.Array;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 
@@ -23,11 +23,17 @@ import org.jdbi.v3.argument.Argument;
 public class ArrayArgument implements Argument {
 
     private final String elementType;
-    private final Object array;
+    private final Object[] array;
 
-    private ArrayArgument(String elementType, Object array) {
+    private ArrayArgument(String elementType, Object newArray) {
         this.elementType = elementType;
-        this.array = array;
+        int length = Array.getLength(newArray);
+        // We can't cast newArray to Object[], so we need to make a new array
+        // of Object[] and copy newArray's content to it
+        this.array = new Object[length];
+        for (int i = 0; i < length; i++) {
+            array[i] = Array.get(newArray, i);
+        }
     }
 
     public static ArrayArgument fromArray(String elementType, Object[] array) {
@@ -44,7 +50,7 @@ public class ArrayArgument implements Argument {
 
     @Override
     public void apply(int position, PreparedStatement statement, StatementContext ctx) throws SQLException {
-        Array sqlArray = statement.getConnection().createArrayOf(elementType, (Object[]) array);
+        java.sql.Array sqlArray = statement.getConnection().createArrayOf(elementType, array);
         statement.setArray(position, sqlArray);
     }
 }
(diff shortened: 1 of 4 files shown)


postgres/src/main/java/org/jdbi/v3/pg/ArrayColumnMapper.java
```diff
@@ -34,17 +34,23 @@ public ArrayColumnMapper(Class<?> componentType, StatementContext ctx) {
...
             return buildFromResultSet(array);
         }
         try {
-            // more efficient, doesn't have to copy an unknown size array
-            return (Object[]) array.getArray();
+            // We can't cast Object[] to componentType, so we need to create a new array of the correct
+            // type and copy the source array content to it.
+            Object[] objectArray = (Object[]) array.getArray();
+            Object componentTypeArray = java.lang.reflect.Array.newInstance(componentType, objectArray.length);
+            for (int i = 0; i < objectArray.length; i++) {
```
qualidafial: Could we skip this loop if the component type of the object array is already correct?

qualidafial:
We can add back `@SingleValueResult` if it's needed. We only removed it since `CollectorFactory` removed the need for it, pre-array support.

postgres/src/main/java/org/jdbi/v3/pg/ArrayColumnMapper.java
arteam: Looks like a reasonable optimization.

postgres/src/test/java/org/jdbi/v3/pg/TestSqlArrays.java
```diff
@@ -87,7 +98,7 @@ public void testIntList() throws Exception {
         UUID[] fetchUuidArray();
 
         @SqlUpdate(U_INSERT)
-        void insertUuidArray(UUID[] u);
+        void insertUuidArray(@Bind("uuids") UUID[] u);
```
arteam: Do I need to compile the class with the debug information to make that work? Currently I get
```
java.lang.UnsupportedOperationException: A parameter was not given a name, and parameter name data is not present in the class file, for: public abstract void org.jdbi.v3.pg.TestSqlArrays$ArrayObject.insertUuidArray(java.util.UUID[]) :: java.util.UUID[] arg0
```

qualidafial:
In fact let's add it as `@SingleValue` or maybe `@SingleRow` so it can double as the annotation for the return value and non-batch parameter on e.g. a batch execute:

    @SqlQuery
    @SingleValue
    List<String> getArrayColumn(...);
    
    @SqlBatch
    int insertBlahs(List<String> names, @SingleValue List<String> tags);

postgres/src/test/java/org/jdbi/v3/pg/TestSqlArrays.java
qualidafial: Yes. This is the default setting when you run the Maven build. In your IDE you need to enable the `-parameters` compiler option.

arteam:
I will take a look about handling `@SingleValue` and return back to you.

postgres/src/main/java/org/jdbi/v3/pg/ArrayArgument.java
```diff
@@ -23,11 +23,17 @@
...
-    private final Object array;
+    private final Object[] array;
 
-    private ArrayArgument(String elementType, Object array) {
+    private ArrayArgument(String elementType, Object newArray) {
         this.elementType = elementType;
-        this.array = array;
+        int length = Array.getLength(newArray);
+        // We can't cast newArray to Object[], so we need to make a new array
+        // of Object[] and copy newArray's content to it
```
stevenschlansker: :+1: good idea

stevenschlansker:
Nice.  I believe this unambiguously improves the code, so I'll merge it even if we need to do followup.
I really like the `@SingleValue` idea.

PR #394, opened 2016-06-03
Add support for JSR-310 time types

arteam:
They correspondent to the SQL TIME type. `OffsetTime` has a limited usefulness, but it's included for the consistency sake.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/mapper/BuiltInMapperFactory.java
+++ b/core/src/main/java/org/jdbi/v3/mapper/BuiltInMapperFactory.java
@@ -22,6 +22,7 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Time;
 import java.sql.Timestamp;
 import java.time.Instant;
 import java.time.LocalDate;
@@ -29,6 +30,7 @@ import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.time.LocalTime;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jdbi/v3/mapper/BuiltInMapperFactory.java
```diff
@@ -163,4 +169,15 @@ private static ZonedDateTime getZonedDateTime(ResultSet r, int i) throws SQLExce
         Timestamp ts = r.getTimestamp(i);
         return ts == null ? null : ZonedDateTime.ofInstant(ts.toInstant(), ZoneId.systemDefault());
     }
+
+    private static OffsetTime getOffsetTime(ResultSet r, int i) throws SQLException {
+        Time time = r.getTime(i);
+        ZoneOffset zoneOffset = Instant.now().atZone(ZoneId.systemDefault()).getOffset();
```
qualidafial: This one is a problem: if somebody's system default zone observes daylight savings time, then offset will shift by an hour every time the zone offset shifts.

Example:

I live in the `United States/Denver` zone, which observes Daylight savings. Because I'm a rank amateur, I run my application server in local time instead of UTC.

During the winter, I store an `OffsetDateTime` of `15:10:05-07:00`. Because JDBC hasn't caught up with JSR-310, it gets saved to the database as just `15:10:05`.

Later, during the summer (after daylight savings has gone into effect), I query that row from the database, and get back `15:10:05`. The zone offset at this time is `-06:00`, so the result is `15:10:05-06:00`.

This makes me wonder whether JDBI could be a little opinionated on this one, and just store all time types in UTC.

arteam: That's a valid concern. We could probably support only `LocalTime`, because  `OffsetTime` is to ambiguous.
We have no means to store/retrieve the time zone, because JDBC doesn't provide such API.

arteam: I've removed support for  `OffsetTime`.

qualidafial: _sigh_

qualidafial:
Thanks @arteam!

I commented somewhere else that the Postgres drivers support `LocalDate`, `LocalTime`, `LocalDateTime` and `OffsetDateTime` in the driver through the `ResultSet.setObject(int, Class)` and `PreparedStatement.setObject` methods. If you'd care to open a PR adding support for those classes to `jdbi3-postgres`, I'd be happy to merge it.

We're also interested in new artifacts to add support for other JDBC drivers and the custom types they support.

PR #396, opened 2016-06-10
travis: don't do separate install / verify runs, it doubles the time

stevenschlansker:
needed?


stevenschlansker:
Nope, same time

PR #401, opened 2016-06-17
Use Postgres driver support for JSR-310 types

qualidafial:


postgres/src/main/java/org/jdbi/v3/pg/Jsr310ArgumentFactory.java
```diff
@@ -0,0 +1,48 @@
...
+import java.time.LocalTime;
+import java.time.OffsetDateTime;
+import java.util.Optional;
+
+import org.jdbi.v3.StatementContext;
+import org.jdbi.v3.argument.Argument;
+import org.jdbi.v3.argument.ArgumentFactory;
+import org.jdbi.v3.argument.ObjectArgument;
+
+public class Jsr310ArgumentFactory implements ArgumentFactory {
```
stevenschlansker: Jackson has started renaming all their classes from `Jsr310` to `JavaTime` now that it's actually part of the JDK.  Should we follow their lead?

PR #406, opened 2016-07-01
@OutParameter annotation for @SqlCall methods

### LINKED ISSUE #405: `@OutParameter` annotation
We should add a declarative annotation for registering out-parameters, as suggested by this user on the mailing list:

https://groups.google.com/d/msg/jdbi/yShp_-7Q7wM/8h99BCSqu5sJ

stevenschlansker:
Fixes #405

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,6 +1,7 @@
 2.74
   - cglib 3.2.2, asm 5.1; fixes codegen for new Java 8 bridge methods
   - @UseStringTemplate3StatementLocator now caches created locators
+  - new @OutParameter annotation for fetching named out params on @SqlCall methods
 
 2.73
   - Allow clearing of bindings in SQLStatement
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/customizers/OutParameter.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/customizers/OutParameter.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


src/test/java/org/skife/jdbi/v2/sqlobject/TestOutParameters.java
```diff
@@ -0,0 +1,95 @@
...
+        OutParameters callStoredProc(@Bind("outParam1") Integer outParam1);
+
+        @SqlUpdate("CREATE OR REPLACE PROCEDURE procOneOUTParameter(outParam1 OUT NUMBER) IS BEGIN outParam1 := 100; END")
+        void insertStoredProc();
+    }
+
+    @SqlStatementCustomizingAnnotation(OutParameter.Factory.class)
+    @Target(ElementType.METHOD)
+    @Retention(RetentionPolicy.RUNTIME)
+    public @interface OutParameter {
```
qualidafial: Could we add this add this annotation to the SQL Object API while we're at it?

stevenschlansker:
@qualidafial After playing with this a bit and looking over the `Call` code, it looks like implementing this in v2 will not be easy.  The code is a bit of a mess and is clearly one of the least-loved corners of the codebase.  This is in no small part due to the frankly piss-poor design of the JDBC `CallableStatement` API.

I am thinking this might need to be deferred to v3, unless someone in particular needs v2 support and is willing to help implement it there.

qualidafial:
Do you have a sense of how you would change the `Call` class in v3?

stevenschlansker:
Well, at the very least, get rid of `SQLStatement` doing some `if (getClass().isSubclassOf(Call.class))` bullshit.  And I think the fact that OutParameters are registered as Parameters that don't ever actually bind to the statement is pretty bogus too..

stevenschlansker:
PS @qualidafial I will be out for the next week in Michigan, so apologies if I'm not checking GitHub much.

stevenschlansker:
I think I was wrong.  It's just not obvious how it works, but if you construct your SQL right (WTF is this `{call myStoredProc()}` BS?) it actually works fine.

qualidafial:
Looks good.

Issue #195, opened 2015-10-22
@Transaction should not be exclusive of @SqlQuery, @SqlUpdate, etc


stevenschlansker:
See #86, right now `@Transaction` and other annotations are all exclusive, this is not so great.

stevenschlansker:
Also add an example of both a single and multiple statement `@Transaction` method.

qualidafial:
Looking at this.

The idea I'm working from is to add a new meta-annotation for decorator-style `Handler`s.

Then the transaction handler would simply be a decorator around the final handler, whether that be the pass-through handler (for default or concrete methods), or the annotation-specific handler for the method.

This also opens up the possibility for supporting other types of decorators--caching comes to mind as a possible use case.

qualidafial:
This is addressed in #411

qualidafial:
Resolved in #411

Issue #194, opened 2015-10-22
@Transaction should somehow support Spring-style "ensure we are part of a transaction" semantics


stevenschlansker:
See #86.  Spec to be written here.

qualidafial:
Planning to tackle this along with #195

qualidafial:
This is addressed in #411

qualidafial:
Resolved in #411

PR #411, opened 2016-07-09
@Transaction ALL THE THINGS

qualidafial:
Addresses issues #194, #195:

* Make `@Transaction` combinable with sql method annotations e.g. `@SqlUpdate`, `@SqlCall`, etc.
* Make nested calls to `@Transaction`-annotated methods result in a single transaction (instead of throwing an exception). The transaction isolation level is determined by the annotation on the outermost (first) method in the call chain.

@stevenschlansker Do you have time to review this before you leave? This is a small PR.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
@@ -13,6 +13,8 @@
  */
 package org.jdbi.v3.sqlobject;
 
+import static java.util.stream.Collectors.toList;
+
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Method;
@@ -21,7 +23,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.function.BiConsumer;
@@ -67,7 +68,9 @@ public enum SqlObjectFactory implements ExtensionFactory<SqlObjectConfig> {
 
         return Stream.of(extensionType.getMethods())
                 .flatMap(m -> Stream.of(m.getAnnotations()))
-                .anyMatch(a -> a.annotationType().isAnnotationPresent(SqlMethodAnnotation.class));
+                .map(a -> a.annotationType())
+                .anyMatch(type -> type.isAnnotationPresent(SqlMethodAnnotation.class) ||
+                        type.isAnnotationPresent(SqlMethodDecoratingAnnotation.class));
     }
 
     /**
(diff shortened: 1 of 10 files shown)


stevenschlansker:
Yes, I'll get a review of this in today.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -67,7 +68,9 @@ public boolean accepts(Class<?> extensionType) {
 
         return Stream.of(extensionType.getMethods())
                 .flatMap(m -> Stream.of(m.getAnnotations()))
-                .anyMatch(a -> a.annotationType().isAnnotationPresent(SqlMethodAnnotation.class));
+                .map(a -> a.annotationType())
+                .anyMatch(type -> type.isAnnotationPresent(SqlMethodAnnotation.class) ||
+                        type.isAnnotationPresent(SqlMethodDecoratingAnnotation.class));
```
qualidafial: This check for `SqlMethodDecoratingAnnotation` might not be necessary. I'm still thinking it through, but I believe that testing for mixins interfaces or `SqlMethodAnnotation` should be sufficient.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -105,38 +108,55 @@ public boolean accepts(Class<?> extensionType) {
...
+        return addDecorators(handler, sqlObjectType, method);
+    }
+
+    private Handler buildBaseHandler(Class<?> sqlObjectType, Method method) {
+        return Stream.of(method.getAnnotations())
+                .map(a -> a.annotationType().getAnnotation(SqlMethodAnnotation.class))
+                .filter(Objects::nonNull)
+                .map(a -> buildFactory(a.value()))
+                .map(factory -> factory.buildHandler(sqlObjectType, method))
+                .findFirst()
```
stevenschlansker: I worry that this depends on the iteration order of annotations on your method, which is inherently undefined.  Should we impose an ordering here, or fail if there are multiple annotations matching?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -105,38 +108,55 @@ public boolean accepts(Class<?> extensionType) {
...
+                .orElseGet(PassThroughHandler::new);
+    }
+
+    private Handler addDecorators(Handler handler, Class<?> sqlObjectType, Method method) {
+        List<HandlerDecorator> decorators = Stream.of(method.getAnnotations())
+                .map(a -> a.annotationType().getAnnotation(SqlMethodDecoratingAnnotation.class))
```
stevenschlansker: Ditto here, the ordering might not be well defined?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
qualidafial: Probably wouldn't hurt to fail on multiple annotations here.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/TransactionDecorator.java
```diff
@@ -21,23 +21,25 @@
...
     }
 
     @Override
     public Object invoke(Supplier<Handle> handle, SqlObjectConfig config, Object target, Object[] args, Method method) throws Exception
     {
         Handle h = handle.get();
+
         if (h.isInTransaction()) {
-            throw new TransactionException("Nested @Transaction detected - this is currently not supported.");
+            // Already in transaction. The outermost @Transaction method determines the transaction isolation level.
```
stevenschlansker: Should this check that the inner txn level matches the outer, and bail if not?  Otherwise you might accidentally execute code intended to be run at `SERIALIZABLE` level at a less strict level like `READ COMMITTED`.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
qualidafial: Correct, but as `HandlerDecorator` is internal and there's only one implementation so far, we don't have to solve this just yet.

I've debated between a `priority` property on the `@SqlMethodDecoratingAnnotation` vs some comparator so clients can set their own priority. No firm decisions yet.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -146,6 +166,16 @@ private HandlerFactory buildFactory(Class<? extends HandlerFactory> factoryClazz
         return factory;
     }
 
+    private static HandlerDecorator buildDecorator(Class<? extends HandlerDecorator> decoratorClass) {
+        HandlerDecorator decorator;
+        try {
+            decorator = decoratorClass.newInstance();
```
qualidafial: Will fix.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/TransactionDecorator.java
qualidafial: Sounds like a good idea. I will fix.

stevenschlansker:
Couple thoughts.  Only real issues: I'm worried that the Decorator is underspecified because the enumeration order of annotations is not well defined.  I'm also thinking that if there are nested `@Transaction` annotations with mismatched isolation levels that should be an error rather than simply running code under an unexpected level.  Other than that looks great!

sqlobject/src/main/java/org/jdbi/v3/sqlobject/TransactionDecorator.java
qualidafial: Is there a monotonic order of strictness, so that I could increase the transaction isolation in the nested call, but not decrease it?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
stevenschlansker: Could consider using the standard `@Priority` annotation instead of introducing our own?

qualidafial: Sure

sqlobject/src/main/java/org/jdbi/v3/sqlobject/TransactionDecorator.java
stevenschlansker: I believe the common 4 are in increasing order of strictness, but I am not sure this means that it is safe to increase mid transaction.  Do databases even support that?  Also, things run under `REPEATABLE READ` or `SERIALIZABLE` may spuriously abort and need to be retried, which code expecting to run at `READ COMMITTED` would not be able to handle.  I'm worried that being too lenient here just doesn't make sense, and lean towards forbidding it unless we can come up with a sensible use case other than trying to make sloppy / broken code work.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
stevenschlansker: I'm OK with merging with this unfixed as long as a followup issue is logged.

stevenschlansker:
This should also add documentation about this behavior! 😉

sqlobject/src/main/java/org/jdbi/v3/sqlobject/TransactionDecorator.java
qualidafial: Ok, I'll add a check to ensure the isolation level matches.

What about a method with `@Transaction(READ_COMMITTED)` that calls a method with a plain `@Transaction`? Should a bare `@Transaction` just go with the flow?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
qualidafial: I'll just fix it now

qualidafial: Hmm.. `java.annotation.Priority` is Java EE

sqlobject/src/main/java/org/jdbi/v3/sqlobject/TransactionDecorator.java
stevenschlansker: Personally I feel any mismatch should just blow up.  The transaction isolation levels affect execution of the code, so anything that suddenly changes it from under you can introduce subtle and hard to reproduce bugs.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
qualidafial: I'm not positive yet on a good solution. Let's hold off on this until it's strictly necessary.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/TransactionDecorator.java
qualidafial: I'm not sure I can do that.

The default transaction isolation level can vary by database vendor, so I can't hardcode a specific level to expect as the "default" in a nested call to bare `@Transaction` methods.

PR #412, opened 2016-07-18
Correctly escape the at-sign in code examples in Javadoc

arteam:
The correct representations of the at-sign(@) in XML is code `&#064;` rather than `&amp;` which correspondents to the ampersand (&) symbol .


qualidafial:
Lol, whoops. Good catch.

PR #415, opened 2016-07-23
Decrease ClasspathStatementLocator GC pressure.

john9x:
Hi!

ClasspathStatementLocator gives essential GC pressure due to many string concatenation for every query even if static SQL objects are used.
```
cache_key = '/' + mungify(ctx.getSqlObjectType().getName() + '.' + name) + ".sql";
```
I propose to use Method object as cache key, this is completely enough for SQL objects.

And also I propose to cache individual SQL statements for SQL object methods cause they are static.

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
+++ b/src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
@@ -25,6 +25,7 @@ import java.io.InputStream;
 import java.util.Collections;
 import java.util.Map;
 import java.util.WeakHashMap;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
 
 /**
@@ -32,7 +33,7 @@ import java.util.regex.Matcher;
  */
 public class ClasspathStatementLocator implements StatementLocator
 {
-    private final Map<String, String> found = Collections.synchronizedMap(new WeakHashMap<String, String>());
+    private final Map<Object, String> found = Collections.synchronizedMap(new WeakHashMap<Object, String>());
 
     /**
      * Very basic sanity test to see if a string looks like it might be sql
@@ -74,20 +75,32 @@ public class ClasspathStatementLocator implements StatementLocator
     @SuppressFBWarnings("DM_STRING_CTOR")
     public String locate(String name, StatementContext ctx)
     {
-        final String cache_key;
+        final Object cache_key;
+        boolean isSqlObjectMethod = false;
         if (ctx.getSqlObjectType() != null) {
-            cache_key = '/' + mungify(ctx.getSqlObjectType().getName() + '.' + name) + ".sql";
+            if (ctx.getSqlObjectMethod() != null) {
+                isSqlObjectMethod = true;
+                cache_key = ctx.getSqlObjectMethod();
+            } else {
+                cache_key = '/' + mungify(ctx.getSqlObjectType().getName() + '.' + name) + ".sql";
+            }
         }
         else {
             cache_key = name;
         }
 
-        if (found.containsKey(cache_key)) {
-            return found.get(cache_key);
+        String cached = found.get(cache_key);
+        if (cached != null) {
+            return cached;
         }
+
 
         if (looksLikeSql(name)) {
             // No need to cache individual SQL statements that don't cause us to search the classpath
+            // But for static SQL object methods caching decreases GC pressure and not threatens memory leaks.
+            if (isSqlObjectMethod) {
+                found.put(cache_key, name);
+            }
             return name;
         }
         final ClassLoader loader = selectClassLoader();
--- a/src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
+++ b/src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
@@ -39,7 +39,7 @@ public class ClasspathStatementLocator implements StatementLocator
         final Class<?> sqlObjectType;
         final Method sqlObjectMethod;
 
-        public CacheKey(String name, Class<?> sqlObjectType, Method sqlObjectMethod) {
+        CacheKey(String name, Class<?> sqlObjectType, Method sqlObjectMethod) {
             this.name = name;
             this.sqlObjectType = sqlObjectType;
             this.sqlObjectMethod = sqlObjectMethod;
@@ -47,14 +47,21 @@ public class ClasspathStatementLocator implements StatementLocator
 
         @Override
         public boolean equals(Object o) {
-            if (this == o) return true;
-            if (o == null || getClass() != o.getClass()) return false;
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
 
             CacheKey cacheKey = (CacheKey) o;
 
-            if (name != null ? !name.equals(cacheKey.name) : cacheKey.name != null) return false;
-            if (sqlObjectType != null ? !sqlObjectType.equals(cacheKey.sqlObjectType) : cacheKey.sqlObjectType != null)
+            if (name != null ? !name.equals(cacheKey.name) : cacheKey.name != null) {
                 return false;
+            }
+            if (sqlObjectType != null ? !sqlObjectType.equals(cacheKey.sqlObjectType) : cacheKey.sqlObjectType != null) {
+                return false;
+            }
             return sqlObjectMethod != null ? sqlObjectMethod.equals(cacheKey.sqlObjectMethod) : cacheKey.sqlObjectMethod == null;
 
         }
--- a/src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
+++ b/src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
@@ -53,22 +53,19 @@ public class ClasspathStatementLocator implements StatementLocator
             if (o == null || getClass() != o.getClass()) {
                 return false;
             }
+            CacheKey that = (CacheKey) o;
+            return eq(this.name, that.name)
+                    && eq(this.sqlObjectType, that.sqlObjectType)
+                    && eq(this.sqlObjectMethod, that.sqlObjectMethod);
+        }
 
-            CacheKey cacheKey = (CacheKey) o;
-
-            if (name != null ? !name.equals(cacheKey.name) : cacheKey.name != null) {
-                return false;
-            }
-            if (sqlObjectType != null ? !sqlObjectType.equals(cacheKey.sqlObjectType) : cacheKey.sqlObjectType != null) {
-                return false;
-            }
-            return sqlObjectMethod != null ? sqlObjectMethod.equals(cacheKey.sqlObjectMethod) : cacheKey.sqlObjectMethod == null;
-
+        private boolean eq(Object left, Object right) {
+            return left == null ? right == null : left.equals(right);
         }
 
         @Override
         public int hashCode() {
-            int result = name != null ? name.hashCode() : 0;
+            int result = name == null ? 0 : name.hashCode();
             result = 31 * result + (sqlObjectType != null ? sqlObjectType.hashCode() : 0);
             result = 31 * result + (sqlObjectMethod != null ? sqlObjectMethod.hashCode() : 0);
             return result;
--- a/src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
+++ b/src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
@@ -22,6 +22,7 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.reflect.Method;
 import java.util.Collections;
 import java.util.Map;
 import java.util.WeakHashMap;
@@ -33,7 +34,41 @@ import java.util.regex.Matcher;
  */
 public class ClasspathStatementLocator implements StatementLocator
 {
-    private final Map<Object, String> found = Collections.synchronizedMap(new WeakHashMap<Object, String>());
+    private static class CacheKey {
+        final String name;
+        final Class<?> sqlObjectType;
+        final Method sqlObjectMethod;
+
+        public CacheKey(String name, Class<?> sqlObjectType, Method sqlObjectMethod) {
+            this.name = name;
+            this.sqlObjectType = sqlObjectType;
+            this.sqlObjectMethod = sqlObjectMethod;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            CacheKey cacheKey = (CacheKey) o;
+
+            if (name != null ? !name.equals(cacheKey.name) : cacheKey.name != null) return false;
+            if (sqlObjectType != null ? !sqlObjectType.equals(cacheKey.sqlObjectType) : cacheKey.sqlObjectType != null)
+                return false;
+            return sqlObjectMethod != null ? sqlObjectMethod.equals(cacheKey.sqlObjectMethod) : cacheKey.sqlObjectMethod == null;
+
+        }
+
+        @Override
+        public int hashCode() {
+            int result = name != null ? name.hashCode() : 0;
+            result = 31 * result + (sqlObjectType != null ? sqlObjectType.hashCode() : 0);
+            result = 31 * result + (sqlObjectMethod != null ? sqlObjectMethod.hashCode() : 0);
+            return result;
+        }
+    }
+
+    private final Map<CacheKey, String> found = Collections.synchronizedMap(new WeakHashMap<CacheKey, String>());
 
     /**
      * Very basic sanity test to see if a string looks like it might be sql
@@ -75,19 +110,9 @@ public class ClasspathStatementLocator implements StatementLocator
     @SuppressFBWarnings("DM_STRING_CTOR")
     public String locate(String name, StatementContext ctx)
     {
-        final Object cache_key;
-        boolean isSqlObjectMethod = false;
-        if (ctx.getSqlObjectType() != null) {
-            if (ctx.getSqlObjectMethod() != null) {
-                isSqlObjectMethod = true;
-                cache_key = ctx.getSqlObjectMethod();
-            } else {
-                cache_key = '/' + mungify(ctx.getSqlObjectType().getName() + '.' + name) + ".sql";
-            }
-        }
-        else {
-            cache_key = name;
-        }
+        final CacheKey cache_key = new CacheKey(name, ctx.getSqlObjectType(), ctx.getSqlObjectMethod());
+        boolean isSqlObjectMethod = ctx.getSqlObjectType() != null && ctx.getSqlObjectMethod() != null;
+
 
         String cached = found.get(cache_key);
         if (cached != null) {
@@ -122,7 +147,7 @@ public class ClasspathStatementLocator implements StatementLocator
             if (in_stream == null) {
                 // Ensure we don't store an identity map entry which has a hard reference
                 // to the key (through the value) by copying the value, avoids potential memory leak.
-                found.put(cache_key, name == cache_key ? new String(name) : name);
+                found.put(cache_key, isSqlObjectMethod ? name : new String(name));
                 return name;
             }
             String sql;
--- a/src/test/java/org/skife/jdbi/v2/sqlobject/TestClasspathStatementLocator.java
+++ b/src/test/java/org/skife/jdbi/v2/sqlobject/TestClasspathStatementLocator.java
@@ -38,6 +38,7 @@ public class TestClasspathStatementLocator {
         handle = dbi.open();
 
         handle.execute("create table something (id int primary key, name varchar(100))");
+        handle.execute("insert into something (id, name) values (6, 'Martin')");
     }
 
     @After
@@ -48,10 +49,24 @@ public class TestClasspathStatementLocator {
 
     @Test
     public void testBam() throws Exception {
-        handle.execute("insert into something (id, name) values (6, 'Martin')");
-
         Something s = handle.attach(Cromulence.class).findById(6L);
         assertThat(s.getName(), equalTo("Martin"));
+    }
+
+    @Test
+    public void testOverride() throws Exception {
+        Something s = handle.attach(SubCromulence.class).findById(6L);
+        assertThat(s.getName(), equalTo("overridden"));
+    }
+
+    @Test
+    public void testCachedOverride() throws Exception {
+        Something s = handle.attach(Cromulence.class).findById(6L);
+        assertThat(s.getName(), equalTo("Martin"));
+
+        // and now make sure we don't accidentally cache the statement from above
+        s = handle.attach(SubCromulence.class).findById(6L);
+        assertThat(s.getName(), equalTo("overridden"));
     }
 
     @RegisterMapper(SomethingMapper.class)
@@ -59,4 +74,7 @@ public class TestClasspathStatementLocator {
         @SqlQuery
         public Something findById(@Bind("id") Long id);
     }
+
+    @RegisterMapper(SomethingMapper.class)
+    static interface SubCromulence extends Cromulence { }
 }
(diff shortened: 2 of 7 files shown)


qualidafial:
Hi @john9x! Thanks for your observations, and for submitting a PR.

Have you done any before-and-after measurements to validate that this change actually reduces GC pressure?

src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -74,20 +75,32 @@ public static boolean looksLikeSql(String sql)
     @SuppressFBWarnings("DM_STRING_CTOR")
     public String locate(String name, StatementContext ctx)
     {
-        final String cache_key;
+        final Object cache_key;
+        boolean isSqlObjectMethod = false;
         if (ctx.getSqlObjectType() != null) {
-            cache_key = '/' + mungify(ctx.getSqlObjectType().getName() + '.' + name) + ".sql";
+            if (ctx.getSqlObjectMethod() != null) {
```
qualidafial: Is there ever a case where `sqlObjectType` is set, but `sqlObjectMethod` is not?

john9x:
I can't give you concrete measurements right now.
I observed GC pressure from ClasspathStatmentLocator during profiling my application in the production environment.
In my case best solution was to replace ClasspathStatementLocator with stub. 

>Is there ever a case where sqlObjectType is set, but sqlObjectMethod is not?

It is just for any case to not get strange NPE.

john9x:
I wrote a naive synthetic test but it shows the difference in GC pressure well.
JDK 1.8.0_60

VM options
`-Xmx16m -verbose:gc -Xloggc:improved.gc.log -XX:+PrintGCDetails -XX:+PrintGCDateStamps`
After ten minute run for old and new versions I loaded GC logs to [GCViewer](https://github.com/chewiebug/GCViewer).

Old version:
Total freed memory: 255M
Number of GC pauses: 84

Improved version:
Total freed memory: 35M
Numer of GC pauses: 11

```
public class ImprovedCslTest {
    public static void main(String[] args) throws Exception {
        int threadsCount = 100;
        List<Thread> threads = new ArrayList<Thread>(threadsCount);
        final StatementLocator csl = new ClasspathStatementLocator();
        final StatementContext ctx1 = new TestingStatementContext(new HashMap<String, Object>()) {
            @Override
            public Class<?> getSqlObjectType() {
                return ImprovedCslTest.class;
            }
            @Override
            public Method getSqlObjectMethod() {
                try {
                    return ImprovedCslTest.class.getMethod("testMethod1");
                } catch (NoSuchMethodException e) {
                }
            }
        };
        final StatementContext ctx2 = new TestingStatementContext(new HashMap<String, Object>()) {
            @Override
            public Class<?> getSqlObjectType() {
                return ImprovedCslTest.class;
            }

            @Override
            public Method getSqlObjectMethod() {
                try {
                    return ImprovedCslTest.class.getMethod("testMethod2");
                } catch (NoSuchMethodException e) {
                }
            }
        };
        final String sql1 = "SELECT some_column, other_column, one_more_column FROM table JOIN USING(id) WHERE a=1 AND b=2 OR c=3 ORDER BY d LIMIT 100500";
        final String sql2 = "INSERT INTO some_table some_column, other_column, one_more_column FROM table JOIN USING(id) WHERE a=1 AND b=2 OR c=3 ORDER BY d LIMIT 100500";
        final AtomicInteger cnt = new AtomicInteger();
        for (int i = 0; i < threadsCount; i++) {
            Thread t = new Thread(new Runnable() {
                @Override
                public void run() {
                    try {
                        Thread.sleep(new Random().nextInt(1000));
                        for (; ; ) {
                            Thread.sleep(1000);
                            if (new Random().nextDouble() < 0.5) {
                                csl.locate(sql1, ctx1);
                            } else {
                                csl.locate(sql2, ctx2);
                            }
                            cnt.incrementAndGet();
                        }

                    } catch (Exception e) {
                        e.printStackTrace();
                    }
                }
            });
            t.setDaemon(true);
            threads.add(t);
        }
        for (Thread thread : threads) {
            thread.start();
        }
        Thread.sleep(10 * 60 * 1000);
        System.gc();
        System.out.println("Calls: " + cnt.intValue());
    }
    public void testMethod1() {   }
    public void testMethod2() {   }
}
```

qualidafial:
:+1: from me.

@stevenschlansker or @arteam, any concerns about this merge? I think using the `Method` as the cache key should be safe, it's the only part I was concerned about but I think the logic is sound.

If we're good to merge this, we should also forward these changes to the v3 branch, as it performs a similar "munge" operation in the new SQL Locators.

arteam:
I've wrote a simple microbenchmark:
```java
@BenchmarkMode(Mode.Throughput)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
public class QueriesBenchmark {

    @State(Scope.Benchmark)
    public static class DB {

        Dao dao;

        public DB() {
            Handle h = new DBI("jdbc:h2:mem:queries-benchmark", "sa", "").open();
            h.execute("create table something (id int primary key, name varchar(100))");

            PreparedBatch batch = h.prepareBatch("insert into something (id, name) values (?,?)");
            batch.add(1, "Brian");
            batch.add(2, "Thom");
            batch.execute();

            dao = h.attach(Dao.class);
        }
    }

    interface Dao {

        @SqlQuery("select name from something order by id")
        List<String> query();
    }

    @Benchmark
    public List<String> measureQuery(DB db) {
        return db.dao.query();
    }


    public static void main(String[] args) throws Exception {
        Options opt = new OptionsBuilder()
                .include(QueriesBenchmark.class.getSimpleName())
                .warmupIterations(10)
                .measurementIterations(5)
                .forks(1)
                .addProfiler(GCProfiler.class)
                .build();


        new Runner(opt).run();

    }
}
```

and got the following results:

Old version
```
Benchmark                                                        Mode  Cnt     Score      Error   Units
QueriesBenchmark.measureQuery                                   thrpt    5   234.525 ±    8.454  ops/ms
QueriesBenchmark.measureQuery:·gc.alloc.rate                    thrpt    5  1941.566 ±   68.475  MB/sec
QueriesBenchmark.measureQuery:·gc.alloc.rate.norm               thrpt    5  8683.776 ±   32.468    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space           thrpt    5  1952.571 ±  330.605  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space.norm      thrpt    5  8731.176 ± 1230.080    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space       thrpt    5     0.112 ±    0.182  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space.norm  thrpt    5     0.502 ±    0.812    B/op
QueriesBenchmark.measureQuery:·gc.count                         thrpt    5    53.000             counts
QueriesBenchmark.measureQuery:·gc.time                          thrpt    5    42.000                 ms
```

New version
```
Benchmark                                                        Mode  Cnt     Score      Error   Units
QueriesBenchmark.measureQuery                                   thrpt    5   240.153 ±  113.416  ops/ms
QueriesBenchmark.measureQuery:·gc.alloc.rate                    thrpt    5  1310.922 ±  621.074  MB/sec
QueriesBenchmark.measureQuery:·gc.alloc.rate.norm               thrpt    5  5728.005 ±    0.002    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space           thrpt    5  1306.050 ±  599.641  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space.norm      thrpt    5  5716.802 ± 1271.977    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space       thrpt    5     0.069 ±    0.101  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space.norm  thrpt    5     0.292 ±    0.321    B/op
QueriesBenchmark.measureQuery:·gc.count                         thrpt    5    37.000             counts
QueriesBenchmark.measureQuery:·gc.time                          thrpt    5    39.000                 ms
```

It looks like that the claim is correct, but it doesn't bring a big performance gain. 

Regarding the implementation, I am not really fond of mixing `String`s and `Method`s in one cache. That's not really obvious and could lead to strange errors. Probably we can split it to two caches?

john9x:
Hmm. It looks like your benchmark for new version was not in a steady state. Too big mean error.
>QueriesBenchmark.measureQuery                                   thrpt    5   240.153 ±  113.416  ops/ms
QueriesBenchmark.measureQuery:·gc.alloc.rate                    thrpt    5  1310.922 ±  621.074  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space           thrpt    5  1306.050 ±  599.641  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space.norm      thrpt    5  5716.802 ± 1271.977    B/op

Your benchmark, my results:
```
# JMH 1.13 (released 2 days ago)
# VM version: JDK 1.8.0_60, VM 25.60-b23
# VM invoker: C:\Program Files\Java\jdk1.8.0_60\jre\bin\java.exe
# VM options: -Didea.launcher.port=7532 -Didea.launcher.bin.path=C:\Program Files (x86)\JetBrains\IntelliJ IDEA 15.0.3\bin -Dfile.encoding=UTF-8
# Warmup: 10 iterations, 1 s each
# Measurement: 5 iterations, 1 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Throughput, ops/time
# Benchmark: org.skife.jdbi.QueriesBenchmark.measureQuery

Old version:

Benchmark                                                        Mode  Cnt     Score      Error   Units
QueriesBenchmark.measureQuery                                   thrpt    5   188,665 ±    4,336  ops/ms
QueriesBenchmark.measureQuery:·gc.alloc.rate                    thrpt    5  1600,308 ±   36,740  MB/sec
QueriesBenchmark.measureQuery:·gc.alloc.rate.norm               thrpt    5  8896,006 ±    0,001    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space           thrpt    5  1606,154 ±  680,094  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space.norm      thrpt    5  8925,767 ± 3650,992    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space       thrpt    5     0,056 ±    0,198  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space.norm  thrpt    5     0,314 ±    1,106    B/op
QueriesBenchmark.measureQuery:·gc.count                         thrpt    5    59,000             counts
QueriesBenchmark.measureQuery:·gc.time                          thrpt    5    40,000                 ms

New version:

Benchmark                                                        Mode  Cnt     Score     Error   Units
QueriesBenchmark.measureQuery                                   thrpt    5   295,221 ±   1,312  ops/ms
QueriesBenchmark.measureQuery:·gc.alloc.rate                    thrpt    5  1598,867 ±   7,189  MB/sec
QueriesBenchmark.measureQuery:·gc.alloc.rate.norm               thrpt    5  5680,004 ±   0,001    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space           thrpt    5  1589,077 ± 145,234  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space.norm      thrpt    5  5645,252 ± 520,939    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space       thrpt    5     0,188 ±   0,225  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space.norm  thrpt    5     0,666 ±   0,800    B/op
QueriesBenchmark.measureQuery:·gc.count                         thrpt    5    89,000            counts
QueriesBenchmark.measureQuery:·gc.time                          thrpt    5    47,000                ms
```
This PR is not about huge performance gain but about GC pressure.
For new version - gc.alloc.rate.norm ~5680 B/op, for old version - 8896 B/op. 

>mixing Strings and Methods in one cache
IMHO it's not a big deal for private cache that isolated in one class.

john9x:
And 1 second measurement time is not enough to get accurate results about GC.
10 seconds benchmark
```
Old version:

Benchmark                                                        Mode  Cnt     Score    Error   Units
QueriesBenchmark.measureQuery                                   thrpt    5   188,732 ±  0,533  ops/ms
QueriesBenchmark.measureQuery:·gc.alloc.rate                    thrpt    5  1595,513 ±  4,643  MB/sec
QueriesBenchmark.measureQuery:·gc.alloc.rate.norm               thrpt    5  8864,565 ±  1,015    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space           thrpt    5  1596,018 ±  7,266  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space.norm      thrpt    5  8867,368 ± 19,893    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space       thrpt    5     0,120 ±  0,038  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space.norm  thrpt    5     0,666 ±  0,208    B/op
QueriesBenchmark.measureQuery:·gc.count                         thrpt    5  2434,000           counts
QueriesBenchmark.measureQuery:·gc.time                          thrpt    5  1350,000               ms


New version:

Benchmark                                                        Mode  Cnt     Score    Error   Units
QueriesBenchmark.measureQuery                                   thrpt    5   298,840 ±  1,692  ops/ms
QueriesBenchmark.measureQuery:·gc.alloc.rate                    thrpt    5  1641,841 ±  9,184  MB/sec
QueriesBenchmark.measureQuery:·gc.alloc.rate.norm               thrpt    5  5761,021 ±  1,637    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space           thrpt    5  1643,517 ± 11,049  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Eden_Space.norm      thrpt    5  5766,906 ± 43,958    B/op
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space       thrpt    5     0,110 ±  0,037  MB/sec
QueriesBenchmark.measureQuery:·gc.churn.PS_Survivor_Space.norm  thrpt    5     0,386 ±  0,129    B/op
QueriesBenchmark.measureQuery:·gc.count                         thrpt    5  2267,000           counts
QueriesBenchmark.measureQuery:·gc.time                          thrpt    5  1310,000               ms
```

stevenschlansker:
@arteam how would you feel about setting your benchmark up as a PR to make a new `benchmark` project?  seems like starting to collect such benchmarks in one place could be useful long term.

I'll review the meat of this PR shortly.

hgschmie:
We had instances of CSL in the past where multiple instances of the same class methods (e.g. through inheritance) referenced different files from the class path. So naively using the method name (which will always resolve to the class where inheritance came from) will not work. 

I want to see a test pass that takes this into account before merging this.

arteam:
@john9x, Thanks for running the benchmark on your environment! I must admit I am not an expert in probability and statistics, so thank you for pointing to flaws in my experimentation methodology. I think your observations are correct, which confirms that the GC pressure could play a major role in the long run.

>  IMHO it's not a big deal for private cache that isolated in one class.
I don't think that's a big issue as well, but in my opinion the 2-cache version would be easier to comprehend without losing in performance. Let's see what @stevenschlansker thinks about it.

@stevenschlansker, I would not object. I don't write microbenchmarks often, because usually performance issues come up on more high levels, but for JDBI it may be doable.

src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -74,20 +75,32 @@ public static boolean looksLikeSql(String sql)
...
+                isSqlObjectMethod = true;
+                cache_key = ctx.getSqlObjectMethod();
+            } else {
+                cache_key = '/' + mungify(ctx.getSqlObjectType().getName() + '.' + name) + ".sql";
```
stevenschlansker: If we are going to worry about object allocation here, why not do something even simpler like using a `new SimpleEntry<>(ctx.getSqlObjectType(), name)` which should be much cheaper than assembling this string?

src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -74,20 +75,32 @@ public static boolean looksLikeSql(String sql)
...
+        if (cached != null) {
+            return cached;
         }
 
+
         if (looksLikeSql(name)) {
             // No need to cache individual SQL statements that don't cause us to search the classpath
+            // But for static SQL object methods caching decreases GC pressure and not threatens memory leaks.
+            if (isSqlObjectMethod) {
+                found.put(cache_key, name);
```
stevenschlansker: I don't see how this is improving the situation much.  You still have to assemble the cache key, which is the "expensive" portion you are trying to avoid -- filling the cache with these saves you the `looksLikeSql` call but nothing else as far as I can see?

stevenschlansker:
The code looks okay, but I am not sure I understand these benchmarking results.  In particular, it looks like JMH is claiming that "alloc rate" is 1.6 +/- 9 GB/s (???)

john9x:
>it looks like JMH is claiming that "alloc rate" is 1.6 +/- 9 GB/s (???)
~1600 MB/sec with mean error 9 MB/sec.  In the Russian locale comma used as decimal mark. :)

>naively using the method name
I propose to use not method name but whole `Method` object.

stevenschlansker:
Using the Method object over the name is likely better, but doesn't address the concern about subclasses either way, correct?  The Method object belongs to the superclass either way.

qualidafial:
I think @hgschmie's observation about multiple subclasses will be important. Different subclasses of the same parent interface will want to search different directories of the classpath.

So we'll need to use a tuple of `(sqlObjectType, sqlObjectMethod | name)` as the cache key to solve this properly.

qualidafial:
I am so glad we got rid of the `looksLikeSql` stuff in v3.

stevenschlansker:
If we just unconditionally use the tuple `(sqlObjectType, sqlObjectMethod, name)` (with some values potentially null) then we would once again have uniform types of cache keys, and could remove the worry about the `Map<Object, String>` not having a great key type.

john9x:
>The Method object belongs to the superclass either way. 

Hmm. May be I don't understand something but how it`s possible?

```
public class MethodTest {

    abstract class SuperDao {
         public abstract void method();
    }

    class Dao extends SuperDao{
        public void method() {};
    }

    @Test
    public void name() throws Exception {
        Method superMethod = SuperDao.class.getMethod("method");
        Method method = Dao.class.getMethod("method");

        Assert.assertTrue(superMethod.getDeclaringClass().equals(SuperDao.class));
        Assert.assertTrue(method.getDeclaringClass().equals(Dao.class));
        Assert.assertFalse(superMethod.equals(method));

    }
}
```

stevenschlansker:
I added the test case in #418, hopefully that shows what we're talking about better.
In your code example you've overridden `method` in the class, so indeed it refers to the subclass's `Method`.  But if you don't override it, it still goes to the superclass.

john9x:
@stevenschlansker Ah! Now I see what you mean.

src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -22,17 +22,53 @@
...
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            CacheKey cacheKey = (CacheKey) o;
+
+            if (name != null ? !name.equals(cacheKey.name) : cacheKey.name != null) return false;
+            if (sqlObjectType != null ? !sqlObjectType.equals(cacheKey.sqlObjectType) : cacheKey.sqlObjectType != null)
```
qualidafial: These naked `if` blocks without braces are causing the build to fail. Need to add braces.

Alternatively, since we depend on Java 7, you could use `java.util.Objects.equals` to simplify this method:

    return Objects.equals(this.name, cacheKey.name)
        && Objects.equals(this.sqlObjectType, cacheKey.sqlObjectType)
        && Objects.equals(this.sqlObjectMethod, cacheKey.sqlObjectMethod);

stevenschlansker:
Unfortunately I believe v2 is still built using Java 6.  We then skip straight to 8 with v3...

qualidafial:
Travis is only building with Java 7 and 8 at the moment.

stevenschlansker:
I think that's just because they don't officially have an `openjdk6` or something like that:
https://github.com/jdbi/jdbi/blob/master/src/build/travis-toolchains.xml
https://github.com/jdbi/jdbi/blob/master/pom.xml#L82
https://github.com/jdbi/jdbi/blob/master/pom.xml#L284-L294

stevenschlansker:
@john9x when you update this PR to fix the braces issue, you may wish to rebase on master to ensure the now merged tests still pass.

john9x:
Also I added couple of fixes to reduce unnecessary memory allocations.

qualidafial:
Oops, sorry @john9x, this slipped through the cracks. I'll get this reviewed hopefully today and (assuming no new issues are identified) merged into master.

Thank you very much taking the time to help fix this.

src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -22,17 +22,60 @@
...
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+
+            CacheKey cacheKey = (CacheKey) o;
+
+            if (name != null ? !name.equals(cacheKey.name) : cacheKey.name != null) {
```
qualidafial: This would be easier to read if the ternary condition were reversed:

    if (!(name == null ? cacheKey.name == null : name.equals(cacheKey.name))) {

Alternatively, add a null-safe equality method and call that in `equals`:

    public boolean equals(Object o) {
      if (this == o) {
        return true;
      }
      if (o == null || getClass() != o.getClass()) {
        return false;
      }
      CacheKey that = (CacheKey) o;
      return eq(this.name, that.name)
          && eq(this.sqlObjectType, that.sqlObjectType)
          && eq(this.sqlObjectMethod, that.sqlObjectMethod);
    }

    private boolean eq(Object left, Object right) {
      return left == null ? right == null : left.equals(right);
    }

src/main/java/org/skife/jdbi/v2/ClasspathStatementLocator.java
```diff
@@ -22,17 +22,60 @@
...
+            if (sqlObjectType != null ? !sqlObjectType.equals(cacheKey.sqlObjectType) : cacheKey.sqlObjectType != null) {
+                return false;
+            }
+            return sqlObjectMethod != null ? sqlObjectMethod.equals(cacheKey.sqlObjectMethod) : cacheKey.sqlObjectMethod == null;
+
+        }
+
+        @Override
+        public int hashCode() {
+            int result = name != null ? name.hashCode() : 0;
```
qualidafial: ditto on readability here:

    int result = name == null ? 0 : name.hashCode();

src/test/java/org/skife/jdbi/v2/sqlobject/TestClasspathStatementLocator.java
```diff
@@ -48,15 +49,32 @@ public void tearDown() throws Exception {
...
     public void testBam() throws Exception {
-        handle.execute("insert into something (id, name) values (6, 'Martin')");
+        Something s = handle.attach(Cromulence.class).findById(6L);
+        assertThat(s.getName(), equalTo("Martin"));
+    }
+
+    @Test
+    public void testOverride() throws Exception {
+        Something s = handle.attach(SubCromulence.class).findById(6L);
+        assertThat(s.getName(), equalTo("overridden"));
```
qualidafial: 👍

src/test/java/org/skife/jdbi/v2/sqlobject/TestClasspathStatementLocator.java
```diff
@@ -48,15 +49,32 @@ public void tearDown() throws Exception {
...
     }
 
     @RegisterMapper(SomethingMapper.class)
     static interface Cromulence {
         @SqlQuery
         public Something findById(@Bind("id") Long id);
     }
+
+    @RegisterMapper(SomethingMapper.class)
```
qualidafial: Huh, I guess we don't look at supertypes for annotations.. we should probably fix that in v3.

qualidafial: Opened #438

qualidafial:
Except for some readability concerns this is ready to merge. @john9x ping me when you get those last few comments addressed and I'll merge this PR.

john9x:
Fixed. 
Actually `equals` and `hashCode` methods was automatically generated by IntelliJ IDEA :)

qualidafial:
Looks great. Thanks!

PR #418, opened 2016-07-26
Add a test showing that subclasses of SqlObject find statements based on the subclass name, not parent class

stevenschlansker:
tag @hgschmie @john9x @qualidafial


stevenschlansker:
Open question: should the locator fall back to the superclass's location if the subclass doesn't have its own?

john9x:
IMHO Do not change current behavior by default (not to fall back to the superclass) but add CSL constructor parameter to change bahavior.

stevenschlansker:
Sounds reasonable, and either way should probably be a followup and let this PR stand alone.

qualidafial:
> Open question: should the locator fall back to the superclass's location if the subclass doesn't have its own?

Let's hold off on this until we have evidence it's actually a problem. Seems speculative just now.

qualidafial:
:shipit:

PR #419, opened 2016-07-28
Correctly handle null values in a response in JDBI3

arteam:
Users often use methods which return only a single result. In this case a returned `null` indicates that a RDBMS didn't find a document which satisfies the passed arguments. But actually the returned document can be `null` itself, if it's a nullable column.

We should expect such values and return them as null/optional values to the user. Currently we throw a `NullPointerException` instead, which is a rather harsh strategy for handling nullable values.


qualidafial:
:shipit:

stevenschlansker:
Good catch!  I'm a little worried that this means "no result" is the same return value as "null result" but this is a strict improvement over today and I don't have a proposal for better off the top of my head.

PR #423, opened 2016-07-29
Expose the SQL method in StatementContext

arteam:
It's useful in timing collectors when we want to track the time of completing of SQL queries of different methods.

See [metrics-jdbi](https://github.com/dropwizard/metrics/tree/3.2-development/metrics-jdbi) as an example.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/StatementContext.java
+++ b/core/src/main/java/org/jdbi/v3/core/StatementContext.java
@@ -13,6 +13,7 @@
  */
 package org.jdbi.v3.core;
 
+import java.lang.reflect.Method;
 import java.lang.reflect.Type;
 import java.sql.Connection;
 import java.sql.PreparedStatement;
@@ -41,6 +42,8 @@ public final class StatementContext
     private PreparedStatement statement;
     private Connection        connection;
     private Binding           binding;
+    private Method            extensionMethod;
+    private Class<?>          extensionType;
     private boolean           returningGeneratedKeys;
     private boolean           concurrentUpdatable;
     private String[]          generatedKeysColumnNames;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/jdbi/v3/core/StatementContext.java
```diff
@@ -208,6 +210,16 @@ Cleanables getCleanables() {
         return cleanables;
     }
 
+    public Method getSqlObjectMethod()
```
qualidafial: Let's name this property `extensionMethod` instead.

qualidafial: As I think about this more, is there anything wrong with making this property just a string, with a generic name to communicate that it identifies the statement context for tracking purposes (logging, metrics)

stevenschlansker:
Personally I'd rather have the `Method` over a `String` unless there's a particular reason to hide it.

qualidafial:
As discussed on the another ticket: because of SQL locators, the same method can run a different SQL statement if you have a DAO type extend some base type:

  package com.foo;

  interface BaseDao {
    @UseClasspathSqlLocator
    @SqlQuery
    list();
  }
  interface MyDao extends BaseDao {}
  interface YourDao extends BaseDao {}

These two types will look for sql in e.g. `com/foo/MyDao/list.sql` and `com/foo/YourDao/list.sql` respectively. Since they may exhibit different performance, it makes sense to track these two methods separately.

Therefore I think if we're going to put `StatementContext.extensionMethod` back in, we should also put in `extensionType` so that instrumentation like timing collectors have the full context.

arteam:
Thanks @qualidafial! This makes sense, I've updated my PR according to your suggestion.

core/src/main/java/org/jdbi/v3/core/StatementContext.java
```diff
@@ -208,6 +211,30 @@ Cleanables getCleanables() {
...
+    {
+        return extensionType;
+    }
+
+    public void setExtensionType(Class<?> extensionType)
+    {
+        this.extensionType = extensionType;
+    }
+
+    public JdbiConfig getConfig() {
```
qualidafial: JdbiConfig isn't a public type

qualidafial: Let's remove this method, then :shipit:

arteam: Good catch! I've updated the PR.

PR #424, opened 2016-08-01
Batch: rewrite SQLException.getNextException into Throwable.addSuppressed

stevenschlansker:
Previously if your batch insert has trouble with e.g. Postgres, you would see:
```
java.sql.BatchUpdateException: Batch entry 1 insert into something (id, name) values (0, 'Keith') was aborted.  Call getNextException to see the cause.
  at org.postgresql.jdbc.BatchResultHandler.handleError(BatchResultHandler.java:136)
```
Now, you get the actual information!
```
java.sql.BatchUpdateException: Batch entry 1 insert into something (id, name) values (0, 'Keith') was aborted.  Call getNextException to see the cause.
  at org.postgresql.jdbc.BatchResultHandler.handleError(BatchResultHandler.java:136)
Suppressed: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "something_pkey"
  Detail: Key (id)=(0) already exists.
```

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Batch.java
+++ b/core/src/main/java/org/jdbi/v3/core/Batch.java
@@ -122,11 +122,25 @@ public class Batch extends BaseStatement
             }
             catch (SQLException e)
             {
-                throw new UnableToExecuteStatementException(e, getContext());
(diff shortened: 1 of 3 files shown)


qualidafial:
LGTM

PR #426, opened 2016-08-01
ClasspathSqlLocator: Replace WeakHashMap with a 10 minute Guava Cache

stevenschlansker:
Current implementation uses a WeakHashMap -- this is bad for a cache
    because the common case, in 'findSqlOnClasspath', uses a computed
    key that doesn't exist outside the scope of the lookup.  So the moment
    a lookup succeeds, the key is immediately thrown away, and is therefore
    collected on the next minor GC.
    
    This also makes the tests flaky I think.  I'm observing intermittent
    failures on Travis.
    
    Guava solves this nicely with the Cache class, configured to expire
    after an interval of disuse rather than relying on GC.  I wish there
    was a lighter weight solution but I didn't see any obvious choices
    in the base JDK libraries.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/locator/ClasspathSqlLocator.java
+++ b/core/src/main/java/org/jdbi/v3/core/locator/ClasspathSqlLocator.java
@@ -13,13 +13,18 @@
  */
 package org.jdbi.v3.core.locator;
 
-import static java.util.Collections.synchronizedMap;
-
 import java.io.IOException;
 import java.io.InputStream;
-import java.util.Map;
+import java.util.AbstractMap;
+import java.util.Map.Entry;
 import java.util.Optional;
-import java.util.WeakHashMap;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+import com.google.common.base.Throwables;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.util.concurrent.UncheckedExecutionException;
 
 import org.antlr.runtime.ANTLRInputStream;
 import org.jdbi.v3.core.internal.SqlScriptParser;
@@ -33,7 +38,10 @@ public final class ClasspathSqlLocator {
 
     private static final SqlScriptParser SQL_SCRIPT_PARSER = new SqlScriptParser((t, sb) -> sb.append(t.getText()));
 
-    private static final Map<String, String> CACHE = synchronizedMap(new WeakHashMap<>());
+    private static final Cache<Entry<ClassLoader, String>, String> CACHE = CacheBuilder.newBuilder()
+            .expireAfterAccess(10, TimeUnit.MINUTES)
+            .build();
+
     private static final String SQL_EXTENSION = ".sql";
 
     private ClasspathSqlLocator() {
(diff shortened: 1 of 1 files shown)


qualidafial:
I've been baffled by this one. I get the test failure in local testing too.

Are we getting any value out of testing the cache? This test was just a carry over from the old StatementLocator tests.

stevenschlansker:
@qualidafial what do you think of this updated PR?  I think the logic in there is bogus, and the test was exposing a real problem.

core/src/main/java/org/jdbi/v3/core/locator/ClasspathSqlLocator.java
```diff
@@ -75,7 +79,11 @@ private static String resourcePathFor(String fullyQualifiedName) {
     }
 
     private static String findSqlOnClasspath(ClassLoader classLoader, String path) {
-        return CACHE.computeIfAbsent(path, p -> readResource(classLoader, p));
+        try {
+            return CACHE.get(path, () -> readResource(classLoader, path));
```
qualidafial: Use `CACHE.getUnchecked` instead if you're just going to rethrow.

stevenschlansker: Unfortunately that's only a method on `LoadingCache` not `Cache` itself -- I can't fathom why!  But we should probably fix the code to include the ClassLoader anyway, so I'll do that now...

qualidafial:
Yeah I could see the weak keys causing that.

This does introduce another hard dependency on Guava though. If you're ok with that, 👍 from me

stevenschlansker:
Grrrr... I'm not really :(
We were already looking to get rid of it and this just makes it harder.  I'll come back with a v2 that doesn't use Guava I guess...

stevenschlansker:
Or we could just shade it in...  Meh, I'll file another issue and defer it.  I just want the build to work for now.

core/src/main/java/org/jdbi/v3/core/locator/ClasspathSqlLocator.java
qualidafial: Just provide the loader function to the `CacheBuilder.build` method as a lambda expression. Moves the lambda from here to the cache declaration

stevenschlansker:
filed #427

core/src/main/java/org/jdbi/v3/core/locator/ClasspathSqlLocator.java
stevenschlansker: Can't do that -- the `classLoader` is not available there, and it's not part of the cache key, so we have to capture it.

qualidafial: Oh, FFS Guava. You can't use a lambda because it's an abstract class. Because of course it is.

qualidafial:
Nevermind my code comment. Just :shipit: as is, assuming the build passes.

stevenschlansker:
I fixed it better by making the ClassLoader part of the key.
Minor issue is that means we would prevent any ClassLoaders from being GCed for up to 10 minutes... probably not a huge issue though.

Gah!  `getUnchecked` wraps *any* thrown exception with `UncheckedExecutionException` so we do actually have to try / catch it anyway.

qualidafial:
Do we expect the content to vary based on classloader? Or just to be _accessible_ depending on classloader? If the latter, then I don't think we need classloader to be part of the cache key.

stevenschlansker:
I do not expect content to vary based on ClassLoader.  But I think it is just asking for trouble to not consider it.  For example imagine an application server with jdbi in its global classloader, running two different versions of the same application.  (Yes this is dumb, but people do dumb things all the time...)

You'd nondeterministically get different SQL depending on which version of the app calls load first!

I think it's better to be overly defensive here rather than leave subtle surprises to bite us later on.

qualidafial:
Fair enough.

Issue #427, opened 2016-08-01
Consider Guava Cache usage in ClasspathSqlLocator


stevenschlansker:
Added it in for now to fix a flaky test case regarding `WeakHashMap` usage.  We need to figure out whether we include Guava as a dependency, shade it in, or come up with a different solution.

qualidafial:
Does maven-shade-plugin do any tree shaking?

stevenschlansker:
I don't think so... we should probably have some way to validate that the shaded jar actually usefully works if we decide to go down that path.

qualidafial:
https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html#minimizeJar

arteam:
What about using [expiringmap](https://github.com/jhalterman/expiringmap)? It provides the same functionality as the Guava cache, but much lighter than Guava. I forgot, do we ship the JDBI core with Guava or not? As I see, the singe dependency on it  is `TypeToken`.

qualidafial:
We'd like to remove the dependency on Guava, to minimize the number of third party dependencies.

Right now all usage of TypeToken is encapsulated within the GenericTypes class, so that we can simply swap it out when we come up with a portable solution.

stevenschlansker:
Yes, that's a good idea -- we can definitely consider expiringmap

qualidafial:
ExpiringMap is included in PR #429. This may or may not ship depending on what comes out of the mailing list discussions.

qualidafial:
Guava cache was added in #426. ExpiringMap is being considered in #429. Closing this issue as duplicate.

PR #429, opened 2016-08-03
Fix string template sql locator / statement rewriter so that template…

qualidafial:
…s can reference other templates in a group file.

### DIFF
--- a/string-template/src/main/java/org/jdbi/v3/stringtemplate/UseStringTemplateSqlLocator.java
+++ b/string-template/src/main/java/org/jdbi/v3/stringtemplate/UseStringTemplateSqlLocator.java
@@ -23,8 +23,11 @@ import java.lang.annotation.Target;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+
+import net.jodah.expiringmap.ExpirationPolicy;
+import net.jodah.expiringmap.ExpiringMap;
 
 import org.antlr.stringtemplate.StringTemplate;
 import org.antlr.stringtemplate.StringTemplateGroup;
@@ -64,7 +67,10 @@ public @interface UseStringTemplateSqlLocator {
     Class<? extends StatementRewriter> value() default ColonPrefixStatementRewriter.class;
 
     class LocatorFactory implements SqlObjectConfigurerFactory {
-        private static final Map<String, Class<?>> TYPE_CACHE = new ConcurrentHashMap<>();
+        private static final Map<String, Class<?>> TYPE_CACHE = ExpiringMap.builder()
+                .expiration(10, TimeUnit.MINUTES)
+                .expirationPolicy(ExpirationPolicy.ACCESSED)
+                .build();
 
         private static final SqlLocator SQL_LOCATOR = (sqlObjectType, method) -> {
             String name = SqlAnnotations.getAnnotationValue(method).orElseGet(method::getName);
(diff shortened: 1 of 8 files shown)


string-template/src/main/java/org/jdbi/v3/stringtemplate/UseStringTemplateSqlLocator.java
```diff
@@ -58,9 +64,20 @@
...
+            StringTemplateGroup group = findStringTemplateGroup(sqlObjectType);
+            if (!group.isDefined(name)) {
+                throw new IllegalStateException("No StringTemplate group " + name + " for class " + sqlObjectType);
+            }
+
+            String typeName = sqlObjectType.getName();
+            TYPE_CACHE.putIfAbsent(typeName, sqlObjectType);
+
+            // Hash character # should never appear in class names.. right?
+            return typeName + "#" + name;
```
qualidafial: This fixes the problem, but is kind of gross.

string-template/src/main/java/org/jdbi/v3/stringtemplate/UseStringTemplateSqlLocator.java
```diff
@@ -89,8 +106,19 @@ public SqlStatementCustomizer createForMethod(Annotation annotation, Class<?> sq
 
         private SqlStatementCustomizer create(UseStringTemplateSqlLocator annotation) {
             StatementRewriter delegate = createDelegate(annotation.value());
-            StatementRewriter rewriter = new StringTemplateStatementRewriter(delegate);
-            return q -> q.setStatementRewriter(rewriter);
+            StatementRewriter locatingRewriter = (sql, params, ctx) -> {
+                int delimiterIndex = sql.indexOf('#');
+                String typeName = sql.substring(0, delimiterIndex);
+                String templateName = sql.substring(delimiterIndex + 1);
```
qualidafial: This is why we can't have nice things

qualidafial:
@stevenschlansker, @arteam, would you mind looking this over?

string-template/src/main/java/org/jdbi/v3/stringtemplate/UseStringTemplateSqlLocator.java
stevenschlansker: Could assert that neither has a `#` in it rather than just hoping

string-template/src/main/java/org/jdbi/v3/stringtemplate/UseStringTemplateSqlLocator.java
```diff
@@ -58,9 +64,20 @@
     Class<? extends StatementRewriter> value() default ColonPrefixStatementRewriter.class;
 
     class LocatorFactory implements SqlObjectConfigurerFactory {
+        private static final Map<String, Class<?>> TYPE_CACHE = new ConcurrentHashMap<>();
```
stevenschlansker: This holds hard references to `Class` objects and will prevent class unloading.

stevenschlansker:
This should probably also use either Guava's Cache or expiringmap, same reasons as in #427

string-template/src/main/java/org/jdbi/v3/stringtemplate/UseStringTemplateSqlLocator.java
qualidafial: Damn, you're right. I will fix.

qualidafial:
I'll roll the expiringmap work into this PR.

string-template/src/main/java/org/jdbi/v3/stringtemplate/UseStringTemplateSqlLocator.java
qualidafial: From what I can tell in the java language spec, we're good. Only UTF letters and digits are valid in class names.

stevenschlansker:
Code looks good.  Will hold just a day or two pending discussion on mailing list.

stevenschlansker:
build breakage fixed by #433

stevenschlansker:
No further messages came through yet, and I think this is better than the existing code, so if we choose on a different direction we'll just PR it separately later.

PR #431, opened 2016-08-03
Expose Handle.isClosed

### LINKED ISSUE #430: Add Handle.isClosed()
I'd like to implement my own transaction handling (yes, I'm masochistic). To do so, I need to know whether the underlying connection is already closed, otherwise `commit()`/`rollback()` will throw an exception. Please add `isClosed()`.

stevenschlansker:
Fixes #430


qualidafial:
:shipit:

PR #434, opened 2016-08-04
Add @GetGeneratedKeys support to @SqlBatch

jhaber:
Right now annotating an `@SqlBatch` method with `@GetGeneratedKeys` has no effect even though the `PreparedBatch` class has support for it. This PR adds very limited support, only allowing `int[]` as a return type. We've had this modification in our fork for over a year without issue. Let me know if that limitation seems reasonable and I can try to add some tests, otherwise I can try to do a more extensive refactor to add more complete support.


stevenschlansker:
Yes, that's a reasonable compromise.  Obviously we'd prefer full support but we can punt on that for v3.

Code looks good, will merge once we have some tests :)

Thanks!

jhaber:
Sounds good, I'll try to push another commit later today

jhaber:
@stevenschlansker I just pushed 263c952bd3f497bfab2984611ce429ef07741479 and it built successfully. This verifies the behavior against two different databases, one using a column name to fetch the generated keys and one using the standard  `Statement.RETURN_GENERATED_KEYS`

stevenschlansker:
thanks!

PR #436, opened 2016-08-05
Work around Eclipse vs javac compiler disagreement

stevenschlansker:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=499197

This commit should be reverted once a fix lands upstream.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/ExtensionRegistry.java
+++ b/core/src/main/java/org/jdbi/v3/core/ExtensionRegistry.java
@@ -55,9 +55,10 @@ class ExtensionRegistry {
         this.factories = new CopyOnWriteArrayList<>();
     }
 
+    @SuppressWarnings({"unchecked", "rawtypes"}) // working around https://bugs.eclipse.org/bugs/show_bug.cgi?id=499197
     ExtensionRegistry(List<Entry<? extends ExtensionConfig<?>>> factories) {
-        this.factories = factories.stream()
-                .map(Entry::copyOf)
+        this.factories = (List) factories.stream()
+                .map(e -> Entry.copyOf((Entry)e))
                 .collect(toCollection(CopyOnWriteArrayList::new));
     }
 


core/src/main/java/org/jdbi/v3/core/ExtensionRegistry.java
```diff
@@ -55,9 +55,10 @@ static ExtensionRegistry copyOf(ExtensionRegistry registry) {
         this.factories = new CopyOnWriteArrayList<>();
     }
 
+    @SuppressWarnings({"unchecked", "rawtypes"}) // working around https://bugs.eclipse.org/bugs/show_bug.cgi?id=499197
     ExtensionRegistry(List<Entry<? extends ExtensionConfig<?>>> factories) {
-        this.factories = factories.stream()
-                .map(Entry::copyOf)
+        this.factories = (List) factories.stream()
+                .map(e -> Entry.copyOf((Entry)e))
```
qualidafial: This is why I didn't bother generifying the `Entry` class originally. Sometimes it's just a ton of work making the compiler happy with generics.

stevenschlansker: I'll play around with de-generifying it and see if that's nicer.

stevenschlansker: Hm.  It's not convincingly better.  Either way you end up having to have one raw type hack (it ends up being on the `factory.attach(extension, config)` where those types are dependent.  So I guess I'm leaning towards the existing format, which in theory will be fixed in Eclipse 4.7, vs de-generifying which unconditionally requires a raw type escape.

qualidafial:
:shipit: once the build passes

core/src/main/java/org/jdbi/v3/core/ExtensionRegistry.java
qualidafial: Or rather, I think I generified Entry but only made `factories` a `List<Entry>` without the nested parameter.

I'm not really picky though, it's an internal class, so anything that does the job is fine.

PR #437, opened 2016-08-08
Remove jdbi v2 openjdk8 build

stevenschlansker:
Now that we use toolchains to select a jdk6, there's
little reason to select jdk8.  Whether we choose 7 or 8
we then immediately switch to 6, so it's just redundant work.


arteam:
LGTM, merged.

PR #440, opened 2016-08-09
Latest basepom, make tests execute faster

stevenschlansker:


pom.xml
```diff
@@ -113,6 +112,7 @@
         <dependency>
             <groupId>log4j</groupId>
             <artifactId>log4j</artifactId>
+            <version>1.2.17</version>
```
qualidafial: Should this dependency be test scoped?

qualidafial:
One comment to look at, then :shipit:

pom.xml
stevenschlansker: The `Log4JLog` class in the compile classpath needs it.  So it's `optional` instead of `test`.

PR #442, opened 2016-08-10
Allow imperative SQL object methods to locate SQL statements in the s…

qualidafial:
…ame classpath directories as declarative methods.

Add methods Handle.getSqlObjectType, Handle.getSqlObjectMethod, and Handle.setSqlObjectContext (sets both
properties at once). These values are forwarded to the StatementContext of any Query, Update, Batch, etc
upon creation of a SQLStatement.

In v3, we may want to hoist the setting of these values to a universal Handler wrapper which gets wrapped
around every method. In this way we set the properties once on the Handle and would not have to expose
setters on the StatementContext.

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/BasicHandle.java
+++ b/src/main/java/org/skife/jdbi/v2/BasicHandle.java
@@ -27,7 +27,6 @@ import org.skife.jdbi.v2.tweak.StatementLocator;
 import org.skife.jdbi.v2.tweak.StatementRewriter;
 import org.skife.jdbi.v2.tweak.TransactionHandler;
 
-import java.lang.reflect.Method;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.Collections;
@@ -53,8 +52,12 @@ class BasicHandle implements Handle
     private final TransactionHandler       transactions;
     private final Connection               connection;
 
-    private Class<?> sqlObjectType;
-    private Method sqlObjectMethod;
+    private final ThreadLocal<SqlObjectContext> sqlObjectContext = new ThreadLocal<SqlObjectContext>() {
+        @Override
+        protected SqlObjectContext initialValue() {
+            return new SqlObjectContext();
+        }
+    };
 
     BasicHandle(TransactionHandler transactions,
                 StatementLocator statementLocator,
@@ -93,7 +96,7 @@ class BasicHandle implements Handle
                                               this,
                                               statementBuilder,
                                               sql,
-                                              new ConcreteStatementContext(globalStatementAttributes, queryRegistry, sqlObjectType, sqlObjectMethod),
+                                              new ConcreteStatementContext(globalStatementAttributes, queryRegistry, sqlObjectContext.get()),
                                               log,
                                               timingCollector,
                                               Collections.<StatementCustomizer>emptyList(),
@@ -261,7 +264,7 @@ class BasicHandle implements Handle
                           statementRewriter,
                           statementBuilder,
                           sql,
-                          new ConcreteStatementContext(globalStatementAttributes, new MappingRegistry(mappingRegistry), sqlObjectType, sqlObjectMethod),
+                          new ConcreteStatementContext(globalStatementAttributes, new MappingRegistry(mappingRegistry), sqlObjectContext.get()),
                           log,
                           timingCollector,
                           foreman,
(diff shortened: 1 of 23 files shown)


qualidafial:
@stevenschlansker would you look this over?

stevenschlansker:
Code seems fine.  It's going to conflict badly with the v3 branch though -- is there a compelling reason to get it in v2, or would it be reasonable as a v3 only fix?

src/main/java/org/skife/jdbi/v2/BasicHandle.java
```diff
@@ -487,4 +490,20 @@ public void registerContainerFactory(ContainerFactory<?> factory)
     {
         this.containerFactoryRegistry.register(factory);
     }
+
+    @Override
+    public void setSqlObjectContext(Class<?> sqlObjectType, Method sqlObjectMethod) {
```
stevenschlansker: Having this public makes me a little uneasy

qualidafial: Yeah. The issue is I need a way to make it so any `Query`, `Update`, etc created in the method call would inherit the sql object type and method.

An alternate approach I considered was to create a decorator `Handle` class, which calls `setSqlObjectType` and `setSqlObjectMethod` on every `Query`, `Update`, etc created by the decorated handle. Then wrap every `Handler` in `SqlObject` in a decorator which wraps the handle in a decorate before calling down to the base handler. It seemed a little heavy handed, but it didn't result in any new API.

stevenschlansker: Meh, that seems really heavy handed.

stevenschlansker:
I guess my only reservation now is whether we could do this feature more cleanly in v3, where we can refactor apis as necessary, and save ourselves some merge pain.  Maybe v2 users don't actually need this (nobody noticed it until now).  But if you want it in v2 then :shipit:

src/main/java/org/skife/jdbi/v2/BasicHandle.java
qualidafial: Actually it's not so bad, I'm halfway done with it. Let me run with this and see where it gets us.

qualidafial:
This is actually a pain at my day job, so I'd prefer to fix it in v2 :)

src/main/java/org/skife/jdbi/v2/BasicHandle.java
qualidafial: Current work in progress here: https://github.com/jdbi/jdbi/commit/9e7aa0b5ff2598755fdaba47d6578dc26a574686

So there's some existing tests that expect the same handle to come back in certain circumstances--putting a wrapper around every handle sort of undermines that assumption.

At this point I think I'd rather stay with the approach on this PR.

qualidafial:
One thing I want to fix: sql object context (type and method) should be thread-local.

qualidafial:
@stevenschlansker Would you look this over again? I took a different tack and moved setting the sql object type / method stuff into the `HandleDing`. This way we only have one call in `SqlObject` that sets the context, which replaced the code scattered through the various handlers that was doing the same thing.

src/main/java/org/skife/jdbi/v2/BasicHandle.java
```diff
@@ -52,6 +52,12 @@
     private final TransactionHandler       transactions;
     private final Connection               connection;
 
+    private final ThreadLocal<SqlObjectContext> sqlObjectContext = new ThreadLocal<SqlObjectContext>() {
```
stevenschlansker: `ThreadLocal.withInitial(SqlObjectContext::new)`?

qualidafial: Nope, Java 6 :(

stevenschlansker: Oh shit this is v2 nevermind...

stevenschlansker:
LGTM.  I'm going to merge -- if you wouldn't mind, would you then merge this into the `jdbi3` branch?  I'll do it if you don't get around to it.  I'm just tired of merge conflicts ;)

qualidafial:
Sure, I'll take care of the merge.

qualidafial:
Oh good lord, I see what you mean

qualidafial:
@stevenschlansker Actually now that I look at it, the SqlLocator concept is only relevant for declarative methods in v3, i.e. You only pass raw SQL to e.g. `Handle.createQuery()` now. Thus a user would just call the sql locator factory method directly:

    package com.foo;

    interface FooDao extends GetHandle {
      default List<Foo> list() {
        String sql = ClasspathSqlLocator.findSqlOnClasspath(FooDao.class, "list");
        return getHandle().createQuery(sql).fold(...);
      }
    }

With that said, it's still valuable to ensure that the extension type / method are being set on the context, for e.g. logging metrics.

qualidafial:
Opened #454 for jdbi3 branch

PR #444, opened 2016-08-10
Move mapper registration to SqlStatement, generalize Register*Mapper

stevenschlansker:
`@GetGeneratedKeys` is going to do actual mapping, which means registering is important on `@SqlUpdate` and `@SqlBatch` as well.  So might as well just move it up.

Extend the generated key mapper to use both column or row mappers, if available.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultGeneratedKeyMapper.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultGeneratedKeyMapper.java
@@ -32,11 +32,16 @@ class DefaultGeneratedKeyMapper implements RowMapper<Object> {
 
     @Override
     public Object map(ResultSet r, StatementContext ctx) throws SQLException {
-        ColumnMapper<?> columnMapper = ctx.findColumnMapperFor(returnType)
-                .orElseThrow(() -> new IllegalStateException("No column mapper for " + returnType));
-
-        return "".equals(columnName)
+        ColumnMapper<?> columnMapper = ctx.findColumnMapperFor(returnType).orElse(null);
+        if (columnMapper != null) {
+            return "".equals(columnName)
                 ? columnMapper.map(r, 1, ctx)
                 : columnMapper.map(r, columnName, ctx);
+        }
+        RowMapper<?> rowMapper = ctx.findRowMapperFor(returnType).orElse(null);
+        if (rowMapper != null) {
+            return rowMapper.map(r, ctx);
+        }
+        throw new IllegalStateException("No column or row mapper for " + returnType);
     }
 }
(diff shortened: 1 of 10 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultGeneratedKeyMapper.java
```diff
@@ -32,11 +32,16 @@
...
-                .orElseThrow(() -> new IllegalStateException("No column mapper for " + returnType));
-
-        return "".equals(columnName)
+        ColumnMapper<?> columnMapper = ctx.findColumnMapperFor(returnType).orElse(null);
+        if (columnMapper != null) {
+            return "".equals(columnName)
                 ? columnMapper.map(r, 1, ctx)
                 : columnMapper.map(r, columnName, ctx);
+        }
+        RowMapper<?> rowMapper = ctx.findRowMapperFor(returnType).orElse(null);
```
qualidafial: Since the mapping registry automatically adapts column mappers into row mappers, you could just skip the column mapper part.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/RegisterBeanMapper.java
```diff
@@ -51,10 +50,8 @@ public SqlStatementCustomizer createForMethod(Annotation annotation, Class<?> sq
         }
 
         private SqlStatementCustomizer create(RegisterBeanMapper annotation) {
-            return statement -> {
-                Query<?> query = (Query<?>) statement;
-                query.registerRowMapper(new BeanMapperFactory(annotation.value()));
-            };
+            return statement ->
+                statement.registerRowMapper(new BeanMapperFactory(annotation.value()));
```
qualidafial: Praise Jibbers, finally. Thank you.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultGeneratedKeyMapper.java
stevenschlansker: Unfortunately if we do that, we break the `columnName` annotation handling passed in on our constructor.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/RegisterRowMapperFactory.java
```diff
@@ -63,15 +63,7 @@ private SqlStatementCustomizer create(RegisterRowMapperFactory ma) {
...
-            return statement -> {
-                if (statement instanceof Query) {
-                    Query<?> q = (Query<?>) statement;
-                    for (RowMapperFactory factory : m) {
-                        q.registerRowMapper(factory);
-                    }
-
-                }
-            };
+            return stmt -> Arrays.asList(m).forEach(stmt::registerRowMapper);
```
qualidafial: There's also a `Stream.of(m).forEach(...)` if you like.

qualidafial:
One last comment to consider, than :shipit:

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/RegisterRowMapperFactory.java
stevenschlansker: I'll just replace the array usage with `List`.  More modern that way.

PR #446, opened 2016-08-10
Add support for mapping InetAddress to/from Postgres 'inet' type

stevenschlansker:


arteam:
Looks good to me. Should work not only for with the Postgres' `inet` type, but also with any text based type.

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
```diff
@@ -86,6 +88,8 @@
         register(map, double.class, Types.DOUBLE, PreparedStatement::setDouble);
         register(map, Float.class, Types.FLOAT, PreparedStatement::setFloat);
         register(map, float.class, Types.FLOAT, PreparedStatement::setFloat);
+        register(map, Inet4Address.class, Types.OTHER, (p, i, v) -> p.setObject(i, v.getHostAddress(), Types.OTHER));
+        register(map, Inet6Address.class, Types.OTHER, (p, i, v) -> p.setObject(i, v.getHostAddress(), Types.OTHER));
```
qualidafial: I'm curious why these are separate, considering the two factories are identical. Couldn't there be just one arg factory for `InetAddress`?

core/src/test/java/org/jdbi/v3/core/argument/TestInetAddress.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.google.common.collect.ImmutableSet;
+
+import org.jdbi.v3.core.PGDatabaseRule;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class TestInetAddress {
+    @Rule
+    public PGDatabaseRule db = new PGDatabaseRule();
```
qualidafial: Could we test this with H2 as well, to validate that this solution isn't limited to just Postgres?

qualidafial:
A few comments to consider, then :shipit:

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
stevenschlansker: The primitive factory doesn't consider type hierarchy, only a lookup of the exact class passed in.  `InetAddress` is an abstract class so it would never match.  We could improve this logic, but it's nonsense for most types (`int` doesn't have a superclass) and it'd be more verbose than just enumerating the implementation types here for now.

qualidafial: I was leaning more toward having an if block in the mapperFor method, similar to what were already doing for Optional. Your method is simpler though, and we can always switch to the if block if it turns out to be a problem.

core/src/test/java/org/jdbi/v3/core/argument/TestInetAddress.java
qualidafial: I did some tests from this branch, and couldn't find a built in argument / mapper implementation that worked on both Postgres and H2. Postgres wants you to call `PreparedStatement.setObject(colNumber, value, Types.OTHER)` whereas H2, which doesn't have an `inet` type, can only use `VARCHAR` and wants you to call `PreparedStatement.setString`.

qualidafial: I think the best we can do for this one is to provide `setString`-based built-in arguments and mappers, and then provide override implementations in `PostgresJdbiPlugin`

arteam:
@stevenschlansker Could you try to adapt the implementation to support H2 as well?

stevenschlansker:
Yes, sorry, this is still on my todo list.

stevenschlansker:
Okay, I updated this PR to test both H2 and Pg, and did some minor refactoring along the way.

Issue #344, opened 2016-05-04
Support <define> query interpolation in default statement rewriter.


qualidafial:
People get tripped up by this all the time. We should just support it out of the box.

qualidafial:
@stevenschlansker @brianm @hgschmie Is there a known performance benefit from caching rewritten statements?

I ask because the cache is currently using the non-rewritten SQL as the cache key, but this doesn't work if we start substituting defined values for e.g. `<foo>` in the rewriter, since the value of `foo` can vary on every call.

My unscientific tests showed no performance difference, regardless of whether the cache was present or not. Do we need it?

qualidafial:
Opened #355 which addresses this issue. Closing in favor of #355.

PR #449, opened 2016-08-12
Use st3 locator implicitly for BindIn

jhaber:
This is something we're probably going to use in our fork but just wanted to get your thoughts on it, potentially for jdbi3. The problem we run into is that people use `@BindIn` without remembering to use `@UseStringTemplate3StatementLocator`. After a bit of debugging they realize what happened and add `@UseStringTemplate3StatementLocator`, but the annotation isn't very granular (can only be applied to an entire DAO) so then some unrelated queries start failing because they contain SQL like `a < b` which the `StringTemplate3StatementLocator` requires you to escape.

I was considering updating `@UseStringTemplate3StatementLocator` to target `ElementType.METHOD` as well, that way you could only annotate methods using `@BindIn` and save some headaches there. But it seems to me that this is all unnecessary if `@BindIn` added the necessary statement locator. The heuristic I'm using for this is whether the `SQLStatement` is still using the default `ClasspathStatementLocator`, if so we apply the `UseStringTemplate3StatementLocator` customizer to the `SQLStatement` using the default annotation options. Let me know what you think of this approach.

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/SQLStatement.java
+++ b/src/main/java/org/skife/jdbi/v2/SQLStatement.java
@@ -215,7 +215,7 @@ public abstract class SQLStatement<SelfType extends SQLStatement<SelfType>> exte
         return statementBuilder;
     }
 
-    protected StatementLocator getStatementLocator()
+    public StatementLocator getStatementLocator()
     {
         return this.locator;
     }
--- a/src/main/java/org/skife/jdbi/v2/sqlobject/customizers/Define.java
+++ b/src/main/java/org/skife/jdbi/v2/sqlobject/customizers/Define.java
@@ -13,10 +13,13 @@
  */
 package org.skife.jdbi.v2.sqlobject.customizers;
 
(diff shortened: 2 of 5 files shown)


qualidafial:
Just as a heads up, we've added `<name>`-style token recognition and define substitution (similar to StringTemplate) in jdbi3 already. It's part of the colon-prefix (and hash-prefix IIRC) SQL grammars.

qualidafial:
My personal preference would be to enhance `ColonPrefixNamedParamStatementRewriter` and `HashPrefixStatementRewriter` to recognize angle-bracketed tokens like `<foo>` as in #344.

This would prevent coupling usage of `@Define` or `@BindIn` with the baggage that StringTemplate brings along, namely escaping `<` and `>` and other tokens that are significant in StringTemplate's grammar.

jhaber:
Cool, sounds like it won't be an issue for jdbi3. In the meantime we'll keep this workaround on our fork

PR #452, opened 2016-08-19
WIP: Use Xmx1g for MAVEN_OPTS

vlsi:
It might increase "documentation" build time for Travis


vlsi:
This seems to have no effect on the build time

qualidafial:
All the same, thanks for trying it out

PR #453, opened 2016-08-19
Split tests, docs, and style checks to their own Travis jobs

vlsi:
This splits Travis jobs into TEST, DOCUMENTATION, and CODE STYLE.
The benefits are:
1) Faster build times as jobs go in parallel
2) More fine grained errors: error in code style would not not terminate the build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: java
 
-jdk:
-  - oraclejdk8
-
 sudo: false
+
+# No much sense in keeping jdbi artifacts cached
+before_cache:
+  rm -rf $HOME/.m2/repository/org/jdbi
 
 cache:
   directories:
@@ -12,5 +13,19 @@ cache:
 before_install:
     - cp src/build/travis-toolchains.xml ~/.m2/toolchains.xml
 
-install: mvn -Ptoolchains -DskipTests=true -Dbasepom.check.skip-all=true -B install
-script: mvn -Ptoolchains -B verify
+install: mvn -Ptoolchains -DskipTests=true -Dbasepom.check.skip-all=true -Dmaven.javadoc.skip=true -B install
+
+matrix:
+  include:
+    - env:
+        - TESTS=Y
+      script: mvn -Dmaven.javadoc.skip=true -Dbasepom.check.skip-basic=true -Dbasepom.check.skip-findbugs=true -Dbasepom.check.skip-pmd=true -Dbasepom.check.skip-checkstyle=true -Ptoolchains -B verify
+      jdk: oraclejdk8
+    - env:
+        - DOCUMENTATION=Y
+      script: mvn -DskipTests=true -Dbasepom.check.skip-all=true -Ptoolchains -B verify
+      jdk: oraclejdk8
+    - env:
+        - CODE_STYLE=Y
+      script: mvn -DskipTests=true -Dmaven.javadoc.skip=true -Ptoolchains -B verify
+      jdk: oraclejdk8


qualidafial:
What happens to the generated site if you run `mvn install` in docs? Those sources are how it's generation javadoc if I'm not mistaken.

vlsi:
Ah, I misunderstood the meaning of `includeDependencySources`

Ok, is it expected to "include sources of antlr/antlr kind of dependencies"?
In other words, what if I add `<dependencySourceInclude>org.jdbi:*</dependencySourceInclude>`?

qualidafial:
Any idea why the documentation build took 19 minutes? In principle, this looks like a good step, but the entire build before only took 17 minutes.

vlsi:
>Any idea why the documentation build took 19 minutes

As far as I could see, Travis job has high delays when it downloads resources for third-party dependencies when building javadoc.
It looks like `dependencySourceInclude` means "try downloading all the dependencies and their javadocs, then apply dependencySourceInclude".
Several dependencies return 404 error, so the doc job spends time trying to download that.

It might make sense trying another options.

By the way: are you using http://javadoc.io/? Theoretically, it could make sense to make "references to external classes refer to http://javadoc.io/".

If not, "third party dependencies can be excluded in the <dependency>, and it would probably improve doc build time a lot".

Another option to try is to start "javadoc task" in "offline mode", so it does not try to access "non-existing javadoc-resources jars"

>but the entire build before only took 17 minutes.

It varies greatly.
https://travis-ci.org/jdbi/jdbi/builds

1394 spring-refresh -- 47 min
1384 inet-address -- 1h 16min
1379 ggk-mappers -- 13min

Well, of course having 3 jobs does impose some overhead (initial resource download is performed 3 times), but Travis does have .m2 cache, so the overhead is not that big.
Other than that, I do not see why "split configuration" would execute longer.
Note, I tried to move "findbugs/checkstyle/etc" into its own job, so I hope there's very little overlap between the jobs.

Additional gain: addition of new jobs becomes straightforward. For instance, it is not that hard to add JDK9 the build matrix.

stevenschlansker:
My main concern is that we are unlikely to merge a PR until all of the checks (code style, findbugs, etc) have completed.  So unless Travis will truly execute our jobs in parallel, and is able to launch all 3 (?) jobs quickly, it may end up being slower to separate the jobs out.  There's a fair amount of fixed overhead for each job.

I wonder if really the problem here is "the build is too slow on Travis" and we should investigate doing offline JavaDoc builds.

I'm not opposed to multiple builds if it does end up working nicely, I'm just very worried that we'd spend a bunch of time tweaking and fiddling with it only to find out that we got minimal or no gain in actual "time until PR goes green".

vlsi:
>  So unless Travis will truly execute our jobs in parallel, and is able to launch all 3 (?) jobs quickly,

Typically, Travis runs up to 5 jobs in parallel.
For instance, in pgjdbc project: https://travis-ci.org/pgjdbc/pgjdbc
"wall clock time" is 15 minutes, however sum of all elapsed times is 1h, so there's like 4x speedup from using multiple jobs.

> it may end up being slower to separate the jobs out

Why?
Typically, the jobs are started roughly at the same time.

>no gain in actual "time until PR goes green".

I'm not following you here. Can you elaborate?

stevenschlansker:
OK, as long as jobs are running simultaneously, that's most of my worry.  To elaborate a bit more on the remaining point:  Let's assume a normal build takes 5 minutes for the sake of argument.  Travis has a "slow build box" 10% of the time, and it takes 15 minutes.  With 1 distinct job, you have a "slow build" 10% of the time.  With 2 jobs, 19%.  3 -- 27%.  And so on.

This assumes that my mental model where Travis slowdowns are related to individual builds and where they run.  But hopefully it makes my point clearer.

If a couple runs shows that this is not a problem, then I've no further issue with it.  But it really does worry me how variable our build times are... :/

stevenschlansker:
Single data point -- notice how the current builds took 7.5, 14, 20 minutes -- so the entire process ran 20 -- which is a bit slower than our usual 15 minute build time.  Obviously with such high variability, a single build is not indicative, but I think it illustrates my point that a single slow step then becomes the limiting factor for the entire 3 step build.

stevenschlansker:
Okay, I ran it a couple more times, and the build seems to hover around 6-7 minutes.  So maybe this really is an improvement.  Will try a couple more, and if I don't find any further issues, will merge.

stevenschlansker:
Thank you for this contribution!

PR #454, opened 2016-08-21
Set context extension method for default/non-abstract methods

### LINKED ISSUE #441: StatementContext.sqlObject* should be populated for imperative methods
Currently we only populate `StatementContext.sqlObjectMethod` and `.sqlObjectType` for declarative methods.

However, if for any reason you need to introduce a concrete/default method that uses the fluent API (folding is a common reason), then using SQL in files on the classpath becomes less convenient.

    package com.foo;

    interface FooDao extends GetHandle {
      default List<Foo> list() {
        return getHandle().createQuery("list") // looks in root directory only
          .fold(...);
      }
    }

Declarative methods in the same interface would look in `/com/foo/FooDao/<name>.sql`, however in an imperative method you have to give a full path. This is annoying. A call to `getHandle().createQuery("methodName")` should look for SQL in the same file as `@SqlQuery("methodName")` or `@SqlQuery ReturnType methodName()`.

qualidafial:
Fixes #441 on v3 branch (rewrite of #442 to jdbi3 branch)

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/ConstantHandleSupplier.java
+++ b/core/src/main/java/org/jdbi/v3/core/ConstantHandleSupplier.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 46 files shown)


qualidafial:
@stevenschlansker Can you review?

Issue #341, opened 2016-05-02
Evaluate SQLObject implementation choices


stevenschlansker:
Currently, we use `cglib` to generate implementation classes.
There's an open question to changing this approach, either
* JDK standard dynamic Proxy instances
* Compile time annotation processor that generates code

qualidafial:
Important to consider: with `java.lang.reflect.Proxy` we are restricted to interfaces only--no classes. With Java 8's default interface methods, it's probably no big deal.

But whichever route we take, let's not expose any implementation details. That way we can pivot in the future, should a better alternative emerge.

qualidafial:
My vote on this would be to use proxies for the time being. If the interface-only limitation turns out to be a problem, we can always migrate to something else.

umcodemonkey:
Being limited to interfaces only is a *huge* deal. We use abstract classes all the time in our SQL object implementations. Default interface methods are not a substitute.

I would also be surprised if the Proxy implementation had similar performance to CGLIB generated code. I would hope to see considerable perf testing on any replacement solution to ensure there isn't any regression on what can be a very perf sensitive part of an application.

Annotation processing could possibly work, but it has its own downsides of potentially depending on the concrete implementation that's generated, and also the additional tooling that is needed to support them.

I'm really surprised this is even an issue. Is there some huge drawback to the CGLIB code (other than debugger tracing, which I'll admit can be challenging) that you hope to address by making a change?

qualidafial:
@umcodemonkey Could you provide examples of sql objects that would not work as an interface?

umcodemonkey:
In many cases it's needed to migrate legacy non-JDBI database classes to JDBI incrementally, without having to do a whole rewrite. In other cases we need to support state in the DAO, either for metrics collection or other in-app bookkeeping. In these two cases, composition could achieve similar goals (having the JDBI generated class be a member of the legacy or stateful class), but that does introduce maintenance overhead in many cases.

I'm still curious to hear what's driving the motivation to replace CGLIB. If there's an email thread someone can point me to, I'd appreciate it.

qualidafial:
This issue is about exploring alternatives to CGLIB to see if there's a better option.

We're simply taking this opportunity (pre-v3 release) to evaluate alternatives and see if other approaches might serve us better.

A few soft motivations I have are:

* Minimize external dependencies
* Minimize JDBI's total footprint (including dependencies)
* Simplify the project code.
* Simplify usage.

[...]

Issue #339, opened 2016-05-02
Dependency injected SQL objects


robfletcher:
It would be useful sometimes to be able to inject dependencies into a (abstract class based) SQL object. Currently it's not possible to have constructor arguments although it seems like it would be reasonable to add varargs to `Handle.attach` to supply values.

qualidafial:
I'm not sure this is a good idea--seems like we're mixing concerns and coalescing layers that ought to stay separate.

Could you provide some examples of objects that would be useful to inject, and how that would make using a SQL object more convenient?

stevenschlansker:
Usually when I am computing "java side", I prefer to do so in a separate class and treat the SqlObject itself strictly as the interface between the database and your code.  I often end up wrapping them:

```
class MyDao {
    @Inject
    MyDao(SomeDependency dep) { ... }

   interface MyDaoSqlObject {
       @SqlQuery(...)
        int doSomething(...);
    }

    void doSomething(int value) {
        myDaoObject.doSomething(dep.computeSomeValue(value));
    }
}
```

The advantage of this is that all of the logic remains in normal Java code - usually no need to step through a proxy layer in a debugger, etc.

Maybe this pattern will help you clean up your code?  Otherwise as qualidafial suggested, please let us know about your use case.

stevenschlansker:
You can now accomplish this in jdbi3 by putting a custom configuration object on the statement context.  Pseudocoding:

```java

class InjectedDeps implements JdbiConfig<InjectedDeps> {
    SomeDependency dep;
    void setSomeDependency(dep);
}

handle.getConfig(InjectedDeps.class).setSomeDependency(me);

interface MyDao extends GetHandle {
    default int computeSomeResult() {
        return getHandle().getConfig(InjectedDeps.class).dep.doSomething();
    }
}
```

I think that means this issue is resolved, let us know if that doesn't work for some reason!

qualidafial:
Whoa, I didn't realize config would support this use case. Nice catch.

pragmasoft-ua:
I think this example worths addition into official documentation

qualidafial:
JdbiConfig documented here: http://jdbi.org/#_jdbiconfig

You can access any configuration object in a SQL object by extending SqlObject and calling getHandle().getConfig(type)

PR #456, opened 2016-08-21
Migrate SqlObject from CGLIB to Java Proxy implementation. 

qualidafial:
Submitted for discussion with the team.

This PR removes CGLIB and ASM from the project. It also restricts SqlObject extensions to public interfaces only.

This was surprisingly straightforward. Only drawback I saw was that extending Cloneable in a SqlObject class stopped working, which forced me to change some tests--pretty sure we don't want our users using `clone()` in their SQL Objects any way.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
@@ -18,8 +18,10 @@ import static java.util.stream.Collectors.toSet;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
+import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.Proxy;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -31,10 +33,6 @@ import java.util.function.BiConsumer;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 
-import net.sf.cglib.proxy.Enhancer;
-import net.sf.cglib.proxy.Factory;
-import net.sf.cglib.proxy.MethodInterceptor;
-
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.extension.ExtensionFactory;
 import org.jdbi.v3.sqlobject.mixins.GetHandle;
@@ -43,11 +41,24 @@ import org.jdbi.v3.sqlobject.mixins.Transactional;
 public enum SqlObjectFactory implements ExtensionFactory<SqlObjectConfig> {
     INSTANCE;
 
-    private static final MethodInterceptor NO_OP = (proxy, method, args, methodProxy) -> null;
+    private static final Object[] NO_ARGS = new Object[0];
+    private static final Method EQUALS_METHOD;
+    private static final Method HASHCODE_METHOD;
+    private static final Method TOSTRING_METHOD;
+
+    static {
+        try {
+            EQUALS_METHOD = Object.class.getMethod("equals", Object.class);
+            HASHCODE_METHOD = Object.class.getMethod("hashCode");
+            TOSTRING_METHOD = Object.class.getMethod("toString");
+        }
+        catch (NoSuchMethodException wat) {
+            throw new IllegalStateException("JVM error", wat);
+        }
+    }
 
     private final Map<Method, Handler> mixinHandlers = new HashMap<>();
     private final ConcurrentMap<Class<?>, Map<Method, Handler>> handlersCache = new ConcurrentHashMap<>();
-    private final ConcurrentMap<Class<?>, Factory> factories = new ConcurrentHashMap<>();
     private final ConcurrentMap<Class<? extends SqlObjectConfigurerFactory>, SqlObjectConfigurerFactory>
             configurerFactories = new ConcurrentHashMap<>();
 
@@ -63,6 +74,14 @@ public enum SqlObjectFactory implements ExtensionFactory<SqlObjectConfig> {
 
     @Override
     public boolean accepts(Class<?> extensionType) {
+        if (!extensionType.isInterface()) {
+            throw new IllegalArgumentException("SQL Objects are only supported for interfaces.");
+        }
+
+        if (!Modifier.isPublic(extensionType.getModifiers())) {
+            throw new IllegalArgumentException("SQL Object types must be public.");
+        }
+
         if (GetHandle.class.isAssignableFrom(extensionType) ||
                 Transactional.class.isAssignableFrom(extensionType)) {
             return true;
@@ -83,30 +102,17 @@ public enum SqlObjectFactory implements ExtensionFactory<SqlObjectConfig> {
      */
     @Override
     public <E> E attach(Class<E> extensionType, SqlObjectConfig config, Supplier<Handle> handle) {
-        Factory f = factories.computeIfAbsent(extensionType, type -> {
-            Enhancer e = new Enhancer();
-            e.setClassLoader(extensionType.getClassLoader());
+        Map<Method, Handler> handlers = methodHandlersFor(extensionType);
 
-            List<Class<?>> interfaces = new ArrayList<>();
-            if (extensionType.isInterface()) {
-                interfaces.add(extensionType);
-            }
-            else {
-                e.setSuperclass(extensionType);
-            }
-            e.setInterfaces(interfaces.toArray(new Class[interfaces.size()]));
-            e.setCallback(NO_OP);
-
-            return (Factory) e.create();
-        });
-
-
-        Map<Method, Handler> handlers = buildHandlersFor(extensionType);
-        MethodInterceptor interceptor = createMethodInterceptor(extensionType, config, handlers, handle);
-        return extensionType.cast(f.newInstance(interceptor));
+        InvocationHandler invocationHandler = createInvocationHandler(extensionType, config, handlers, handle);
+        return extensionType.cast(
+                Proxy.newProxyInstance(
+                        extensionType.getClassLoader(),
+                        new Class[]{extensionType},
+                        invocationHandler));
     }
 
-    private Map<Method, Handler> buildHandlersFor(Class<?> sqlObjectType) {
+    private Map<Method, Handler> methodHandlersFor(Class<?> sqlObjectType) {
         return handlersCache.computeIfAbsent(sqlObjectType, type -> {
             final Map<Method, Handler> handlers = new HashMap<>();
 
@@ -188,17 +194,24 @@ public enum SqlObjectFactory implements ExtensionFactory<SqlObjectConfig> {
         return decorator;
     }
 
-    private MethodInterceptor createMethodInterceptor(Class<?> sqlObjectType,
+    private InvocationHandler createInvocationHandler(Class<?> sqlObjectType,
                                                       SqlObjectConfig baseConfig,
                                                       Map<Method, Handler> handlers,
                                                       Supplier<Handle> handle) {
-        return (proxy, method, args, methodProxy) -> {
-            Handler handler = handlers.get(method);
-
-            // If there is no handler, pretend we are just an Object and don't open a connection (Issue #82)
-            if (handler == null) {
-                return methodProxy.invokeSuper(proxy, args);
+        return (proxy, method, args) -> {
+            if (EQUALS_METHOD.equals(method)) {
+                return proxy == args[0];
             }
+
+            if (HASHCODE_METHOD.equals(method)) {
+                return System.identityHashCode(proxy);
+            }
+
+            if (TOSTRING_METHOD.equals(method)) {
+                return sqlObjectType + "@" + Integer.toHexString(System.identityHashCode(proxy));
+            }
+
+            Handler handler = handlers.get(method);
 
             SqlObjectConfig config = baseConfig.createCopy();
             forEachConfigurerFactory(sqlObjectType, (factory, annotation) ->
@@ -206,7 +219,7 @@ public enum SqlObjectFactory implements ExtensionFactory<SqlObjectConfig> {
             forEachConfigurerFactory(method, (factory, annotation) ->
                     factory.createForMethod(annotation, sqlObjectType, method).accept(config));
 
-            return handler.invoke(handle, config, proxy, args, method);
+            return handler.invoke(handle, config, proxy, args == null ? NO_ARGS : args, method);
         };
     }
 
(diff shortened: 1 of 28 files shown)


qualidafial:
One way to address #341

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -43,11 +41,24 @@
...
+    static {
+        try {
+            EQUALS_METHOD = Object.class.getMethod("equals", Object.class);
+            HASHCODE_METHOD = Object.class.getMethod("hashCode");
+            TOSTRING_METHOD = Object.class.getMethod("toString");
+        }
+        catch (NoSuchMethodException wat) {
+            throw new IllegalStateException("JVM error", wat);
+        }
+    }
```
qualidafial: A bit of copypasta here from `OnDemandExtensions` class

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -63,6 +74,14 @@ public SqlObjectConfig createConfig() {
...
     @Override
     public boolean accepts(Class<?> extensionType) {
+        if (!extensionType.isInterface()) {
+            throw new IllegalArgumentException("SQL Objects are only supported for interfaces.");
+        }
+
+        if (!Modifier.isPublic(extensionType.getModifiers())) {
+            throw new IllegalArgumentException("SQL Object types must be public.");
+        }
+
```
qualidafial: More copypasta

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -188,25 +194,32 @@ private static HandlerDecorator buildDecorator(Class<? extends HandlerDecorator>
...
+                return proxy == args[0];
+            }
+
+            if (HASHCODE_METHOD.equals(method)) {
+                return System.identityHashCode(proxy);
+            }
 
-            // If there is no handler, pretend we are just an Object and don't open a connection (Issue #82)
-            if (handler == null) {
-                return methodProxy.invokeSuper(proxy, args);
```
qualidafial: Seems like this line was already redundant, with the lazy handle stuff in place.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -188,25 +194,32 @@ private static HandlerDecorator buildDecorator(Class<? extends HandlerDecorator>
...
+            Handler handler = handlers.get(method);
+
             SqlObjectConfig config = baseConfig.createCopy();
             forEachConfigurerFactory(sqlObjectType, (factory, annotation) ->
                     factory.createForType(annotation, sqlObjectType).accept(config));
             forEachConfigurerFactory(method, (factory, annotation) ->
                     factory.createForMethod(annotation, sqlObjectType, method).accept(config));
 
-            return handler.invoke(handle, config, proxy, args, method);
+            return handler.invoke(handle, config, proxy, args == null ? NO_ARGS : args, method);
```
qualidafial: Yep, `InvocationHandler` gets passed `null` if there are no arguments. ಠ_ಠ

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestMapBinder.java
```diff
@@ -141,7 +141,7 @@ public Result map(ResultSet r, StatementContext ctx)
         }
     }
 
-    static class Result
+    public static class Result
```
qualidafial: I was surprised by this one. IIRC I got an illegal access exception when the class was non-public

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
arteam: Sorry for a dumb question, but why do we need here handle `Object` methods? It looks like if we omit those lines, all tests pass. As I see, `OnDemandExtensions` already takes care of them.

qualidafial: Extensions created with Handle.attach will not use the on-demand wrapper.

We probably want some equivalent tests for attached SQL objects. If those pass without these handlers, I'm happy to get rid of them.

qualidafial: @arteam Yep, you're right. I forgot there were already equals/hashCode/toString/finalize handlers elsewhere in the code.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestMapBinder.java
qualidafial: If the class is package-private, then the query runs, the result set is mapped to a `Result` object, and then at the very end, trying to return the `Result` throws an `IllegalAccessError`. Bizarre.

qualidafial:
@arteam @stevenschlansker Any concerns or objections to merging this?

stevenschlansker:
Will review today hopefully

qualidafial:
This ticket would render #339 moot (dependency-injected SQL objects).

stevenschlansker:
Should we reach out to the community to assess how many people actually have instance fields / state in their SqlObject classes?  the code itself LGTM, I mostly worry about unintended consequences for our users...

umcodemonkey:
@stevenschlansker I added some feedback to #341.

stevenschlansker:
Thanks, let's hold off on merging this until we get some consensus in the other issue.  And probably at least some basic performance tests.

qualidafial:
Roger. There are some test improvements I'd like to keep though--I'll open a separate PR for that.

chrisbouchard:
As a user, I can say that I do take advantage of some abstract class features.

* I have several DAOs with non-abstract methods. I know that Java 8 default methods would offer a similar feature, but my company only started allowing Java 8 for development in the past few months (and not even for all projects).
* We've also created a Spring bean factory for autowiring into JDBI classes (for use in those non-abstract methods), which requires fields.

I could probably get around this by separating my implementation into a facade class delegating to a JDBI sql object, but I like that I can currently use JDBI to "fill in the blanks".

qualidafial:
> I have several DAOs with non-abstract methods. I know that Java 8 default methods would offer a similar feature, but my company only started allowing Java 8 for development in the past few months (and not even for all projects).

This discussion is for JDBI v3, which will require Java 8. So..

> We've also created a Spring bean factory for autowiring into JDBI classes (for use in those non-abstract methods), which requires fields.

We don't use Spring where I work. Could you elaborate? Maybe give a toy example?

chrisbouchard:
Our goal is that our DAOs should not return partial results -- the publicly accessable methods should return fully-constructed objects. However, this requires orchestrating DAOs. We are able to load JDBI sql objecs as Spring beans via a custom annotation (`@JdbiRepository`) and inject other beans via Spring dependency injection.

Here is a sample DAO that we might write:

```java
@JdbiRepository
public abstract class FooDaoImpl {
    @Autowired BarDao barDao;

    public Foo findFooById(int id) {
        Foo foo = findFooByIdWithoutBars(id);
        List<Bar> bars = barDao.findBarsForFooId(id);
        foo.setBars(bars);
        return foo;
    }

    @SqlQuery(...)
    protected abstract Foo findFooByIdWithoutBars(@Bind("id") id);
}
```

qualidafial:
@chrisbouchard If you're only injecting other SQL objects, then we've got you covered. Here's an equivalent example using an interface, without fields:

    @JdbiRepository
    public interface FooDaoExt extends FooDao, GetHandle {
        @Override
        default Foo findFooById(int id) {
            Foo foo = findFooByIdWithoutBars(id);

            BarDao barDao = getHandle().attach(barDao);
            List<Bar> bars = barDao.findBarsForFooId(id);
            foo.setBars(bars);
            return foo;
        }

        @SqlQuery(...)
        protected abstract Foo findFooByIdWithoutBars(@Bind("id") id);
    }

chrisbouchard:
@qualidafial I did not know about `@CreateSqlObject`. Since `BarDao` is the public (non-JDBI) interface, presumably I would have to use
```java
@CreateSqlObject BarDaoImpl getBarDao();
```

Unfortunately Java 8 does not allow `protected` or `private` methods in interfaces (another reason it's nice to have abstract classes). So if `FooDaoImpl` gets a `BarDaoImpl`, there's no access control to stop it calling `BarDaoImpl`'s "internal" methods.

qualidafial:
Per http://mail.openjdk.java.net/pipermail/jdk9-dev/2015-March/001981.html private interface methods will be supported in Java 9.

chrisbouchard:
I had heard about that, and I am looking forward to Java 9. However, I don't realistically expect to use any of its features at my currently company for a year at least. This past year we got most of our users upgraded to Java 7 (!) and allowed Java 8 for server-side code.

qualidafial:
As I stated previously, this issue is discussing implementation choices for JDBI v3, which _requires_ Java 8.

If you're on Java 7, then stay on JDBI v2.x, which still uses CGLIB.

qualidafial:
That said, private and protected methods is a good example. We'll definitely have to weigh that as a factor when making the final determination.

chrisbouchard:
Is there a plan to maintain v2 after v3 is released? At any rate, my point was that Java 9 is way off for many of us, even as we get onto Java 8.

qualidafial:
We're not going to suddenly dump the majority of our users when v3 gets released. :)

I plan to continue limited support on v2 for a couple years--enough to give folks time to migrate. By "limited" I mean just bug fixes and trivial new features. We'll definitely continue accepting pull requests.

Once the team agrees that it's time to discontinue development on v2, we'll make an announcement on the mailing list.

stevenschlansker:
Yes, I agree -- we're not dropping v2 immediately, but we'll be focusing new features and improvements on v3.

For me the major pain point with cglib is that it tends to be fragile and break a fair amount -- we had at least two or three releases after Java 8 came out exposing weird edge cases where cglib couldn't enhance various new types and blew up with thoroughly unhelpful error messages -- https://github.com/jdbi/jdbi/issues/340

That's the main reason I think it'd be nice to be on the officially supported Proxy code.

Reading into my crystal ball a bit, I suspect the combination of need for private methods and the probably dismal performance is going to keep us on cglib, at least until Java 9... but we still need to get some microbenchmarks going :)

arteam:
I also wrote about fragility of CGLib in a related issue. I think we should
go the proxy generation approach. It's pretty stable and used by many
frameworks in the Java land. I don't think performance would be a big
concern, because a proxy call should be quite optimized by a JVM. I will
try to write some microbenchmarks to back this argument.
Am 06.09.2016 18:46 schrieb "Steven Schlansker" <notifications@github.com>:

arteam:
Ping. 

I think we should go ahead and merge this pull request. Despite its controversy, I think the general
view on it is positive among the JDBI maintainers. We can then release a new RC to allow the early JDBI3 users to test their approach against the version with a new proxy approach and see if there are some unexpected bugs or use cases in the wild ((like private methods in interfaces). If we see a negative feedback or some blocking issues from the users, we could rethink our approach.

qualidafial:
@stevenschlansker Do you think we should merge this?

stevenschlansker:
@chrisbouchard @umcodemonkey would you two be wiling to work with us to do a test run of refactoring your code into interfaces?  I don't want to hold this PR in purgatory forever, but so far 100% of the non-developer feedback has been hesitant at best, so I'm skeptical of merging just yet.  I'm pretty confident we can get things into a reasonable shape with a little bit of work though.

umcodemonkey:
If it was a small piece of work for us to do, then I'd be happy to help you vet the change, but it's going to be a pretty significant refactoring for us to support it, which is part of the reason why I'm so hesitant about it. There is a lot of other goodness in JDBI 3 to which our access will be denied when this PR lands, at least in the near future.

I could dive deeper into other technical reasons why I'm still dubious, but it's your project, and I shall resist bike-shedding it. JDBI 2, despite not having the nice features I see coming in v3, works fine for us for now; I also understand that projects have to move forward in the directions that are needed to not be stuck in the past.

stevenschlansker:
OK, seems like we're not going to reach total accord here in the short term.  So here's my proposal:

Let's bravely move forward with this change.  My opinion is that separating the "jdbi generated" part of your code from "my own state" leads to better design, and use of an official API lets us cut out a crufty and invasive dependency that has caused problems in the past.

That said, I'm very sensitive to not leaving our existing users out in the cold -- even if "it's our project" the project is nothing without the users, and we have to have a credible upgrade path from v2 to v3.  The nice thing is that anything that works with the Proxy implementation will work with the CGLIB implementation as well.  So if we switch to Proxy now (there will never be a more opportune moment) and later find out that we can't make it stick (users can't migrate, other problems crop up, whatever) it's an easy revert that doesn't break API.

I'll leave this up for a day or so for people to consider / object and then do the merge.  @umcodemonkey when you just can't live without v3 features and want to switch, we'll happily work with you to get the code migrated.  If it ends up being a hugely difficult project or the code doesn't end up better for it, we will revert this change and move back to CGLIB.

Sound fair to everyone?

Oduig:
To anyone finding this via Google, a little more details on how to migrate state+behavior from v2 to v3.
It wasn't immediately obvious to me how to handle state+behavior, but in hindsight the solution is simple.

Before:
```
public class SqlUserStore implements UserStore {
  private final UserCache cache;
  
  public SqlUserStore() {
    this.cache = new UserCache();
  }

  public User createUser() {
    long id = this.insertUser();
    User user = new User(id);
    cache.register(user);
    return user;
  }

  @Override
  @SqlUpdate("INSERT INTO users VALUES (...)")
  @GetGeneratedKeys
  long insertUser();
}

// In your initialization logic
UserStore userStore = database.jdbi.onDemand(SqlUserStore.class)
```

After:
```
public class SqlUserStore implements UserStore {
  private final SqlUsers sqlUsers;
  private final UserCache cache;
  
  public SqlUserStore(SqlUsers sqlUsers) {
    this.sqlUsers = sqlUsers;
    this.cache = new UserCache();
  }

  public User createUser() {
    long id = sqlUsers.insertUser();
    User user = new User(id);
    cache.register(user);
    return user;
  }
}

public interface SqlUsers {
  @Override
  @SqlUpdate("INSERT INTO users VALUES (...)")
  @GetGeneratedKeys
  long insertUser();
}

// In your initialization logic
UserStore userStore = new SqlUserStore(database.jdbi.onDemand(SqlUsers.class))
```

Result: separation of queries into a separate interface away from logic/state. I like it.

stevenschlansker:
Our decision here turned out to be quite prescient.  cglib isn't going to work going forward, at least according to This Guy On The Internet:
https://dzone.com/articles/jdk-11-and-proxies-in-a-world-past-sunmiscunsafe

Issue #438, opened 2016-08-08
Honor sqlobject annotations in supertypes


qualidafial:
Noticed this while reviewing #415. If a sql object type has an annotation on the type e.g. `@RegisterMapper(FooMapper.class)`, then subtypes have to redeclare those annotations. They are not inherited.

```
@RegisterMapper(MyMapper.class)
interface FooDao {
  ...
}

interface BarDao extends FooDao {}
```

In this example, the `@RegisterMapper(MyMapper.class)` is not picked up when opening a SQL object of type `BarDao`.

qualidafial:
Looking into this.

Getting SQL Object to process annotations on supertypes was easy. Getting it process annotations from super methods is a different matter.

There doesn't appear to be an inexpensive way to reflectively answer the question: does this method override a supertype method, and can I get the method object for the supertype?

On one hand, I'm annoyed that this is so difficult.

On the other hand, it's probably for the best. Overriding methods can totally replace the behavior of the overridden method. In that sense, having to redeclare any annotations needed on the overriding method is probably excusable.

qualidafial:
Tracked in #956

PR #458, opened 2016-08-25
Expose SqlObject dispatch machinery for public use

### LINKED ISSUE #455: Expose SqlObject dispatch machinery for public use
I propose making the following SqlObject types public:

* `Handler`
* `@SqlMethodAnnotation` and `HandlerFactory`
* `@SqlMethodDecoratingAnnotation` and `HandlerDecorator`

This will allow users and library authors to create their own SQL method handlers and decorators, which will in turn make caching feasible as a bolt-on option for users who want it, although there may be other use cases we haven't thought of.

We should also introduce a way of specifying order for when there are multiple decorators. My gut tells me that a static `@Priority` annotation is not the best approach, since users might want different ordering in different scenarios.

qualidafial:
Fixes #455

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/Handler.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/Handler.java
@@ -17,7 +17,20 @@ import java.lang.reflect.Method;
 
 import org.jdbi.v3.core.HandleSupplier;
 
-interface Handler
-{
-    Object invoke(HandleSupplier handle, SqlObjectConfig config, Object target, Object[] args, Method method) throws Exception;
+/**
+ * Implements the contract of a SQL Object method.
+ */
+public interface Handler {
+    /**
+     * Executes a SQL Object method, and returns the result.
+     *
+     * @param target the SQL Object instance being invoked
+     * @param method the method being invoked on the SQL Object.
+     * @param args   the arguments that were passed to the method.
+     * @param config the SQL Object configuration.
+     * @param handle a (possibly lazy) Handle supplier.
+     * @return the method return value, or null if the method has a void return type.
+     * @throws Exception any exception thrown by the method.
+     */
+    Object invoke(Object target, Method method, Object[] args, SqlObjectConfig config, HandleSupplier handle) throws Exception;
 }
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
@@ -100,9 +100,12 @@ public enum SqlObjectFactory implements ExtensionFactory<SqlObjectConfig> {
             return (Factory) e.create();
         });
 
+        SqlObjectConfig instanceConfig = config.createCopy();
+        forEachConfigurerFactory(extensionType, (factory, annotation) ->
+                factory.createForType(annotation, extensionType).accept(instanceConfig));
 
         Map<Method, Handler> handlers = buildHandlersFor(extensionType);
-        MethodInterceptor interceptor = createMethodInterceptor(extensionType, config, handlers, handle);
+        MethodInterceptor interceptor = createMethodInterceptor(extensionType, instanceConfig, handlers, handle);
         return extensionType.cast(f.newInstance(interceptor));
     }
 
@@ -189,7 +192,7 @@ public enum SqlObjectFactory implements ExtensionFactory<SqlObjectConfig> {
     }
 
     private MethodInterceptor createMethodInterceptor(Class<?> sqlObjectType,
-                                                      SqlObjectConfig baseConfig,
+                                                      SqlObjectConfig instanceConfig,
                                                       Map<Method, Handler> handlers,
                                                       HandleSupplier handle) {
         return (proxy, method, args, methodProxy) -> {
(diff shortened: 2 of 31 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -158,9 +160,18 @@ private Handler buildBaseHandler(Class<?> sqlObjectType, Method method) {
...
     private Handler addDecorators(Handler handler, Class<?> sqlObjectType, Method method) {
-        List<HandlerDecorator> decorators = Stream.of(method.getAnnotations())
-                .map(a -> a.annotationType().getAnnotation(SqlMethodDecoratingAnnotation.class))
-                .filter(Objects::nonNull)
+        List<Class<? extends Annotation>> annotationTypes = Stream.of(method.getAnnotations())
+                .map(Annotation::annotationType)
+                .filter(type -> type.isAnnotationPresent(SqlMethodDecoratingAnnotation.class))
+                .collect(toList());
+
+        DecoratorOrder order = method.getAnnotation(DecoratorOrder.class);
```
stevenschlansker: Should this also consider the `method.getDeclaringClass()` and potentially superinterfaces?  If you actually have to specify an order, it will be inconvenient to do so on every method.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestSqlMethodDecoratingAnnotations.java
```diff
@@ -0,0 +1,138 @@
...
+        dao.unordered();
+
+        List<String> invocations = TestSqlMethodDecoratingAnnotations.invocations.get();
+        assertTrue(invocations.contains("foo"));
+        assertTrue(invocations.indexOf("foo") < 2);
+
+        assertTrue(invocations.contains("bar"));
+        assertTrue(invocations.indexOf("bar") < 2);
+
+        assertTrue(invocations.indexOf("method") == 2);
```
stevenschlansker: nit: check size of `invocations`, what if it accidentally does
```
foo
bar
method
method
```

sqlobject/src/main/java/org/jdbi/v3/sqlobject/Handler.java
```diff
@@ -17,7 +17,20 @@
...
+     *
+     * @param target the SQL Object instance being invoked
+     * @param method the method being invoked on the SQL Object.
+     * @param args   the arguments that were passed to the method.
+     * @param config the SQL Object configuration.
+     * @param handle a (possibly lazy) Handle supplier.
+     * @return the method return value, or null if the method has a void return type.
+     * @throws Exception any exception thrown by the method.
+     */
+    Object invoke(Object target, Method method, Object[] args, SqlObjectConfig config, HandleSupplier handle) throws Exception;
```
stevenschlansker: Should this take a statement context or the like?  Potentially instead of `SqlObjectConfig`?  Seems that a third party `Handler` would more likely need to access its own configuration, or lookup arguments / mappers / etc.

stevenschlansker:
Nice!  Left a couple of thoughts for you to consider, then :shipit:

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
qualidafial: That's a good idea. I'll make it so you can annotate the sql object type.

There's already issue #438 for considering the annotations on supertypes--let's discuss that part of your suggestion there.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/Handler.java
qualidafial: The statement context doesn't exist until you create a statement from the innermost handle, e.g. `handle.get().createQuery("select ...")`. What about the JdbiConfig?

stevenschlansker: `JdbiConfig` could make sense, except we'd probably have to make it public and move a fair amount of `StatementContext` code into it, and then leave delegate methods behind.

I wonder if it makes sense to move the `StatementContext` construction to *before* the `Handler` chain is invoked -- if you actually want to change the behavior of the statement, say register a dynamic mapper or tweak anything aside from the actual parameters, that seems like the natural place to make 'statement-local' changes.

qualidafial: TBH I don't see why we would need this. Do you have a specific use case in mind?

qualidafial: Just as a refresher: The reason we added `SqlObjectConfig` was so that the extension could have its own place to store configuration. e.g. `SqlLocator` is now owned by sqlobject, and core knows nothing about it.

Generically, `ExtensionConfig` exists so that third-party extensions can have their own place to store any configuration that core knows nothing about.

qualidafial: > Seems that a third party Handler would more likely need to access its own configuration, or lookup arguments / mappers / etc.

qualidafial: I'm going to go ahead and merge this as is. If you feel strongly about exposing the `StatementContext` then let's discuss that on another ticket.

PR #463, opened 2016-08-30
Rename HandleSupplier.get() to .getHandle(), add documentation.

qualidafial:
Just a small refactor so that documentation is more complete.

HandleSupplier used to implement Supplier<Handle> but now we just declare out own method for this, with our own JDBI-specific documentation.


qualidafial:
@arteam Would you review?

arteam:
LGTM!

PR #465, opened 2016-08-31
Fix AbstractMethodError being swallowed in pass-through methods.

### LINKED ISSUE #464: AbstractMethodError are swallowed in pass-through methods
If an `AbstractMethodError` is thrown while evaluating a pass-through method, it is caught by `PassThroughHandler#invoke`, but a new `AbstractMethodError` is thrown instead without any of the original context ([link to source](https://github.com/jdbi/jdbi/blob/c8aa6a0/src/main/java/org/skife/jdbi/v2/sqlobject/PassThroughHandler.java#L37)). It turns out my `ResultSetMapper` was throwing `AbstractMethodError` because of a JDBC driver mismatch, but JDBI just reported that my method "didn't make sense".

qualidafial:
Fixes #464.


arteam:
LGTM as well.

PR #469, opened 2016-09-01
ConstructorMapper: ignore extra rows in the result set

stevenschlansker:
needed for the upcoming JoinRowMapper

PR #470, opened 2016-09-01
Join row mapper inintial commit

stevenschlansker:
even has docs!  requires #469 to be reviewed first

### DIFF
--- a/core/src/test/java/org/jdbi/v3/core/mapper/JoinRowMapperTest.java
+++ b/core/src/test/java/org/jdbi/v3/core/mapper/JoinRowMapperTest.java
@@ -1,0 +1,149 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.mapper;
+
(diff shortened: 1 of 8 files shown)


core/src/test/java/org/jdbi/v3/core/mapper/JoinRowMapperTest.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Test
+    public void testJoin() throws Exception
+    {
+       // tag::multimap[]
+        Multimap<User, Article> joined = HashMultimap.create();
+        h.createQuery("SELECT * FROM user NATURAL JOIN author NATURAL JOIN article")
+            .map(JoinRowMapper.forTypes(User.class, Article.class))
+            .forEach(jr -> joined.put(jr.get(User.class), jr.get(Article.class)));
```
qualidafial: Multimap is pretty much the greatest

docs/src/adoc/joins.adoc
```diff
@@ -0,0 +1,39 @@
+= Join Mapping
+
+Joining multiple tables together is a very common database
+task.  It is also where the mismatch between the relational
+model and Java's object model starts to rear its ugly head.
+
+Here we present a couple of strategies for retrieving
+results from more complicated rows.
+
+== JoinRowMapper
```
qualidafial: This would be a good place to describe folds as well. Maybe show the new sexy way first, and then show them how to "drop down" to folds if join mapping doesn't suffice.

qualidafial:
This is an elegant, delightful fix for the join problem.

Would be great if we also added a `@RegisterJoinRowMapper({Foo.class, Bar.class, ...})` annotation so SQL objects could return `JoinRow`s.

💯 👍 :shipit:

qualidafial:
I'm done reviewing this. @arteam did you have anything you wanted to add?

stevenschlansker:
Addressed the feedback.  Will merge sometime tomorrow if there's no further comments.

arteam:
Sorry for a lack of feedback from me. This LGTM.

PR #472, opened 2016-09-02
Type-safe version of DBIFactoryBean

robfletcher:
I updated `DBIFactoryBean` to specify a generic type so it can be used to autowire by type. I had to update the Spring dependency so picked the lowest version that provided `FactoryBean<T>`. In the newer version of Spring it's also necessary to explicitly proxy the `DataSource` used by `DBIFactoryBean` for transaction-awareness.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,7 +82,7 @@
         <project.build.targetJdk>1.6</project.build.targetJdk>
         <basepom.maven.version>2.0.9</basepom.maven.version>
         <dep.antlr.version>3.4</dep.antlr.version>
-        <dep.spring.version>2.0.1</dep.spring.version>
+        <dep.spring.version>3.0.0.RELEASE</dep.spring.version>
         <basepom.check.fail-all>true</basepom.check.fail-all>
 
         <basepom.test.fork-count>8</basepom.test.fork-count>
@@ -118,7 +118,21 @@
 
         <dependency>
             <groupId>org.springframework</groupId>
-            <artifactId>spring</artifactId>
+            <artifactId>spring-beans</artifactId>
+            <version>${dep.spring.version}</version>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-jdbc</artifactId>
+            <version>${dep.spring.version}</version>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-tx</artifactId>
             <version>${dep.spring.version}</version>
             <optional>true</optional>
         </dependency>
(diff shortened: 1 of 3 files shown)


qualidafial:
@stevenschlansker You seem to be the most knowledgeable in the team on Spring. You wanna look at this?

arteam:
If people are interested in the Spring integration for the v2 branch, I think we could try to backport this [commit](https://github.com/jdbi/jdbi/commit/e81bbe0f33c787365e602da3ebd0bf494e4b636a) from v3. My biggest concern is that it's probably not a binary-compatible change, so it may affect people who use old versions of Spring.

qualidafial:
This is where I think it makes sense to have separate artifacts per dependency major version (or even minor, if they break compat), e.g. `jdbi3-spring3`, `jdbi3-spring4`

stevenschlansker:
@robfletcher Thanks for the contribution -- unfortunately this is a compatibility breaking change, so at the very least I think we need to consult with other Spring users to make sure we aren't going to break too much stuff.  Another option would be to take our upcoming `jdbi3` release for a test drive, `jdbi3-spring` should be much more up to date.

stevenschlansker:
I've reached out to the mailing list to see if anyone objects.  We'll hold this PR for a couple of days pending comment, but the code LGTM.

robfletcher:
@stevenschlansker understood. FWIW I can imagine a `jdbi-spring` artifact, or series of artifacts, might make sense. Although I guess there's a question of how old a version of Spring do you actually want to support? Spring 3.0 is almost 7 years old now!

One other thing I'd like to do is auto-wire type things like `ResultSetMapperFactory` and `ArgumentFactory` instances into the DBI factory.

stevenschlansker:
Right -- I'm not at all saying we should continue to support Spring 2, what I am more worried about is if someone is using Spring 4 with our `DBIFactoryBean`, upgrading JDBI to be compiled against Spring 3 may break their usage, even though they've never touched Spring 2 at all.  I'm hopeful that all our Spring users will prefer to have up to date dependencies and are OK with a minor build breakage as long as it doesn't happen often, so I would prefer to just go with this commit and avoid the complexity of managing separate artifacts, but maybe someone else will disagree...

stevenschlansker:
We'd also totally love a PR for those features as well.  And I think they don't break compatibility so we could include them no problem.

robfletcher:
Yeah, I think the `FactoryBean<T>` interface and anything I'd need to touch to implement auto-wiring (basically just the `@Autowired` annotation) are unlikely to result in breaking changes going forward.

seancarroll:
I'm currently using Spring 4.3.2 with JDBI 2.73 and would be happy to help test

stevenschlansker:
@seancarroll If you would rebuild your project with the latest `2.76-SNAPSHOT` and make sure there's no ugly surprises that'd be fantastic, thanks :)

seancarroll:
I'm pretty sure I didn't do it in the most efficient way possible but I did rebuild with the latest and ran my unit/integration tests. I didn't encounter any issues.

stevenschlansker:
Thanks @seancarroll I was pretty sure it'd be an easy upgrade but it's always good to check!  Thanks @robfletcher for the contribution.

Issue #473, opened 2016-09-04
Support arrays in BindIn, besides Collection


TheRealMarnes:
The current code of BindIn assumes the value it was passed is a Collection<?>:
`Collection<?> coll = (Collection<?>) arg;`

Could an instanceof check be added to handle not only Collection<?>, but also arrays like Object[]? This would allow us to pass arrays and varargs to BindIn, which would be very convenient in some cases.

After determining the argument is an array, it can be cast and then copied into a (Array)List, either manually or with any of the utility classes like Arrays.asList or Collections.addAll, depending on what sort of compatibility you strive for.

If the argument is neither a Collection nor an array, an IllegalArgumentException could be thrown.

Thanks

PR #474, opened 2016-09-04
Support arrays in BindIn, besides Collection #473

TheRealMarnes:
https://github.com/jdbi/jdbi/issues/473

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
+++ b/src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
@@ -24,62 +24,114 @@ import org.skife.jdbi.v2.sqlobject.SqlStatementCustomizingAnnotation;
 import java.lang.annotation.Annotation;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
+import java.lang.reflect.Array;
 import java.lang.reflect.Method;
-import java.sql.SQLException;
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
-import java.util.List;
 
+/**
+ * Binds an Iterable or array/varargs of any type to a placeholder as a comma-separated list (e.g. for WHERE X IN (...) query statements).
+ *
+ * Throws IllegalArgumentException if the argument is not an array or Iterable. How null and empty collections are handled can be configured with onEmpty:EmptyHandling - throws IllegalArgumentException by default.
+ *
+ * Don't forget to add @UseStringTemplate3StatementLocator to your query class and use the correct placeholder format:
+ *  {@literal @}SqlQuery("SELECT * FROM THINGS WHERE ID IN ({@literal <}ids{@literal >})")
+ *  abstract Object[] foo({@literal @}BindIn("ids") int[] ids);
+ *  ids = [1, 2, 3] -> SELECT * FROM THINGS WHERE ID IN (1,2,3)
+ */
 @Retention(RetentionPolicy.RUNTIME)
 @SqlStatementCustomizingAnnotation(BindIn.CustomizerFactory.class)
 @BindingAnnotation(BindIn.BindingFactory.class)
 public @interface BindIn
 {
+    /**
+     * placeholder in your query to be replaced with comma-separated list
+     */
     String value();
+
+    /**
+     * what to do when the argument is null or empty
+     */
+    EmptyHandling onEmpty() default EmptyHandling.THROW;
 
     final class CustomizerFactory implements SqlStatementCustomizerFactory
     {
-
         @Override
-        public SqlStatementCustomizer createForMethod(Annotation annotation, Class sqlObjectType, Method method)
+        public SqlStatementCustomizer createForMethod(final Annotation annotation, final Class sqlObjectType, final Method method)
         {
-            throw new UnsupportedOperationException("Not supported on method!");
+            throw new UnsupportedOperationException("Not supported on method");
         }
 
         @Override
-        public SqlStatementCustomizer createForType(Annotation annotation, Class sqlObjectType)
+        public SqlStatementCustomizer createForType(final Annotation annotation, final Class sqlObjectType)
         {
             throw new UnsupportedOperationException("Not supported on type");
         }
 
         @Override
-        public SqlStatementCustomizer createForParameter(Annotation annotation,
-                                                         Class sqlObjectType,
-                                                         Method method,
-                                                         Object arg)
+        public SqlStatementCustomizer createForParameter(final Annotation annotation, final Class sqlObjectType, final Method method, final Object arg)
         {
-            Collection<?> coll = (Collection<?>) arg;
-            BindIn in = (BindIn) annotation;
-            final String key = in.value();
-            final List<String> ids = new ArrayList<String>();
-            for (int idx = 0; idx < coll.size(); idx++) {
-                ids.add("__" + key + "_" + idx);
+            final BindIn bindIn = (BindIn) annotation;
+
+            final int size;
+            if (arg == null)
+            {
+                switch (bindIn.onEmpty())
+                {
+                    case VOID:
+                        // skip argument iteration altogether to output nothing at all
+                        size = 0;
+                        break;
+                    case NULL:
+                        // output exactly 1 value: null
+                        size = 1;
+                        break;
+                    case THROW:
+                        throw new IllegalArgumentException("argument is null; null was explicitly forbidden on this instance of BindIn");
+                    default:
+                        throw new IllegalStateException(EmptyHandling.valueNotHandledMessage);
+                }
+            } else
+            {
+                size = Util.size(arg);
             }
 
-            StringBuilder names = new StringBuilder();
-            for (Iterator<String> i = ids.iterator(); i.hasNext();) {
-                names.append(":").append(i.next());
-                if (i.hasNext()) {
+            if (size == 0)
+            {
+                switch (bindIn.onEmpty())
+                {
+                    case VOID:
+                        // output nothing - taken care of with size = 0
+                        break;
+                    case NULL:
+                        // output null - handle in Binder
+                        break;
+                    case THROW:
+                        throw new IllegalArgumentException("argument is empty; emptiness was explicitly forbidden on this instance of BindIn");
+                    default:
+                        throw new IllegalStateException(EmptyHandling.valueNotHandledMessage);
+                }
+            }
+
+            final String key = bindIn.value();
+
+            // generate and concat placeholders
+            final StringBuilder names = new StringBuilder();
+            for (int i = 0; i < size; i++)
+            {
+                if (i > 0)
+                {
                     names.append(",");
                 }
+                names.append(":__").append(key).append("_").append(i);
             }
             final String ns = names.toString();
 
             return new SqlStatementCustomizer()
             {
                 @Override
-                public void apply(SQLStatement q) throws SQLException
+                public void apply(final SQLStatement q)
                 {
                     q.define(key, ns);
                 }
@@ -90,21 +142,139 @@ public @interface BindIn
     class BindingFactory implements BinderFactory<BindIn>
     {
         @Override
-        public Binder build(BindIn in)
+        public Binder build(final BindIn bindIn)
         {
-            final String key = in.value();
+            final String key = bindIn.value();
 
-            return new Binder<Annotation, Iterable>()
+            return new Binder<Annotation, Object>()
             {
                 @Override
-                public void bind(SQLStatement q, Annotation bind, Iterable coll)
+                public void bind(final SQLStatement q, final Annotation bind, final Object arg)
                 {
-                    int idx = 0;
-                    for (Object s : coll) {
-                        q.bind("__" + key + "_" + idx++, s);
+                    if (arg == null || Util.size(arg) == 0)
+                    {
+                        switch (bindIn.onEmpty())
+                        {
+                            case VOID:
+                                // output nothing, end now
+                                break;
+                            case NULL:
+                                // output null
+                                q.bind("__" + key + "_0", (String) null);
+                                break;
+                            case THROW:
+                                final Exception inner = new IllegalArgumentException("argument is null; null was explicitly forbidden on this instance of BindIn");
+                                throw new IllegalStateException("Illegal argument value was caught too late. Please report this to the jdbi developers.", inner);
+                            default:
+                                throw new IllegalStateException(EmptyHandling.valueNotHandledMessage);
+                        }
+                    } else
+                    {
+                        // replace placeholders with actual values
+                        final Iterator it = Util.toIterator(arg);
+                        for (int i = 0; it.hasNext(); i++)
+                        {
+                            q.bind("__" + key + "_" + i, it.next());
+                        }
                     }
                 }
             };
         }
+    }
+
+    final class Util
+    {
+        private Util()
+        {
+        }
+
+        static Iterator toIterator(final Object obj)
+        {
+            if (obj == null)
+            {
+                throw new IllegalArgumentException("cannot make iterator of null");
+            }
+
+            if (obj instanceof Iterable)
+            {
+                return ((Iterable) obj).iterator();
+            }
+
+            if (obj.getClass().isArray())
+            {
+                if (obj instanceof Object[])
+                {
+                    return Arrays.asList((Object[]) obj).iterator();
+                } else
+                {
+                    return new ReflectionArrayIterator(obj);
+                }
+            }
+
+            throw new IllegalArgumentException(getTypeWarning(obj.getClass()));
+        }
+
+        static int size(final Object obj)
+        {
+            if (obj == null)
+            {
+                throw new IllegalArgumentException("cannot get size of null");
+            }
+
+            if (obj instanceof Collection)
+            {
+                return ((Collection) obj).size();
+            }
+
+            if (obj instanceof Iterable)
+            {
+                final Iterable<?> iterable = (Iterable) obj;
+
+                int size = 0;
+                for (final Object x : iterable)
+                {
+                    size++;
+                }
+
+                return size;
+            }
+
+            if (obj.getClass().isArray())
+            {
+                return Array.getLength(obj);
+            }
+
+            throw new IllegalArgumentException(getTypeWarning(obj.getClass()));
+        }
+
+        private static String getTypeWarning(final Class type)
+        {
+            return "argument must be one of the following: Iterable, or an array/varargs (primitive or complex type); was " + type.getName() + " instead";
+        }
+    }
+
+    /**
+     * describes what needs to be done if the passed argument is null or empty
+     */
+    enum EmptyHandling
+    {
+        /**
+         * output "" (without quotes, i.e. nothing)
+         *
+         * select * from things where x in ()
+         */
+        VOID,
+        /**
+         * output "null" (without quotes, as keyword), useful e.g. in postgresql where "in ()" is invalid syntax
+         *
+         * select * from things where x in (null)
+         */
+        NULL,
+        /**
+         * throw IllegalArgumentException
+         */
+        THROW;
+
+        static final String valueNotHandledMessage = "EmptyHandling type on BindIn not handled. Please report this to the jdbi developers.";
     }
 }
--- a/src/test/java/org/skife/jdbi/v2/sqlobject/BindInNullTest.java
+++ b/src/test/java/org/skife/jdbi/v2/sqlobject/BindInNullTest.java
@@ -1,0 +1,161 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.skife.jdbi.v2.sqlobject;
+
+import org.h2.jdbcx.JdbcDataSource;
+import org.hamcrest.CoreMatchers;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.skife.jdbi.v2.DBI;
+import org.skife.jdbi.v2.Handle;
+import org.skife.jdbi.v2.Something;
+import org.skife.jdbi.v2.logging.FormattedLog;
+import org.skife.jdbi.v2.sqlobject.stringtemplate.UseStringTemplate3StatementLocator;
+import org.skife.jdbi.v2.tweak.SQLLog;
+import org.skife.jdbi.v2.unstable.BindIn;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+import static org.skife.jdbi.v2.unstable.BindIn.EmptyHandling.NULL;
+import static org.skife.jdbi.v2.unstable.BindIn.EmptyHandling.VOID;
+
+public class BindInNullTest
+{
+    private static Handle handle;
+
+    @BeforeClass
+    public static void init()
+    {
+        final JdbcDataSource ds = new JdbcDataSource();
+        ds.setURL("jdbc:h2:mem:" + UUID.randomUUID());
+        final DBI dbi = new DBI(ds);
+        dbi.registerMapper(new SomethingMapper());
+        handle = dbi.open();
+
+        handle.execute("create table something (id int primary key, name varchar(100))");
+        handle.execute("insert into something(id, name) values(1, null)");
+        handle.execute("insert into something(id, name) values(2, null)");
+        handle.execute("insert into something(id, name) values(3, null)");
+        handle.execute("insert into something(id, name) values(4, null)");
+
+        handle.execute("insert into something(id, name) values(5, 'bla')");
+        handle.execute("insert into something(id, name) values(6, 'null')");
+        handle.execute("insert into something(id, name) values(7, '')");
+    }
+
+    @AfterClass
+    public static void exit()
+    {
+        handle.execute("drop table something");
+        handle.close();
+    }
+
+    // TODO run the following 2 tests against a postgresql db
+    // due to queries being parameterized with ?, we can't really test that the query's final form contains "in (null)", i.e. that EmptyHandling.NULL works as intended on a db like postgresql that demands the "in (null)" syntax
+    // all we can test is that the query returns no rows, as specified by sql's general design
+    // h2 accepts either form without causing trouble
+
+    @Test
+    public void testSomethingByIterableHandleNullWithNull()
+    {
+        final SomethingByIterableHandleNull s = handle.attach(SomethingByIterableHandleNull.class);
+
+        final List<Something> out = s.get(null);
+
+        Assert.assertEquals(0, out.size());
+    }
+
+    @Test
+    public void testSomethingByIterableHandleNullWithEmptyList()
+    {
+        final SomethingByIterableHandleNull s = handle.attach(SomethingByIterableHandleNull.class);
+
+        final List<Something> out = s.get(new ArrayList<Object>());
+
+        Assert.assertEquals(0, out.size());
+    }
+
+    @UseStringTemplate3StatementLocator
+    private interface SomethingByIterableHandleNull
+    {
+        @SqlQuery("select id, name from something where name in (<names>)")
+        List<Something> get(@BindIn(value = "names", onEmpty = NULL) Iterable<Object> ids);
+    }
+
+    //
+
+    @Test
+    public void testSomethingByIterableHandleVoidWithNull()
+    {
+        final TestSqlLogger logger = new TestSqlLogger();
+        handle.setSQLLog(logger);
+        final SomethingByIterableHandleVoid s = handle.attach(SomethingByIterableHandleVoid.class);
+
+        final List<Something> out = s.get(null);
+        final List<String> log = logger.getLog();
+
+        Assert.assertEquals(0, out.size());
+        Assert.assertEquals(1, log.size());
+        Assert.assertThat(log.get(0), CoreMatchers.containsString(" where id in ();"));
+    }
+
+    @Test
+    public void testSomethingByIterableHandleVoidWithEmptyList()
+    {
+        final TestSqlLogger logger = new TestSqlLogger();
+        handle.setSQLLog(logger);
+        final SomethingByIterableHandleVoid s = handle.attach(SomethingByIterableHandleVoid.class);
+
+        final List<Something> out = s.get(new ArrayList<Object>());
+        final List<String> log = logger.getLog();
+
+        Assert.assertEquals(0, out.size());
+        Assert.assertEquals(1, log.size());
+        Assert.assertThat(log.get(0), CoreMatchers.containsString(" where id in ();"));
+    }
+
+    @UseStringTemplate3StatementLocator
+    private interface SomethingByIterableHandleVoid
+    {
+        @SqlQuery("select id, name from something where id in (<ids>);")
+        List<Something> get(@BindIn(value = "ids", onEmpty = VOID) Iterable<Object> ids);
+    }
+
+    private class TestSqlLogger extends FormattedLog implements SQLLog
+    {
+        final List<String> log = new ArrayList<String>();
+
+        @Override
+        protected boolean isEnabled()
+        {
+            return true;
+        }
+
+        @Override
+        protected void log(final String msg)
+        {
+            log.add(msg);
+        }
+
+        public List<String> getLog()
+        {
+            // make a copy just to guarantee integrity
+            return new ArrayList<String>(log);
+        }
+    }
+}
--- a/src/test/java/org/skife/jdbi/v2/sqlobject/BindInTest.java
+++ b/src/test/java/org/skife/jdbi/v2/sqlobject/BindInTest.java
@@ -1,0 +1,222 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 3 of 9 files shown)


src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
```diff
@@ -85,6 +86,42 @@ public void apply(SQLStatement q) throws SQLException
                 }
             };
         }
+
+		private static Collection<?> tryToCast(Object arg) {
+			if (arg instanceof Collection) {
+				return (Collection<?>) arg;
+			}
+			if (arg instanceof Object[]) {
```
qualidafial: I'd probably like to see this the other way around: convert collections to an array and bind from the array. You can use `java.lang.reflect.Array.getLength(array)` and `Array.get(array, index)` and then you don't have to have all this code testing for different array types.

Also, there's a checkstyle error failing the build, because your code editor is using tabs instead of spaces for indentation. This project uses spaces.

TheRealMarnes:
@qualidafial sorry, I only meant the pull request as a rough draft that needs to be fine-tuned before fully merging, not as a ready-to-go submission... There's a million aspects to how a piece of code is written, like logging, exception catching and throwing, message style, considerations like which situations to check first, or like you said indeed maybe to use an array instead of a list entirely, etc. I can't possibly know how you guys like the code to be at every level, just wanted to write this as a quick POC

TheRealMarnes:
Isn't using reflection kinda bad btw? It's slow and hacky, while the type check may be verbose but it's fast and clean

qualidafial:
Java reflection hasn't been "slow" for years. Certainly it's slower than static code, but not by the crippling factor that it used to be. Moreover as seldom as `@BindIn` is used in practice, this case is not likely to become a performance bottleneck.

If you can demonstrate that use of reflection here adversely affects performance in real-world scenarios, I'd be happy to revisit. But absent any evidence of performance issues, I'd rather use code that is maintainable and easy to understand.

TheRealMarnes:
Fair enough, I'll patch it up some more. I'm not saying it'll be perfect though, someone who's more in on things in jdbi's specific domain should finetune it thoroughly anyway

qualidafial:
Just a heads up: unless a project is using compile-time annotation processors (e.g. Dagger DI), most libraries that make use of annotations are doing so through reflection. Nearly all annotation-driven APIs I'm aware of do their magic through reflection--JDBI included.

stevenschlansker:
That's a little different though.  Accessing annotations via reflection is once per statement, whereas mapping like this will be once per value converted.  So it will be many orders of magnitude more common.  I still lean towards the readable version first, then we can iterate and improve it as we find bottlenecks, but it's quite reasonable to consider the performance cost.  And thanks @TheRealMarnes for giving this a go, don't worry at all if you are iteratively improving it -- that's what code review is for :)

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
```diff
@@ -85,6 +86,42 @@ public void apply(SQLStatement q) throws SQLException
...
+				return Arrays.asList((Object[]) arg);
+			}
+			// necessary because of how java handles primitives...
+			if (arg instanceof boolean[]) {
+				return Arrays.asList((boolean[]) arg);
+			}
+			if (arg instanceof byte[]) {
+				return Arrays.asList((byte[]) arg);
```
stevenschlansker: Are you sure this construct actually works?  I believe this is actually creating a `List<byte[]>` which is not what you want.

stevenschlansker:
Needs some test cases.  I suspect that adding tests will reveal that the `Arrays.asList` approach won't work -- it won't do the dirty work of boxing primitives or the like.

TheRealMarnes:
I'm going to put the casting/reflection utility method in a utility class so it can be cleanly tested. Where do you guys want it to be (package and/or class)?

stevenschlansker:
I think there's a `.util` package that might naturally host it.  Other than that, no particular preference, just come up with a good name for it :)

TheRealMarnes:
Dunno, that package seems to be (and the package.html even says so) meant for utilities for the end users. This utility is something that should stay out of the end user's way. Maybe I should make a util_private package?

stevenschlansker:
Or just a different class in the same package -- that way you can make it package-private and not visible to end users.

TheRealMarnes:
Not the cleanest from a project files pov, but that does make it completely invisible to users. I like that idea more :P

TheRealMarnes:
For some reason I can't get the code to compile locally, in intellij or maven, because of missing dependencies (org.skife.jdbi.rewriter package among others). I'll just push and see if jenkins starts building it or anything. I'm not very familiar with github or this project

src/main/java/org/skife/jdbi/v2/unstable/Objects.java
```diff
@@ -0,0 +1,27 @@
...
+    private Objects() {}
+
+    static Object[] toArray(final Object obj) {
+        if (obj instanceof Collection) {
+            return ((Collection) obj).toArray();
+        }
+
+        if (obj.getClass().isArray()) {
+            final int length = Array.getLength(obj);
+            final Object[] out = new Object[length];
```
qualidafial: You only need to copy the array if `obj.getClass().getComponentType().isPrimitive()`. Otherwise you can just cast to `Object[]`

TheRealMarnes:
Thanks for the catch, I'd forgotten to re-add that

TheRealMarnes:
Come to think of it, why not support Iterable<?>?

stevenschlansker:
Supporting `Iterable` makes a lot of sense.  You could even consider supporting `Iterator` too, although this starts to get more complex :)

Your build errors due to missing `rewriter` package is likely because IntelliJ doesn't know to build the ANTLR grammars -- this is handled in the `generate-sources` Maven phase: https://github.com/jdbi/jdbi/blob/master/pom.xml#L321-L332  So I'm a little surprised that Maven doesn't work -- what command do you invoke?  what error do you get?

Eclipse knows how to figure this out (with a plugin) but unfortunately I don't know how to get IntelliJ set up right.

TheRealMarnes:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-toolchains-plugin:1.1:toolchain (default) on project jdbi: Cannot find matching toolchain definitions for the following toolchain types:
[ERROR] jdk [ vendor='sun' version='1.6' ]
[ERROR] Please make sure you define the required toolchains in your ~/.m2/toolchains.xml file.

Yay... but wait, maven says java 6, the travis file says jdk7? Maybe the jdk version numbering is a little wonky and it's not like I've looked it up, but shouldn't it be the same?

stevenschlansker:
Ah, yeah this needs to be documented better.  You have to configure the Maven Toolchains on your dev box:
http://maven.apache.org/plugins/maven-toolchains-plugin/usage.html
http://maven.apache.org/plugins/maven-toolchains-plugin/toolchains/jdk.html

Travis only has openjdk7 type, but it actually has Java 6 still -- and jdbi v2 is based on Java 6, so we use Toolchains to make sure you use the right one:
https://github.com/jdbi/jdbi/blob/master/src/build/travis-toolchains.xml

arteam:
Yes, that is a rather confusing message. You need to set up Maven
toolchains locally to point to a JDK which can compile Java 6 code. I think
we moved this functionality to a specific profile in JDBI3, just for the
reason a to simplify the development setup for the people who want to
contribute.
Am 06.09.2016 22:01 schrieb "TheRealMarnes" <notifications@github.com>:

stevenschlansker:
Also worth noting -- if you don't have Java 6 installed, you don't actually have to.  You can point your jdk6 toolchain at a Java 8 just fine.  The downside is if you accidentally use any new features, you won't notice on a local build - but Travis will catch any errors you make.

TheRealMarnes:
Yeah, pointing it at my jdk8 is what I'm gonna go with...

TheRealMarnes:
Ok, now I run into runtime type exceptions when I run a full test of @BindIn (with varargs for starters)... That'll be for another time :/

stevenschlansker:
Please leave a note of any combinations you find that don't work!  We may not fix them all up for jdbi v2, but we are reworking a lot of this for v3, so we'd appreciate failing tests as issues or even a PR for the v3 branch 😄

TheRealMarnes:
For some reason I can't get a screenshot attached here... Anyway, I've pushed the test class for BindIn, it passes on Iterable but fails on varargs and Iterator. The error is

java.lang.ClassCastException: [I cannot be cast to java.lang.Iterable
	at org.skife.jdbi.v2.unstable.BindIn$BindingFactory$1.bind(BindIn.java:98)

TheRealMarnes:
For some reason I can't get a screenshot attached here... Anyway, I've pushed the test class for BindIn, it passes on Iterable but fails on varargs and Iterator. The error is

java.lang.ClassCastException: [I cannot be cast to java.lang.Iterable
	at org.skife.jdbi.v2.unstable.BindIn$BindingFactory$1.bind(BindIn.java:98)

TheRealMarnes:
For some reason I can't get a screenshot attached here... Anyway, I've pushed the test class for BindIn, it passes on Iterable but fails on varargs and Iterator. The error is

java.lang.ClassCastException: [I cannot be cast to java.lang.Iterable
	at org.skife.jdbi.v2.unstable.BindIn$BindingFactory$1.bind(BindIn.java:98)

qualidafial:
Something's broken in Github's UI so I can't leave a comment on the specific line, but in BindIn.java:

* Line 64: the contents of `argArray` are never used--just the length. The conversion to array is wasted here--it would be better to have an `Objects.size(arrayOrIterable)` method
* Lines 97-107: This is where the rubber hits the road for `@BindIn`: this is where you need to iterate  the iterable / array and bind each value to the named parameter introduced in `BindIn.CustomizerFactory.createForParameter`. Currently the code is still assuming the object is an `Iterable`, but not checking for arrays.

qualidafial:
Something's broken in Github's UI so I can't leave a comment on the specific line, but in BindIn.java:

* Line 64: the contents of `argArray` are never used--just the length. The conversion to array is wasted here--it would be better to have an `Objects.size(arrayOrIterable)` method
* Lines 97-107: This is where the rubber hits the road for `@BindIn`: this is where you need to iterate  the iterable / array and bind each value to the named parameter introduced in `BindIn.CustomizerFactory.createForParameter`. Currently the code is still assuming the object is an `Iterable`, but not checking for arrays.

TheRealMarnes:
Isn't a cast just a manual override of the reference type that doesn't actually involve any real work (like a true conversion would)?

qualidafial:
If you're talking about array casts, you're mostly correct. Array casts are checked at runtime, e.g. you can cast a `String[]` to `Object[]`, and an `Object[]` reference that points to a `String[]` can be converted back to `String[]`. However casting a `int[]` to `long[]` will throw a `ClassCastException`.

stevenschlansker:
Notably, primitive arrays do not descend from `Object[]` -- there is no common type to `String[]` and `int[]` other than `Object` itself.  That's why the `java.lang.reflect.Array` methods all take `Object` rather than an array type, and is also one reason why writing code to handle both is otherwise quite painful.  Java Arrays are covered in warts...

TheRealMarnes:
Yes, I know,  it's just... sorry, I'm having a big brainfart atm (6 hours of sleep in total over 2 nights might be related to that). I lost track since Matthew's comment about the line numbers, but I just dug around again, looked a bit further than my nose (I'm notorious for not doing that) and checked things out in debug mode again, and I think I my train of thought is back on rails.

What's happening is that Bindifier.bind is trying to pass BindIn's arguments (varargs, iterator, ...) to Binder.bind, which expects an iterable, and that's where it goes wrong. I think I was being misled a bit by the stacktrace, I'm used to writing code in such a way that type errors are impossible even at runtime, so it didn't occur to me the cast exception was coming from the method parameters themselves, and that's how I lost track.

TheRealMarnes:
It works for varargs :) Now to make it work with iterator

TheRealMarnes:
Aaahhh and now it's failing on iterator with "no parameter found for placeholder", because the iterator arg becomes an empty array... BindIn first checks its size and later has to iterate it again to get the contents, but checking its size already spent its iteration...

stevenschlansker:
Yeah, maybe not worth it.  Feel free to cut `Iterator` support, I think `Iterable` is enough.  If you actually want to do `Iterator` you probably have to dump it into a temporary array, which is not great.

TheRealMarnes:
Yeah, since iterators can't be reset and we can't really link the 2 parts of BindIn to share the original params, I think there's logically no way around it. I don't think there's much of a point to it anyway since the iterator needs to be completely iterated to fill up the query, so it's not like using an iterator instead of a collection/array is saving anything.

src/main/java/org/skife/jdbi/v2/unstable/Objects.java
```diff
@@ -0,0 +1,75 @@
...
+
+                return out;
+            }
+        }
+
+        throw new IllegalArgumentException("argument must be one of the following: Iterable, or an array/varargs (primitive or complex type)");
+    }
+
+    static int arraySize(final Object obj) {
+        if (obj instanceof Iterable) {
```
qualidafial: It may be worthwhile to check for `Collection` first, and return `((Collection)obj).size()`.

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
```diff
@@ -25,82 +25,72 @@
...
+                ids.add("__" + key + "_" + i);
             }
 
-            StringBuilder names = new StringBuilder();
-            for (Iterator<String> i = ids.iterator(); i.hasNext();) {
-                names.append(":").append(i.next());
-                if (i.hasNext()) {
+            final StringBuilder names = new StringBuilder();
+            for (final Iterator<String> it = ids.iterator(); it.hasNext(); ) {
+                names.append(":").append(it.next());
```
qualidafial: This is sort of orthogonal to this PR, but it would create less objects if the previous for-loop were collapsed into this one:

    for (int i = 0; i < size; i++) {
      if (id > 0) {
        names.append(",");
      }
      names.append(":__").append(key).append('_').append(i);
    }

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
```diff
@@ -25,82 +25,72 @@
...
-        public SqlStatementCustomizer createForMethod(Annotation annotation, Class sqlObjectType, Method method)
-        {
+        public SqlStatementCustomizer createForMethod(final Annotation annotation, final Class sqlObjectType, final Method method) {
             throw new UnsupportedOperationException("Not supported on method!");
         }
 
         @Override
-        public SqlStatementCustomizer createForType(Annotation annotation, Class sqlObjectType)
-        {
+        public SqlStatementCustomizer createForType(final Annotation annotation, final Class sqlObjectType) {
```
qualidafial: While I share your love of K&R brace style, we're standardized on Allman brace style (open brace on the next line) on this project.

More to the point though, changing lines just for formatting makes the real history of changes opaque when you need to see the reason a line of code got to be like it is. Can you revert these lines that are just making whitespace changes?

src/test/java/org/skife/jdbi/v2/sqlobject/TestBindIn.java
```diff
@@ -0,0 +1,123 @@
...
+        in.add(2);
+
+        s.get(in.iterator());
+    }
+
+    @UseStringTemplate3StatementLocator
+    private interface SomethingIterator {
+        @SqlQuery("select id, name from something where id in (<ids>)")
+        List<Something> get(@BindIn("ids") Iterator<Integer> ids);
+    }
```
qualidafial: What happens if you pass an empty array/collection to a `@BindIn`?

qualidafial:
Good tests.

qualidafial:
👍 from me, after addressing my comment about whitespace changes, and adding a test to characterize what happens when you pass an empty array / iterable for a `@BindIn` parameter.

TheRealMarnes:
Do we want to allow passing an empty argument to BindIn? Technically there's nothing wrong with it and neutral code that only crashes when it needs to is nice to have, but functionally it makes no sense and is probably an error the user wants to know about...

TheRealMarnes:
I think things are done. Thoughts, anyone?

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
```diff
@@ -90,21 +100,98 @@ public void apply(SQLStatement q) throws SQLException
...
+                    for (final Object s : arr)
+                    {
                         q.bind("__" + key + "_" + idx++, s);
                     }
                 }
             };
         }
     }
+
+    final class Util
```
stevenschlansker: nit: this class should be static

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
```diff
@@ -90,21 +100,98 @@ public void apply(SQLStatement q) throws SQLException
...
+    {
+        private Util()
+        {
+        }
+
+        static Object[] toArray(final Object obj)
```
stevenschlansker: Could consider converting to `Iterator` instead here -- returning `Arrays.asList(obj).iterator()` for non-primitive arrays, `Iterable.iterator()` for iterators, and a lazy `Iterator` over `Array.get()` for primitive arrays.

That would save having to assemble the whole array at once, you would lazily iterate over it.

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
```diff
@@ -90,21 +100,98 @@ public void apply(SQLStatement q) throws SQLException
...
+                    for (int i = 0; i < length; i++)
+                    {
+                        out[i] = Array.get(obj, i);
+                    }
+
+                    return out;
+                }
+            }
+
+            throw new IllegalArgumentException("argument must be one of the following: Iterable, or an array/varargs (primitive or complex type)");
```
stevenschlansker: Could consider appending `obj.getClass()` to the reason here, so you know what type was actually sent in, in case it isn't what you were expecting.

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
TheRealMarnes: isn't an inner class on an interface public static by default? Intellij points out that static is redundant when I try it, and classes outside the package can see BindIn.Util (suggesting the implicit public)

src/test/java/org/skife/jdbi/v2/sqlobject/BindInTest.java
```diff
@@ -0,0 +1,184 @@
...
+import java.util.UUID;
+
+public class BindInTest
+{
+    private static Handle handle;
+
+    @BeforeClass
+    public static void init()
+    {
+        final JdbcDataSource ds = new JdbcDataSource();
```
stevenschlansker: Could consider the `H2DatabaseRule` instead of doing manual setup.

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
TheRealMarnes: What would that save though? We're going to iterate over the entire bunch of elements every time anyway

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
TheRealMarnes: good point

stevenschlansker:
Nice tests!

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
stevenschlansker: Oops, missed that it was on an annotation.  My bad.

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
stevenschlansker: Consider binding a `List<MyBean>` -- right now, it'll assemble an intermediate `MyBean[]` which is only then used to iterate over to bind.  You could bind directly from the `Iterator` and remove the temporary copy.  It's a relatively small point, so it's an optional fixup if you like :)

stevenschlansker:
Consider the comments above (mostly the one about building intermediate arrays), then LGTM!

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
TheRealMarnes: Hmm, yeah, I suppose it does create a little less cruft/shortcuts a few steps... I mildly dislike the idea of passing iterators around because I prefer method arguments to be "natural" and obvious rather than obfuscated/indirect, but if I'm honest with myself that's not much of a good argument anyway. I'll patch it

src/test/java/org/skife/jdbi/v2/sqlobject/BindInTest.java
TheRealMarnes: Sorry, what?

TheRealMarnes:
Squash everything down to 1 commit then?

History gon' be rewritten guuuud...

stevenschlansker:
I'm not a stickler for exactly one commit all the time, but having history polluted with multiple rounds of checkstyle fixups isn't really helpful :)

I'm a fan of "as many commits as reasonably make sense in isolation" -- which, for a relatively self-contained new feature like this, is usually one or two commits IMO.

src/test/java/org/skife/jdbi/v2/sqlobject/BindInTest.java
stevenschlansker: https://github.com/jdbi/jdbi/blob/jdbi3/core/src/test/java/org/jdbi/v3/core/H2DatabaseRule.java

used like this

https://github.com/jdbi/jdbi/blob/jdbi3/core/src/test/java/org/jdbi/v3/core/TestUuid.java#L26-L27

TheRealMarnes:
Makes sense

*presses garbage compactor button*

TheRealMarnes:
CustomizerFactory is final, but BindingFactory is not. Is this intentional?

src/test/java/org/skife/jdbi/v2/sqlobject/BindInTest.java
TheRealMarnes: but that's v3, this code is in v2

stevenschlansker: Gah, I'm sorry, I keep getting my wires crossed here...

TheRealMarnes: :P

TheRealMarnes:
I added some null handling, exceptions and a tad of logging for the sake of not saddling users up with mysterious behavior. Is it overkill?

src/main/java/org/skife/jdbi/v2/unstable/ObjectIterator.java
```diff
@@ -0,0 +1,51 @@
...
+            return Array.get(arr, index++);
+        } else {
+            throw new ArrayIndexOutOfBoundsException("only " + size + " elements available");
+        }
+    }
+
+    @Override
+    public void remove()
+    {
+        // no
```
TheRealMarnes: throw new UnsupportedOperationException()?

qualidafial:
@stevenschlansker For future reference, we can do squash merges in Github without forcing contributors to squash changes by hand. It's an option under the dropdown on the `Merge pull request` button.

qualidafial:
+1 from me

arteam:
Great pull requests and review! +1 from me as well.

TheRealMarnes:
It's not very important, but I just realized there's no test for Util.size(collection) and some of the test names are kinda poor

TheRealMarnes:
Behavior of setting combinations of allowNull and allowEmpty needs to be tested and documented with a truth table

I'd suggest not merging yet, I'll work on this again tonight

TheRealMarnes:
Is everyone alright with the functional ideas? I feel like I've made up half the functionality of BindIn on my own as I went along now, I'd like to make sure every piece of the design (not just technically but also functionally) is thought-out, deliberate, and agreed upon by the jdbi team.

The fact that there's a debug log statement for silent null passage (the only one in all of jdbi from what I can tell, slf4j will handle it safely with a noop logger in the worst case, but still) and an exception for unexpected behavior?

The separate options for allownull/empty: do we want the code to be strict or flexible automatically, or is configurable like this the best way? Should the default be strict instead of flexible? Not throwing exceptions unless really necessary is not a bad thing, but silent failures that get swept under the rug are just as bad. Where is the line between non-interrupting failure and throwing an IAE?

qualidafial:
I feel like you went further with the `allowNull()` and `allowEmpty()` than I intended with my comment about characterization tests.

For background, there was a report (on the mailing lists I believe) a few weeks ago of a problem using `@BindIn` with an empty collection, which was apparently a valid use case for this specific user. I have not verified this, but the user reported that in Postgres you can't just say `WHERE foo IN ()`, it has to be `WHERE foo IN (null)` to represent an empty list.

So my intention was just to get some tests in place that capture whatever the current behavior is when the `@BindIn` parameter is empty. Sorry if this wasn't clear.

TheRealMarnes:
Well, null and empty arguments need to be handled some way or other (at least definitely null), no? And since use cases vary, I basically wanted to make it possible to let the user choose between making them illegal cases or letting them pass quietly, with a sane default. If I removed that, we'd either have undefined behavior or unwanted exceptions.

Maybe what I could do to take it down a notch is combine the null and empty cases into a single case and provide a way to attach a nullness/emptiness handler the user can override to do what he wants and that to defaults to something sane? Thats what your report story inspires me to do anyway

My main concern was really just if with the growing complexity of BindIn I hadnt made any bad implementation/design choices, if maybe it wouldnt work make a better tool if it worked differently

TheRealMarnes:
Tests added, now just thinking of ways to reduce the overengineering/cruft a bit

TheRealMarnes:
So I got an idea yesterday to hopefully simplify things a bit, and came up with EmptyHandling (currently pushed and building, I can revert it if it's disliked)

Basically you'd use BindIn as
`@BindIn(value = "x", onEmpty = THROW)`
to throw an IAE, with VOID to just output empty parentheses "()" if the argument is null or empty, and NULL to output "(null)" (looking at you, postgres guy in the mailing list)

The code got kind of longer, I think, but in terms of complexity I think it got simpler

TheRealMarnes:
I asked 2 friends who've been working as java devs for years to pretend to be end users and compare the booleans implementation to the new EmptyHandling one, and they both prefer the new one. I think we should stick to that. It seems cleaner, clearer and more extensible anyway, having an enum to express use cases.

Review/confirmation and thoughts, please? I didn't get to figuring out how to test the NULL case since I need to check the query itself for that, but I guess I'll find a way tonight. Maybe through a fake slf4j sql logger that asserts the query contents.

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
```diff
@@ -90,21 +127,120 @@ public void apply(SQLStatement q) throws SQLException
...
+
+        private static String getTypeWarning(final String type)
+        {
+            return "argument must be one of the following: Iterable, or an array/varargs (primitive or complex type); was " + type + " instead";
+        }
+    }
+
+    enum EmptyHandling
+    {
+        VOID, // output "()"
```
stevenschlansker: If these were Javadoc comments, it'd appear as an IDE tooltip.

src/main/java/org/skife/jdbi/v2/unstable/ObjectIterator.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package org.skife.jdbi.v2.unstable;
+
+import java.lang.reflect.Array;
+import java.util.Iterator;
+
+/**
+ * Implements Iterator for unidentified arrays that have been cast to Object. Note that its elements will be returned as Object, primitives included (will be autoboxed).
+ */
+class ObjectIterator implements Iterator<Object>
```
stevenschlansker: This class actually iterates over arrays, not any arbitrary `Object` (even though that's not proven at compile time) so IMO the name is a little confusing.  Maybe `ReflectionArrayIterator` or something like that?

src/main/java/org/skife/jdbi/v2/unstable/BindIn.java
TheRealMarnes: will do

src/main/java/org/skife/jdbi/v2/unstable/ObjectIterator.java
TheRealMarnes: agreed

stevenschlansker:
This is looking great, the code is very readable!  Final comment: you could add a line to `RELEASE_NOTES` to announce the improvement.  Consider the last few comments, and then we can merge it.  Thanks for your persistence with this PR!

TheRealMarnes:
Working on the final detail, but hit a snag. I'm writing two tests to verify that when setting onEmpty = NULL, the query does actually contain "null" where the placeholder was. I thought to use an SQLLog to do that, but I forgot the logger only gets a parameterized query with ? everywhere, without the parameter values. How can I get around this?

qualidafial:
Since `@BindIn` is just defining what `<ids>` gets turned into, you could use a mock `Query` and verify that `Query.define("ids", expected)` gets called.

TheRealMarnes:
Ah, right, I hadn't noticed a mock system was available, I'll take a look tonight, thanks

TheRealMarnes:
The binding methods on SQLStatement are final... it's making mocking pretty difficult, but I might get there still.

stevenschlansker:
If you need to make them not final, that's an OK change to make.

TheRealMarnes:
Alright, thanks

TheRealMarnes:
Well, I'm not quite happy with the EmptyHandling.NULL tests, but I don't think it can get any better. They verify that null is given as argument value to the query, and that's the closest I could get. After that, the most useful thing I could find was the parameterized query with ? and its bound arguments (the null value). I think after that point we can really only cross our fingers and presume the jdbc driver (or whatever) translates a null parameter to the null sql keyword internally, I don't think you can actually get the resulting sql string.

stevenschlansker:
Don't worry about 100% unit test coverage.  As long as there's some sort of functional test making sure the feature behaves as specified that's likely good enough.

TheRealMarnes:
Well yeah, but in this case we have a test that clearly shows a feature not working as intended. I'm honestly not sure what to do with that now. The tests show null is definitely bound to the query and debug view shows jdbc will execute a query with ? bound to null, yet the query doesn't seem to turn out as I expect. Unless h2 doesn't like "x in (null)" too much?

stevenschlansker:
Oh, I misunderstood the problem.  It's entirely possible the behavior of that construct varies between h2 and pg.  I can take a bit of a look later tonight possibly if you haven't figure it out by then.

TheRealMarnes:
Well, I did just boot up one of my own projects that runs an h2 db with mock data, and this happened:
vivaldi_2016-09-13_23-47-07

So it seems it's a problem with h2.

stevenschlansker:
I think that looks correct?  `x IN (null)` should select no rows, even if some values are null, because `null != null`

stevenschlansker:
http://stackoverflow.com/questions/6362112/in-clause-with-null-or-is-null

TheRealMarnes:
Then I guess the tests actually mean it's working up to specs?...

I suppose what it really needs is an integration test with postgres instead of h2.

TheRealMarnes:
Shall I just fix and break down the null handling test in light of this lesson, push it as such, and consider it finally done then? You guys can figure out a way to test the null handling some other time then, and in the meantime mark EmptyHandling.NULL as experimental/untested...

stevenschlansker:
What exactly doesn't work right?  With `EmptyHandling.NULL`, if you pass in say a `new String[0]`, the behavior I expect is that it binds as `x IN (null)` and returns no rows.  Which part of this is broken?

src/test/java/org/skife/jdbi/v2/sqlobject/BindInNullTest.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Test
+    public void testSomethingByIterableHandleNullWithEmptyList()
+    {
+        final SomethingByIterableHandleNull s = handle.attach(SomethingByIterableHandleNull.class);
+
+        final List<Something> out = s.get(new ArrayList<Object>());
+
+        Assert.assertEquals(4, out.size());
```
stevenschlansker: I expect `out` to be empty here, yes?

TheRealMarnes:
Well I guess up to there it's working indeed. The thing I'd still like to verify is that it in fact does output the null keyword in the query so that it ends up being "x in (null)", but like I said earlier I don't think we can test for that...

Yeah sorry, my thoughts are a mess right now anyway. I guess everything is in order?

src/test/java/org/skife/jdbi/v2/sqlobject/BindInNullTest.java
```diff
@@ -0,0 +1,157 @@
...
+        handle.execute("drop table something");
+        handle.close();
+    }
+
+    @Test
+    public void testSomethingByIterableHandleNullWithNull()
+    {
+        final SomethingByIterableHandleNull s = handle.attach(SomethingByIterableHandleNull.class);
+
+        final List<Something> out = s.get(null);
```
stevenschlansker: This might be considered a programming error -- should we really assume `null` is `empty list` instead of a coding error?

stevenschlansker:
Looking at the failing test, I think the test is wrong.  An empty input should match no rows no matter what data is in the table.

TheRealMarnes:
> Looking at the failing test, I think the test is wrong. An empty input should match no rows no matter what data is in the table.
Yes, I'm changing that test atm. I mean the functionality itself is apparently working as intended.

qualidafial:
> The thing I'd still like to verify is that it in fact does output the null keyword in the query so that it ends up being "x in (null)", but like I said earlier I don't think we can test for that...

You could `Mockito.spy` on the statement rewriter, and check that the rewritten query is as expected.

I feel this is getting out of hand though. I'm good to just merge this.

stevenschlansker:
Yes, I think the existing test (once fixed) is sufficient.  No need for `spy`ing.

TheRealMarnes:
agreed

stevenschlansker:
Merged, thanks!

TheRealMarnes:
Just for my information, around when are you planning to release 2.77? I kinda need the bindin fix in a project I'm working on (hence how I found the bug and why I did any of this to begin with)

stevenschlansker:
@TheRealMarnes sorry was on vacation and lost track of it, just did the release :)

TheRealMarnes:
Awesome, thanks

PR #484, opened 2016-09-14
Add case sensitivity option to DefaultMapper

### LINKED ISSUE #483: Column name case sensitivity
Classes like the `DefaultMapper` lower-case the column names from the `ResultSet`. This does't work well for databases like Postgres, where the column names are case sensitive (so if you have a table with columns containing uppercase characters, you need to remember and use the correct column case).

A workaround is supplying a custom map function which doesn't lowercase the columns such as this:
```
.map((index, r, ctx) -> {
    ResultSetMetaData m = r.getMetaData();
    Map<String, Object> row = new HashMap<>();
    for (int i = 1; i <= m.getColumnCount(); i++) {
        row.put(m.getColumnLabel(i), r.getObject(i));
    }
    return row;
})
```

However, this seems largely duplicative of the `DefaultMapper` implementation. Would you take a PR that adds a util `CaseSensitiveMapper` which has this functionality?

hpx7:
Fixes #483

### DIFF
--- a/src/main/java/org/skife/jdbi/v2/DefaultMapper.java
+++ b/src/main/java/org/skife/jdbi/v2/DefaultMapper.java
@@ -24,10 +24,20 @@ import java.util.Map;
 
 public class DefaultMapper implements ResultSetMapper<Map<String, Object>>
 {
+    private boolean foldCase;
+
+    public DefaultMapper(){
+        this(true);
+    }
+
+    public DefaultMapper(boolean foldCase) {
+        this.foldCase = foldCase;
+    }
+
     @Override
     public Map<String, Object> map(int index, ResultSet r, StatementContext ctx)
     {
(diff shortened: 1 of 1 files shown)


src/main/java/org/skife/jdbi/v2/DefaultMapper.java
```diff
@@ -24,10 +24,20 @@
 
 public class DefaultMapper implements ResultSetMapper<Map<String, Object>>
 {
+    private boolean foldCase;
```
stevenschlansker: this should be `final`

stevenschlansker:
Fixing missing final and merging.  Thanks for the contribution!

hpx7:
Thanks! When do you think this could be out in a release?

stevenschlansker:
Released as `2.76`

PR #485, opened 2016-09-14
SnakeCaseColumnNameStrategy: simplify logic, handle digits in names

### LINKED ISSUE #481: support for camel-case conversion have a bug
@stevenschlansker  hi, ,i am test the camel case,but I found a bug, spring JDBC beanPropertyRowMapper is have a problem, for this like 'user_name_2' this variable has a problem. I found, from inside the sql2o sql2o replacement inside under camel method, named hope you can accept. thank you
```
 public class UnderscoreToCamelCase {
    public static String convert(String  underscore){
        if(underscore==null || underscore.isEmpty()) return underscore;
        return convert00(underscore);
    }

    private static String convert00(String  underscore){
        char[] chars = underscore.toCharArray();
        int write=-1,len=chars.length;
        boolean upper=false;
        for (int read = 0; read < len; ++read) {
            char c = chars[read];
            if('_'==c){
                upper = true;
                continue;
            }
            if(upper){
                upper = false;
                chars[++write]=Character.toUpperCase(c);
            } else {
                chars[++write]=Character.toLowerCase(c);
            }
        }
        return new String(chars, 0, ++write);
    }
}
```

stevenschlansker:
fixes #481, @kkrgwbj could you please test this fixes your problem?

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/util/bean/SnakeCaseColumnNameStrategy.java
+++ b/core/src/main/java/org/jdbi/v3/core/util/bean/SnakeCaseColumnNameStrategy.java
@@ -30,23 +30,7 @@ public class SnakeCaseColumnNameStrategy implements ColumnNameMappingStrategy {
 
     @Override
     public boolean nameMatches(String propertyName, String sqlColumnName) {
-        // Convert the property name from camel-case to underscores syntax. Freely adapted from Spring
-        // BeanPropertyRowMapper.
-        StringBuilder propertyNameWithUnderscores = new StringBuilder();
-        propertyNameWithUnderscores.append(propertyName.substring(0, 1));
-        for (int i = 1; i < propertyName.length(); i++) {
-            // Do case comparison using strings rather than chars (avoid to deal with non-BMP char handling).
-            String s = propertyName.substring(i, i + 1);
-            String slc = s.toLowerCase(locale);
-            if (!s.equals(slc)) {
-                // Different cases: tokenize.
-                propertyNameWithUnderscores.append("_").append(slc);
-            }
-            else {
-                propertyNameWithUnderscores.append(s);
-            }
-        }
-        return propertyNameWithUnderscores.toString().equals(sqlColumnName.toLowerCase(locale));
+        return sqlColumnName.replace("_", "").toLowerCase(locale).equals(propertyName.toLowerCase(locale));
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jdbi/v3/core/util/bean/SnakeCaseColumnNameStrategy.java
```diff
@@ -30,23 +30,7 @@ public SnakeCaseColumnNameStrategy(Locale locale) {
...
-            if (!s.equals(slc)) {
-                // Different cases: tokenize.
-                propertyNameWithUnderscores.append("_").append(slc);
-            }
-            else {
-                propertyNameWithUnderscores.append(s);
-            }
-        }
-        return propertyNameWithUnderscores.toString().equals(sqlColumnName.toLowerCase(locale));
+        return sqlColumnName.replace("_", "").toLowerCase(locale).equals(propertyName.toLowerCase(locale));
```
stevenschlansker: This is significantly more permissive now in the presence of e.g. double underscore `__` but that doesn't seem likely to be harmful.

qualidafial: It would be simpler (and probably faster) if you used:

    return sqlColumnName.replace("_", "").equalsIgnoreCase(propertyName);

stevenschlansker: Yeah, I'm waffling back and forth between the approaches.  The downside of `equalsIgnoreCase` is it relies on the default locale, so it is hard to specify behavior, it depends on the environment you run the code in.  But maybe that's OK?

qualidafial: Hmm. Looks like my suggestion ignores locales. Scratch that.

stevenschlansker: Here's what the code looks like if we do it that way.  I'm having trouble figuring out which way is better.  I wish there was an `equalsIgnoreCase(String, Locale)` :sad:

qualidafial: In the absence of a demonstrated problem the other way, let's go with the simple solution.

We can always enhance it later to use locales.

qualidafial:
👍 from me assuming the build passes.

stevenschlansker:
It looks like a more complete solution might use `java.text.Collator` which I just learned about while researching this issue... probably not worth the complexity for right now but let's keep it in mind.

qualidafial:
Build passed. :shipit: !

Issue #488, opened 2016-09-15
SQLObject should throw exception when an abstract SQL method has no SQL method annotation


qualidafial:
Related to #477. This is the inverse case.

qualidafial:
Fix in #491

PR #491, opened 2016-09-18
Do sanity checks on SQL object attach: non-default method with no ann…

### LINKED ISSUE #477: SQLObject should throw exception when a non-abstract SQL method has a SQL method annotation.
Since a SQL method annotation implies a specific implementation, it is an error to annotate a concrete (default) method with a SQL method annotation e.g. `@SqlUpdate`

qualidafial:
…otation, default method with sql method annotations should both throw exceptions

Fixes #477, #488

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
@@ -14,7 +14,6 @@
 package org.jdbi.v3.sqlobject;
 
 import static java.util.stream.Collectors.toList;
-import static java.util.stream.Collectors.toSet;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
@@ -132,10 +131,10 @@ public enum SqlObjectFactory implements ExtensionFactory<SqlObjectConfig> {
     }
 
     private Handler buildBaseHandler(Class<?> sqlObjectType, Method method) {
-        Set<Class<?>> sqlMethodAnnotations = Stream.of(method.getAnnotations())
+        List<Class<?>> sqlMethodAnnotations = Stream.of(method.getAnnotations())
                 .map(Annotation::annotationType)
                 .filter(type -> type.isAnnotationPresent(SqlMethodAnnotation.class))
-                .collect(toSet());
+                .collect(toList());
 
         if (sqlMethodAnnotations.size() > 1) {
             throw new IllegalStateException(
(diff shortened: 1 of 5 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -145,13 +144,25 @@ private Handler buildBaseHandler(Class<?> sqlObjectType, Method method) {
                             method.getName(),
                             sqlMethodAnnotations));
         }
+        if (method.isDefault()) {
+            if (!sqlMethodAnnotations.isEmpty()) {
+                throw new IllegalStateException(String.format(
+                        "Method %s.%s has mutually exclusive default implementation and a SQL method annotation (must be one or the other)",
```
qualidafial: Hmm, might be good to mention the class of the offending annotation.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -133,10 +131,10 @@ private Handler buildMethodHandler(Class<?> sqlObjectType, Method method) {
     }
 
     private Handler buildBaseHandler(Class<?> sqlObjectType, Method method) {
-        Set<Class<?>> sqlMethodAnnotations = Stream.of(method.getAnnotations())
+        List<Class<?>> sqlMethodAnnotations = Stream.of(method.getAnnotations())
```
arteam: Could you, please, elaborate why we need here a list rather than a set?

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestSqlObject.java
```diff
@@ -65,15 +66,16 @@ public void testPassThroughMethod() throws Exception
         assertThat(c, equalTo(new Something(3, "Cora")));
     }
 
-    @Test
+    @Test(expected = AbstractMethodError.class)
```
arteam: I think it would be great to verify the message of the exception as well.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
qualidafial: Aesthetic only. `list.get(0)` feels slightly less gross than `set.iterator().next()` ¯\\\_(ツ)\_/¯

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -145,13 +143,27 @@ private Handler buildBaseHandler(Class<?> sqlObjectType, Method method) {
...
+            return new DefaultMethodHandler();
+        }
 
         return sqlMethodAnnotations.stream()
                 .map(type -> type.getAnnotation(SqlMethodAnnotation.class))
                 .map(a -> buildFactory(a.value()))
                 .map(factory -> factory.buildHandler(sqlObjectType, method))
                 .findFirst()
-                .orElseGet(PassThroughHandler::new);
+                .orElseThrow(() -> new AbstractMethodError(String.format(
```
stevenschlansker: It feels weird for one error path to throw `IllegalStateException` and another to throw `AbstractMethodError`.  I get why, and I can't think of anything better off the top of my head, but it caught my eye.

qualidafial: We could throw a different exception here. The two cases are opposite sides of the same coin, so I could see `IllegalStateException` being fine here.

PR #498, opened 2016-09-21
Reuse a Postgres instance during tests

arteam:
Starting up and shutting down a Postgres instance are rather heavy operations. As a consequence of this, unit tests for the Postgres module takes a considerable amount of time (1 min). We could speed this process up, by reusing `PostgresDbRule` on the unit class level, and recreating the relevant table before every test. This should be a more fast process (~15 sec), because dropping and creating tables are relatively cheap operations.


stevenschlansker:
The embedded postgres driver is supposed to prepare instances intelligently so you don't have to spin up a new DB for every individual test case, without doing this.

There's no problem with making this sort of change but it might be worth taking a look at the code to understand why that is apparently not working.

(I'm on vacation for the next week)

Issue #478, opened 2016-09-11
@BindBean should be able to ignore unknown properties


qualidafial:
Migrating a project from Hibernate to JDBI3. When using `@BindBean` on an object with a collection property, `@BindBean` expects that collection to map to something. We need some kind of equivalent to JPA's `@Transient` annotation, or a property on the `@BindBean` annotation itself to let JDBI know it can ignore certain properties.

qualidafial:
Workaround for now: extend `GetHandle` and drop down to the fluent interface in SQL objects. 👎

miere:
Hi @qualidafial. How you doing?

Here at @sizebay we are migrating our main API - which makes heavy use of JDBI2 - to JDBI3 once only the version 3 have _really_ solved the annoying ```@Transaction``` behavior that exists on v2. During the migration process we have faced a lot of problems with "bean"s that have Collection attributes. I do believe it is a feature regression and thus I would consider this issue a bug.

Bellow is the code I did as workaround for this situation (it made sense on my case, once I don't use Collections inside _beans_ as argument for my queries). I hope it help someone else too...
```java
// this class is exposed as SPI for JdbiPlugin
public class ListJDBIArgumentFactory implements ArgumentFactory, JdbiPlugin {

	@Override
	public Optional<Argument> build(Type type, Object o, StatementContext statementContext) {
		return Optional.of( new NullArgument(0) );
	}

	@Override
	public void customizeDbi(Jdbi dbi) {
		dbi.registerArgumentFactory(this);
	}
}
```
As you've suggested, something like ```@Transient``` would be nice at first look. You can apply the same behavior on ```transient```(the modifier) fields . I would also suggest to let 'Java Collection's attribute NULL by default and let developers to change this behavior if needed.

Regards

qualidafial:
We're actually working in another issue on supporting array and collection arguments as SQL arrays, so your approach here would sort of undermine the SQL array support.

I'm considering a few options for resolving this `@BindBean` issue:

1. Just ignore bean properties for which we don't have a mapper. This is similar to what `BeanMapper` does--it only throws an exception if it can't find _any_ columns to map.
2. Introduce an some kind of `@Ignore` annotation and skip any bean property where the getter or setter has that annotation.
3. Add options to the `@BindBean` annotation e.g. `@BindBean(exclude={"things", "widgets"})`.

[...]

Issue #494, opened 2016-09-18
Unify @BindBean and SqlStatement.bindFromProperties


qualidafial:
The `@BindBean` binding annotation and `SqlStatement.bindFromProperties` (which uses `BeanPropertyArguments`) are doing the same thing, although BeanPropertyArguments is arguably the better approach, given its lazy approach.

We should enhance `BeanPropertyArguments` to support an object name prefix (as in `@BindBean.value()`) and then just use `SqlStatement.bindFromProperties` directly in the binding annotation handler.

qualidafial:
While we're at it.. can we rename `bindFromProperties` to just `bindBean`?

qualidafial:
Fixed in #501

PR #501, opened 2016-09-25
Unify @BindBean and SqlStatement.bindFromProperties

qualidafial:
* Enhance BeanPropertyArguments to support bean name prefix like `@BindBean`
* Rename `SqlStatement.bindFromProperties()` to `.bindBean()`
* Rename `SqlStatement.bindFromMap()` to `.bindMap()`
* Modify `@BindBean` to use `SqlStatement.bindBean` instead of its own parallel implementation.
  This also makes `@BindBean` bindings lazy instead of eager, which fixes issue #478
* Modify @BindMap:
  * Rename `BindMap.value()` to `.keys()`
  * Rename `BindMap.prefix()` to `.value()` (to agree with `@BindBean`)
  * Rename `BindMap.implicitKeyStringConversion()` to `.convertKeys()`
  * Use `SqlStatement.bindMap` instead of its own parallel implementation.


qualidafial:
Fixes #494, #478

qualidafial:
@arteam Would you mind reviewing this PR?

core/src/main/java/org/jdbi/v3/core/BeanPropertyArguments.java
```diff
@@ -26,53 +26,54 @@
...
-                                                                             name, bean), e, ctx);
+                    try
+                    {
+                        return ctx.findArgumentFor(
+                                descriptor.getReadMethod().getGenericReturnType(),
+                                descriptor.getReadMethod().invoke(bean));
+                    }
+                    catch (IllegalAccessException e)
+                    {
+                        throw new UnableToCreateStatementException(String.format("Access exception invoking getter for " +
```
arteam: Could we write a test for this exception?

core/src/main/java/org/jdbi/v3/core/SqlStatement.java
```diff
@@ -308,31 +308,41 @@ public SelfType bind(String name, Argument argument)
...
      */
-    public SelfType bindFromProperties(Object o)
+    public SelfType bindBean(Object bean)
     {
-        return bindNamedArgumentFinder(new BeanPropertyArguments(o, getContext(), getArgumentRegistry()));
+        return bindNamedArgumentFinder(new BeanPropertyArguments(null, bean, getContext()));
+    }
+
+    /**
+     * Binds named parameters from JavaBean properties on o, with the given prefix.
```
qualidafial: Done

core/src/main/java/org/jdbi/v3/core/BeanPropertyArguments.java
qualidafial: Done

PR #502, opened 2016-09-25
DefaultMethodHandler performance: cache MethodHandles.Lookup instance…

qualidafial:
…s per declaring class, and method handles per default method handler.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultMethodHandler.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultMethodHandler.java
@@ -13,55 +13,65 @@
  */
 package org.jdbi.v3.sqlobject;
 
+import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.jdbi.v3.core.HandleSupplier;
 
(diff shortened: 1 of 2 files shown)


qualidafial:
@arteam Would you please review this?

arteam:
LGTM

sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultMethodHandler.java
```diff
@@ -13,55 +13,65 @@
...
 
 import org.jdbi.v3.core.HandleSupplier;
 
-class DefaultMethodHandler implements Handler
-{
-    DefaultMethodHandler()
-    {
-    }
+class DefaultMethodHandler implements Handler {
+    private static final Map<Class<?>, MethodHandles.Lookup> privateLookups = new ConcurrentHashMap<>();
```
stevenschlansker: Since this is static, this likely breaks class unloading.

qualidafial: You're right. I'll submit a follow-up that uses weak keys.

qualidafial: Looks like `SqlObjectFactory` is also caching classes in `ConcurrentHashMap`.

qualidafial: Added card https://github.com/jdbi/jdbi/projects/1#card-304665 to project board

PR #503, opened 2016-09-26
Make `@Define` default to the parameter name, if available.

qualidafial:
@arteam Would you mind reviewing this?

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/Define.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/Define.java
@@ -19,6 +19,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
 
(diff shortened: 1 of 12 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/Define.java
```diff
@@ -37,15 +38,27 @@
...
-            Define d = (Define) annotation;
-            final String key = d.value();
+            Define define = (Define) annotation;
+
+            String name = define.value();
+            if (name.isEmpty()) {
+                if (param.isNamePresent()) {
+                    name = param.getName();
+                } else {
+                    throw new UnsupportedOperationException("A @Define parameter was not given a name, "
```
arteam: Could we test this exception? I think it would be useful.

arteam:
LGTM, exception one comment about testing the new exception message.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/Define.java
qualidafial: Not without a lot of trouble--this is configured in the POM. We'd have to spin up a separate artifact with a different compiler configuration just to test this. I'm not sure it's worth it.

PR #505, opened 2016-09-27
Throw exception if SQL object default methods/parameters are annotate…

### LINKED ISSUE #504: SqlObject should throw exception if default methods have customizing/binding annotations
Default methods do not do anything with these annotations, making them no-ops. We should add some runtime validation, and throw exceptions when users try to use these annotations, so they do not assume they are doing anything.

* `SqlMethodDecoratingAnnotation`s like `@Transaction` are fine
* `SqlStatementCustomizingAnnotation`s like `@Define` should be forbidden.
* `BindingAnnotation`s like `@Bind` should be forbidden.

qualidafial:
…d with binding or statement customizing annotations.

Fixes #504


qualidafial:
@stevenschlansker or @arteam Would one of you mind reviewing this?

stevenschlansker:
Nice!

PR #506, opened 2016-09-28
[WIP] Postgres' hstore support

arteam:
This PR is to track progress and getting a feedback for implementing a support for Postgres' `hstore` type.

### DIFF
--- a/postgres/src/main/java/org/jdbi/v3/postgres/HStoreArgumentFactory.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/HStoreArgumentFactory.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.postgres;
+
+import org.jdbi.v3.core.StatementContext;
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.argument.ArgumentFactory;
+import org.jdbi.v3.core.util.GenericTypes;
+
+import java.lang.reflect.Type;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * An argument factory which binds Java's {@link Map} to Postgres' hstore type.
+ */
+public class HStoreArgumentFactory implements ArgumentFactory {
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Optional<Argument> build(Type type, Object value, StatementContext ctx) {
+        if (Map.class.isAssignableFrom(GenericTypes.getErasedType(type))) {
+            return Optional.of((i, p, cx) -> p.setObject(i, value));
+        }
+        return Optional.empty();
+    }
+}
--- a/postgres/src/main/java/org/jdbi/v3/postgres/HStoreColumnMapper.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/HStoreColumnMapper.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.postgres;
+
+import org.jdbi.v3.core.StatementContext;
+import org.jdbi.v3.core.mapper.ColumnMapper;
+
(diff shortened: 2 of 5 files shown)


postgres/src/main/java/org/jdbi/v3/postgres/HStoreArgumentFactory.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class HStoreArgumentFactory implements ArgumentFactory {
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Optional<Argument> build(Type type, Object value, StatementContext ctx) {
+        if (value instanceof Map) {
+            // Unfortunately, the PostgreSQL driver doesn't support mapping from Java maps to the hstore type.
+            // Therefore, we convert the map to a string representation and pass it to a PostgreSQL server.
+            return Optional.of((i, p, cx) -> p.setObject(i, toString((Map<String, ?>) value), Types.OTHER));
```
qualidafial: The Postgres driver supports `hstore` out of the box. You can just call `setObject` with the map:

    statement.setObject(col, map);

Also the Postgres driver thoroughly escapes keys and values--see `HStoreConverter` in the Postgres JDBC jar.

Alternatively you could use `PGobject`:

    PGobject obj = new PGobject();
    obj.setType("hstore");
    obj.setValue(HStoreConverter.toString(map));
    statement.setObject(col, obj);

postgres/src/main/java/org/jdbi/v3/postgres/HStoreArgumentFactory.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+        return Optional.empty();
+    }
+
+    private static String toString(Map<String, ?> map) {
+        return map.entrySet().stream()
+                .map(e -> "\"" + e.getKey() + "\" => \"" + e.getValue() + "\"")
```
stevenschlansker: Probably need some kind of escaping here, or at least an assertion that the key / value don't contain `"`

postgres/src/main/java/org/jdbi/v3/postgres/HStoreColumnMapper.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * A column mapper which maps Postgres' hstore type to Java's {@link Map}.
+ */
+public class HStoreColumnMapper implements ColumnMapper<Map<String, String>> {
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Map<String, String> map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
+        return (Map<String, String>) r.getObject(columnNumber);
```
qualidafial: It's unfortunate we can't use `getObject(columnNumber, Map.class)` here for some added type safety.

qualidafial:
We use `hstore` at my work, so I can't wait to use this.

postgres/src/main/java/org/jdbi/v3/postgres/HStoreArgumentFactory.java
arteam: Wow, I didn't see that! Neat that it's already implemented in a much better way!

postgres/src/main/java/org/jdbi/v3/postgres/HStoreArgumentFactory.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * An argument factory which binds Java's {@link Map} to Postgres' hstore type.
+ */
+public class HStoreArgumentFactory implements ArgumentFactory {
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Optional<Argument> build(Type type, Object value, StatementContext ctx) {
+        if (value instanceof Map) {
```
arteam: Unfortunately, we don't have a generic information about the type. The type which is passed to argument factories is derived from the value by a simple `getClass` call, which returns only the erased type.

postgres/src/main/java/org/jdbi/v3/postgres/HStoreArgumentFactory.java
arteam: Thanks! I solved the issue by delegating this task to the Postgres JDBC driver, which already implements converting with escaping.

postgres/src/main/java/org/jdbi/v3/postgres/HStoreColumnMapper.java
arteam: Yeah, I think the Postgres driver could actually support `hstore` for this method as well.

postgres/src/main/java/org/jdbi/v3/postgres/HStoreArgumentFactory.java
qualidafial: It depends.

SQL objects should pass the full type information from the method parameter (or bean property, in the case of `@BindBean`).

However if you're using the fluent interface, then `type` could be `Map.class`, `Object.class` (for nulls), or `Map<String,String>` depending on which method you used to bind the parameter.

arteam: Hmm, I didn't consider the issue from this angle. Let me check and get back to you.

arteam: I gave a little bit thought to this question, and I think we should not limit API to the `Map<String, String>` type. The PostgreSQL JDBC driver supports any maps during inserts, and we are just proxy calls to it. By removing the restriction on the `<String, String>` generic parameter we support all types of insertions in JDBI and provide a good developer experience.

qualidafial: Fair enough.

Can we still change this `instanceof` condition to `Map.class.isAssignableFrom(GenericTypes.getErasedType(type)`? That will allow this argument factory to also capture nulls passed to Map arguments in SQL objects.

arteam: Sure, I've updated the PR.

PR #508, opened 2016-09-30
Use synchronized WeakHashMaps for singleton caches, so classes can be garbage collected.

qualidafial:


qualidafial:
@stevenschlansker Could you review please?

PR #510, opened 2016-10-04
Map null results to absent optional values in SqlObject API

arteam:
Currently we use `Optional` in the SqlObject API to indicate the absence of result in a response. But many developers expect that null values will map to optional absent values as well.
Unfortunately, we don't have two different optional abstraction to represent this use cases, so we have to mix these use cases. On the pragmatic point of view, this change provides better developer experience, because developers can effectively work with nullable values inside the `SqlObjectAPI`. Also, it consistent with the logic when results are presented as plain null values.


qualidafial:
LGTM

PR #513, opened 2016-10-11
Rename `createStatement` to `createUpdate`

stevenschlansker:
`Statement` is ambiguous -- it actually returns
an `Update` object (there is the abstract class
`SqlStatement`, but that covers other types too).

Other option would be to rename the `Update` class
to `Statement`, but that's too close to
`SqlStatement`.


pom.xml
```diff
@@ -216,7 +216,7 @@
             <dependency>
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
-                <version>19.0</version>
+                <version>20.0-rc1</version>
```
qualidafial: 😐

PR #509, opened 2016-10-04
Move tests to AssertJ

arteam:
AssertJ is modern testing library which allows to write test assertions
which are easy to read and understand. It supports Java 8 features and
allows to perform assertions in the chain style.

This changes adopts AssertJ as JDBI's testing library and removes
all usages of Hamcrest assertions, which are fine, but are rather
limited.

PR #515, opened 2016-10-11
Temporarily disable guava dependency check

stevenschlansker:
Fallout from #509 -- assertJ depends on guava 20.0-rc1 (annoying!)
We can just hope they release before we do? *g*

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -341,6 +341,13 @@
                 <groupId>org.assertj</groupId>
                 <artifactId>assertj-guava</artifactId>
                 <version>3.0.0</version>
+                <!-- XXX: https://github.com/joel-costigliola/assertj-guava/issues/33 -->
+                <exclusions>
+                  <exclusion>
+                    <groupId>com.google.guava</groupId>
+                    <artifactId>guava</artifactId>
+                  </exclusion>
+                </exclusions>
             </dependency>
 
             <dependency>


qualidafial:
`assertj-guava` is only a test dependency. I'd say this is one of those scenarios where adding an `<exclusion>` is justified.

stevenschlansker:
Ugh.  The root cause I think is:
```xml
 <dependency>
<version>[${guava.version},)</version>
</dependency>
```

from `assertj-guava`.

There's an open issue https://github.com/joel-costigliola/assertj-guava/issues/33

stevenschlansker:
Aaaaaand you can't declare an exception -- you have to name a specific version:
```
      <exception>
        <groupId>log4j</groupId>
        <artifactId>log4j</artifactId>
        <expectedVersion>1.2.13</expectedVersion>
        <resolvedVersion>1.2.16</resolvedVersion>
      </exception>
```
so the moment Guava's upstream version changes, our exception becomes totally ineffective. (╯°□°）╯︵ ┻━┻)

stevenschlansker:
I guess we'll just disable it entirely for now.  Sound good?

qualidafial:
Build still broken :(

Should we get rid of assertj-guava until this is resolved upstream? We can modify the few tests that use Guava assertions.

qualidafial:
See my workaround in #516. Adding an `<exclusion>` of `com.google.guava:guava` to the `assertj-guava` managed dependency fixed the problem for me.

PR #522, opened 2016-10-14
Remove final modifier from StatementContext

arteam:
`StatementContext` is exposed publicly from `TimingCollector`, so it makes sense to allow users to mock it to test their `TimingCollector` implementations.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/StatementContext.java
+++ b/core/src/main/java/org/jdbi/v3/core/StatementContext.java
@@ -33,7 +33,7 @@ import org.jdbi.v3.core.mapper.RowMapper;
  * to all statement customizers. This makes it possible to parameterize the processing of
  * the tweakable parts of the statement processing cycle.
  */
-public final class StatementContext
+public class StatementContext
 {
     private final JdbiConfig config;
     private final ExtensionMethod extensionMethod;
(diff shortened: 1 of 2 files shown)


hgschmie:
So for the purpose of one very specific and questionable way of testing implementations we unconditionally open extending this class?  I am not sure that this is the right call.

Why can a StatementContext not be constructed and a real instance used for a test case?

arteam:
> Why can a StatementContext not be constructed and a real instance used for a test case?

I would do that, but unfortunately all setter methods in `StatementContext` have the private-protected access level and can't be invoked outside of the `org.jdbi.v3.core` package.

arteam:
I think allowing to subclass `StatementContext` doesn't do any harm, because its internal fields have the private access level and can't be accessed by subclasses directly.

stevenschlansker:
I guess it comes down, do we have to expose `StatementContext` as public API or not?  Right now it's used in e.g. `Argument#apply(int, PreparedStatement, StatementContext)` and `ColumnMapper#map(ResultSet, int, StatementContext)`.  If I am to write a custom `Argument` implementation, it seems that I need to be able to at least implement or mock the `StatementContext` object for test purposes.

Maybe `StatementContext` needs to be an interface?  Or we make the setters public, or we try to hide it better.

arteam:
Originally it was an interface with `ConcreteStatementContext` as an implementation.

arteam:
#534 as a follow-up.

qualidafial:
I still favor making the class non-final, over going back to interface and impl types.

Let's add a public static factory method, make the no-arg constructor protected, and add a big bold disclaimer that says the class is non-final for testing purposes, and that extending it is not supported.

arteam:
I think it doesn't make much sense to create the static factory method because `JdbiConfig` is package-private, so the class can't be created outside of the `org.jdbi.v3.core` package. But we can make the default constructor package-private. The only place where it's created outside of its package is `TestColonStatementRewriter`, but we can use a mock there.

stevenschlansker:
Thanks for the discussion everyone!

PR #523, opened 2016-10-14
[WIP] Support for enum Sets as bit strings

arteam:
This is a WIP issue for tracking support for mapping Java's enum sets to Postgres' bit strings.

Motivation:
This feature be very useful in scenarios when the parent entity has a lot of one-to-many relationships, and as a consequence joins become painful. If the universe of possible values for the relationship is finite and low,  it could make sense to inline the relationship as a bit string (bitmap).

### DIFF
--- a/postgres/src/main/java/org/jdbi/v3/postgres/EnumSetArgument.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/EnumSetArgument.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.postgres;
+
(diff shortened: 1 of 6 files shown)


qualidafial:
There's a bit of inertia against this approach already in the code base.

* The built in mappers and arguments in core map enums by name, not ordinal. We do have an enum-by-ordinal mapper, but it is not installed by default.
* TypedEnumArgumentFactory (installed by default in postgres bundle) also maps enum arguments by name.

I think supporting `EnumSet` is a great idea, but I'm not comfortable coupling that support to a `bit(n)` column type.

A user might (with equal validity) want to represent an `EnumSet` in an `int[]` or a `varchar[]` or even `my_enum[]` column.

All of this is to say, I think these new mappers / argument factories could be valuable to folks, but I don't think they should be installed as part of the Postgres bundle.

It might even be worthwhile to emulate what we did in core for supporting enums by name vs by ordinal. See e.g. `EnumMapper`, `EnumByOrdinalMapperFactory`, `EnumByNameMapperFactory`.

arteam:
Okay, let's put it on hold.

stevenschlansker:
Hi, we are deleting the `jdbi3` branch so this PR will need to be re-opened.  Sorry!

qualidafial:
Follow up, after #1257 reminded me about this issue.

PR #1160 (type qualifier annotations a-la JSR 330) will make it possible to register mappers and argument factories for a _qualified_ type. This means you could annotated a type as e.g. `@VarBit EnumSet<MyEnum>` which would make mapping a `bit(n)` field to an enum unambiguous. I'm still doubtful that this would be widely used enough to put in the core library, but it would make implementing it and keeping those mappers separate from the default enum mapping pretty trivial.

Food for thought.

Issue #526, opened 2016-10-17
Streams don't quite make up for v2's Query.fold


qualidafial:
In the process of converting my v2 project to v3, I've repeatedly found myself wishing to have `Query.fold` back--or something similar to it.

When we removed it, the idea was that Java 8 streams would fill the niche. They're very handy, but not all use cases are served by it. In particular:

* For model-based folds (`JoinRowMapper` is great for this), `Stream.reduce(U, BiFunction<U,T,U>, BinaryOperator<U>)` does the job, but the `BinaryOperator` parameter is pointless, since this is not a parallel stream. This is one instance where I think the JDK over-engineered streams for parallelism, which hurts usability for sequential streams.
* We still need a fold with raw access to the raw `ResultSet`. Not all use cases fit in the map-reduce approach.

I propose we add the following API:

```java
@FunctionalInterface
interface ResultSetAccumulator<U> {
    U reduce(U identity, ResultSet rs, StatementContext ctx);
}

class Query {
    ...
    <U> U reduce(U identity, ResultSetAccumulator<U> accumulator);
    <U> U reduce(U identity, BiFunction<U, T, U> accumulator);
}
```

stevenschlansker:
I'll give this one a go, unless you've already started.

qualidafial:
All yours

qualidafial:
I guess `ResultBearing` is a better home for the new methods.

umcodemonkey:
Random question from the peanut gallery: does `Stream.map(Function<T, U>).reduce(U, BinaryOperator<U>)` not achieve the same goal? Or are there use cases where this would produce incorrect results? Not considering the awkward syntax here, just correctness.

umcodemonkey:
Never mind; I missed the comment about needing access to the raw `ResultSet`. Probably could be worked around, but inconvenient to do so.

qualidafial:
@umcodemonkey it depends on the use case. If your row type is the same as a reduced type then `Stream.reduce(U, BinaryOperator<U>)` works fine.

However in my case my result type is generally some kind of collection. e.g. say your row type is `JoinRow` containing a `Foo` and `Bar`, and you want a `ListMultimap<Foo,Bar>` back from the method. That's when the `BinaryOperator` parameter (in this case, a `BinaryOperator<ListMultimap<Foo,Bar>>`) is superfluous. It's only for parallelism, and our stream is strictly sequential.

umcodemonkey:
[...]

PR #530, opened 2016-10-18
ResultBearing: add reduce method, and docs

stevenschlansker:
Here's a partial pass at #526 

It's nice shorthand for the `.stream().reduce(...)` version which takes the unnecessary parallel combiner.


qualidafial:
LGTM

PR #531, opened 2016-10-18
SqlObject: map to boolean, improve code structure

### LINKED ISSUE #525: @SqlUpdate should support boolean return type
I believe this was supported in v2? Not sure.

At any rate, it's often handy to have a boolean return type from a `@SqlUpdate` method, which returns true if any rows were modified.

stevenschlansker:
Fixes #525

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/UpdateHandler.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/UpdateHandler.java
@@ -21,8 +21,8 @@ import org.jdbi.v3.core.HandleSupplier;
 import org.jdbi.v3.core.Update;
 import org.jdbi.v3.core.exception.UnableToCreateStatementException;
 import org.jdbi.v3.core.mapper.RowMapper;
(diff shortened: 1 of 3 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/UpdateHandler.java
```diff
@@ -37,11 +37,16 @@
         boolean isGetGeneratedKeys = method.isAnnotationPresent(GetGeneratedKeys.class);
 
         Type returnType = GenericTypes.resolveType(method.getGenericReturnType(), sqlObjectType);
-        if (!isGetGeneratedKeys && returnTypeIsInvalid(method.getReturnType()) ) {
-            throw new UnableToCreateSqlObjectException(invalidReturnTypeMessage(method, returnType));
+        if (!isGetGeneratedKeys) {
```
qualidafial: If you invert this condition (and swap the if/else blocks) then inner if/else chain below can make one if/else chain instead of a nested one.

qualidafial:
Looks like a few tests failed (`TestCreateSqlObjectException`) over the changed wording of the exception

PR #534, opened 2016-10-19
Make `StatementContext` an interface

arteam:
And create a concrete implementation of it, which is package-private accessible for internal JDBI components. This allows to restrict the amount of components which can change statement context's state, but at the same time to keep a public interface which allows to read the state. This also make statement contexts more testable because the use code which interacts with statement contexts can mock the interface.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/BaseStatement.java
+++ b/core/src/main/java/org/jdbi/v3/core/BaseStatement.java
@@ -26,9 +26,9 @@ abstract class BaseStatement
 {
     final JdbiConfig config;
     private final Collection<StatementCustomizer> customizers = new ArrayList<>();
-    private final StatementContext context;
+    private final ConcreteStatementContext context;
 
-    BaseStatement(JdbiConfig config, StatementContext context)
+    BaseStatement(JdbiConfig config, ConcreteStatementContext context)
     {
         this.config = config;
         this.context = context;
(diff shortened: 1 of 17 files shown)


christophercurrie:
If it's now an interface, why do so many classes need references to it's concrete implementation?

arteam:
> If it's now an interface, why do so many classes need references to it's concrete implementation?

The reason for that is that the statement classes (`Query`, `PreparedBatch`, `Update`) acquire the context using the `getContext` method of `BaseStatement`. They expect an implementation which 
allows to change context's internals.

qualidafial:
My 2 cents: the reason we abandoned the separate interface/implementations for DBI (now Jdbi), Handle, and StatementContext is because our internal code was a sea of casts, which we came to view as a design smell.

Once we switched to classes, we could play around with method visibility and hide things appropriately. 

If testability is a concern, I think making the classes non-final is the least worst option available to us. We can add a disclaimer to the docs e.g. `This class is non-final for internal unit testing purposes. Subclassing for production use is not supported, and may break at any time without notice.`

Then add a static factory method to the class, make the constructors package private or protected, and make all internal code use the static method instead of the constructor.

arteam:
I think both ways are fine from the point of view of improving testability (maybe removing the `final` modifier from `StatementContext` is easier and more sane). As I see, the interface approach is more close to the spirit of the classic OOP design.

qualidafial:
@stevenschlansker do you care to weigh in?

stevenschlansker:
Closed in favor of #522

PR #540, opened 2016-10-25
Statement execution - increase flexibility around reducing, improve cleanup

### LINKED ISSUE #526: Streams don't quite make up for v2's Query.fold

stevenschlansker:
StatementExecutor: allow customization of how jdbi runs a statement
ResultBearing: change main `iterator()` interface to instead use `execute(StatementExecutor)`
ReduceRow: introduce a jdbi "view" over a `ResultSet`
ResultBearing: introduce `reduceRows` and `reduceResultSet`
ResultSetResultIterator: simplify interface greatly by removing unused parameters

fixes #526

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/BaseStatement.java
+++ b/core/src/main/java/org/jdbi/v3/core/BaseStatement.java
@@ -13,6 +13,7 @@
  */
 package org.jdbi.v3.core;
 
+import java.io.Closeable;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.util.ArrayList;
@@ -22,7 +23,7 @@ import org.jdbi.v3.core.Cleanables.Cleanable;
 import org.jdbi.v3.core.exception.UnableToExecuteStatementException;
 import org.jdbi.v3.core.statement.StatementCustomizer;
 
-abstract class BaseStatement
+abstract class BaseStatement implements Closeable
 {
     final JdbiConfig config;
     private final Collection<StatementCustomizer> customizers = new ArrayList<>();
@@ -32,7 +33,6 @@ abstract class BaseStatement
     {
         this.config = config;
         this.context = context;
-        addCustomizer(new StatementCleaningCustomizer());
     }
 
     final ArgumentRegistry getArgumentRegistry() {
@@ -45,6 +45,11 @@ abstract class BaseStatement
     public final StatementContext getContext()
     {
         return context;
+    }
+
+    protected void addCleanable(Cleanable cleanable)
+    {
+        getContext().getCleanables().add(cleanable);
     }
 
     protected void addCustomizers(final Collection<StatementCustomizer> customizers)
@@ -86,31 +91,10 @@ abstract class BaseStatement
         }
     }
 
-    protected final void cleanup()
+    @Override
+    public void close()
     {
-        for (StatementCustomizer customizer : customizers) {
-            try {
-                customizer.cleanup(context);
-            }
-            catch (SQLException e) {
-                throw new UnableToExecuteStatementException("Could not clean up", e, context);
-            }
-        }
-    }
-
-    protected void addCleanable(final Cleanable cleanable)
-    {
-        context.getCleanables().add(cleanable);
-    }
-
-    class StatementCleaningCustomizer implements StatementCustomizer
-    {
-        @Override
-        public final void cleanup(final StatementContext ctx)
-            throws SQLException
-        {
-            context.getCleanables().clean();
-        }
+        getContext().close();
     }
 }
 
--- a/core/src/main/java/org/jdbi/v3/core/ResultBearing.java
+++ b/core/src/main/java/org/jdbi/v3/core/ResultBearing.java
@@ -28,8 +28,28 @@ import org.jdbi.v3.core.util.StreamConsumer;
 
 public interface ResultBearing<T> extends Iterable<T>
 {
+    /**
+     * Execute the statement.  The given {@code StatementExecutor}
+     * decides the execution strategy and return type.
+     *
+     * Most users will not use this method directly.
+     *
+     * @param executor the StatementExecutor to use
+     * @return the produced results
+     */
+    <R> R execute(StatementExecutor<T, R> executor);
+
+    /**
+     * Stream all the rows of the result set out
+     * with an {@code Iterator}.  The {@code Iterator} must be
+     * closed to release database resources.
+     * @return the results as a streaming Iterator
+     */
     @Override
-    ResultIterator<T> iterator();
+    default ResultIterator<T> iterator()
+    {
+        return execute(ResultSetResultIterator<T>::new);
+    }
 
     /**
      * Get the only row in the result set.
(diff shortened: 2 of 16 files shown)


stevenschlansker:
@qualidafial @arteam this might not be done yet, but does it look like it's going in the right direction?

core/src/main/java/org/jdbi/v3/core/BaseStatement.java
```diff
@@ -47,6 +47,11 @@ public final StatementContext getContext()
         return context;
     }
 
+    protected void addCleanable(Cleanable cleanable)
```
qualidafial: Should be package private if `Cleanable` type is non-public.

core/src/main/java/org/jdbi/v3/core/ReduceRow.java
```diff
@@ -0,0 +1,101 @@
...
+
+import org.jdbi.v3.core.exception.UnableToExecuteStatementException;
+
+/**
+ * A ReduceRow is a view over a single row of a {@code ResultSet}
+ * that uses the current statement context to map rows and
+ * columns to values.  It is not valid past the return of
+ * the method that receives it.
+ */
+public class ReduceRow
```
qualidafial: This is brilliant. Now you have me wondering if ReduceRow and JoinRow are the same thing.

If so, I wonder if we couldn't just make ReduceRow (possibly with a different name) be the new default mapper. :thinking_face:

core/src/main/java/org/jdbi/v3/core/ReduceRow.java
```diff
@@ -0,0 +1,101 @@
...
+    ReduceRow(ResultSet rs, StatementContext ctx)
+    {
+        this.rs = rs;
+        this.ctx = ctx;
+    }
+
+    /**
+     * Use a row mapper to extract a type from the current ResultSet row.
+     */
+    public <T> T getRow(Class<T> rowClass)
```
qualidafial: Wants a `T getRow(GenericType<T>)` method too. Ditto for `getColumn`

core/src/main/java/org/jdbi/v3/core/ResultBearing.java
```diff
@@ -156,4 +163,49 @@ default T findOnly() {
...
+    default <U> U reduceRows(U identity, BiFunction<U, ReduceRow, U> accumulator) {
+        return execute((mapper, rs, ctx) -> {
+            try {
+                ReduceRow rr = new ReduceRow(rs, ctx);
+                U result = identity;
+                while (rs.next()) {
+                    result = accumulator.apply(result, rr);
+                }
+                return result;
+            } catch (SQLException e) {
```
qualidafial: If `StatementExecutor` threw SQLException, you could catch it once in the execute method above.

core/src/main/java/org/jdbi/v3/core/ResultSetAccumulator.java
```diff
@@ -0,0 +1,22 @@
...
+ * limitations under the License.
+ */
+package org.jdbi.v3.core;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+
+public interface ResultSetAccumulator<T>
+{
+    T apply(T in, ResultSet rs) throws SQLException;
```
qualidafial: Feels like we still want the context argument.

core/src/main/java/org/jdbi/v3/core/ReduceRow.java
qualidafial: In fact if we had `getRow(RowMapper<T>)` and `getColumn(int|String, ColumnMapper)` methods then JoinRow would be redundant.

qualidafial: If we made JoinRow look like ReduceRow (lazy mapping, map row or column types, no need to register types joined on row) then I see a very strong case for JoinRow being the default mapped typed instead of Map<String,Object>

core/src/main/java/org/jdbi/v3/core/ReduceRow.java
stevenschlansker: Yes, I had the same thought, I think those are both good ideas.
I really dislike the current `Map<String, Object>` default.

stevenschlansker: I don't think we can make this the default mapper.  It is actually not valid to "map" to a `ReduceRow` -- since it is a view, not a real object that holds its data.

I still think the current `Map<String, Object>` default sucks though.

core/src/main/java/org/jdbi/v3/core/ReduceRow.java
stevenschlansker: `JoinRow` cannot be lazy because the `ResultSet` is iterated concurrently and may have advanced by the time anybody looks at it.

stevenschlansker: Concurrently might not be the right word.  But by the time you retrieve an element from your `List` or whatever container you've got, the `ResultSet` is advanced and potentially even closed.

core/src/main/java/org/jdbi/v3/core/ReduceRow.java
qualidafial: Oh well.

One final nit: The name `ReduceRow` might be too coupled to the present use case. There's nothing about this class that is specific to a `reduce` operation.

core/src/main/java/org/jdbi/v3/core/ReduceRow.java
qualidafial: Yeah, you're probably right. Oh well. 😞

core/src/main/java/org/jdbi/v3/core/ReduceRow.java
stevenschlansker: `ResultView`?

core/src/main/java/org/jdbi/v3/core/BaseStatement.java
stevenschlansker: Just being consistent with the other methods.  You can't use it either way.

stevenschlansker:
Okay, I addressed code review and added more javadoc.  Mind taking one last look?  Feel free to merge if you're happy with it.

core/src/main/java/org/jdbi/v3/core/ReduceRow.java
qualidafial: SGTM

qualidafial:
Reviewing.

core/src/main/java/org/jdbi/v3/core/BaseStatement.java
qualidafial: Oh right, didn't notice the class visibility ¯\\\_(ツ)\_/¯

core/src/main/java/org/jdbi/v3/core/RowView.java
```diff
@@ -0,0 +1,129 @@
...
+        return rowClass.cast(getRow((Type) rowClass));
+    }
+
+    /**
+     * Use a row mapper to extract a type from the current ResultSet row.
+     */
+    @SuppressWarnings("unchecked")
+    public <T> T getRow(GenericType<T> rowClass)
+    {
+        return (T) getRow((Type) rowClass);
```
qualidafial: Should be `rowClass.getType()`

core/src/main/java/org/jdbi/v3/core/RowView.java
```diff
@@ -0,0 +1,129 @@
...
+        return type.cast(getColumn(column, (Type) type));
+    }
+
+    /**
+     * Use a column mapper to extract a type from the current ResultSet row.
+     */
+    @SuppressWarnings("unchecked")
+    public <T> T getColumn(String column, GenericType<T> type)
+    {
+        return (T) getColumn(column, (Type) type);
```
qualidafial: Ditto: should be `type.getType()`

core/src/main/java/org/jdbi/v3/core/RowView.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+    /**
+     * Use a column mapper to extract a type from the current ResultSet row.
+     */
+    @SuppressWarnings("unchecked")
+    public <T> T getColumn(int column, GenericType<T> type)
+    {
+        return (T) getColumn(column, (Type) type);
```
qualidafial: Ditto

qualidafial:
Damn fine work. Merging.

PR #544, opened 2016-10-27
Fix class leak: use synchronized weak hash map instead of ConcurrentH…

qualidafial:
…ashMap, so class keys can be garbage-collected.


arteam:
LGTM

Issue #529, opened 2016-10-17
@BindIn.value() should default to the parameter name


qualidafial:
For parity with `@Bind` and `@Define`, and soon `@DefineIn`

qualidafial:
Fixed in #545

PR #545, opened 2016-10-27
Make `@BindIn` defined attribute default to parameter name if undefin…

qualidafial:
…ed. Clean up javadoc for Bind, BindIn, Define, DefineIn.

Fix for #529


qualidafial:
@arteam Could you review?

postgres/src/test/java/org/jdbi/v3/postgres/TestSqlArrays.java
```diff
@@ -180,4 +185,36 @@ public void testIntList() throws Exception {
...
+    @RegisterRowMapper(SomethingMapper.class)
+    public interface WhereInDao {
+        @SqlUpdate("create table something(id int, name text)")
+        void createTable();
+
+        @SqlBatch("insert into something(id, name) values (:id, :name)")
+        void insert(@BindBean Something... somethings);
+
+        @SqlQuery("select * from something where id = any(:ids) order by id")
+        List<Something> getByIds(int... ids);
```
qualidafial: I threw this test in, wanted to see array arguments worked without having to use `@BindIn`

qualidafial:
I think I'm ready to move `@BindIn` out of the unstable package too.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/BindIn.java
```diff
@@ -132,9 +152,20 @@ public SqlStatementCustomizer createForParameter(final Annotation annotation, fi
         @Override
         public Binder<BindIn, Object> build(final BindIn bindIn)
         {
-            final String key = bindIn.value();
-
             return (q, param, index, bind, arg) -> {
+                String name = bindIn.value();
+                if (name.isEmpty()) {
+                    if (param.isNamePresent()) {
+                        name = param.getName();
```
stevenschlansker: This logic duplicates `Bind`'s parameter handling -- probably worth pulling it out into a util class?  Seems likely we'll want to use it again.

string-template/src/test/java/org/jdbi/v3/sqlobject/BindInTest.java
```diff
@@ -65,6 +65,25 @@ public static void exit()
...
+        final List<Something> out = s.get(1, 2);
+
+        assertThat(out).hasSize(2);
+    }
+
+    @UseStringTemplateStatementRewriter
+    public interface SomethingWithExplicitAttributeName
+    {
+        @SqlQuery("select id, name from something where id in (<ids>)")
+        List<Something> get(@BindIn("ids") int... ids);
```
stevenschlansker: If the point here is to test the explicit setting of `ids`, it should probably differ from the parameter name?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/BindIn.java
```diff
@@ -36,14 +36,20 @@
 import org.jdbi.v3.sqlobject.SqlStatementCustomizingAnnotation;
 
 /**
- * Binds an Iterable or array/varargs of any type to a placeholder as a comma-separated list (e.g. for WHERE X IN (...) query statements).
+ * Binds each value in the annotated {@link Iterable} or array/varargs argument, and defines a named attribute as a
+ * comma-separated list of each bound parameter name. Common use cases:
  *
- * Throws IllegalArgumentException if the argument is not an array or Iterable. How null and empty collections are handled can be configured with onEmpty:EmptyHandling - throws IllegalArgumentException by default.
+ * <pre>
+ * &#64;SqlQuery("SELECT * FROM THINGS WHERE ID IN (&lt;ids&gt;)")
```
arteam: Could you convert `things` in the lower case? I think this would be easier for the reader to distinguish between SQL statements and database objects.

PR #547, opened 2016-10-28
`@Bind` no longer needs the option to switch binder.

### LINKED ISSUE #542: @Bind.binder() superfluous?
The layer of indirection from the `@Bind` annotation, through `BindFactory`, which just instantiates the class from `Bind.binder()` (defaults to `DefaultObjectBinder`) looks to me like a vestigial tail from before the `@BindingAnnotation` meta-annotation was introduced. Can we get rid of it in v3?

stevenschlansker:
You can create custom `@BindingAnnotation`s instead.
Fixes #542


sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultObjectBinder.java
```diff
@@ -21,13 +21,19 @@
 import org.jdbi.v3.core.SqlStatement;
 import org.jdbi.v3.core.util.GenericTypes;
 
-class DefaultObjectBinder implements Binder<Bind, Object>
+class DefaultObjectBinder implements BinderFactory<Bind, Object>, Binder<Bind, Object>
```
stevenschlansker: This is a bit funky.

stevenschlansker:
This along the lines of what you were thinking?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultObjectBinder.java
```diff
@@ -21,13 +21,19 @@
...
 {
     DefaultObjectBinder()
     {
     }
 
     @Override
+    public Binder<Bind, Object> build(Bind annotation)
+    {
+        return this;
```
qualidafial: Could do what the other binder factories are doing, and turn the `bind` method into a lambda that we return from the `build` method.

stevenschlansker:
Okay, updated the PR.  I think it's non-controversial but I'll leave it here for a day or three in case you want to re-review.

PR #548, opened 2016-10-28
H2 array integration improvements, support `Set<T>` types, bare configuration

stevenschlansker:
Three small changes in one:
- Support for configuring core features that aren't bound to an extension (RFC, do you like this?)
- Use that feature to support H2 array binding, introduce a plugin to configure it
- Add support for mapping `Set<T>` and friends

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/ConfigOnlyExtension.java
+++ b/core/src/main/java/org/jdbi/v3/core/ConfigOnlyExtension.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core;
+
+import org.jdbi.v3.core.extension.ExtensionConfig;
+import org.jdbi.v3.core.extension.ExtensionFactory;
+
+class ConfigOnlyExtension<C extends ExtensionConfig<C>> implements ExtensionFactory<C>
+{
+    private final Class<C> configClass;
+
+    ConfigOnlyExtension(Class<C> configClass)
+    {
+        this.configClass = configClass;
+    }
+
+    @Override
+    public C createConfig() {
+        try {
+            return configClass.getConstructor().newInstance();
+        } catch (ReflectiveOperationException e) {
+            throw new IllegalStateException("Unable to instantiate " + configClass, e);
+        }
+    }
+
+    @Override
+    public boolean accepts(Class<?> extensionType) {
+        return extensionType == configClass;
+    }
+
+    @Override
+    public <E> E attach(Class<E> extensionType, C config, HandleSupplier handle) {
+        return extensionType.cast(config);
+    }
+}
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -651,8 +651,14 @@ public class Handle implements Closeable
                 .orElseThrow(() -> new NoSuchExtensionException("Extension not found: " + extensionType));
     }
 
-    public <C extends ExtensionConfig<C>> void configureExtension(Class<C> configClass, Consumer<C> consumer) {
+    public <C extends ExtensionConfig<C>> Handle configureExtension(Class<C> configClass, Consumer<C> consumer) {
         config.extensionRegistry.configure(configClass, consumer);
+        return this;
+    }
+
+    public <C extends ExtensionConfig<C>> Handle configure(Class<C> configClass, Consumer<C> consumer) {
+        registerExtension(new ConfigOnlyExtension<C>(configClass));
+        return configureExtension(configClass, consumer);
     }
 
     public ExtensionMethod getExtensionMethod() {
--- a/core/src/main/java/org/jdbi/v3/core/Jdbi.java
+++ b/core/src/main/java/org/jdbi/v3/core/Jdbi.java
@@ -366,9 +366,16 @@ public class Jdbi
         return this;
     }
 
-    public <C extends ExtensionConfig<C>> Jdbi configureExtension(Class<C> configClass, Consumer<C> consumer) {
+    public <C extends ExtensionConfig<C>> Jdbi configureExtension(Class<C> configClass, Consumer<C> consumer)
+    {
         config.extensionRegistry.configure(configClass, consumer);
         return this;
+    }
+
+    public <C extends ExtensionConfig<C>> Jdbi configure(Class<C> configClass, Consumer<C> consumer)
+    {
+        registerExtension(new ConfigOnlyExtension<C>(configClass));
+        return configureExtension(configClass, consumer);
     }
 
     /**
--- a/core/src/main/java/org/jdbi/v3/core/Jdbi.java
+++ b/core/src/main/java/org/jdbi/v3/core/Jdbi.java
@@ -73,6 +73,13 @@ public class Jdbi
     {
         Objects.requireNonNull(connectionFactory, "null connectionFactory");
         this.connectionFactory = connectionFactory;
+
+        baseConfiguration();
+    }
+
+    private void baseConfiguration()
+    {
+        configure(SqlArrayConfiguration.class, c -> {});
     }
 
     /**
--- a/core/src/main/java/org/jdbi/v3/core/SqlArrayArgument.java
+++ b/core/src/main/java/org/jdbi/v3/core/SqlArrayArgument.java
@@ -44,8 +44,18 @@ class SqlArrayArgument<T> implements Argument {
 
     @Override
     public void apply(int position, PreparedStatement statement, StatementContext ctx) throws SQLException {
-        java.sql.Array sqlArray = statement.getConnection().createArrayOf(typeName, array);
-        ctx.getCleanables().add(sqlArray::free);
-        statement.setArray(position, sqlArray);
+        ArgumentStyle argumentStyle = ctx.findConfiguration(SqlArrayConfiguration.class).getArgumentStyle();
+        switch (argumentStyle) {
+        case SQL_ARRAY:
+            java.sql.Array sqlArray = statement.getConnection().createArrayOf(typeName, array);
+            ctx.getCleanables().add(sqlArray::free);
+            statement.setArray(position, sqlArray);
+            break;
+        case OBJECT_ARRAY:
+            statement.setObject(position, array);
+            break;
+        default:
+            throw new UnsupportedOperationException("Unknown array argument style " + argumentStyle);
+        }
     }
 }
--- a/core/src/main/java/org/jdbi/v3/core/SqlArrayArgument.java
+++ b/core/src/main/java/org/jdbi/v3/core/SqlArrayArgument.java
@@ -16,8 +16,9 @@ package org.jdbi.v3.core;
 import java.lang.reflect.Array;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
-import java.util.List;
+import java.util.Collection;
 
+import org.jdbi.v3.core.SqlArrayConfiguration.ArgumentStyle;
 import org.jdbi.v3.core.argument.Argument;
 import org.jdbi.v3.core.argument.SqlArrayType;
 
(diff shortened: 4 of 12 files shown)


core/src/main/java/org/jdbi/v3/core/ConfigOnlyExtension.java
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core;
+
+import org.jdbi.v3.core.extension.ExtensionConfig;
+import org.jdbi.v3.core.extension.ExtensionFactory;
+
+class ConfigOnlyExtension<C extends ExtensionConfig<C>> implements ExtensionFactory<C>
```
stevenschlansker: This is the thing I'm most curious for review of.  The intention here is to provide an extensible way for *core* jdbi to customize, e.g. cache sizes or other "indirect" tuneables, in this case "sql array binding strategy".

Other option would be to have a `ObjectArrayArgumentFactory` that binds `Object[]` instead of `sql.Array` for H2's benefit.  Might be simpler overall, but I think the concept here is interesting.

core/src/main/java/org/jdbi/v3/core/StatementContext.java
```diff
@@ -155,6 +157,13 @@ public Object getAttribute(String key)
         return config.collectorRegistry.findCollectorFor(type);
     }
 
+    public <C extends ExtensionConfig<C>> C findConfiguration(Class<C> configType)
```
stevenschlansker: This is kinda weird too.  Do we actually expect to have multiple configurations loaded ever?  Shouldn't just the most specific one win?  I'm not sure the `configure(Class, Consumer)` API makes sense, why not just `getConfig(Class)` that returns the most specific one.

core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -651,8 +651,14 @@ public TransactionIsolationLevel getTransactionIsolationLevel() {
...
 
-    public <C extends ExtensionConfig<C>> void configureExtension(Class<C> configClass, Consumer<C> consumer) {
+    public <C extends ExtensionConfig<C>> Handle configureExtension(Class<C> configClass, Consumer<C> consumer) {
         config.extensionRegistry.configure(configClass, consumer);
+        return this;
+    }
+
+    public <C extends ExtensionConfig<C>> Handle configure(Class<C> configClass, Consumer<C> consumer) {
+        registerExtension(new ConfigOnlyExtension<C>(configClass));
+        return configureExtension(configClass, consumer);
```
qualidafial: If we're looking for arbitrary configuration storage, then it probably makes sense to separate that out from extensions. That is, extensions can have configuration, but they can fetch whatever configuration they're interested in, rather than expecting core to understand what each extension needs.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -73,6 +73,13 @@ private Jdbi(ConnectionFactory connectionFactory)
     {
         Objects.requireNonNull(connectionFactory, "null connectionFactory");
         this.connectionFactory = connectionFactory;
+
+        baseConfiguration();
+    }
+
+    private void baseConfiguration()
+    {
+        configure(SqlArrayConfiguration.class, c -> {});
```
qualidafial: Our mapper, argument, and collector registries might fit in this paradigm. :thinking:

core/src/main/java/org/jdbi/v3/core/ConfigOnlyExtension.java
qualidafial: Per my comment in `Handle`, I think if we're going to use configuration objects this way, then it makes sense to make config a standalone concept on which extensions might rely, in which case this class would be superfluous.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -366,11 +373,18 @@ public Jdbi registerExtension(ExtensionFactory<?> extensionFactory)
...
-    public <C extends ExtensionConfig<C>> Jdbi configureExtension(Class<C> configClass, Consumer<C> consumer) {
+    public <C extends ExtensionConfig<C>> Jdbi configureExtension(Class<C> configClass, Consumer<C> consumer)
+    {
         config.extensionRegistry.configure(configClass, consumer);
         return this;
     }
 
+    public <C extends ExtensionConfig<C>> Jdbi configure(Class<C> configClass, Consumer<C> consumer)
+    {
+        registerExtension(new ConfigOnlyExtension<C>(configClass));
```
qualidafial: Repeated calls for the same config class will register duplicate (and useless) extensions.

core/src/main/java/org/jdbi/v3/core/SqlArrayArgument.java
```diff
@@ -36,15 +37,25 @@
...
-        statement.setArray(position, sqlArray);
+        ArgumentStyle argumentStyle = ctx.findConfiguration(SqlArrayConfiguration.class).getArgumentStyle();
+        switch (argumentStyle) {
+        case SQL_ARRAY:
+            java.sql.Array sqlArray = statement.getConnection().createArrayOf(typeName, array);
+            ctx.getCleanables().add(sqlArray::free);
+            statement.setArray(position, sqlArray);
+            break;
+        case OBJECT_ARRAY:
+            statement.setObject(position, array);
```
qualidafial: Nifty

core/src/main/java/org/jdbi/v3/core/SqlArrayConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+package org.jdbi.v3.core;
+
+import org.jdbi.v3.core.extension.ExtensionConfig;
+
+public class SqlArrayConfiguration implements ExtensionConfig<SqlArrayConfiguration>
+{
+    public enum ArgumentStyle
+    {
+        SQL_ARRAY, OBJECT_ARRAY
```
qualidafial: I'd probably call it `JAVA_ARRAY` instead of `OBJECT_ARRAY`, since we also support primitive arrays.

core/src/main/java/org/jdbi/v3/core/StatementContext.java
qualidafial: `configType` might be an interface, in which case it's possible there are multiple registered config types that match it.

I have to think about this for a bit.

core/src/main/java/org/jdbi/v3/core/vendor/H2DatabasePlugin.java
```diff
@@ -0,0 +1,31 @@
...
+import org.jdbi.v3.core.SqlArrayConfiguration;
+import org.jdbi.v3.core.SqlArrayConfiguration.ArgumentStyle;
+import org.jdbi.v3.core.spi.JdbiPlugin;
+
+public class H2DatabasePlugin implements JdbiPlugin
+{
+    @Override
+    public void customizeDbi(Jdbi dbi)
+    {
+        dbi.configure(SqlArrayConfiguration.class, c -> c.withArgumentStyle(ArgumentStyle.OBJECT_ARRAY));
```
qualidafial: Nice.

qualidafial: We could use this configuration stuff for reflective mapper column name matching strategies too.

core/src/main/java/org/jdbi/v3/core/SqlArrayConfiguration.java
qualidafial: Hmm.. I may have my wires crossed on this one. Let me dig into this a little before you spend time on it.

core/src/main/java/org/jdbi/v3/core/StatementContext.java
stevenschlansker: In the case where we have an interface with multiple implementations, the `Consumer<C>` must be the most general type.  So unless your configuration interfaces' differ only in the implementation of methods, you would have no way to invoke any additional functionality on the more specific types.  e.g. if your configuration takes 
```java
interface SuperConfig { void setFoo(int); }
interface SubConfig extends SuperConfig { void setBar(int); }
```

then if you ever want to invoke `setBar` you *have* to configure `SubConfig.class` explicitly.
Do you have a good example where stacking configurations makes it simpler?  I wonder if a simple `configure(ConfigType)` and `getConfig(Class<ConfigType>)` might suffice and be easier to work with.

qualidafial: I think we should just have `getConfig(Class<T extends ExtensionConfig>): T` and only match by exact class. Might make sense to turn ExtensionConfig into an abstract class too.

I've thought back on how this class came about in the extensions PR, and had to acknowledge that this idea of configuration by instanceof was speculative. We don't have any concrete reason to hold to it.

core/src/main/java/org/jdbi/v3/core/SqlArrayArgumentFactory.java
```diff
@@ -23,19 +23,31 @@
...
+        if (Collection.class.isAssignableFrom(erasedType)) {
+            return guessElementType(type, (Collection<?>) value)
                     .flatMap(ctx::findArrayTypeFor)
-                    .map(arrayType -> new SqlArrayArgument(arrayType, (List) value));
+                    .map(arrayType -> new SqlArrayArgument<>(arrayType, (Collection) value));
         }
         return Optional.empty();
     }
+
+    private Optional<? extends Type> guessElementType(Type type, Collection<?> value) {
```
stevenschlansker: Right now if you `bind("ids", new ArrayList<>(1, 2, 3))`, `findGenericParameter` returns `empty` because the runtime type is erased.  Guessing based on the elements is kinda shitty, and we fall back to `Object`, but I'm not sure what else to do.  (You can be explicit with `bindByType` but asking our users to do that for every array sucks.)

qualidafial: Would it be worth it to add `SqlStatement.bindArray(elementType, array)` methods, with e.g. overloads for `elementType` as `Class<T>` or `GenericType<T>`, and `array` as `T...` or `Collection<T>`?

qualidafial: If the element type was a `Class<T>` then users could bind the array without having to use `GenericType`:

```java
statement.bindArray(int.class, new ArrayList<>(1, 2, 3));
```

On principle, I'm a little leery of going by the first element. I think if we can't reliably figure out what the type is, then the user should have to tell us.

stevenschlansker: It used to work, though, and it's pretty convenient.... I agree that inferring from the first element though is dangerous, since an empty list may then break.

core/src/main/java/org/jdbi/v3/core/ConfigOnlyExtension.java
qualidafial: To elaborate on this comment and others:

There is nothing extension-specific about `ExtensionConfig`. We introduced it because extensions needed _somewhere_ to keep configuration. But if you think about it, we already maintain config objects all over the place: the registries. What if we promoted configuration objects to first-class citizens?

Proposal:

* Rename existing `JdbiConfig` to `ConfigRegistry`. Make it public.
* Rename `ExtensionConfig` to `JdbiConfig`.
* Modify the argument registry, mapping registry, collector registry, and extension registry to implement `JdbiConfig`. It might be valuable to make their `createChild()` method implementations use a chain-of-responsibility pattern to minimize copied objects at runtime.
* Fold the statement rewriter, timing collector, and statement attributes into a new config type. `SqlStatementConfig` perhaps?
* Refactor `ConfigRegistry` so it is just a registry of `JdbiConfig` types, similar to the other registries. A `JdbiConfig` of a given type is instantiated automatically the first time it is accessed.
* Replace `ConfigRegistry.copyOf` with `ConfigRegistry.createChild` for uniformity with `JdbiConfig`. Possibly also make `ConfigRegistry` use chain of responsibility, to defer copying parent config objects until needed.
* Refactor extensions so that they get the `ConfigRegistry`, not just not just the one config type. To that end, remove the `C extends ExtensionConfig<C>` generic parameter, and the `createConfig` method from `ExtensionFactory`.
* Add accessor methods to `Jdbi`, `Handle`, `SqlStatement`, and `StatementContext`
* Refactor SQL Object so that `Handler` has access to the entire config registry, not just `SqlObjectConfig`.

The cool thing about this is that anybody can add their own new config concepts, and then everything can access that configuration with the same ease as the built-in config like the mapping registry.

@stevenschlansker @arteam @brianm What do you think?

qualidafial: We could also have a config type for our reflective mappers, that would allow a global configuration of e.g. which `ColumnNameMappingStrategy` to use.

qualidafial: Ping?

stevenschlansker: @qualidafial sorry I'm not responding as quickly, it's been a shitty few weeks for various reasons...  I like the idea of refactoring around a `ConfigRegistry`, I think it's a more natural fit than the current API.

qualidafial: I'll start experimenting with this tonight

qualidafial:
Please see #556

qualidafial:
The contents of this PR was included in #556

stevenschlansker:
Thanks!

PR #552, opened 2016-11-08
Add column name prefix option to reflective mappers

### LINKED ISSUE #532: Give reflection-based row mappers a column name prefix option
Applies to: `BeanMapper`, `ConstructorMapper`, `FieldMapper`.

The specific use case I have in mind is with join row mappers:

``` java
class File {
  int id;
  String name;
  String contents;
}

class Directory {
  int id;
  String name;
  List<File> files;
}
```

I can't use reflection-based mappers with `JoinRowMapper` here, because the classes share the same property names. However if I could get a mapper that applies a column name prefix e.g. `f_` and `d_`, then I could exploit those prefixes to disambiguate those common property names:

Fluent:

``` java
handle.createQuery("select d.id d_id, d.name d_name, " +
    "f.id f_id, f.name f_name, f.contents f_contents " +
    "from directories d left join files f on d.id = f.directory_id")
    .registerRowMapper(BeanMapperFactory.forType("f_", File.class))
    .registerRowMapper(BeanMapperFactory.forType("d_", Directory.class))
    .map(JoinRowMapper.forTypes(Directory.class, File.class))
    .list();
```

SQL Object:

``` java
@RegisterBeanMapper(value={File.class,Directory.class}, prefix={"f_", "d_"})
@RegisterJoinRowMapper({File.class, Directory.class})
@SqlQuery("select d.id d_id, d.name d_name, " +
    "f.id f_id, f.name f_name, f.contents f_contents " +
    "from directories d left join files f on d.id = f.directory_id")
List<JoinRow> listFiles();
```

qualidafial:
* Add column name prefix option to BeanMapper, FieldMapper, and ConstructorMapper.
* Remove BeanMapperFactory in favor of BeanMapper.of factory methods like ConstructorMapper.
* Add FieldMapper.of factory methods.
* Add RegisterFieldMapper SQL object annotation.

Fixes #532


qualidafial:
Amended usage from #532:

Fluent:

```java
handle.createQuery("select d.id d_id, d.name d_name, " +
    "f.id f_id, f.name f_name, f.contents f_contents " +
    "from directories d left join files f on d.id = f.directory_id")
    .registerRowMapper(BeanMapper.of(File.class, "f_"))
    .registerRowMapper(BeanMapper.of(Directory.class, "d_"))
    .map(JoinRowMapper.forTypes(Directory.class, File.class))
    .list();
```

SQL Object:

```java
@RegisterBeanMapper(value={File.class,Directory.class}, prefix={"f_", "d_"})
@RegisterJoinRowMapper({File.class, Directory.class})
@SqlQuery("select d.id d_id, d.name d_name, " +
    "f.id f_id, f.name f_name, f.contents f_contents " +
    "from directories d left join files f on d.id = f.directory_id")
List<JoinRow> listFiles();
```

qualidafial:
@stevenschlansker @arteam would either of you be able to review this?

core/src/main/java/org/jdbi/v3/core/mapper/BeanMapper.java
```diff
@@ -93,6 +136,16 @@ public T map(ResultSet rs, StatementContext ctx)
         for (int i = 1; i <= metadata.getColumnCount(); ++i) {
             String name = metadata.getColumnLabel(i);
 
+            if (prefix.length() > 0) {
```
stevenschlansker: `!prefix.isEmpty()` might be trivially more idiomatic

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBeanMapper.java
```diff
@@ -0,0 +1,250 @@
...
+        h.createUpdate("create table testBean (valueType varchar(50))").execute();
+        dao = h.attach(TestDao.class);
+    }
+
+    @Test
+    public void testMapBean() {
+        h.createUpdate("insert into testBean (valueType) values ('foo')").execute();
+
+        List<TestBean> beans = dao.listBeans();
+        assertThat(beans).extracting(TestBean::getValueType).containsExactly(ValueType.valueOf("foo"));
```
stevenschlansker: mind blown

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBeanMapper.java
```diff
@@ -0,0 +1,250 @@
...
+            return getHandle().createQuery(
+                    "select f.id f_id, f.name f_name, " +
+                            "d.id d_id, d.name d_name, d.contents d_contents " +
+                            "from folders f left join documents d " +
+                            "on f.id = d.folder_id " +
+                            "where f.id = :folderId")
+                    .bind("folderId", folderId)
+                    .registerRowMapper(BeanMapper.of(Folder.class, "f_"))
+                    .registerRowMapper(BeanMapper.of(Document.class, "d_"))
+                    .reduceRows(Optional.<Folder>empty(), (folder, row) -> {
```
stevenschlansker: If `reduce` were still called `fold` this test case would be very hard to read :P

qualidafial: Yo dawg, I heard you like folding folders

arg20:
@qualidafial In this example given above:
```
handle.createQuery("select d.id d_id, d.name d_name, " +
    "f.id f_id, f.name f_name, f.contents f_contents " +
    "from directories d left join files f on d.id = f.directory_id")
    .registerRowMapper(BeanMapper.of(File.class, "f_"))
    .registerRowMapper(BeanMapper.of(Directory.class, "d_"))
    .map(JoinRowMapper.forTypes(Directory.class, File.class))
    .list();
```
What if I have another query that retrieves files only so I don't add the prefix `f_`?

Once I do: `.registerRowMapper(BeanMapper.of(File.class))` (no prefix), won't this stop working? In other words, for the same `File` class, how can I register mappers with and without the prefix and alternate between them depending on the query?

qualidafial:
@arg20 the last registered mapper wins. So if you register a mapper with no prefix on the Jdbi (i.e. global mapper), you can override it on a SqlStatement by registering another.

Also note that you should prefer BeanMapper.factory(...) over .of(...) for registering mappers.

arg20:
@qualidafial thanks for the response, one last question, does that apply to SqlObjects as well? if I `@RegisterRowMapper` on a specific method, it will override the default for that method call but any other method will still use the default?

qualidafial:
@arg20 That's correct. Configuration is isolated from method to method and from statement to statement, so they can't change each other.

arg20:
@qualidafial Perfect, thanks for the extensive explanation, I really appreciate it, this was exactly how I needed jdbi to behave. Keep up the good work, will contribute if possible.

PR #555, opened 2016-11-10
Make recursive calls to an on-demand extension reentrant.

### LINKED ISSUE #550: On-demand extensions should be reentrant within the same thread
``` java
interface MyDao extends GetHandle, Transactional<MyDao> {
  default void doSomething() { ... }
}

MyDao myDao = jdbi.onDemand(MyDao.class);
myDao.useTransaction((dao, status) -> {
  myDao.doSomething(); // myDao, not dao
});
```

The `TransactionConsumer` lambda should have used the `dao` parameter, not the `myDao` variable from outside the lambda. Thus any calls to `myDao` inside the consumer callback will not be part of the transaction as expected. This is an easy error to make.

The on-demand extension wrapper should remember the materialized extension per thread, and simply forward those calls through to the existing extension instance when somebody makes a recursive call to an on-demand extension.

qualidafial:
Fixes #550 

@stevenschlansker code review?

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
+++ b/core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
@@ -35,6 +35,8 @@ public class OnDemandExtensions {
     }
 
     public static <E> E create(Jdbi dbi, Class<E> extensionType) {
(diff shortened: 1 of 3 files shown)


stevenschlansker:
Nice!

PR #556, opened 2016-11-10
Massive refactor for generic, extensible configuration support.

qualidafial:
See https://github.com/jdbi/jdbi/pull/548#discussion_r86909321 for background.

This branch is a major refactor of JDBI. This PR subsumes the work on PR #548, and generalizes the existing `ExtensionConfig` concept into `JdbiConfig`:

The former `JdbiConfig` class has been renamed to `ConfigRegistry`, and `ExtensionConfig` class has in turn been named to `JdbiConfig`.

`JdbiConfig` is the base interface for all configuration types:

```java
interface JdbiConfig<This extends JdbiConfig<This>> {
    // allows for inheritance of configuration from parent to child configs
    This createChild();
}
```

`ConfigRegistry` keeps a cache of config objects, and creates them as needed. For this reason, all implementations of `JdbiConfig` must provide a public no-arg constructor.

When a child object is created by JDBI, e.g. when `Jdbi.open()` opens a handle, a child config registry is created. All config objects in the child registry inherit from the corresponding config object in the parent registry.

Thus the configuration for a SQL statement executed by a SQL object is inherited from the SQL method, then the SQL object, then the JDBI or handle that opened the extension.

Other API changes:
* Moved many classes around, in pursuit of package-by-feature.
* Converted all existing registries to config classes, and made them public.
* Introduced `Configurable` interface, collecting all the common registration/configuration methods into the interface as default implementations. Normalized all registration method names in the process. There is also a single non-default method `getConfig()` which returns the config registry itself.
* Made `Jdbi`, `Handle`, `HandleSupplier`, `BaseStatement` all implement `Configurable`, and removed now-duplicated method implementations.
* Split up `MappingRegistry` into `RowMappers` and `ColumnMappers` config classes.
* Split up `ArgumentRegistry` into `Arguments` and `SqlArrayTypes` config classes.
* Renamed `ExtensionRegistry` to `Extensions`.
* Renamed `CollectorRegistry` to `JdbiCollectors` (to avoid conflict with `java.util.stream.Collectors`)
* Added `SqlStatements` config class, which holds defined attributes, statement rewriter, and timing collector.
* Renamed `SqlObjectConfig` to `SqlObjects`.
* Made `Cleanable` public, added public `StatementContext.addCleanable` method.
* Refactored `HandleSupplier`, replacing `getExtensionMethod` and `setExtensionMethod` with `invokeInContext`, which takes an extension method, a config registry, and a task, and runs the task using that config registry and extension method on the handle.
* Refactored `Handler` interface, removing `SqlObjectConfig` argument since you can get configuration out of the `HandleSupplier` now.
* Renamed `JdbiPlugin.customizeDbi(Jdbi dbi)` to `customizeJdbi(Jdbi jdbi)`.
* Copied over features from PR #548 as proof of concept.
* Converted several annotations from `@SqlStatementCustomizingAnnotation` into `@ConfiguringAnnotation`: `@OverrideStatementRewriterWith`, `@Register*`, `@Use*SqlLocator`, `@UseStringTemplateStatementRewriter`. This makes these annotations work on all SQL Object methods, including default methods.
* Made `SqlObjectFactory` not a singleton :tada:
* Modified factory classes e.g. `RowMapperFactory` to accept a `ConfigRegistry` instead of a `StatementContext`. This way it's valid to create and cache a row mapper for a given type at the `Jdbi` layer, and have it reusable across different contexts.


qualidafial:
At this point, I've succeeded in doing the renames and converting all the registries to config objects. I've also made the config registry and all the other registries use the chain-of-responsibility pattern (where the child registry delegates to its parent if it has nothing on a particular type) instead of duplicating entire registries at every step.

However I've hit a wall trying to make `SqlObjectConfig` just another config type, and removing it as an argument from `ExtensionFactory.build()` and `Handler.invoke`. I hope that by spelling this out to the group that one of us will get a bright idea to work out the issue.

Currently, a `SqlObjectConfig` instance is created at the `Jdbi` level when the extension is registered. This config object is cloned when an extension is created. That clone is then cloned again in turn for each method call. Thus global configuration is inherited from parent to child, but individual method calls will have a separate configuration object they can modify in isolation.

The problem is that, when a `Handle` is created (e.g. for an extension), the handle inherits configuration directly from the `Jdbi` instance. Any extension instance configuration has forked away from the configuration that the `Handle` will get.

I've modified `LazyHandle` to make the lazily generated handle inherit from the extension config. This gets us partway there:

```
Jdbi.config (config1)
+- Extension
   +- config2 (inherits from config1)
       +- config-per-method (inherits from config2)
   +- lazy handle
      +- config3 (inherits from config2)
```

But there's still a disconnect when it gets to the method level.

I'm considering adding a `Handle.createChild(): Handle` method. A child handle would be essentially the same handle with the same lifecycle as the parent, but with a forked configuration. By the same token, `HandleSupplier` would get a `createChild()` method with the same semantics. Thus:

```
Jdbi.config (config1)
+- Extension
   +- lazy handle for instance (inherits config from config1)
   +- lazy handle per method (inherits config from lazy handle for instance)

LazyHandle
+- config (root inherits from config1, children inherit from parent)
```

One possible simplification with this is that the ExtensionMethod could be just another bit of configuration. Since the configuration is forked for each method, there would be no need to manage thread locals.

It seems a bit weird, but it's my best idea at the moment. Suggestions?

stevenschlansker:
If the problem is that there are two separate trees of configuration, shouldn't we just forbid that?  Why does the `ExtensionRegistry` have to keep configuration at all?  It could keep a `Supplier<JdbiConfig<C>>` of defaults to apply if none is otherwise applied.

qualidafial:
The original motivation is to see whether/how this configuration idea works, and whether other parts of the API can be simplified by it.

In the process of this work, I came across a few corner cases that I also hope to address as part of this PR:

Most annotations don't work on default methods, and I'm kind of annoyed by that. IMO this usage should be valid:

```java
@RegisterBeanMapper(Foo.class)
interface FooDao {
  @SqlQuery("select * from foo")
  List<Foo> listFoos(int id)

  @RegisterBeanMapper(Bar.class)
  @RegisterJoinRowMapper({Foo.class, Bar.class})
  List<Foo> listFoosWithBars(int a, String b) {
    return getHandle().createQuery(
            "select ... from foo left join bar on foo.id = bar.foo_id where ...")
       .bind("a", a)
       .bind("b", b)
       .reduce(...);
  }
}
```

If extension methods could have distinct but related handles per method invocation, then these registering annotations could be `SqlObjectConfiguringAnnotation`s instead of `SqlStatementCustomizingAnnotation`s--making them work on all methods, default or not.

I also noticed something about our existing design: because all methods on a SQL Object instance share the same handle, it's possible for a default method to modify the behavior of other methods by modifying configuration directly on `getHandle()`. Examples:

* A global mapper is registered for a given type, but an override is registered for that type by a default method--successive calls to other methods that map to that type will use the override mapper.
* A default method sets a different statement rewriter on the handle before doing its thing. Successive calls to other methods will use the new rewriter and possibly fail as a result.

Again, if there was a "root" `Handle` per instance and a separate child handle per method invocation, there would be no possibility of configuration bleed.

qualidafial:
I've abandoned the `Handle.createChild()` approach, now looking into `Handle.useConfig(ConfigRegistry, HandleCallback)` to swap out config on the Handle. Fingers crossed.

qualidafial:
Bazinga! Got it working locally, in the process of migrating the registering annotations (and any other non-statement-specific ones I find) from sql statement customizing annotations to configuring annotations.

Will push code tomorrow.

qualidafial:
Need to add tests for the scenarios mentioned in https://github.com/jdbi/jdbi/pull/556#issuecomment-260239007:

* [ ] Default methods benefit from e.g. `@RegisterRowMapper` annotations on a type or method.
* [ ] Directly modifying the handle (e.g. setting a new statement rewriter) in one SQL object method should not affect the configuration in other methods.

qualidafial:
@stevenschlansker @arteam

Ahem.. I may have, uh, gotten a little carried away. This PR branch is currently +2112, −1766 lines vs `jdbi3` branch.

Could either of you review this? Pretty please?

stevenschlansker:
Thanks for the work on this.  I'll try to review this weekend, will take a bit of reading!

qualidafial:
Follow up changes I'd like to make after this gets merged:
* Add a config class for reflective mappers. Store the column name mapping strategies on that config instead of directly in `BeanMapper`, `FieldMapper`, and `ConstructorMapper`.
* Look through our other argument and mappers, and move any embedded configuration to a configuration object, where it makes sense to allow end users to define it globally.

qualidafial:
What do you think of the `<This>` parameter name for self-returning methods?

stevenschlansker:
I'm OK with that change.  I was fine with `SelfType` to start with, though.
As long as it's consistent across the codebase.

core/src/main/java/org/jdbi/v3/core/mapper/RowMappers.java
```diff
@@ -0,0 +1,106 @@
...
+            return Optional.of(cached);
+        }
+
+        Optional<RowMapper<?>> mapper = findFirstPresent(
+                () -> factories.stream()
+                        .flatMap(factory -> toStream(factory.build(type, config)))
+                        .findFirst(),
+                () -> config.findColumnMapperFor(type)
+                        .map(SingleColumnMapper::new),
+                // FIXME possible to taint parent cache with child config's mappers?
```
stevenschlansker: I think the real problem here is that the cache is keyed only on the `type`, not the `config`, even though that's part of the computation.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -0,0 +1,124 @@
...
+import org.jdbi.v3.core.SqlStatement;
+import org.jdbi.v3.core.StatementContext;
+import org.jdbi.v3.core.TimingCollector;
+import org.jdbi.v3.core.rewriter.ColonPrefixStatementRewriter;
+import org.jdbi.v3.core.rewriter.StatementRewriter;
+
+public final class SqlStatements implements JdbiConfig<SqlStatements> {
+
+    private final Map<String, Object> attributes;
+    private volatile StatementRewriter statementRewriter;
```
stevenschlansker: The use of `volatile` in config classes seems to be inconsistent.  We should decide exactly how thread safe configuration classes are and specify it in the interface.

I vote for, "config classes are thread safe for access / cache purposes but not for re-configuration; any configuration changes must be applied to a `Configurable` before it is handed off to other threads", or something like that?

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestRegisteredMappersWork.java
```diff
@@ -106,6 +116,25 @@ public void testBeanMapperFactory() throws Exception
     }
 
     @Test
+    public void testBeanMapperFactoryDefaultMethod() throws Exception
+    {
+        BeanMappingDao bdb = db.getSharedHandle().attach(BeanMappingDao.class);
+        bdb.createBeanTable();
+
+        Bean lima = new Bean();
```
stevenschlansker: amazing

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestSqlMethodAnnotations.java
```diff
@@ -69,7 +69,7 @@ public void testCustomAnnotation() {
         class Factory implements HandlerFactory {
             @Override
             public Handler buildHandler(Class<?> sqlObjectType, Method method) {
-                return (obj, m, args, config, handle) -> {
+                return (obj, m, args, handle) -> {
```
stevenschlansker: couldn't this just be `-> "foo"` ?

core/src/main/java/org/jdbi/v3/core/HandleSupplier.java
```diff
@@ -13,25 +13,17 @@
...
-    void setExtensionMethod(ExtensionMethod extensionMethod);
-
+public interface HandleSupplier extends Configurable<HandleSupplier> {
     /**
      * Returns a handle, possibly creating it lazily. A Handle holds a database connection, so extensions should only
      * call this method in order to interact with the database.
      */
     Handle getHandle();
+
+    <V> V invokeInContext(ExtensionMethod extensionMethod, ConfigRegistry config, Callable<V> task) throws Exception;
```
qualidafial: I'd say no, because that might create an unnecessary handle instance through the call to `getHandle()`. That's why `HandleSupplier` exists, really--to allow for lazy creation. Otherwise we could just use `Handle` directly.

core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Obtain an argument for given value in the given context
+     *
+     * @param type  the type of the argument.
+     * @param value the argument value.
+     * @param config the config registry, for composition
+     * @return an Argument for the given value.
+     */
+    public Optional<Argument> findFor(Type type, Object value, ConfigRegistry config) {
+        return findFirstPresent(
```
qualidafial: I considered that, but I don't think we should. `ArgumentFactory.build` takes both a type and a value argument. I don't think we can assume that just because a factory returns an argument for a particular type, that it will return one for any value of that type.

But that might be a good thing to follow up on: what if Arguments were like mappers, and didn't hold values but instead received a value as a parameter and write them to a prepared statement on demand? That would definitely allow us to cache and optimize in batches.

I'll open up a separate issue to discuss this further.

core/src/main/java/org/jdbi/v3/core/mapper/RowMappers.java
qualidafial: This is the thing that makes me question whether config inheritance is a good idea.

In this case, if the cache is keyed on both type and config, then because of the calls from child to parent, a `Jdbi`-level config (1) could end up keeping a SQL object method-level config (4) alive in memory long after that SQL object has been discarded:

```
Jdbi       -> config (1)
^                ^
|                | (inherits from, delegates to)
|                |
Handle     -> config (2)
^                ^
|                | (inherits from, delegates to)
|                |
Sql Object -> config (3)
^                ^
|                | (inherits from, delegates to)
|                |
Method     -> config (4)
```

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestRegisteredMappersWork.java
qualidafial: Heh, can't take credit for that--it's just a copypasta and modification of a preexisting test. 😆

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
qualidafial: Agreed, I'll add this to `JdbiConfig` in the form of `implementers of this interface should`, and then make sure it's enforced in all our own config classes.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestSqlMethodAnnotations.java
qualidafial: yep, will fix

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
qualidafial: So per your suggested javadoc, this means that config class fields need _not_ be volatile?

core/src/main/java/org/jdbi/v3/core/JdbiConfig.java
```diff
@@ -13,48 +13,17 @@
...
+ *
+ * @param <This> A "This" type. Should always be the configuration class.
+ */
+public interface JdbiConfig<This extends JdbiConfig<This>> {
+    /**
+     * Returns a copy of this configuration object. Changes to the copy should not modify the original.
+     *
+     * @return a copy of this configuration object.
+     */
+    This createChild();
```
qualidafial: Opening a separate issue to discuss this post-merge.

qualidafial:
Leeeeeerooooy Jenkins!!!

core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -51,166 +40,50 @@
...
-    protected final JdbiConfig config;
-    protected StatementBuilder statementBuilder;
-    private boolean closed = false;
-    protected final TransactionHandler transactions;
-    protected final Connection connection;
+    private final TransactionHandler transactions;
+    private final Connection connection;
 
-    private ThreadLocal<ExtensionMethod> extensionMethod = new ThreadLocal<>();
+    private ThreadLocal<ConfigRegistry> config;
```
stevenschlansker: I think this is worthy of mentioning in a comment.  I probably would have figured this out eventually but it's at least a little subtle :)

core/src/main/java/org/jdbi/v3/core/mapper/RowMappers.java
stevenschlansker: It might be possible to fix this by ensuring that you never cache in a parent -- you only cache in the "original" level you do a lookup.  But I'm leaning more and more towards moving to a simpler solution without the chain of responsibility...

core/src/main/java/org/jdbi/v3/core/JdbiConfig.java
qualidafial: Yep I'm leaning that way too. Let's move this discussion to #564

PR #567, opened 2016-11-24
WIP: Refactor arguments API to facilitate caching.

qualidafial:
This makes the argument more like mappers, in that arguments now represent _types_ instead of specific _values_ of that type.

In fact it might make sense to rename them to `ArgumentMapper` or `ArgMapper` or something.

Changes:
* ArgumentFactory: removed value parameter. Provides arguments based on type only.
* Argument: added value parameter. An instance no longer represents an argument value, but an argument type. The value is now provided to the Argument as a method parameter.

To do:
* [x] Actually do the caching part
* [x] Add registration for inferred argument types, like we do with concretely typed mappers.
* [ ] Give the test changes a once-over to make sure the tests are still validating the same rules.
* [ ] Give the whole thing a once over to make sure we actually want this.
* [x] See if we can add the `<?>` to the return type on `ArgumentFactory.build()`.


qualidafial:
This seems like a pretty major change just to support argument caching in batches.

It's been an interesting thought exercise, but I think the amount of change it takes to support this one small use case just isn't worth it.

Perhaps there's some small incidental improvements that might be worth reproducing, but looking at the bigger picture, I think merging this would be a mistake.

Unless I hear otherwise from the team, I'm planning to abandon this PR.

stevenschlansker:
Okay.  Sorry to send you down a rabbit hole! 😞

qualidafial:
¯\\\_(ツ)\_/¯

PR #569, opened 2016-11-24
Instruct Eclipse to not build asciidoc

stevenschlansker:
There's no plugin for it so this just causes build errors.
No effect in other environments.


qualidafial:
LGTM

PR #570, opened 2016-11-24
Remove pom.xml <developers> section

stevenschlansker:
It's a little out of date.

I don't want to diminish past contributions,
but we don't keep this up to date, and I feel
it doesn't add much value.

So the most fair resolution seems to be to delete it :)


qualidafial:
LGTM

PR #572, opened 2016-11-24
Clone configs instead of chaining them

### LINKED ISSUE #564: Consulting parent config with child config registry can pollute parent caches
Context: https://github.com/jdbi/jdbi/pull/556#discussion_r89422438

If we can't prevent cache pollution, we should probably ditch the chain-of-reponsibility pattern from child-to-parent config objects, and just clone config like we were doing before.

If we do that we should rename `JdbiConfig.createChild()` to `createCopy()` or maybe just `copy()`.

qualidafial:
Fixes #564


qualidafial:
@stevenschlansker This PR is pretty straightforward. Did you want to review this before I merge it?

PR #573, opened 2016-11-24
Jdbi3 reflection mapper config

qualidafial:


qualidafial:
Whoops--thought I had switched target branch

PR #576, opened 2016-11-25
Promote BindIn, DefineIn to core

### LINKED ISSUE #562: Add @DefineIn, @BindIn behavior into core
* [ ] Add `SqlStatement.bindList(String attributeName, Iterable<? extends Object> values)`. Copy implementation from `@BindIn`, then make the annotation just an alias for that new method.
* [ ] Add `SqlStatement.defineList(String attributeName, Iterable<? extends Object> values)`. Copy implementation from `@DefineIn`, then make the annotation just an alias for that new method.
* [ ] Rename `DefineIn` to `DefineList`, and `BindIn` to `BindList`-- to steer away from the assumption that the annotations are only for `WHERE IN` clauses.
* [ ] Make sure the core methods have javadoc examples analagous to what's in `@BindIn`, `@DefineIn` javadoc.

qualidafial:
Fixes #562

* [x] Add `SqlStatement.bindList(String attributeName, List<?> values)`. Make the annotation use this method directly.
* [x] Add `SqlStatement.defineList(String attributeName, List<?> values)`. Make the annotation use this method directly.
* [x] Rename `DefineIn` to `DefineList`, and `BindIn` to `BindList`-- to steer away from the assumption that the annotations are only for `WHERE IN` clauses.
* [x] Make sure the core methods have javadoc examples analagous to what's in `@BindIn`, `@DefineIn` javadoc.


core/src/main/java/org/jdbi/v3/core/SqlStatement.java
```diff
@@ -1174,6 +1177,62 @@ public final This bindBySqlType(int position, Object value, int sqlType)
...
+    /**
+     * Bind a parameter for each value in the given list, and defines an attribute as the comma-separated list of
+     * parameter references (using colon prefix).
+     *
+     * @param key    attribute name
+     * @param values values that will be comma-spliced into the defined attribute value.
+     * @return this
+     * @throws IllegalArgumentException if the list is empty.
+     */
+    public final This bindList(String key, List<?> values) {
```
stevenschlansker: Should this accept `Collection` or maybe even `Iterable`?  It looks like `@BindList` supports them.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/BindList.java
```diff
@@ -0,0 +1,178 @@
...
+     */
+    EmptyHandling onEmpty() default EmptyHandling.THROW;
+
+    final class Factory implements SqlStatementCustomizerFactory {
+        @Override
+        public SqlStatementCustomizer createForParameter(Annotation annotation, Class<?> sqlObjectType, Method method, Parameter param, Object arg) {
+            final BindList bindList = (BindList) annotation;
+            final String name = ParameterUtil.getParameterName(bindList, bindList.value(), param);
+
+            if (arg == null || Util.size(arg) == 0) {
```
stevenschlansker: Kinda crappy to potentially walk the entire list just to see if it's empty or not here.

core/src/main/java/org/jdbi/v3/core/SqlStatement.java
qualidafial: Sure. Also it probably makes sense to add an overloaded version with varargs.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizers/BindList.java
qualidafial: Good point. I'll add a `Util.isEmpty(Object)` method so this can short circuit.

PR #577, opened 2016-11-25
Expose Script.getStatements()

### LINKED ISSUE #575: Expose org.skife.jdbi.v2.Script.getStatements()
Is it possible to expose the org.skife.jdbi.v2.Script.getStatements() method (making it public)? It is so much useful, but currently private.

I have a requirement to parse a user supplied SQL script before further actions. I am expecting below code to do it
```
Script script = handle.createScript(scriptText);
List<String> statements = script.getStatements();   //  not doable now as the method is private
```
If the script contains both select statement and insert/delete/update statements, I need to throw an error telling user (on UI) this is not supported and the script must contain either single select statement, or a couple of insert/delete/update statements.

If the script contains single select statement, I need to execute it using below JDBI code, and pass the resultSet back to user (on UI).
```
Query<Map<String, Object>> query = handle.createQuery(scriptText);
List<Map<String, Object>> resultSet = query.list();
```
If the script contains only a couple of insert/delete/update statements, I need to execute the script using below JDBI code
```
Script script = handle.createScript(scriptText);
int[] returnValues = script.execute();
```
and display some information to user (on UI) like below (depending on the type of statement)
```
10 rows inserted
2 rows deleted
...
```
It would also be great if there is another way to achieve the same requirement without introducing much complexity or inelegance. Thanks in advance for any advice.

stevenschlansker:
Fixes #575

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,6 +1,7 @@
 2.78
   - @BindIn: fix handling of empty lists on Postgres
   - clear SqlObject ThreadLocals on close, fixes leak on e.g. webapp reload
+  - expose Script.getStatements()
 
 2.77
   - Improved BindIn functionality: can now process Iterables and arrays/varargs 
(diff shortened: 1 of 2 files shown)


qualidafial:
LGTM

PR #581, opened 2016-11-27
@SingleValue for prepared batch parameters

stevenschlansker:
Also increases use of `ReflectionArrayIterator` so you can bind e.g. `int[]` without hitting a nasty `ClassCastException`.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/BatchHandler.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/BatchHandler.java
@@ -16,7 +16,6 @@ package org.jdbi.v3.sqlobject;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/BatchHandler.java
```diff
@@ -193,20 +193,14 @@ public StatementContext getContext() {
...
-            else if (arg instanceof Iterator) {
-                extras.add((Iterator<?>) arg);
-                foundIterator = true;
-            }
-            else if (arg.getClass().isArray()) {
-                extras.add(Arrays.asList((Object[])arg).iterator());
+        for (int paramIdx = 0; paramIdx < method.getParameterCount(); paramIdx++) {
+            final boolean singleValue = method.getParameters()[paramIdx].isAnnotationPresent(SingleValue.class);
+            final Object arg = args[paramIdx];
+            if (ReflectionArrayIterator.isIterable(arg) && !singleValue) {
```
qualidafial: Meh: could swap these two conditions, so that `!singleValue` can short-circuit the `isIterable` call.

PR #583, opened 2016-11-27
remove TransactionStatus interface

### LINKED ISSUE #407: TransactionCallback interface exposes a bit too much
In the `TransactionCallback` interface, you get both a `Handle` as well as a `TransactionStatus` object.  `TransactionStatus.setRollbackOnly` is what you're meant to call if you want to roll back your transaction, not `Handle.rollback`.  The provision of both objects _as-is_ provides the possibility of doing it wrong without realizing it.  I would like to suggest that we at the very least provide some further method documentation (perhaps just on `Handle`?).  More than that, though, bifurcating the `Handle` interface might be the clearest.  There's the normal one you can get from an `IDBI`, and perhaps a more restricted one you get in your `TransactionCallback.inTransaction` call.

stevenschlansker:
per #407, its only feature, 'setRollbackOnly', just adds a different way
of doing what you can already do through other means.

Either call 'handle.rollback()' and do further cleanup yourself,
or throw an exception of your choice.  But there's no reason to
have a third option of "set a flag to have us throw an exception for you".

Bonus cleanup: JDBC has Savepoints.  We renamed them Checkpoints
for no apparent reason.  Let's ditch a vocab term.

Fixes #407
cc @pennello


core/src/main/java/org/jdbi/v3/core/transaction/TransactionHandler.java
```diff
@@ -28,55 +28,55 @@
 public interface TransactionHandler
 {
     /**
-     * Called when a transaction is started
+     * Called when a transaction is started.
```
qualidafial: The wording on this and other methods in the class seems to imply that the handler is only being _informed_ about the change on the transaction, when in reality the handler is _responsible_ for carrying out that change.

I think something like `Begins a transaction on the given handle` would be more accurate.

core/src/main/java/org/jdbi/v3/core/transaction/TransactionStatus.java
```diff
@@ -1,25 +0,0 @@
...
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-package org.jdbi.v3.core.transaction;
-
-/**
- * Allows rolling back a transaction in a TransactionCallback
- */
-public interface TransactionStatus
```
qualidafial: Goodnight, sweet prince

stevenschlansker:
Yeah, we'll put some examples in the transaction docs.

PR #584, opened 2016-11-27
Forbid most Transactional methods on OnDemand extensions. 

### LINKED ISSUE #416: Using Transactional with onDemand makes it easy to do nonsensical things
From the mailing list:

> For transaction support I've currently chosen to implement the Transactional interface and manually call dao.begin(), dao.commit() and dao.rollback() and my UserResource has a few transactional and a few non-transactional methods:

``` java
public interface UserDao implements Transactional<UserDao> { 
}

public class UserResource {

   public void insertUser(User user) {
      userDao.insertUser(user);
   }

   public void updateUser(User user) {
      try {
         userDao.begin();
         userDao.updateUser(user);
         userDao.commit(); 
      catch (Exception e) {
         userDao.rollback();
      }
   }
}
```

This usage doesn't work, but it's not clear to an outsider, and we should make it blow up or harder to misuse.

stevenschlansker:
Fixes #416

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
+++ b/core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
@@ -22,6 +22,7 @@ public class OnDemandExtensions {
     private static final Method EQUALS_METHOD;
     private static final Method HASHCODE_METHOD;
     private static final Method TOSTRING_METHOD;
+    private static final ThreadLocal<Boolean> IN_ON_DEMAND_EXTENSION = ThreadLocal.withInitial(() -> false);
 
     static {
         try {
@@ -30,7 +31,7 @@ public class OnDemandExtensions {
             TOSTRING_METHOD = Object.class.getMethod("toString");
         }
         catch (NoSuchMethodException wat) {
-            throw new IllegalStateException("JVM error", wat);
+            throw new ExceptionInInitializerError(wat);
         }
     }
 
@@ -56,10 +57,12 @@ public class OnDemandExtensions {
                 }
                 return dbi.withExtension(extensionType, extension -> {
                     threadExtension.set(extension);
+                    IN_ON_DEMAND_EXTENSION.set(true);
                     try {
                         return method.invoke(extension, args);
                     }
                     finally {
+                        IN_ON_DEMAND_EXTENSION.remove();
                         threadExtension.remove();
                     }
                 });
@@ -72,6 +75,11 @@ public class OnDemandExtensions {
         return extensionType.cast(
                 Proxy.newProxyInstance(
                         extensionType.getClassLoader(),
-                        new Class[]{extensionType}, handler));
+                        new Class<?>[] { extensionType }, handler));
+    }
+
+    public static boolean inOnDemandExtension()
+    {
+        return IN_ON_DEMAND_EXTENSION.get();
     }
 }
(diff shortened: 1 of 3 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/TransactionalHelper.java
```diff
@@ -26,13 +27,13 @@
...
+            h.put(Transactional.class.getMethod("begin"), guard(new BeginHandler()));
+            h.put(Transactional.class.getMethod("commit"), guard(new CommitHandler()));
+            h.put(Transactional.class.getMethod("rollback"), guard(new RollbackHandler()));
 
-            h.put(Transactional.class.getMethod("savepoint", String.class), new SavepointHandler());
-            h.put(Transactional.class.getMethod("releaseSavepoint", String.class), new ReleaseSavepointHandler());
-            h.put(Transactional.class.getMethod("rollbackToSavepoint", String.class), new RollbackSavepointHandler());
+            h.put(Transactional.class.getMethod("savepoint", String.class), guard(new SavepointHandler()));
+            h.put(Transactional.class.getMethod("releaseSavepoint", String.class), guard(new ReleaseSavepointHandler()));
+            h.put(Transactional.class.getMethod("rollbackToSavepoint", String.class), guard(new RollbackSavepointHandler()));
```
qualidafial: The `rollback`, `savepoint`, `releaseSavepoint`, and `rollbackToSavepoint` methods should all be valid within an `inTransaction` or `useTransaction` call.

core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
```diff
@@ -72,6 +75,11 @@
         return extensionType.cast(
                 Proxy.newProxyInstance(
                         extensionType.getClassLoader(),
-                        new Class[]{extensionType}, handler));
+                        new Class<?>[] { extensionType }, handler));
+    }
+
+    public static boolean inOnDemandExtension()
+    {
+        return IN_ON_DEMAND_EXTENSION.get();
```
qualidafial: I think we need something more closely scoped to the instance.

Example: if the on-demand instance has a default method that calls a `@CreateSqlObject` that creates a `Transactional` type, then calls transactional methods internally, that should be valid usage.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/mixins/TestTransactionalOnDemand.java
```diff
@@ -0,0 +1,75 @@
...
+        txn.rollbackToSavepoint("somewhere");
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void testReleaseSavepoint() throws Exception
+    {
+        txn.releaseSavepoint("somewhere");
+    }
+
+    public interface Txn extends Transactional<Txn>
```
qualidafial: I'd like to see tests showing that `inTransaction` and `useTransaction` still work as expected.

qualidafial:
A simpler solution might be to add a check in `LazyHandleSupplier.close()`: before closing the handle, check if the handle is in a transaction. If so, rollback the transaction, close the handle, and throw an exception with an informative message.

This should catch users who call `Handle.begin()` without an accompanying `commit()` or `rollback()`, as well as those who call `Transactional.begin()` on an on-demand SQL object.

qualidafial:
@stevenschlansker You mind if I take a stab at this? I have an approach in mind that should be pretty unobtrusive.

stevenschlansker:
Go for it.

qualidafial:
Opened PR #597

PR #585, opened 2016-11-27
Specializing row mappers

qualidafial:
Added method `RowMapper<T>.memoize(ResultSet, StatementContext): RowMapper<T>`.

This improves performance of mapping since the memoized mapper only scans the column names and performance column-to-(field/parameter/property) mapping and column mapper lookup once for the entire result set.

Audited all StatementExecutors, and all other callers of `RowMapper.map` to ensure that the row mapper is memoized once before parsing the result set.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/RowView.java
+++ b/core/src/main/java/org/jdbi/v3/core/RowView.java
@@ -16,8 +16,12 @@ package org.jdbi.v3.core;
 import java.lang.reflect.Type;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.jdbi.v3.core.exception.UnableToExecuteStatementException;
+import org.jdbi.v3.core.mapper.ColumnMapper;
+import org.jdbi.v3.core.mapper.RowMapper;
 import org.jdbi.v3.core.util.GenericType;
 
 /**
@@ -29,6 +33,9 @@ public class RowView
 {
     private final StatementContext ctx;
     private final ResultSet rs;
+
+    private final Map<Type, RowMapper<?>> rowMappers = new ConcurrentHashMap<>();
+    private final Map<Type, ColumnMapper<?>> columnMappers = new ConcurrentHashMap<>();
 
     RowView(ResultSet rs, StatementContext ctx)
     {
@@ -59,12 +66,23 @@ public class RowView
     public Object getRow(Type type)
     {
         try {
-            return ctx.findRowMapperFor(type)
-                    .orElseThrow(() -> new UnableToExecuteStatementException("No row mapper for " + type, ctx))
-                    .map(rs, ctx);
+            return rowMapperFor(type).map(rs, ctx);
         } catch (SQLException e) {
             throw new UnableToExecuteStatementException(e, ctx);
         }
+    }
+
+    private RowMapper<?> rowMapperFor(Type type) throws SQLException {
+        if (rowMappers.containsKey(type)) {
+            return rowMappers.get(type);
+        }
+
+        RowMapper<?> mapper = ctx.findRowMapperFor(type)
+                .orElseThrow(() -> new UnableToExecuteStatementException("No row mapper for " + type, ctx))
+                .memoize(rs, ctx);
+        rowMappers.put(type, mapper);
+
+        return mapper;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/jdbi/v3/core/RowView.java
```diff
@@ -59,14 +66,25 @@
...
-                    .orElseThrow(() -> new UnableToExecuteStatementException("No row mapper for " + type, ctx))
-                    .map(rs, ctx);
+            return rowMapperFor(type).map(rs, ctx);
         } catch (SQLException e) {
             throw new UnableToExecuteStatementException(e, ctx);
         }
     }
 
+    private RowMapper<?> rowMapperFor(Type type) throws SQLException {
+        if (rowMappers.containsKey(type)) {
```
stevenschlansker: Any reason not to use `computeIfAbsent` here?

core/src/main/java/org/jdbi/v3/core/mapper/RowMapper.java
```diff
@@ -30,10 +30,24 @@
...
+     * Returns a row mapper instance with the details of the result set memoized for performance. Overriding this
+     * method is optional--the default implementation simply returns {@code this}.
+     *
+     * @param rs  the result set to memoize over
+     * @param ctx the statement context to memoize over
+     * @return a row mapper equivalent to this one, possibly memoized.
+     * @throws SQLException if anything goes wrong go ahead and let this percolate, jDBI will handle it
+     * @see org.jdbi.v3.core.mapper.reflect.BeanMapper for an example of memoization.
+     */
+    default RowMapper<T> memoize(ResultSet rs, StatementContext ctx) throws SQLException {
```
stevenschlansker: I'm not sure if `memoize` is the right name here.  Usually it specifically refers to a pure function where we remember the return value, which isn't quite what's going on here.  Maybe `specialize` or `beginMapping` or something like that?

stevenschlansker: Also, did you consider e.g. making the `RowMapperFactory` take the `ResultSet` on its factory method?  It'd probably be a bigger change, but it would remove the odd "return this" default implementation, and the need for many `map(...) { return memoize(...).map(...); }` overrides.

qualidafial: I originally did it as `specialize` and renamed it to `memoize`, but you're right that it's not the right name for this pattern. I'll rename it back.

qualidafial: So it feels like we want an additional type between the factory and the mapper. I hope to find a pattern that doesn't burden users with the specializing aspect if they don't need it.

One thought is to have `RowMapper` have a single method `bindTo(ResultSet,StatementContext): RowSupplier<T>`, where RowSupplier is like Supplier but can throw SQLException.

qualidafial: I don't think we can, or should, pull this up to the factories. There's an implicit order that we find out what the user wants:
* What type they want to map to (before the query executes)
* What columns are available to map (after the query executes)

Existing methods like `Query.map(RowMapper)` are called before the query executes, so unless we want to take away that convenience, this is a nonstarter.

I did a spike on my previous comment, (see the `jdbi3-row-suppliers` branch) and ended up with this:

```java
interface RowMapperFactory<T> {
  Optional<RowMapper<T>> build(Type type, ConfigRegistry config);
}

interface RowMapper<T> {
  RowSupplier<T> bind(ResultSet rs, StatementContext ctx) throws SQLException;
}

interface RowSupplier<T> {
  T get() throws SQLException;
}
```

This approach meant I had to convert lambda mappers in tests to higher order e.g. `(rs, ctx) -> rs.getString("foo")`  became `(rs, ctx) -> () -> rs.getString("foo")`. In one case a test broke without any compiler errors, because the existing lambda returned null. That's a papercut I'd rather not introduce.

So my next idea will be to revert `RowMapper` change, rename it, and and introduce a new, parallel mapping pipeline. The new pipeline will be the "official" pipeline and the old mapper will be adapted transparently to work in the new pipeline:

```java
interface RowBinderFactory<T> {
  Optional<RowBinder<T>> build(Type type, ConfigRegistry config);
}

interface RowBinder<T> {
  RowSupplier<T> bind(ResultSet rs, StatementContext ctx) throws SQLException;
}

interface RowSupplier<T> {
  T get() throws SQLException;
}

interface RowMapper<T> { // adapted into RowBinder when registered
  T map(ResultSet rs, StatementContext ctx) throws SQLException;
}
```

These new class names are just working names--I'm open to suggestions.

One possible drawback is that you can mix up `Query.registerRowBinder` and `registerRowMapper` and use a lambda, it's possible the compiler won't catch it because the two types take the same arguments.

Thoughts?

stevenschlansker: Thanks for looking into this, although all that said I'm not sure that adding a parallel set of interfaces is actually any simpler.  I feel pretty strongly that we should have one set of interfaces that we support for this.  If that means we have to go with the existing "specialize" approach, I'd probably prefer that to allowing two totally separate ways of doing it.

qualidafial: I was about to say pretty much the same thing.

I feel the performance aspect is important enough that we need to provide something. But it's not important enough to warrant making JDBI less usable for the other 99% of cases.

I'll merge this after doing one last once-over.

PR #587, opened 2016-11-30
Move Oracle-specific stuff to jdbi3-oracle module.

qualidafial:
* [x] Hide oracle module behind a profile so end-users can build JDBI out of the box, without having to set up Oracle DB in a VM, or set up the Oracle maven repository.
* [x] Set up `~/.m2/settings.xml` in a secure file in TravisCI so build can talk to maven.oracle.com.
* ~See if there's any way to build JDBI with a live Oracle database in TravisCI. Non-critical but would be nice.~ moving this to a separate issue
* [x] Write guide on setting up for builds and unit tests with Oracle
* [x] Write guide on how to create `travis-settings.xml` and `travis-settings-security.xml` from your own `~/.m2/settings[-security].xml` files.


qualidafial:
@stevenschlansker @arteam This is ready for review.

.travis.yml
```diff
@@ -12,20 +12,22 @@ cache:
...
       jdk: oraclejdk8
     - env:
         - DOCUMENTATION=Y
-      script: mvn -DskipTests=true -Dbasepom.check.skip-all=true -Ptoolchains -B verify
+      script: mvn -DskipTests=true -Dbasepom.check.skip-all=true -Ptoolchains -Poracle -B verify
       jdk: oraclejdk8
     - env:
         - CODE_STYLE=Y
-      script: mvn -DskipTests=true -Dmaven.javadoc.skip=true -Ptoolchains -B verify
+      script: mvn -DskipTests=true -Dmaven.javadoc.skip=true -Ptoolchains -Poracle -B verify
```
stevenschlansker: I believe triggering a profiles with e.g. `-Poracle` will *suppress* all other profiles activation (i.e. it specifies the list of active profiles, rather than solely triggering activation)
Are you sure that doesn't accidentally nuke anything else?  Also, how does it get activated during a release build?

docs/pom.xml
```diff
@@ -175,6 +175,15 @@
 
     <profiles>
         <profile>
+          <id>oracle</id>
```
stevenschlansker: This spacing isn't consistent with the rest of the file

oracle/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+            <groupId>com.oracle.jdbc</groupId>
+            <artifactId>ojdbc7</artifactId>
+            <version>12.1.0.2</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <optional>true</optional> <!-- for @Beta annotation -->
```
stevenschlansker: If we're going to do this, should we just have our own `@Beta` annotation rather than leave an orphaned Guava one in?

oracle/src/main/java/org/jdbi/v3/oracle/OracleReturning.java
```diff
@@ -56,34 +58,33 @@
...
      * {@link BeanMapper} uses the names in the result set
      *
      * @param mapper Must use only positional access to the result set
      */
-    public OracleReturning(RowMapper<ResultType> mapper)
-    {
+    public OracleReturning(RowMapper<ResultType> mapper) {
         this.mapper = mapper;
         try {
             this.oraclePS = Class.forName("oracle.jdbc.OraclePreparedStatement");
```
stevenschlansker: Do we still need to do all this reflective bullshit now that it's in a separate artifact?

oracle/src/main/java/org/jdbi/v3/oracle/OracleReturning.java
```diff
@@ -102,37 +102,31 @@ public void beforeExecution(PreparedStatement stmt, StatementContext ctx) throws
...
+                } else {
+                    this.stmt = candidate;
                 }
-                else {
-                    this.stmt = candidate ;
-                }
-            }
-            catch (Exception e) {
+            } catch (Exception e) {
                 throw new IllegalStateException("Statement is not an OraclePreparedStatement, nor" +
```
stevenschlansker: "norone" is my new favorite word

.travis.yml
qualidafial: Per http://maven.apache.org/guides/introduction/introduction-to-profiles.html, you can define multiple profiles and mix and match them as needed. My understanding is that a profile with an activation rule will obey that rule unless that specific profile is overridden on the command line.

With that said, it might make sense to introduce some build scripts at the project root, to take out the guesswork:
* Snapshot build (excluding oracle) - usable by end users who don't want to install an Oracle VM.
* Snapshot build (including oracle)
* Release build (including oracle) - do the extra bookkeeping and publish new documentation.

oracle/pom.xml
qualidafial: Eh, we can just take it out. There's already a commend in that class's javadoc that says it's beta.

qualidafial:
That's a good point--you're going to need to set up an Oracle SSO account in your `settings.xml` and `settings-security.xml` to deploy releases.

I will add a README with instructions on how this is done.

arteam:
Looks good to me. Nice to see progress in the direction of supporting Oracle-specific features.

oracle/src/main/java/org/jdbi/v3/oracle/OracleReturning.java
```diff
@@ -56,83 +58,65 @@
...
-        if (!oraclePS.isAssignableFrom(stmt.getClass())) {
+        if (stmt instanceof OraclePreparedStatement) {
+            this.stmt = OraclePreparedStatement.class.cast(stmt);
+        } else {
             try {
-                final Method get_delegate = stmt.getClass().getMethod("getDelegate");
-                final Object candidate = get_delegate.invoke(stmt);
-                if (!oraclePS.isAssignableFrom(candidate.getClass())) {
+                // If it's an instance of Apache commons-dbcp DelegatingPreparedStatement
+                final Method getDelegate = stmt.getClass().getMethod("getDelegate");
```
stevenschlansker: Does commons-dbcp do the `java.sql.Wrapper` interface?  If so, that might be a more portable way to do this.

qualidafial: `Statement` itself extends `Wrapper`, and dbcp appears to implement it right, so.. yes.

qualidafial: In fact we can just call `stmt.unwrap(OraclePreparedStatement.class)` instead of all this instanceof malarkey.

oracle/src/main/java/org/jdbi/v3/oracle/OracleReturning.java
```diff
@@ -56,83 +57,52 @@
...
-                throw new IllegalStateException("Statement is not an OraclePreparedStatement, nor" +
-                                                "one which we know how to find it from", e);
-            }
-        }
-        else {
-            this.stmt = stmt ;
+        if (!stmt.isWrapperFor(OraclePreparedStatement.class)) {
+            throw new IllegalStateException("Statement is not an instance of, nor a wrapper of, OraclePreparedStatement");
         }
+        this.stmt = stmt.unwrap(OraclePreparedStatement.class);
```
stevenschlansker: Is it safe to stash the statement in an instance variable?  That seems to assume that this customizer will never be used in a reentrant or concurrent fashion?

qualidafial: I was thinking the same thing, but it's implied in the class's API. The `getReturnedResults` and `registerReturnParam` method imply one-off usage. It is a bit weird though.

qualidafial: It's worth noting the reason this class existed in the first place is because `OraclePreparedStatement` doesn't play nice with `getX(String,...)` methods. It wants the column number.

Since the reflective mappers are now all using column index, this class might not have a reason to exist any more. But we need tests to validate that assumption.

qualidafial: This use case for `OracleReturning`, and the request for supporting multiple result sets, has me wondering whether `StatementExecutor` ought to take a `PreparedStatement` (post execution) in lieu of the `RowMapper` and `ResultSet`:

```java
@FunctionalInterface
public interface StatementExecutor<R>
{
    R execute(PreparedStatement stmt, StatementContext ctx) throws SQLException;
}
```

Although I'm realizing now that `StatementExecutor` is a bit of a misnomer: the interface isn't executing the statement--it has already executed by this point.

Since `StatementExecutor` no longer receives the mapper, and `ResultBearing` has no access to it, then the current `ResultBearing.iterator()` method cannot be default--but we can just push it down to `Query`. `GeneratedKeys` (the only other implementation) already overrides `iterator()`.

This also suggests to me that we might want to extract a super-interface of `ResultBearing` that has the `execute(StatementExecutor)` and possibly `getContext()` methods. Thus `ResultBearing` would continue to operate for everything iterable (single result set), vs `StatementExecuting` (bleh) which only implies that the answer came from that `PreparedStatement` in some way.

This would allow us to bring other `SqlStatement` classes (`Update`, `Call`, probably not `Batch`) into the same fold as `Query`. We provide a default set of the most useful result strategies, and other folks can introduce their own strategies for esoteric scenarios.

It also allows third-party solutions for extracting results in esoteric scenarios, e.g. multiple result sets.

qualidafial: Holy shit, this could annihilate the need for `Query`/`Update` separation.

qualidafial:
I've added a `CONTRIBUTING.md` file outlining how to set up one's workstation to talk to the Oracle Maven repository, and run the Oracle unit tests locally. It's in the diff, but you can see the markdown rendered [here](https://github.com/jdbi/jdbi/blob/1b3ed6c5e2833eb743ba48fb6b312e7d254e9e89/CONTRIBUTING.md)

qualidafial:
I've added a README.md in `src/build` outlining how to generate the encrypted files for Travis builds. The markdown is in `src/build/README.md`, but you can read it rendered [here](https://github.com/jdbi/jdbi/blob/e43f22942c379432e15bef71aceed9ccee4ef1db/src/build/README.md)

@stevenschlansker Does this address your concerns about the encrypted files?

qualidafial:
Merging this. @stevenschlansker if you have any further feedback, let's deal with that as a follow up ticket.

Note that this merge means you'll have to rewrite OracleReturning class a bit in #595. Possibly it could be made into a `ResultProducer` instead of a statement customizer?

stevenschlansker:
Yup, sorry I didn't reply, missed the notification.

Issue #592, opened 2016-12-02
Refactor StatementExecutor to accept PreparedStatement instead of mapper and result set


qualidafial:
Between work on #587 (`OracleReturning`), and the request for supporting multiple result sets, these have me thinking that `StatementExecutor` ought to take a `PreparedStatement` (post execution) in lieu of the `RowMapper` and `ResultSet`:

```java
@FunctionalInterface
public interface StatementExecutor<R>
{
    R execute(PreparedStatement stmt, StatementContext ctx) throws SQLException;
}
```

`StatementExecutor` is itself a misnomer: the interface isn't executing the statement--it has already executed by this point. We should rename it to e.g. `ResultReturner` with method name `returnResult`, or something like that.

This gives a uniform execution plan between `Update`, `Query`, `Call`, and possibly even `Batch`. Each class would declare their own methods that accepted `ResultReturner`, but their return types could vary as appropriate for the class. For example, `Batch` might have `execute(ResultReturner<R>): List<R>`, with one result for each statement in the batch. `Update`, `Query`, and `Call` would probably all have `execute(ResultReturner<R>): R`.

Since this allows third-party result returners, this finally gives end users the option to pull results out of a statement using whatever pants-on-head bonkers database feature floats their boat. :trollface: 

By way of refactoring:

Since `StatementExecutor` would no longer receive the mapper, and `ResultBearing` has no access to it, then the current `ResultBearing.iterator()` method cannot be default--but we can just push it down to `Query`. `GeneratedKeys` (the only other implementation) already overrides `iterator()`.

This also suggests that we should extract a sub-interface of `ResultBearing` called `ResultIterable`. `ResultBearing` would retain the `execute(StatementExecutor)` and `getContext()` methods, and `ResultIterable` would get the rest.

Thus `ResultIterable` would continue to operate for everything iterable (`Query`, `GeneratedKeys`, ..`OracleReturning`?).  Whereas `ResultBearing` would be for classes like `Update` and `Call`, which only implies that the result came from the executed `PreparedStatement` in some way.

`Query` and `GeneratedKeys` would implement `ResultIterator`, and `Update` and `Call` could be updated to implement `ResultBearing`.

This is a relatively small change that I think will give us better longevity in 3.x for supporting new use cases as we learn about them.

qualidafial:
Tracked in #595

PR #595, opened 2016-12-03
StatementExecutor -> ResultProducer

stevenschlansker:
First stab at #592


stevenschlansker:
Is this along the lines of what you are thinking?  I didn't refactor `Call` or `Update` yet.

core/src/main/java/org/jdbi/v3/core/ResultProducer.java
```diff
@@ -13,25 +13,24 @@
...
-     * @param mapper the mapper the user's code told you to use
-     * @param rs the result set
+     * @param stmt the prepared statement
+     * @param rs the result set to iterate
      * @param ctx the statement context
      * @return an object of the type your caller expects
      * @throws SQLException sadness
      */
-    R execute(RowMapper<T> mapper, ResultSet rs, StatementContext ctx) throws SQLException;
+    R produce(PreparedStatement stmt, ResultSet rs, StatementContext ctx) throws SQLException;
```
qualidafial: Intention here was to remove the `ResultSet` parameter. Result producers can just call `getResultSet` on the statement any way.

core/src/main/java/org/jdbi/v3/core/ResultProducer.java
```diff
@@ -13,25 +13,24 @@
...
 
+import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 
-import org.jdbi.v3.core.mapper.RowMapper;
-
 /**
- * A StatementExecutor generates the mapped results of a {@link SqlStatement},
+ * A {@code ResultProducer} generates the mapped results of a {@link SqlStatement},
```
qualidafial: Result isn't necessarily mapped--e.g. could be update count.

qualidafial:
I'm good with merging this as is, and coming back later for the Call and Batch refactors.

stevenschlansker:
Slow down cowboy!  I want to see if removing the ResultSet parameter is helpful first, then we can merge :)

core/src/main/java/org/jdbi/v3/core/ResultProducer.java
stevenschlansker: Unless it's generated keys... this is turning out to be harder to remove than I thought.

qualidafial: Where I see this going long term: provide a class `ResultProducers` with static factory methods for all kinds of result producers:

```java
stmt.execute(returningMappedResults(Contact.class))
stmt.execute(returningMappedGeneratedKeys(Contact.class))
stmt.execute(returningOracleDml(Contact.class))
stmt.execute(returningUpdateCount())
```

etc

qualidafial: I was also thinking about making `ResultIterable` its own type, and rather than a subtype of `ResultBearing`. That way result producers like e.g. `returningMappedResults` could return a  `ResultIterable`.

Thus:

```java
stmt.execute(returningMappedResults(Contact.class)).findFirst()
stmt.execute(returningMappedResults(Contact.class)).list()
stmt.execute(returningMappedResults(Contact.class)).collect(toSet())
```

Also here's how I could see multiple return result sets working with this:

```
stmt.execute(returningMultipleResultSets(Contact.class, Account.class, Invoice.class))
    .listOfLists();

stmt.execute(
  returningMappedResults(Contact.class).into(contactConsumer),
  returningMappedResults(Account.class).into(accountConsumer),
  returningMappedResults(Invoice.class).into(invoiceConsumer))
```

These are just musings, but I hope they give you an idea of the motivation, and where this could go.

qualidafial:
@stevenschlansker I have some bandwidth to help on this if you'd like.

qualidafial:
I attempted a spike to change `OracleReturning` into a `ResultProducer<ResultIterable<T>>`. This would make usage much more fluent:

```java
RowMapper<Thing> mapper = ...;
handle.createStatement(
          "insert into something (id, name) values (17, 'Brian') returning id into ?")
      .execute(returningOracleDml(mapper)
                   .registerReturnParam(1, OracleTypes.INTEGER))
      // execute() returned ResultIterable<Thing>
      .stream() // or any other ResultIterable method
```

One thing that's in the way is that `OracleReturning` wants access to the prepared statement _before_ it executes, in order to register return parameters. So this would require another change to the execution plan, where the result producer is actually responsible for executing the statement.

Which.. means that `StatementExecutor` would be an apt name again.. ¯\\\_(ツ)\_/¯

I've also had a change of heart on making `Batch` implement `ResultIterable`--the return type doesn't fit. We can outfit Batch with a method that _accepts_ a `ResultProducer` but the return type would be an array, iterator, iterable, or stream of the produced results.

stevenschlansker:
Do you think we should merge this as-is and continue hacking on it?  Or would you like to take another pass at it?  I kept fiddling for another hour or two and didn't get anything I really liked.  But I don't think we are done yet either.

qualidafial:
First I'd like to merge #609 (remove `ResultProducer` type parameter `<T>` and `ResultSet` method parameter) into this PR branch, but after that I think we're good to merge.

qualidafial:
For parity with `ResultIterator`, I think we could push `ResultBearing.getContext()` down to `ResultIterable`.

qualidafial:
> One thing that's in the way is that OracleReturning wants access to the prepared statement before it executes, in order to register return parameters. So this would require another change to the execution plan, where the result producer is actually responsible for executing the statement.

Ha. Yeah I'm retracting that comment now. We've already got `StatementCustomizer` for this.

So I think for the `OracleReturning` case, it would probably be best to split it into a `StatementCustomizer` that registers return parameters, and a `ResultProducer` that produces a `ResultIterable` from `OraclePreparedStatement.getReturnResultSet()`.

I'll open a separate ticket for it.

PR #597, opened 2016-12-04
Throw TransactionException if handle is closed while in a transaction.

### LINKED ISSUE #416: Using Transactional with onDemand makes it easy to do nonsensical things

qualidafial:
Fixes #416

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -22,6 +22,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
+import org.jdbi.v3.core.exception.TransactionException;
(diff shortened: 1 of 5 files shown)


qualidafial:
@stevenschlansker @arteam This is ready for review

PR #612, opened 2016-12-10
{Bean,Field}Mapper: throw if no columns match, add strict validation

### LINKED ISSUE #417: Extend BeanMapper to better verify bean properties are set
If you typo, BeanMapper silently returns you a null bean.
We can do better -- include two new options, `ignoreUnknownProperties` and `requireAllFieldsSet` that blow up if we have extra / missing fields, respectively?

stevenschlansker:
fixes #417


core/src/main/java/org/jdbi/v3/core/mapper/RowMapperFactory.java
```diff
@@ -33,4 +33,19 @@
...
+     * Create a RowMapperFactory from a given {@link RowMapper} that matches
+     * a {@link Type} exactly.
+     *
+     * @param type the type to match with ==
+     * @param the mapper to return
+     *
+     * @return the factory
+     */
+    public static RowMapperFactory of(Type type, RowMapper<?> mapper) {
+        return (t, ctx) -> t == type
```
qualidafial: Better to use `t.equals(type)` so parameterized types can satisfy test.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/FieldMapper.java
```diff
@@ -125,6 +119,19 @@ public T map(ResultSet rs, StatementContext ctx) throws SQLException {
             fields.add(field);
         }
 
+        if (columnNumbers.isEmpty() && metadata.getColumnCount() > 0) {
+            throw new IllegalArgumentException(String.format("Mapping bean type %s " +
```
qualidafial: Maybe "Mapping fields for type %s" instead of "bean"?

core/src/main/java/org/jdbi/v3/core/mapper/RowMapperFactory.java
```diff
@@ -33,4 +33,19 @@
...
+    /**
+     * Create a RowMapperFactory from a given {@link RowMapper} that matches
+     * a {@link Type} exactly.
+     *
+     * @param type the type to match with ==
+     * @param the mapper to return
+     *
+     * @return the factory
+     */
+    public static RowMapperFactory of(Type type, RowMapper<?> mapper) {
```
qualidafial: This seems so obvious in retrospect.

Could we add this to `ColumnMapperFactory` as well, and add convenience methods registration methods for this pattern to `Configurable`, `RowMappers`, and `ColumnMappers`?

PR #613, opened 2016-12-10
Batch and PreparedBatch: use statement builder.

### LINKED ISSUE #178: StatementBuilder isn't used by PreparedBatch and Batch.
Hi, as title says, org.skife.jdbi.v2.PreparedBatch and org.skife.jdbi.v2.Batch do not use the StatementBuilder to create the statements, is it done by design or is it a defect?

I'm opening this issue because we implemented a custom statement builder to change flags on the  java.sql.Connection of the statement before creating it, but in case of batch statements those flags are not set since the builder is not used.

stevenschlansker:
Fixes #178


stevenschlansker:
actually, one sec, I can do better

stevenschlansker:
ok this one should be good to go

PR #616, opened 2016-12-13
Refactor fluent pipeline: ResultBearing -> ResultSetIterable -> ResultIterable<T>

### LINKED ISSUE #610: Split OracleReturning into a StatementCustomizer and a ResultProducer
Depends on #595 getting merged.

OracleReturning usage is a little awkward.

I think we should split it into two types:
* A `StatementCustomizer` that registers return parameters through `OraclePreparedStatement.registerReturnParameter()`
* A `ResultProducer<ResultIterable<T>>` that produces a result from the `OraclePreparedStatement.getReturnResultSet()` result set.

```java
RowMapper<Thing> mapper = ...;
handle.createStatement(
          "insert into something (id, name) values (17, 'Brian') returning id into ?")
      .addCustomizer(OracleReturnParameters.registerReturnParam(1, OracleTypes.INTEGER))
      .execute(OracleReturningDml.returningDml(mapper)) // => ResultIterable<Thing>
      ... // chain calls to ResultIterable methods e.g. stream(), findFirst(), etc
```

### LINKED ISSUE #614: ResultIterable wants to be distinct from ResultBearing
This is informed by my work on #610, trying to write a `ResultProducer` for `OracleReturning`.

I think `ResultIterable` would be a great standalone return type for `ResultProducer`, but the fact that it extends `ResultBearing` means that it must implement the `execute(ResultProducer)` method, which seems off.

Most of `ResultIterable`'s methods (iterator, findOnly, findFirst, stream, useStream, withStream, list, collect, reduce) imply that some kind mapping is taking place, but not necessarily with a row mapper.

There are also two methods `reduceRows` and `reduceResultSet` that do not imply a mapper, but _do_ imply iteration of a `ResultSet`.

I think we might want to split these interfaces out and make them all standalone: `ResultBearing`, `ResultSetIterable`, and `ResultIterable` (which would no longer extend `ResultBearing`). Classes could implement any combination of them as appropriate.

```java
ResultBearing {
  <U> U execute(ResultProducer<U> producer);
}

ResultSetIterable {
  <U> U reduceRows(U seed, BiFunction<U, RowView, U> accumulator);

  <U> U reduceResultSet(U seed, ResultSetAccumulator<U> accumulator);
}

ResultIterable<T> {
  // all the remaining methods
}
```

I might also look into whether we really need that `getContext()` method. It seems out of place.

qualidafial:
Refactor for ResultBearing(ResultProducer) -> ResultSetIterable(mapper) -> ResultIterable<T> fluent API pipeline

Fixes #614 (ResultIterable wants to be distinct from ResultBearing)

This PR further develops the ResultBearing / ResultProducer<R> execution stategy.

* ResultProducer's `PreparedStatement` parameter changed to `Supplier<PreparedStatement>` so chained calls with lazy intermediate types like `ResultSetIterable` or `ResultIterable` can defer executing the statement until the "terminal" method call--like streams.
* ResultProducers class with static factory methods
  * returningUpdateCount() produces PreparedStatement.getUpdateCount()
  * returningResults() produces ResultSetIterable over PreparedStatement.getResultSet()
  * returningGeneratedKeys(String...) produces ResultSetIterable over PreparedStatement.getGeneratedKeys()
* ResultSetIterable is a decorator over JDBC ResultSet with convenience methods:
  * withResultSet() passes the result set to the callback and returns the callback result, while masking SQLException. This is the only abstract method--all others are written in terms of this method.
  * map(), mapTo(), mapToBean(), and mapToMap() return a `ResultIterable<T>`
  * reduceRows(), and reduceResultSet() were moved to this interface, and still have the same semantics
  * static method `of(Supplier<ResultSet>, StatementContext)` produces a ResultSetIterable over the given result set and context.
* `ResultIterable<T>`:
  * Removed the reduceRows(), reduceResultSet() methods. (They're in ResultSetIterable now)
  * static method `of(Supplier<ResultSet>, RowMapper<T>, StatementContext)` produces a ResultIterable over the given result set.
  * static method `of(ResultIterator<T>)` produces a ResultIterable over the given iterator.
* Query now implements ResultBearing and ResultSetIterable, but not `ResultIterable<T>`. This means:
  * There is no generic `<T>` parameter on Query
  * There is no default mapper
  * Methods inherited from ResultSetIterable must be called last, after methods like e.g. setFetchSize(). In practice this means mapTo(type) must always be the second-to-last chained call, right before e.g. .list()
* SQL object methods can now return a `ResultIterable<T>` instead of `Query<T>`.
* Renamed DefaultMapper to MapMapper.
* Changed Handle.select(String,Obj[]) to return Query. I kinda like this: handle.select("select * from contact where id = ?", 1).mapTo(Contact.class).list()
  * Maybe we could collapse Handle.select() and Handle.createQuery() into a single method
* Removed `GeneratedKeys<T>` in favor of `ResultIterable<T>`. Methods that used to return `GeneratedKeys<T>` now return `ResultIterable<T>`.
* PreparedBatch now implements ResultBearing and ResultSetIterable.
  * Existing executeAndGenerateKeys methods were refactored to use the new fluent pipeline, e.g. execute(returningGeneratedKeys()).mapTo(keyType)
* Refactored OracleReturning into static factory methods that produce 1) a StatementCustomizer (that registers the return parameters) and 2) a ResultProducer (which produces a ResultSetIterable for the return result set). (Fixes #610)
  * Now with tests!
* Update implements ResultBearing
  * executeAndReturnGeneratedKeys(mapper, ...columns) now implemented as execute(returningGeneratedKeys(columns).map(mapper), so these are just convenience methods for public API now, instead of being magic.

Cosmetic changes:
* Renamed Statement.addStatementCustomizer() to addCustomizer(), and pulled it up to BaseStatement
* Made BaseStatement.addCleanable() public
* ConfigRegistry: Added non-wildcard findColumnMapperFor() and findRowMapperFor() methods that accept Class<T> or GenericType<T> and return a mapper for that type.
* ColumnMappers and RowMappers: added non-wildcard findFor() methods.
* StatementContext: added non-wildcard findColumnMapperFor(), findRowMapperFor() methods.

~An outstanding concern is that resource disposal gets a little trickier. At the moment, a ResultSet is opened in the execute(returningResultSet()) method call, and not until e.g. .mapTo(type).list() does the result set get closed.~

* ~One option is to make ResultBearing produce a lazy result, which would only execute the statement on a "terminal" operation (like how JDK8 streams work).~
* ~Another is to sandbox ResultBearing.execute(ResultProducer), e.g. by adding a ResultCallback callback, which would receive the result value and be able to perform further transformations inside the callback, and then we can clean up resource when the callback returns.~


qualidafial:
I have an idea for controlling resource leaks: make `ResultProducer` accept a `Supplier<PreparedStatement>` instead of `PreparedStatement`. Pass that laziness down the line to `ResultSetIterable` and `ResultIterable`. That way, only the final call to e.g. `ResultIterable.list()` actually causes the statement to be executed, not the middle call to e.g. `execute(returningResults())`.

qualidafial:
ResultIterable should override Iterable.forEach to ensure that the iterator is explicitly closed.

qualidafial:
@stevenschlansker @arteam This is ready for review, when you get some time.

It's a bit of a monster, but I think this will be so, so awesome.

qualidafial:
@stevenschlansker @arteam any chance I could get a review on this?

arteam:
Wow, that's a quite big PR! Let me go trough.

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -0,0 +1,84 @@
...
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+
+import org.jdbi.v3.core.StatementContext;
+
+/**
+ * Yo dawg, I heard you like maps, so I made you a mapper that maps rows into {@code Map&lt;String,Object&gt;}. Map
```
arteam: A+ comment, would read again.

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Override
+    public RowMapper<Map<String, Object>> specialize(ResultSet rs, StatementContext ctx) throws SQLException {
+        ResultSetMetaData m = rs.getMetaData();
+        int columnCount = m.getColumnCount();
+        String[] columnNames = new String[columnCount+1];
+
+        for (int i = 1; i <= columnCount; i++) {
+            String key = m.getColumnName(i);
```
arteam: If `SQLException` was unchecked, that would be great.

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -0,0 +1,84 @@
...
+            }
+            if (foldCase) {
+                alias = alias.toLowerCase(Locale.ROOT);
+            }
+
+            columnNames[i] = alias;
+        }
+
+        return (r, c) -> {
+            Map<String, Object> row = new HashMap<>();
```
arteam: We could we create a map with a proper size, we know how many elements we will insert to it.

core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
```diff
@@ -87,6 +88,32 @@ public ColumnMappers register(ColumnMapperFactory factory) {
      * @param config the configuration registry
      * @return a ColumnMapper for the given type, or empty if no column mapper is registered for the given type.
      */
+    @SuppressWarnings("unchecked")
+    public <T> Optional<ColumnMapper<T>> findFor(Class<T> type, ConfigRegistry config) {
+        ColumnMapper<T> mapper = (ColumnMapper<T>) findFor((Type) type, config).orElse(null);
```
qualidafial: Yeah, the type inference breaks down as soon as you have a nested wildcard.

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
qualidafial: If only.

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
qualidafial: Done.

core/src/main/java/org/jdbi/v3/core/ResultProducers.java
```diff
@@ -0,0 +1,107 @@
...
+     */
+    public static ResultProducer<ResultSetIterable> returningResults() {
+        return (supplier, ctx) -> ResultSetIterable.of(getResultSet(supplier, ctx), ctx);
+    }
+
+    private static Supplier<ResultSet> getResultSet(Supplier<PreparedStatement> supplier, StatementContext ctx) {
+        return () -> {
+            try {
+                ResultSet rs = supplier.get().getResultSet();
+                if (rs != null) {
```
stevenschlansker: Should it be an error to call this method if you can't actually get the result set?  Feels like it strongly indicates misuse / coding error.

core/src/main/java/org/jdbi/v3/core/ResultProducers.java
```diff
@@ -0,0 +1,107 @@
...
+     * Result producer that returns a {@link ResultSetIterable} over the statement-generated keys.
+     *
+     * @param generatedKeyColumnNames optional list of generated key column names.
+     * @return ResultSetIterable of generated keys
+     * @see PreparedStatement#getGeneratedKeys()
+     */
+    public static ResultProducer<ResultSetIterable> returningGeneratedKeys(String... generatedKeyColumnNames) {
+        return (supplier, ctx) -> {
+            String[] columnNames = Stream.of(generatedKeyColumnNames)
+                    .filter(name -> name != null && !name.isEmpty())
```
stevenschlansker: This filter seems surprising to me.  If nothing else, it means that result set column numbers no longer align with the key column names array index.  Do people actually pass null or empty strings?  I would just as soon forbid it and throw an exception.

core/src/main/java/org/jdbi/v3/core/ResultSetIterable.java
```diff
@@ -0,0 +1,212 @@
...
+import org.jdbi.v3.core.mapper.RowMapper;
+import org.jdbi.v3.core.mapper.RowMapperFactory;
+import org.jdbi.v3.core.mapper.SingleColumnMapper;
+import org.jdbi.v3.core.mapper.reflect.BeanMapper;
+import org.jdbi.v3.core.util.GenericType;
+
+/**
+ * Provides access to the contents of a {@link ResultSet}
+ */
+public interface ResultSetIterable {
```
stevenschlansker: Feels odd to have a `ResultSetIterable` that does not `implements Iterable` -- maybe the name should change slightly?

core/src/main/java/org/jdbi/v3/core/ResultSetIterable.java
```diff
@@ -0,0 +1,212 @@
...
+        };
+    }
+
+    /**
+     * Invokes the callback with the result set, and returns the value returned by the callback.
+     * @param callback the callback
+     * @param <R> result type returned by the callback.
+     * @return the value returned by the callback.
+     */
+    <R> R withResultSet(ResultSetCallback<R> callback);
```
stevenschlansker: Not sure about the name here, mentioned it above.

core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
```diff
@@ -87,6 +88,32 @@ public ColumnMappers register(ColumnMapperFactory factory) {
...
+    /**
+     * Obtain a column mapper for the given type.
+     *
+     * @param type the target type to map to
+     * @param config the configuration registry
+     * @return a ColumnMapper for the given type, or empty if no column mapper is registered for the given type.
+     */
+    @SuppressWarnings("unchecked")
+    public <T> Optional<ColumnMapper<T>> findFor(GenericType<T> type, ConfigRegistry config) {
+        ColumnMapper<T> mapper = (ColumnMapper<T>) findFor(type.getType(), config).orElse(null);
```
stevenschlansker: Is this `ofNullable(x.orElse(null))` redundant?

core/src/main/java/org/jdbi/v3/core/BaseStatement.java
```diff
@@ -55,24 +55,36 @@ protected StatementBuilder getStatementBuilder()
...
 
-    void addCleanable(Cleanable cleanable)
+    /**
+     * Registers the given {@link Cleanable} to be executed when this statement is closed.
+     *
+     * @param cleanable the cleanable to register
+     * @return this
+     */
+    @SuppressWarnings("unchecked")
+    public final This addCleanable(Cleanable cleanable)
```
qualidafial: I think this is a vestigial tail left over from some intermediate work. I'll make it private again.

core/src/main/java/org/jdbi/v3/core/BaseStatement.java
```diff
@@ -55,24 +55,36 @@ protected StatementBuilder getStatementBuilder()
...
     {
-        getContext().getCleanables().add(cleanable);
+        getContext().addCleanable(cleanable);
+        return (This) this;
```
qualidafial: Or we could hoist the `typedThis` stuff up from `SqlStatement` to `BaseStatement`

core/src/main/java/org/jdbi/v3/core/Query.java
```diff
@@ -79,89 +64,8 @@
...
-    {
-        return new Query<>(getConfig(),
-                getParams(),
-                mapper,
-                getHandle(),
-                getStatementBuilder(),
-                getSql(),
-                getContext(),
-                getStatementCustomizers());
+    public <R> R withResultSet(ResultSetCallback<R> callback) {
```
qualidafial: This method comes from `ResultSetIterable`. The intent with the name was to match existing methods around the API, e.g. `Jdbi.withHandle(HandleCallback)`, `ResultIterable.withStream(StreamCallback)`, etc.

Originally I also had a `useResultSet` method but nothing ended up using it, so I torched it.

I'm open to a different name, but I personally lean toward the `with` method for the consistency with other methods that take callbacks.

core/src/main/java/org/jdbi/v3/core/ResultProducers.java
qualidafial: The null check lives in `ResultSetResultIterator`, but it could probably live here.

core/src/main/java/org/jdbi/v3/core/ResultProducers.java
qualidafial: So there's one place in sqlobject that's being lazy: `UpdateHandler` just passes through `GetGeneratedKeys.columnName()` without checking for default values.

You're right that this is sketchy--I'll update `UpdateHandler` to play nicer, and get rid of this filter.

core/src/main/java/org/jdbi/v3/core/ResultSetIterable.java
qualidafial: I had the same thought, but couldn't think of any good alternative names.

* `Result`
* `JdbiResult`
* `ResultSetResult` (from the office of redundancy office)

core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
qualidafial: It's a generics thing. The compiler won't let me cast from `Optional<ColumnMapper<?>>` to `Optional<ColumnMapper<T>>`. But if I unwrap the `ColumnMapper<?>`, it _will_ let me cast that to `ColumnMapper<T>`. I think javac just gives up on nested wildcards.

core/src/main/java/org/jdbi/v3/core/ResultSetIterable.java
qualidafial: Let's tackle this in a follow-up ticket.

core/src/main/java/org/jdbi/v3/core/PreparedBatch.java
```diff
@@ -60,16 +59,27 @@
...
+
     /**
      * Execute the batch
      *
      * @return the number of rows modified or inserted per batch part.
      */
     public int[] execute() {
-        return (int[]) internalBatchExecute(null, null);
+        // short circuit empty batch
+        if (parts.isEmpty()) {
```
qualidafial: I'm going to go ahead and merge this. If you feel strongly about the empty behavior, then let's discuss it in a follow-up ticket.

qualidafial:
Hold my beer!

PR #617, opened 2016-12-13
Let Travis build ignore Oracle for pull requests from other repositor…

qualidafial:
…ies, so access keys can't leak from build logs


stevenschlansker:
It might have been a bit simpler to set a `PROFILES=` once from the Oracle vars and reuse that rather than if-ing every install block

qualidafial:
I tried that--but apparently my travis-fu is not strong

PR #627, opened 2016-12-18
SqlObject mixins: rework as subinterfaces with default methods

### LINKED ISSUE #607: Idea: Allow users to define their own mixins, via @Mixin annotation
If we introduced a `@Mixin` annotation for mixin interfaces, this would allow users to provide default method implementations with access to the `HandleSupplier`.

This way we can add `@Mixin` to `GetHandle` and `Transactional`, and then they'd no longer need special handling inside `SqlObjectFactory`--instead they'd just be standard mixins.

stevenschlansker:
Fixes #607


stevenschlansker:
This might conflict badly with your package-by-feature branch.  If so we can hold off on merging this, probably an easier rebase than the other way around.

qualidafial:
> This might conflict badly with your package-by-feature branch. If so we can hold off on merging this, probably an easier rebase than the other way around.

Yeah I saw this PR come in and first I went, "cool!" and then, "uh oh."

sqlobject/src/main/java/org/jdbi/v3/sqlobject/mixins/SqlObjectMixin.java
```diff
@@ -16,20 +16,34 @@
...
 import org.jdbi.v3.core.HandleConsumer;
+import org.jdbi.v3.core.HandleSupplier;
 
 /**
- * A mixin interface to define a method for obtaining the Handle attached to a sql object
+ * SqlObject base mixin interface.  Custom mixins may be written as subinterfaces
+ * of this class with default methods.
  */
-public interface GetHandle
+public interface SqlObjectMixin<This extends SqlObjectMixin<This>>
```
qualidafial: Per my comment belong on `getThis()`, I don't think this interface should have the `This` generic parameter.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/mixins/SqlObjectMixin.java
```diff
@@ -16,20 +16,34 @@
...
-     *
-     * @return the handle
+     * @return the {@link HandleSupplier} bound to this sql object.
+     */
+    HandleSupplier getHandleSupplier();
+
+    /**
+     * @return the handle open in the current sql object context.
      */
     Handle getHandle();
```
qualidafial: I'm wondering if mixins even need to be a thing. e.g.

```java
@SqlMethodAnnotation(GetHandle.Factory.class)
@interface GetHandle {
  class Factory implements HandlerFactory {
    public Handler buildHandler(Class<?> sqlObjectType, Method method) {
      return (target, args, handleSupplier) -> handleSupplier.getHandle();
    }
  }
}

interface SqlObjectMixin {
  @GetHandle Handle getHandle();
}

interface Transactional<This extends Transactional<This>> extends SqlObjectMixin {
  ...
}
```

With this approach, the concept of mixins would be implied by inheritance, and `SqlObjectFactory` wouldn't need any special case code for mixins.

stevenschlansker: What can you do with a mixin here that a default method can't do?  I think they are equally powerful, and the default method is more direct.

stevenschlansker: The more I think about it, the more I question the value of mixins.  The only one we have in core can be rewritten as default methods trivially, and I'm having trouble thinking of a use case.  If you want to implement arbitrary methods you have `@SqlMethodHannotation`.  If you want to implement a single method, you can just make it default.

qualidafial: > The more I think about it, the more I question the value of mixins.

This is exactly where I'm going.

`GetHandle.getHandle` could be annotated with a custom SqlMethodAnnotation, and then there's nothing special about GetHandle any more--it's just an interface whose SQL methods are already defined.

stevenschlansker:
Okay, I think this is pretty good now.  It essentially removes the concept of mixins entirely.  Now we call the superinterface `SqlObject`.  Also includes rewriting most `Handler` implementations as lambdas, saving another few hundred loc!

stevenschlansker:
:boom:

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObject.java
```diff
@@ -18,14 +18,15 @@
...
 
 /**
- * A mixin interface to define a method for obtaining the Handle attached to a sql object
+ * SqlObject base interface.  All synthesized implementations implement
+ * this interface, even if the declaration doesn't extend {@code SqlObject}.
+ *
+ * Custom mixins may be written as subinterfaces of this class with default methods.
  */
-public interface GetHandle
+public interface SqlObject
```
qualidafial: omg this name was staring us in the face wasn't it

qualidafial:
woot

PR #631, opened 2016-12-20
Simplify Statement constructor hierarchy greatly.  What a rat's nest.

### LINKED ISSUE #628: Refactor BaseStatement subclass constructors to consolidate duplicate code in factory methods
Batch, Call, PreparedStatement, Query, and Update all have static factory `create(Handle,String sql)` methods that extract stuff from the handle and basically do the same things.

e.g. all of them clone the handle config, and create a `StatementContext` with that config.

We should refactor these constructors to consolidate all this code down into `BaseStatement` constructor.

stevenschlansker:
Fixes #628


stevenschlansker:
Also cleans up the `PreparedBatch` interface and removes `PreparedBatchPart` entirely!

core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
```diff
@@ -126,8 +108,8 @@ public ResultSetIterable executeAndReturnGeneratedKeys(String... columnNames) {
 
     @Override
     public <R> R execute(ResultProducer<R> producer) {
-        if (parts.isEmpty()) {
-            throw new IllegalStateException("No PreparedBatchParts to execute");
+        if (bindings.isEmpty()) {
```
qualidafial: It might be good to add a sanity check to validate that `getBinding()` is empty, to catch users who forget to call `add()`. Mockito does something similar e.g. if you call `verify(mock)` with chaining a method call on the mock.

stevenschlansker: I had the same thought in the shower this morning!  Although after considering it more, I have decided to just `add()` implicitly the final batch if any bindings are present.  If there are any bindings actually missing, the statement will fail to prepare anyway.  If the binding is actually valid, it's unambiguous to just `add()` it, and it feels like we shouldn't `throw` in a situation where we can unambiguously fix the situation...

If you strongly disagree with this decision I'm happy to revisit it.

PR #635, opened 2016-12-23
JdbiConfig: banish redundant methods.

stevenschlansker:
They are convenient, but they create unnecessary collaborators in the API and seem of dubious value to me.  Plus, -90 loc :)

### DIFF
--- a/core/src/test/java/org/jdbi/v3/core/argument/TestArgumentFactory.java
+++ b/core/src/test/java/org/jdbi/v3/core/argument/TestArgumentFactory.java
@@ -100,7 +100,7 @@ public class TestArgumentFactory
         public Optional<Argument> build(Type expectedType, Object value, ConfigRegistry config) {
             if (expectedType == Name.class || value instanceof Name) {
                 Name nameValue = (Name) value;
-                return config.findArgumentFor(String.class, nameValue.getFullName());
+                return config.get(Arguments.class).findFor(String.class, nameValue.getFullName(), config);
             }
             return Optional.empty();
         }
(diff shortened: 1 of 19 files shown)


core/src/test/java/org/jdbi/v3/core/argument/TestArgumentFactory.java
```diff
@@ -100,7 +100,7 @@ public void testOnPreparedBatch() throws Exception
         public Optional<Argument> build(Type expectedType, Object value, ConfigRegistry config) {
             if (expectedType == Name.class || value instanceof Name) {
                 Name nameValue = (Name) value;
-                return config.findArgumentFor(String.class, nameValue.getFullName());
+                return config.get(Arguments.class).findFor(String.class, nameValue.getFullName(), config);
```
qualidafial: It does feel weird to pass the config to something you just got _from_ the config. I've debated whether the config registry  ought to optionally inject itself into config types.

guava/src/main/java/org/jdbi/v3/guava/GuavaMappers.java
```diff
@@ -51,7 +52,7 @@ public static ColumnMapperFactory columnFactory() {
 
             if (ImmutableList.class.equals(erasedType)) {
                 return GenericTypes.findGenericParameter(type, ImmutableList.class)
-                        .flatMap(config::findColumnMapperFor)
+                        .flatMap(t -> config.get(ColumnMappers.class).findFor(type, config))
```
qualidafial: This might be splitting hairs, but it should probably be `findFor(t, config)` (`t` instead of `type`)

core/src/test/java/org/jdbi/v3/core/argument/TestArgumentFactory.java
stevenschlansker: I like it.  It can be a constructor argument.

stevenschlansker: Actually, the setter is a nicer API.

qualidafial: Either way works. If we choose setter, it should be an interface that a config class can optionally implement so configs that don't need it can ignore it.

stevenschlansker: Default method!

qualidafial:
I like it.

PR #639, opened 2016-12-23
Move CollectorFactory from core to sqlobject

qualidafial:
I think this is a better home for `CollectorFactory` and friends.

I had to make guava depend on sqlobject, which required moving a test from sqlobject to guava to avoid a circular dependency.


stevenschlansker:
My main complaint about this is that it means that if you want to e.g. use core and guava, you pull in sqlobject.  Is there any chance that we could make it an optional dependency somehow?

qualidafial:
I had it optional, but `GuavaPlugin` registers collector factories. Possibly it could be optional again if the plugin used some runtime `Class.forName` tests.

I'm on the fence about this PR now.

stevenschlansker:
I'm on the fence too.  I totally understand wanting to pull it out, but I'm not sure that this is a concrete improvement.  It'd be possible to make it optional but adding a bunch of reflective magic doesn't feel like much of an improvement either :(

qualidafial:
Let's close it then. I'm glad we investigated it, but as you said it's not a concrete improvement.

PR #640, opened 2016-12-23
Bug fix: SQL method with primitive result throws NPE if query returns no rows

qualidafial:
Also added some documentation on `-parameters` compiler flag to `CONTRIBUTING.md`.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/ResultReturner.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/ResultReturner.java
@@ -29,6 +29,8 @@ import org.jdbi.v3.core.mapper.RowMapper;
 import org.jdbi.v3.core.generic.GenericTypes;
 import org.jdbi.v3.sqlobject.SingleValue;
 
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
 abstract class ResultReturner
 {
     public Object map(Method method, Query q)
@@ -70,7 +72,7 @@ abstract class ResultReturner
     static ResultReturner forMethod(Class<?> extensionType, Method method)
     {
         Type returnType = GenericTypes.resolveType(method.getGenericReturnType(), extensionType);
-        Class<?> returnClass = GenericTypes.getErasedType(returnType);
+        Class<?> returnClass = getErasedType(returnType);
         if (Void.TYPE.equals(returnClass)) {
             throw new IllegalStateException(String.format(
                     "Method %s#%s is annotated as if it should return a value, but the method is void.",
@@ -154,7 +156,7 @@ abstract class ResultReturner
             if (collector != null) {
                 return bearer.collect(collector);
             }
-            return bearer.findFirst().orElse(null);
+            return bearer.findFirst().map(Object.class::cast).orElseGet(() -> DefaultResult.forType(returnType));
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/ResultReturner.java
```diff
@@ -177,7 +179,7 @@ protected Type elementType(StatementContext ctx)
         @Override
         protected Object result(ResultIterable<?> bearer, StatementContext ctx)
         {
-            return bearer.findFirst().orElse(null);
+            return bearer.findFirst().map(Object.class::cast).orElseGet(() -> DefaultResult.forType(returnType));
```
qualidafial: Java compiler barfs without `.map(Object.class::cast)`, since it can't convert `capture of ?` to `Object`.

Bleh.

stevenschlansker:
I'm not sure how I feel about this behavior.  I agree that NPE is bad behavior, but returning a faked up valid value seems pretty bad too.

Is there a particular reason we shouldn't change the behavior instead to be something like a `NoRowsReturnedException`?  If the user actually wants it to be optional, they can return a boxed type or `Optional`.

qualidafial:
This is causing problems in live code. I keep getting an NPE with this code:

```java
@SqlQuery("select 1 from users where id = :id")
boolean exists(int id);
```

qualidafial:
I would also argue that returning the default value is implied if you specify a primitive return type. If you need to tell "no results" and "result with zero" apart you can always return a wrapper type.

qualidafial:
Some other examples:

```java
@SqlQuery("select 1 from users where name = :name and id <> :id")
boolean nameTaken(@Bind String name, @Bind int id);

@SqlQuery("select 1 from users where email = :email and id <> :id")
boolean emailTaken(@Bind String email, @Bind int id);
```

This is holding up some of my personal work. Do you mind if I push a stripped down version of this with just the boolean support?

stevenschlansker:
Ah, I see.  I was imagining something like this:

```java
@SqlQuery("SELECT id FROM users WHERE name=?")
int findByName(String name);
```
which would then return a valid but incorrect value.  It'd be an easy mistake to make.  That's why I was hesitating.

You *could* replace `SELECT 1` with `SELECT count(1) > 0` and then it works without the change.  But I see why you want this.

Did you consider the implications on queries where you actually do return a boolean?  e.g.

```java
@SqlQuery("SELECT is_cool FROM users WHERE id=?")
boolean isUserCool(int id);
```

This code will now return `false` even if the user doesn't exist.

If you've considered cases like this and think this is still an improvement, go ahead and ship it.
Personally, I somewhat lean towards making the user be super explicit with their SQL (if they always want to have a result, the query should always produce a row) and not make up magic return values.  Feels like the nasty JDBC `getInt()` followed by `wasNull`.  But I will not veto this so if you are convinced go for it.

stevenschlansker:
The more I think about this, the more I think it introduces surprises.  What's your objection to just changing to e.g. `SELECT count(*) > 0`?  The intent is more clear, there's less magic, and it should work just as well?  It's also more efficient if there's more than one row, as you won't copy down a result set only to throw it away.

qualidafial:
Stand by, I'll try the `count(1) > 0` thing

qualidafial:
Yep, that works.

Aside: I'm gonna release another alpha from `jdbi3` branch.

qualidafial:
I'll come back and modify this PR to throw an explicit exception as you suggest.

PR #642, opened 2016-12-25
Add support for binding multiple columns in an in clause.

gshakhn:
BindBeanList allows you to create queries similar to:
`select * from things where (key1, key2) in ((1,2),(3,4))`

This is useful to finding rows that match a list of tuples.

Notes:
1) I believe this is the same issue as https://groups.google.com/forum/#!topic/jdbi/AiJvndhLNk8
2) There is an unsafe `Optional.get` in `SqlStatement.bindBeanList`. I'm not sure what to do if it's empty. Happy to hear suggestions.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
@@ -1226,9 +1226,9 @@ public abstract class SqlStatement<This extends SqlStatement<This>> extends Base
      * Examples:
      * <pre>
      *
-     * List&lt;Things&gt; thingsToSearchBy = ...
+     * List&lt;ThingKey&gt; thingKeys = ...
      * List&lt;Thing&gt; things = handle.createQuery("select * from things where (id, foo) in (&lt;thingKeys&gt;)")
-     *     .bindBeanList("thingKeys", thingsToSearchBy)
+     *     .bindBeanList("thingKeys", thingKeys, Arrays.asList("id", "foo"))
      *     .mapTo(Contact.class)
      *     .list();
      * </pre>
@@ -1237,7 +1237,7 @@ public abstract class SqlStatement<This extends SqlStatement<This>> extends Base
      * @param values list of values that will be comma-spliced into the defined attribute value.
      * @param propertyNames list of properties that will be invoked on the values.
      * @return this
-     * @throws IllegalArgumentException if the list is empty.
+     * @throws IllegalArgumentException if the list of values or properties is empty.
      */
     public final This bindBeanList(String key, List<?> values, List<String> propertyNames) {
         if (values.isEmpty()) {
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
@@ -1238,8 +1238,9 @@ public abstract class SqlStatement<This extends SqlStatement<This>> extends Base
      * @param propertyNames list of properties that will be invoked on the values.
      * @return this
      * @throws IllegalArgumentException if the list of values or properties is empty.
+     * @throws UnableToCreateStatementException If a property can't be found on an value or we can't find a Argument for it.
      */
-    public final This bindBeanList(String key, List<?> values, List<String> propertyNames) {
+    public final This bindBeanList(String key, List<?> values, List<String> propertyNames) throws UnableToCreateStatementException {
         if (values.isEmpty()) {
             throw new IllegalArgumentException(
                     getClass().getSimpleName() + ".bindBeanList was called with no values.");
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -1217,6 +1217,66 @@ public final This bindList(String key, List<?> values) {
...
+     * and defines an attribute as the comma-separated list of parameter references (using colon prefix).
+     *
+     * Used to create query similar to:
+     * select * from things where (id, foo) in ((1,'abc'),(2,'def'),(3,'ghi'))
+     * <p>
+     * Examples:
+     * <pre>
+     *
+     * List&lt;Things&gt; thingsToSearchBy = ...
+     * List&lt;Thing&gt; things = handle.createQuery("select * from things where (id, foo) in (&lt;thingKeys&gt;)")
```
stevenschlansker: Would it make sense to define two keys so you can say `SELECT * FROM things WHERE (<columns>) IN (<values>)` and not have to repeat the list of property names / columns twice?  (Once in the SQL, once as a parameter here)

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -1217,6 +1217,66 @@ public final This bindList(String key, List<?> values) {
...
+
+            names.append("(");
+            for (int propertyIndex = 0; propertyIndex < propertyNames.size(); propertyIndex++) {
+                if (propertyIndex > 0) {
+                    names.append(",");
+                }
+                String propertyName = propertyNames.get(propertyIndex);
+                String name = "__" + key + "_" + valueIndex + "_" + propertyName;
+                names.append(':').append(name);
+                bind(name, beanProperties.find(propertyName).get());
```
stevenschlansker: This should probably be an `orElseThrow` with a descriptive message instead of `get`.

stringtemplate3/src/test/java/org/jdbi/v3/sqlobject/BindBeanListTest.java
```diff
@@ -0,0 +1,217 @@
...
+    @Test
+    public void testSomethingByIterableWithIterable()
+    {
+        final SomethingByIterable s = handle.attach(SomethingByIterable.class);
+
+        final List<Something> out = s.get(() -> Arrays.asList(new SomethingKey(1, "1"),
+                new SomethingKey(2, "2"))
+                .iterator());
+
+        assertThat(out).hasSize(2);
```
stevenschlansker: should probably actually check the results here, not just the size

stringtemplate3/src/test/java/org/jdbi/v3/sqlobject/BindBeanListTest.java
```diff
@@ -0,0 +1,217 @@
...
+    @Test
+    public void testSomethingByArrayWithNonEmptyArray()
+    {
+        final SomethingByVarargs s = handle.attach(SomethingByVarargs.class);
+
+        final List<Something> out = s.get(
+                new SomethingKey(1, "1"),
+                new SomethingKey(2, "2"));
+
+        assertThat(out).hasSize(2);
```
stevenschlansker: should check results here, not just size

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
stevenschlansker: Oh, this might be tricky if your column naming strategy is anything other than directly the same as the Java bean property name.  Maybe this won't work.

gshakhn: Just realized that the documentation is invalid (bindBeanList takes three parameters), so going to fix that.

gshakhn: I'm not sure what you mean about defining 2 keys. Can you give me an example on how you envision it looking?

I agree that it's a bit of duplication to define the keys twice. As you mentioned, that's hard to solve if the bean properties don't match the DB column names. At the very least, I think the current API on the branch is required for that use case. Could definitely extend the API to make the code more DRY in the case where bean properties match the DB column names though.

stevenschlansker: Yeah, I don't think there's a good fix here unless we improve the column name mapping further.  Ignore my comment here, it will just have to be duplicated for now :)

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
gshakhn: Will do.

stringtemplate3/src/test/java/org/jdbi/v3/sqlobject/BindBeanListTest.java
gshakhn: I was cribbing off `BindListTest`, which only had the size assertions. I'll add result assertions to both places.

gshakhn:
@stevenschlansker: Thanks for the feedback! I think I've addressed all of it. Let me know if I missed something or if you have more thoughts.

stevenschlansker:
Very nice :)

qualidafial:
Cool! Thanks for contributing. :+1:

gshakhn:
Thanks for the quick merge. Have a happy new year!

PR #648, opened 2016-12-27
merge ConfigurerFactory into StatementCustomizers, rework parameter customization

### LINKED ISSUE #647: Can ConfigurerFactory and SqlStatementCustomizerFactory be merged?

stevenschlansker:
Turns out the main reason for the ConfigurerFactory vs StatementCustomizer is
simply because StatementCustomizer isn't quite the right abstraction.

Move ConfigRegistry access up to Handler creation.  Pass it in to the customizer's
create* methods.  This allows you to mutate it once upfront, and choose whether or
not to customize the method further.

Fix createForParameter signature to make more sense

Fixes #647


qualidafial:
So.. when I suggested merging configurers and statement customizers, I was rather thinking in the other direction.

I've only had a few minutes to look over the diff, but the main problem I see is that statement customizers don't work on default methods, e.g. if you annotate a sql object type with `@RegisterRowMapper` then default methods already have that mapper registered at the time they run. This was a big win in the switch to configurers that I'm not keen to let go of. There were other reasons but they slip my mind at the moment.

I'll hop on this later for a full review, but I wanted to put this out there before you spend too much time on this approach.

qualidafial:
To elaborate a little, I was thinking we could convert the statement customizers into config types e.g. `Queries`. `Updates`, etc. thus `Queries.setFetchSize(int)` and so on.

Gotta run, I'll circle back later.

stevenschlansker:
Yeah, I didn't mark this as ready for review yet because the `default` method customizers are still broken.  I think that is possible to fix as well.  Could you think a bit more about the "other reasons" you have to object to this approach?  I think it is a very clean way to think about it -- "customizers are created during Handler creation and may make config changes then, and are invoked on every statement created from that Handler".

Assuming we can fix default methods, what else don't you like?

stevenschlansker:
Sorry, didn't mean to step on your toes like that!  I'll pull out the less controversial stuff into a separate PR and then circle back here.

qualidafial:
Sorry if I overreacted.

I'm already working on renaming `ConfigurerFactory` to `Configurer` and having it just take a `ConfigRegistry` directly instead of returning a `Consumer<ConfigRegistry>`.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/config/RegisterCollectorFactory.java
```diff
@@ -19,48 +19,44 @@
...
+    class Factory implements SqlStatementCustomizerFactory {
         @Override
-        public Consumer<ConfigRegistry> createForType(Annotation annotation, Class<?> sqlObjectType) {
-            return create((RegisterCollectorFactory) annotation);
-        }
-
-        private Consumer<ConfigRegistry> create(RegisterCollectorFactory annotation) {
-            List<CollectorFactory> factories = new ArrayList<>();
-            for (Class<? extends CollectorFactory> type : annotation.value()) {
+        public SqlStatementCustomizer createForType(ConfigRegistry registry, Annotation annotation, Class<?> sqlObjectType) {
```
qualidafial: Opened #660

qualidafial:
@stevenschlansker What do you want to do with this PR?

The refactoring where `StatementCustomizerFactory.createForMethod` delegates to `createForType` could be done separately, and it doesn't need to hold up beta.

There's also the `StatementCustomizerFactory.createForParameter` refactor with the new type customizer that accepts the annotated argument--this could be broken out into a separate PR.

I'm personally fine with having the two `StatementCustomizerFactory` and `Configurer` APIs how they are now, but if you think we can improve, I'm open to discussion on it.

One thing has been nagging me: I'm not sure I like putting statement-specific settings into `JdbiConfig` classes. In my mind, this invites users to set config at that `Jdbi` or `Handle` level that is inappropriate to share between statements, e.g. `SqlStatements.setGeneratedKeysColumnNames` or `setConcurrentUpdatable`. It may have been a mistake to move that out of `StatementContext`.

PR #650, opened 2016-12-29
Fix oracle build

qualidafial:


oracle12/pom.xml
```diff
@@ -74,4 +74,18 @@
...
     </dependencies>
+
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-surefire-plugin</artifactId>
+                    <configuration>
+                        <forkCount>1</forkCount>
```
qualidafial: Have to spin up a whole VM for oracle builds--impractical to try spinning the database up and down for each test class.

pom.xml
```diff
@@ -60,6 +60,7 @@
 
         <basepom.check.fail-all>true</basepom.check.fail-all>
 
+        <basepom.release.profiles>basepom.oss-release,oracle</basepom.release.profiles>
```
qualidafial: @stevenschlansker This should prevent accidentally missing the `oracle` profile on release builds. You'll need to have followed the steps in `CONTRIBUTING.md` (at least the `settings.xml` and `settings-security.xml` stuff) before you can release again.

PR #652, opened 2016-12-29
Rename maven artifacts in each pom.xml, and not just the module (dire…

qualidafial:
…ctory) names.


qualidafial:
I'll just merge this, assuming the build passes. We've already gone over this at length in the prior issue/PR.

PR #653, opened 2016-12-29
OutParameters.getXyz(int) methods should use 0-based indices, like ev…

### LINKED ISSUE #643: OutParameters are 1-based
While the in parameters of a Call are 0-based the OutParameters are 1-based. Consider the following stored procedure (HSQLDB).

```sql
CREATE procedure plus1inout (IN arg int, OUT res int)
BEGIN ATOMIC
  SET res = arg + 1;
END
/;
```

you would call this using the following code

```java
OutParameters outParameters = handle.createCall("call plus1inout(?, ?);")
        .bind(0, argument)
        .registerOutParameter(1, Types.INTEGER)
        .invoke();
return outParameters.getInt(2);
```

notice how for binding we use 1 base the index of the out parameter but for retrieving we use 2 as the index for the out parameter.

qualidafial:
…erything else in JDBI

Fixes #643

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/Call.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/Call.java
@@ -94,7 +94,11 @@ public class Call extends SqlStatement<Call>
             OutParameters out = new OutParameters();
             for ( OutParamArgument param : params ) {
                 Object obj = param.map((CallableStatement)stmt);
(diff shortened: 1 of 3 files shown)


qualidafial:
@marschall thanks for bringing this to our attention. This is fixed in v3.

marschall:
@qualidafial you're welcome, thank you for fixing so quickly

PR #654, opened 2016-12-29
Add jdbi3-stringtemplate4 artifact.

qualidafial:


stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateSqlLocator.java
```diff
@@ -0,0 +1,117 @@
...
+     * Locates SQL for the given type and name. Example: Given a type <code>com.foo.Bar</code> and a name of
+     * <code>baz</code>, loads a StringTemplate group file from the resource named <code>com/foo/Bar.sql.stg</code> on
+     * the classpath, and returns the template with the given name from the group.
+     *
+     * @param type the type that "owns" the given StringTemplate group file. Dictates the filename of the
+     *             StringTemplate group file on the classpath.
+     * @param name the template name in the StringTemplate group.
+     * @return the located SQL.
+     */
+    public static ST findStringTemplate(Class<?> type, String name) {
```
qualidafial: This method differs from the one in ST3 since ST4 doesn't let you get the original template string back out.

I feel like the ST3 one should also return a template instance, instead of a string, since a template could reference another template in the group.

brianm:
The jdbi-st4 package isn't spaghetti :-)

qualidafial:
That was a "copypasta" joke, since the jdbi3-st4 artifact is basically a copy-paste of jdbi3-st3 :)

qualidafial:
@arteam @stevenschlansker Could I get a code review?

stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateSqlLocator.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    private static String toString(InputStream inputStream) throws IOException {
+        final int bufferSize = 1024;
+        final char[] buffer = new char[bufferSize];
+        final StringBuilder out = new StringBuilder();
+        Reader in = new InputStreamReader(inputStream, "UTF-8");
+        for (; ; ) {
+            int rsz = in.read(buffer, 0, buffer.length);
+            if (rsz < 0){
```
arteam: Maybe we could use a while loop and move this condition to the top.

stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateSqlLocator.java
```diff
@@ -0,0 +1,117 @@
...
+
+        try (InputStream is = openStream(type.getClassLoader(), path)) {
+            return new STGroupString(toString(is));
+        }
+        catch (IOException e) {
+            throw new RuntimeException("Unable to read StringTemplate group file at " + path + " on classpath", e);
+        }
+    }
+
+    private static String toString(InputStream inputStream) throws IOException {
```
qualidafial: I could use Guava, but we've had a longstanding goal to eliminate that dependency if possible.

Maybe it won't be a big deal once we get shaded jars working again.

stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateSqlLocator.java
qualidafial: Will do, this was lifted directly from StackOverflow ¯\\\_(ツ)\_/¯

qualidafial:
Thanks @arteam

brianm:
Right, and I think we'd be better off starting with the much slimmer jdbi-st4 :-)

qualidafial:
Most of this PR is tests--there's only 4 classes with a total of 345 LOC in src/main (including javadoc) for the new artifact. That said, if you feel like we could slim it down I'm definitely open to it. Did you have anything specific in mind?

stevenschlansker:
I think Brian is referring to https://github.com/jdbi/jdbi-st4 maybe?

PR #657, opened 2016-12-31
Remove TransactionCallback, TransactionConsumer in favor of HandleCal…

qualidafial:
…lback, HandleConsumer

`TransactionCallback` and `TransactionConsumer` were recently modified to take only a `Handle` argument, which is the same signature (aside from method names) as `HandleCallback` and `HandleConsumer`.

This PR removes the transaction callbacks in favor of the handle ones.

Also adds a bit of missing javadoc on `Jdbi` transaction methods.


stevenschlansker:
Nice

PR #661, opened 2017-01-02
Coalesce StatementConfiguration, SqlStatements config classes

### LINKED ISSUE #658: Fold StatementConfiguration into SqlStatements?
They seem to serve the same purpose, but for different bits of data.

qualidafial:
Fixes #658


core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -26,6 +27,10 @@
     private StatementRewriter statementRewriter;
     private TimingCollector timingCollector;
 
+    private boolean returningGeneratedKeys;
+    private String[] generatedKeysColumnNames;
```
arteam: I wonder if it makes sense to declare the generated columns as a `List` rather than an array. We could then force its immutability via `unmodifiable` decorators rather than copying the array.

qualidafial: Actually, now that I look at it, it's not necessary to clone internally since the getter/setter always make a defensive clone. The array is kept private and never mutated, so can be shared between SqlStatements instances.

qualidafial:
@stevenschlansker You introduced `StatementConfiguration`--are you good with coalescing these types?

qualidafial:
If it comes back to you, we can always revisit. Merging.

Issue #601, opened 2016-12-06
Evaluate SqlObject hardcoding of types


stevenschlansker:
Currently, there's a number of cases where SqlObject implementations hardcode types e.g. `ResultBearing` or `Iterable` for return values.  We should consider whether this list of types should be hardcoded, and if not, how to handle it better.

brianm:
It would be better to use the `ContainerFactory` dynamic matching chain via its `accepts` stuff to determine if it is a container.

There is some mucking around between containing and mapping that works differently between core and sqlobject right now, that impedence should probably be brought closer in line

brianm:
I would NOT do this before release as it will lead to API and semantic changes. If we release with current behavior, we need to maintain current behavior.

stevenschlansker:
Do you mean "not do it after release"?  I think you're right that we might make semantic changes, so we should move this to a blocker, to at least consider it before deferring.

qualidafial:
@stevenschlansker Could you give an example of where this is hardcoded?

qualidafial:
Does this need to hold up beta? If I understand this issue correctly, these are internal classes that can be refactored at any time without affecting compatibility.

stevenschlansker:
https://github.com/jdbi/jdbi/blob/jdbi3/sqlobject/src/main/java/org/jdbi/v3/sqlobject/ResultReturner.java#L80

stevenschlansker:
I don't think a fix necessarily needs to hold a beta, but the evaluation maybe should?

qualidafial:
Line 80 in ResultReturner is a bug: is should be looking for `ResultIterable`, not `ResultBearing` now

qualidafial:
Looking into this.

qualidafial:
Opened #663.

qualidafial:
Resolved in #663

PR #663, opened 2017-01-03
ResultReturner refactor

qualidafial:
* Remove ArrayResultReturner in favor of ArrayCollectorFactory in core
* Move rowMapperType method to CustomizingStatementHandler
* Rename method parameters (bearer -> iterable)

Addresses #601 (Evaluate SqlObject hardcoding of types)

We have ResultReturners hardcoded for several categories of return type.

I believe each of them is appropriate at the SqlObject level, except for Array, which this PR promotes to core as ArrayResultReturner.

All the other ResultReturners fall into two categories:
1. The return type is a core object that exists before the Collector comes into play, or represents the result as a whole rather than a series of elements that can be collected.
2. The return type has special semantics in SqlObject, where using a collector makes no sense

* `void` - specific to SqlObject, nothing to collect
* `ResultIterable` - comes before the collector
* `Iterator` - comes before the collector (the stream itself comes from this iterator)
* `Stream` - comes before the collector
* `@SingleValue` - SqlObject-specific exception that avoids using a collector

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/collector/ArrayCollectorFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/collector/ArrayCollectorFactory.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


qualidafial:
@stevenschlansker You filed the original issue. Did you want to weigh in on this before I merge?

stevenschlansker:
I'll try to get through the review queue in the next day or so.  Been a shitty week again sorry :/

core/src/main/java/org/jdbi/v3/core/collector/ArrayCollectorFactory.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public Optional<Type> elementType(Type containerType) {
+        return Optional.ofNullable(getErasedType(containerType).getComponentType());
+    }
+
+    @Override
+    public Collector<?, ?, ?> build(Type containerType) {
+        Class<?> componentType = getErasedType(containerType).getComponentType();
+        return Collector.of(
```
stevenschlansker: :+1:

PR #665, opened 2017-01-04
Remove HandyMapThing

arteam:
Looks like it's not so much handy after all. It's used only in one place, where we can use a singleton map. As a result, this change removes mostly dead code and promotes using the standard JDK collections instead custom ones.


arteam:
Sorry, didn't catch up with upstream.

PR #668, opened 2017-01-11
Add support for readOnly in Handle and @Transaction

### LINKED ISSUE #91: Use readOnly connections when possible

stevenschlansker:
Fixes #91

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/transaction/Transaction.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/transaction/Transaction.java
@@ -39,33 +39,58 @@ import org.jdbi.v3.sqlobject.SqlMethodDecoratingAnnotation;
 @Target({ElementType.METHOD})
 @SqlMethodDecoratingAnnotation(Transaction.Decorator.class)
 public @interface Transaction {
+    /**
+     * @return the transaction isolation level.  If not specified, invoke with the default isolation level.
+     */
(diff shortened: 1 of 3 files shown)


stevenschlansker:
cc @diwakergupta

sqlobject/src/main/java/org/jdbi/v3/sqlobject/transaction/Transaction.java
```diff
@@ -62,10 +73,21 @@ public Handler decorateHandler(Handler base, Class<?> sqlObjectType, Method meth
...
                 HandleCallback<Object, Exception> callback = th -> base.invoke(target, args, handle);
 
-                if (isolation == TransactionIsolationLevel.INVALID_LEVEL) {
-                    return h.inTransaction(callback);
-                } else {
-                    return h.inTransaction(isolation, callback);
+                final boolean flipReadOnly = readOnly != h.isReadOnly();
+                if (flipReadOnly) {
+                    h.setReadOnly(readOnly);
+                }
```
qualidafial: Can we add a check to the `if (h.isInTransaction())` block above? I think it's appropriate to throw an exception if the already-started transaction is read-write, but the present annotation's `readOnly()` is true.

We can probably ignore the inverse case.

Issue #370, opened 2016-05-25
Do a full pass through the code before jdbi3 beta releases


stevenschlansker:
Things to look for:
- Rename all `Jdbi` instances to e.g. `db` or something?
- Anonymous classes that should be lambdafied
- Anything named `thing` should get a better name.
- Make sure all tests are isolated and parallel-safe

qualidafial:
>  Anything named `thing` should get a better name.

For tests and documentation, I think the contact list as an approachable example. You have contacts table and you don't have to get fancy on fields--just use a name, or firstname / lastname--whatever's relevant for the test/example. Then when you need to do joins, you add extra tables for phone numbers, emails, addresses, and so on.

qualidafial:
> Anything named thing should get a better name.

All the things name `Thing` are in tests.

I'm working on renaming `Jdbi` variables / fields to `db`.

qualidafial:
@stevenschlansker So there are a bunch of tests that have both `DatabaseRule` and `Jdbi` fields. Up until now all our test rules have typically been named `db`. What do you think about `Jdbi db` and e.g. `H2DatabaseRule dbRule` for names?

qualidafial:
@stevenschlansker Also, what about renames like `DatabaseRule.getJdbi()` to `getDb()`, or `JdbiPlugin.customizeJdbi()` to `customizeDb()`?

For that matter.. what about `Jdbi` itself?

qualidafial:
Opened #669 to track rename stuff.

qualidafial:
I've crawled the entire javadoc, and have opened Github issues for each problem I found.

Most of them are no-brainers: trivial renames, changing visibility of types or members. I'll start submitting PRs for the obvious ones.

There are a couple that are less obvious, and warrant group discussion:

* #676 Do we like the pattern of returning `Optionals` from factories?
* #692 Refactor `SqlStatementCustomizer` like we did `Configurer`?

So put me down for "completed" on this issue. @stevenschlansker @arteam please let us know when you're satisfied with your pass-through so we can close out this issue.

stevenschlansker:
Thanks for taking the time to file those issues.  I think I agree with your list.  I'll muse a bit more but none of the things that bug me right now really block the beta.  So consider me good to go after we burn down the task list

qualidafial:
Once those last two PRs (#708 and #711) are merged, I am GO for beta release

qualidafial:
All beta-blocking issues are closed.

[...]

PR #669, opened 2017-01-11
Rename Jdbi fields/variables to db, DatabaseRule fields to dbRule

qualidafial:
Carved out from #370.

* [x] Rename `Jdbi` fields and variables to `db`.
* [x] Rename `DatabaseRule` fields to `dbRule` (formerly `db`)
* [ ] Rename `JdbiPlugin.customizeJdbi()` to `customizeDb()`?
* [ ] Rename `DatabaseRule.getJdbi()` to `getDb()`?
* [ ] Rename `Jdbi` itself to `Database` or maybe `Db`?

@jdbi/contributors What does everyone think about the last three renames posed above?


stevenschlansker:
I'm :+1: to the 2 renames you checked at the top.  I'm less enthusiastic about e.g. `Jdbi` -> `Database`.  If you think abstractly about what a 'database' is, the existing `Jdbi` is not.  It is much closer to a `DataSource`.  In particular, you may have multiple `Jdbi` instances pointing to the same abstract 'database' and as such naming it that just seems confusing to me.

Issue #600, opened 2016-12-05
Ensure java.time.{Period,Interval} map to Postgres interval type


stevenschlansker:
Better solution for #598

qualidafial:
Ha, you beat me to it.

stevenschlansker:
Check out `Duration` as well.

Issue #598, opened 2016-12-05
Not replaced values in PostgreSQL's interval


raderio:
```
public interface testRepository {
    @SqlUpdate("INSERT INTO test (code, expire_at) VALUES (:code, NOW() + interval ':days days')")
    void add(@Bind("code") String code, @Bind("days") long days);
}
```

rewritten
`"INSERT INTO sessions (code, expire_at) VALUES (?, NOW() + interval ':days days')"
`

but it should be 
`"INSERT INTO sessions (code, expire_at) VALUES (?, NOW() + interval '? days')"`

qualidafial:
This is correct behavior. JDBI should not mess with anything in quotes inside your query.

Try concatenation and see if that works:

```sql
NOW() + interval (? || ' days')
```

raderio:
What is the reason to not mess with anything in quotes inside your query?

qualidafial:
This is part of the SQL standard. http://stackoverflow.com/a/26555612

raderio:
Ok. Thank you.

qualidafial:
Out of curiosity, did concatenation work?

raderio:
nope
`PSQLException: ERROR: syntax error at or near "$2"`

qualidafial:
Hmm. What happens if you bind the entire interval string:

```
now() + interval ?
```

And then bind the string `2 days` to that argument?

stevenschlansker:
I doubt that sort of concatenation is going to work without nasty casting.  Something like `(:date + ' days')::interval`.  Probably a better solution is to use `java.time.Period` and set up the argument binding correctly.  I filed a followup for that.

DanVanAtta:
FWIW, this worked for me:
``` @SqlUpdate("delete from login where created < now() - :days * '1 day'::interval")```

From: https://stackoverflow.com/questions/52480996/jdbi-interval-postgresql

PR #670, opened 2017-01-13
Adds support for Postgres interval ↔︎ Period & Duration types.

pennello:
I believe this addresses https://github.com/jdbi/jdbi/issues/600 and https://github.com/jdbi/jdbi/issues/598.

cc @stevenschlansker @rmacdonaldsmith


postgres/src/test/java/org/jdbi/v3/postgres/IntervalTestCommon.java
```diff
@@ -0,0 +1,41 @@
...
+            // Can be durations.
+            h.execute("insert into intervals(id, foo) values(1, interval '1 day 15:00:00')");
+            h.execute("insert into intervals(id, foo) values(2, interval '40 days 22 minutes')");
+
+            // Can be intervals.
+            h.execute("insert into intervals(id, foo) values(3, interval '2 years -3 months 40 days')");
+            h.execute("insert into intervals(id, foo) values(4, interval '7 days')");
+
+            // Can be neither.
+            h.execute("insert into intervals(id, foo) values(5, interval '10 years -3 months 100 seconds')");
```
qualidafial: I have mixed feeling about this shared setup. We do this already in e.g. `PgDatabaseRule` and `H2DatabaseRule` and I'm not convinced it's a beneficial pattern.

This is just a personal preference, but I find tests much easier to follow when the sample data is right there in the test, so you don't have to go rooting around in other classes to understand _why_ a test expects a specific result.

Shared setup like this tends to unnecessarily couple a ton of tests to the same setup. I personally would rather duplicate the setup code in each test, so that's it's obvious why a test might fail without having to go searching.

postgres/src/test/java/org/jdbi/v3/postgres/TestDuration.java
```diff
@@ -0,0 +1,145 @@
...
+                .findOnly();
+        assertThat(d).isEqualTo(testDuration);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testInvalidDuration() {
+        handle.createQuery("select foo from intervals where id=?")
+                .bind(0, 5)
+                .mapTo(Duration.class)
+                .findOnly();
```
qualidafial: Here is an example of a test that is difficult to understand in isolation, because the critical data (what is in interval id=5?) is in `IntervalTestCommon` instead of in a setup method or a one-off in the test itself.

postgres/src/main/java/org/jdbi/v3/postgres/DurationArgumentFactory.java
```diff
@@ -0,0 +1,77 @@
...
+ * An {@link IllegalArgumentException} will be thrown in these cases.
+ * The handling of the second is subject to revision in the future; for example, it would be reasonable to have a
+ * configurable truncation option.
+ */
+public class DurationArgumentFactory implements ArgumentFactory {
+    @Override
+    public Optional<Argument> build(Type type, Object value, ConfigRegistry config) {
+        if (!(value instanceof Duration)) {
+            return Optional.empty();
+        }
```
qualidafial: Opened #671 to make it easier to implement argument factories that handle null / subclasses.

postgres/src/test/java/org/jdbi/v3/postgres/IntervalTestCommon.java
pennello: Fair.  It is a little hard to follow, and I don't know if we really gain much by making the setup "common".  Refactored in 9c32d37ec7bc92bc894e49544264f497550e5552.

postgres/src/main/java/org/jdbi/v3/postgres/PeriodArgumentFactory.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Postgres version of argument factory for {@link Period}.
+ */
+public class PeriodArgumentFactory implements ArgumentFactory {
+    @Override
+    public Optional<Argument> build(Type type, Object value, ConfigRegistry config) {
+        if (!(value instanceof Period)) {
+            return Optional.empty();
+        }
```
pennello: Updated in 7f102f4a97e6ab0fdb89a0ba72cbbf3c5167f876.

postgres/src/main/java/org/jdbi/v3/postgres/DurationArgumentFactory.java
pennello: Ah, good point!  I'm not sure why I implemented it that way.  I think I've implemented it better in 7f102f4a97e6ab0fdb89a0ba72cbbf3c5167f876.  Let me know if you see further issue there.  I'm also not quite sure how to test this update.

postgres/src/test/java/org/jdbi/v3/postgres/TestDuration.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testDurationTooPrecise() {
+        handle.insert("insert into intervals(id, foo) values(?, ?)",
+                14, Duration.ofNanos(100));
+    }
+
+    // We guard against reading intervals with seconds too big or too small (i.e., more extreme than Long minimum and
+    // maximum values), but it's unclear how to actually create such intervals in Postgres.
```
pennello: Via an `insert` statement directly, I can't create an `interval` with, e.g., > 2^63 - 1 seconds.

    => select interval '9223372036854775808 seconds';
    ERROR:  interval field value out of range: "9223372036854775808 seconds"
    LINE 1: select interval '9223372036854775808 seconds';

So that won't quite work (I tried this while trying to think of a test here).

I _could_ construct one of the column mappers _directly_ and _directly_ created a `PGInterval` instance and try to pass it in with a mocked `ResultSet`, but this seemed not quite worth it either (`ResultSet` has a zillion methods to implement).  So I just opted for the comment.

Further thoughts, reactions?  Open to suggestions.

postgres/src/test/java/org/jdbi/v3/postgres/TestDuration.java
pennello: Yeah, that's fair.  Refactored in 9c32d37ec7bc92bc894e49544264f497550e5552.

postgres/src/test/java/org/jdbi/v3/postgres/IntervalTestCommon.java
pennello: Fair.  It is a little hard to follow, and I don't know if we really gain much by making the setup "common".  Refactored in 9c32d37ec7bc92bc894e49544264f497550e5552.

postgres/src/main/java/org/jdbi/v3/postgres/PeriodArgumentFactory.java
pennello: Updated in 7f102f4a97e6ab0fdb89a0ba72cbbf3c5167f876.

postgres/src/main/java/org/jdbi/v3/postgres/DurationArgumentFactory.java
pennello: Ah, good point!  I'm not sure why I implemented it that way.  I think I've implemented it better in 7f102f4a97e6ab0fdb89a0ba72cbbf3c5167f876.  Let me know if you see further issue there.  I'm also not quite sure how to test this update.

postgres/src/test/java/org/jdbi/v3/postgres/TestDuration.java
pennello: Yeah, that's fair.  Refactored in 9c32d37ec7bc92bc894e49544264f497550e5552.

postgres/src/main/java/org/jdbi/v3/postgres/DurationArgumentFactory.java
pennello: Oh whoops.  I missed your suggestion of the right way to return `null`.  Fixed in 5ace9385f059734ea6aae3bc99c3b36ee31ab7cc.

postgres/src/main/java/org/jdbi/v3/postgres/PeriodArgumentFactory.java
pennello: Oh whoops.  I missed your suggestion of the right way to return `null`.  Fixed in 5ace9385f059734ea6aae3bc99c3b36ee31ab7cc.

postgres/src/test/java/org/jdbi/v3/postgres/TestDuration.java
qualidafial: If it can't be stored in Postgres, I don't think that it's worth testing the mapper for it.

Testing the argument factory--eh.

qualidafial:
Looks great. Thanks @pennello

pennello:
Cool.  Thanks @qualidafial!

PR #673, opened 2017-01-18
ReflectionArrayIterator: rename to IterableLike, since it handles more than arrays and doesn't necessarily use reflection

stevenschlansker:
BindListUtil: fold into IterableLike
Steal the (excellent) test cases

(scrapping #648 for parts)


qualidafial:
Looks like a single test, `BindBeanListTest.testSomethingByIterator`, is failing. :shipit: once tests are fixed.

Issue #683, opened 2017-01-20
Upgrade to Guava 21, and remove our homegrown Guava Collector implementations


qualidafial:
Guava 21 upgrades to JDK 8, and provides their own Collector implementations for Guava collection types.

We can get rid of our own collector implementations, in favor of those now included in Guava.

arteam:
Good idea. Will happily try out the new Guava version.

stevenschlansker:
tracked as #697

PR #697, opened 2017-01-21
Upgrade to Guava 21

arteam:
Guava 21 is a release which requires Java 8 and provide some additional features for computability between Guava and Java 8 collection frameworks. For instance, we can reuse collectors for immutable lists and sets. They are now provided by Guava itself. Unfortunately, it looks like we can't replace the collections that collects data to Guava's Optional. Guava provides an optional collectors, but it collects
data to Java8's Optional. A pity.

Guava 21 release notes: https://github.com/google/guava/wiki/Release21

Reference issue: #683

### DIFF
--- a/guava/src/main/java/org/jdbi/v3/guava/GuavaCollectors.java
+++ b/guava/src/main/java/org/jdbi/v3/guava/GuavaCollectors.java
@@ -24,7 +24,6 @@ import java.util.stream.Collector;
 import org.jdbi.v3.core.collector.CollectorFactory;
 
 import com.google.common.base.Optional;
-import com.google.common.collect.ImmutableCollection;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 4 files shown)


arteam:
Hm, it looks like the Oracle's Maven repo is down right now. Will try to restart the build later.

qualidafial:
The build is failing because the last alpha release was made without the `oracle` profile, so all the versions were updated except the Oracle module.

stevenschlansker:
Gah.

qualidafial:
I'm not sure how it happened, honestly. I thought that setting the oss.release-profiles property would take care of it.

qualidafial:
At any rate, updating the version in Oracle pom should fix the build.

stevenschlansker:
I wonder if that only affects the forked Maven, and takes effect too late (i.e. after the version updates) or something like that.

Maybe we need to have the module active unconditionally but just skip the build unless the profile is active?

qualidafial:
Part of me wonders why I bother to support Oracle when they can't be arsed to put their jar on central.

stevenschlansker:
Yeah, seriously :(  I just pushed a fix to `jdbi3`.

stevenschlansker:
What the hell is the point of hiding the damn driver.  You'd think they would want people to use their software.

qualidafial:
Reasons!

stevenschlansker:
@arteam I rebased this branch.

stevenschlansker:
```
[ERROR] /home/travis/.m2/repository/com/google/guava/guava/21.0/guava-21.0.jar(com/google/common/collect/Multimap.class): warning: Cannot find annotation method 'value()' in type 'CompatibleWith': class file for com.google.errorprone.annotations.CompatibleWith not found
```

Whoa.

stevenschlansker:
https://github.com/google/guava/issues/2721

PR #702, opened 2017-01-21
TransactionIsolationLevel: rename INVALID to UNKNOWN.

### LINKED ISSUE #695: Rename TransactionIsolationLevel.INVALID_LEVEL to UNSPECIFIED?

stevenschlansker:
fixes #695 

I chose this over `UNSPECIFIED` because it is specified, just not by us (it's either the driver or the db)


stevenschlansker:
It also is used to map any unknown types that the driver might support but we don't know about.  So I'm going to stick with it.

PR #698, opened 2017-01-21
Make Duration/Period factories inherit AbstractArgumentFactory

arteam:
Doing so, we can avoid additional checks for confirming factory to the right type and make the factories more generic-friendly.

References #686

PR #703, opened 2017-01-21
Convert a couple more ArgumentFactory implementations to use AbstractArgumentFactory

stevenschlansker:
based on feedback on #698 by @pennello

### DIFF
--- a/postgres/src/main/java/org/jdbi/v3/postgres/DurationArgumentFactory.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/DurationArgumentFactory.java
@@ -13,14 +13,13 @@
  */
 package org.jdbi.v3.postgres;
 
-import org.jdbi.v3.core.argument.AbstractArgumentFactory;
-import org.jdbi.v3.core.argument.Argument;
-import org.jdbi.v3.core.argument.NullArgument;
-import org.jdbi.v3.core.config.ConfigRegistry;
-import org.postgresql.util.PGInterval;
-
 import java.sql.Types;
 import java.time.Duration;
+
+import org.jdbi.v3.core.argument.AbstractArgumentFactory;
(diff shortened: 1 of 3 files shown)


postgres/src/main/java/org/jdbi/v3/postgres/DurationArgumentFactory.java
```diff
@@ -13,15 +13,14 @@
...
+
 import org.jdbi.v3.core.argument.AbstractArgumentFactory;
 import org.jdbi.v3.core.argument.Argument;
-import org.jdbi.v3.core.argument.NullArgument;
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.postgresql.util.PGInterval;
 
-import java.sql.Types;
-import java.time.Duration;
-
```
pennello: This project should standardize on an import order.  ; ]

stevenschlansker: Yeah, we should...

stevenschlansker:
It didn't look like it was as amenable to being converted.  In particular it doesn't get rid of the bulk of the work, which is selecting the correct `java.sql.Type` to use.

pennello:
@stevenschlansker 
> It didn't look like it was as amenable to being converted. In particular it doesn't get rid of the bulk of the work, which is selecting the correct java.sql.Type to use.

Got it.  Then, perhaps it's worth a comment there to guard against future would-be developers following its pattern without really understanding the implementation differences?

postgres/src/main/java/org/jdbi/v3/postgres/DurationArgumentFactory.java
qualidafial: wut

Issue #706, opened 2017-01-22
hashCode on SqlObject acquires connecftion, fails if DBI is closed


q3aiml:
It would be nice if calling hashCode on a SqlObject did not attempt to acquire a connection that HashCodeHandler does not need. Looking at the tests it seems this is fixed with jdbi 3. Would a PR for jdbi 2 be considered, perhaps similar to how finalize is handled?

Thanks!

Example error when the DAO is stored in a Map and removal is attempted after DBI is closed (or other connection issues):
```
! java.sql.SQLException: HikariDataSource HikariDataSource (example) has been closed.
! at com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.java:79)
! at org.skife.jdbi.v2.DataSourceConnectionFactory.openConnection(DataSourceConnectionFactory.java:34)
! at org.skife.jdbi.v2.DBI.open(DBI.java:211)
! ... 61 common frames omitted
! Causing: org.skife.jdbi.v2.exceptions.UnableToObtainConnectionException: java.sql.SQLException: HikariDataSource HikariDataSource (example) has been closed.
! at org.skife.jdbi.v2.DBI.open(DBI.java:229)
! at org.skife.jdbi.v2.sqlobject.OnDemandHandleDing.getOrCreateLocalDing(OnDemandHandleDing.java:67)
! at org.skife.jdbi.v2.sqlobject.OnDemandHandleDing.retain(OnDemandHandleDing.java:62)
! at org.skife.jdbi.v2.sqlobject.SqlObject.invoke(SqlObject.java:223)
! at org.skife.jdbi.v2.sqlobject.SqlObject$3.intercept(SqlObject.java:133)
! at com.example.ExampleDao$$EnhancerByCGLIB$$39f2831a.hashCode(<generated>)
! at java.util.concurrent.ConcurrentHashMap.replaceNode(ConcurrentHashMap.java:1106)
! at java.util.concurrent.ConcurrentHashMap.remove(ConcurrentHashMap.java:1097)
```

qualidafial:
Huh, I thought that was fixed a long time ago. Do you still see th is problem while using the most recent version (2.78)?

q3aiml:
[...]

PR #707, opened 2017-01-22
Calling hashCode, equals, or toString on SqlObjects should not connect

q3aiml:
For issue #706.


qualidafial:
I don't have push rights to your branch so I coded up a fix and opened up a separate PR that includes your tests, and my fix: #714

Issue #694, opened 2017-01-21
Should @GetGeneratedKeys take multiple column names?


qualidafial:

qualidafial:
Extending it to take multiple column names was easy--in fact it made the annotation agree even better with core.

I realized that `GetGeneratedKeys.value(): Class<? extends RowMapper>` is redundant to `UseRowMapper`. So I removed the value property, renamed `columnName()` to `value()`, and made it a `String[]`.

I found a weird thing which is that registering row mappers now works fine for `@GetGeneratedKeys`, but `@UseRowMapper` doesn't work with it.

I'm experimenting.

qualidafial:
Tracked in #708

PR #708, opened 2017-01-22
Refactor GetGeneratedKeys

qualidafial:
Addresses #694 

* Remove GetGeneratedKeys.value(): Class<? extends RowMapper> in favor of @UseRowMapper
* Rename GetGeneratedKeys.columnName() to value(), and change it from String to String[] (multiple column names)

After looking into it, I got rid of the `UseRowMapper.Factory` annotation handler. It turns out that `SqlBatch` was already doing its own custom handling, and `SqlUpdate` needed to implement its own custom handling too. So I inlined it into the `SqlQuery` handler.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/ResultReturner.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/ResultReturner.java
@@ -21,7 +21,6 @@ import java.util.stream.Stream;
 
 import org.jdbi.v3.core.generic.GenericTypes;
 import org.jdbi.v3.core.result.ResultIterable;
-import org.jdbi.v3.core.result.ResultBearing;
 import org.jdbi.v3.core.statement.StatementContext;
 import org.jdbi.v3.sqlobject.SingleValue;
 
@@ -33,17 +32,6 @@ import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
  */
 abstract class ResultReturner
 {
-    /**
-     * Given a {@link ResultBearing}, map to the element
-     * type and construct the result container.
-     * @param r the ResultBearing
-     * @return the filled result container
-     */
-    public Object map(ResultBearing r, StatementContext ctx)
-    {
-        return result(r.mapTo(elementType(ctx)), ctx);
-    }
-
     /**
      * If the return type is {@code void}, swallow results.
      * @param extensionType
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/SqlQuery.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/SqlQuery.java
@@ -18,9 +18,12 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.lang.reflect.Method;
+import java.lang.reflect.Type;
 
 import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.result.ResultIterable;
 import org.jdbi.v3.core.statement.Query;
+import org.jdbi.v3.core.statement.StatementContext;
 import org.jdbi.v3.sqlobject.Handler;
 import org.jdbi.v3.sqlobject.HandlerFactory;
 import org.jdbi.v3.sqlobject.SqlMethodAnnotation;
@@ -58,7 +61,15 @@ public @interface SqlQuery {
 
         @Override
         void configureReturner(Query q, SqlObjectStatementConfiguration cfg) {
-            cfg.setReturner(() -> magic.map(q, q.getContext()));
+            cfg.setReturner(() -> {
+                StatementContext ctx = q.getContext();
+                Type elementType = magic.elementType(ctx);
+                UseRowMapper useRowMapper = getMethod().getAnnotation(UseRowMapper.class);
+                ResultIterable<?> iterable = useRowMapper == null
+                        ? q.mapTo(elementType)
+                        : q.map(rowMapperFor(useRowMapper));
+                return magic.result(iterable, ctx);
+            });
         }
 
         @Override
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/UseRowMapper.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/UseRowMapper.java
@@ -13,26 +13,18 @@
  */
 package org.jdbi.v3.sqlobject.statement;
 
-import java.lang.annotation.Annotation;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
-import java.lang.reflect.Method;
 
 import org.jdbi.v3.core.mapper.RowMapper;
-import org.jdbi.v3.core.result.ResultBearing;
-import org.jdbi.v3.core.statement.UnableToCreateStatementException;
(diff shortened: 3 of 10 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/ResultReturner.java
```diff
@@ -34,17 +33,6 @@
...
     /**
-     * Given a {@link ResultBearing}, map to the element
-     * type and construct the result container.
-     * @param r the ResultBearing
-     * @return the filled result container
-     */
-    public Object map(ResultBearing r, StatementContext ctx)
-    {
-        return result(r.mapTo(elementType(ctx)), ctx);
-    }
```
qualidafial: Inlined into SqlQuery handler

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestGetGeneratedKeysPostgres.java
```diff
@@ -83,4 +90,63 @@ public void testBatch() throws Exception {
...
+        @SqlUpdate("insert into something(name) values (:name)")
+        @GetGeneratedKeys
+        @UseRowMapper(IdCreateTimeMapper.class)
+        IdCreateTime insert(String name);
+
+        @SqlBatch("insert into something(name) values (:name)")
+        @GetGeneratedKeys
+        @UseRowMapper(IdCreateTimeMapper.class)
+        List<IdCreateTime> insertBatch(String... name);
+    }
```
qualidafial: Probably should add a separate test that registers a mapper instead of `@UseRowMapper`, just to be thorough.

qualidafial: Also.. should probably specify the column names.. d'oh

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/UseRowMapper.java
```diff
@@ -13,49 +13,23 @@
...
-            RowMapper<?> mapper;
-            try {
-                mapper = mapperAnnotation.value().newInstance();
-            } catch (InstantiationException | IllegalAccessException e) {
-                throw new UnableToCreateStatementException("Could not create mapper " + mapperAnnotation.value().getName(), e, null);
-            }
-
-            final ResultReturner returner = ResultReturner.forMethod(sqlObjectType, method);
-            return q -> q.getConfig(SqlObjectStatementConfiguration.class)
-                    .setReturner(() -> returner.result(((ResultBearing) q).map(mapper), q.getContext()));
```
qualidafial: If we can find a shared pattern between SqlBatch, SqlQuery, and SqlUpdate I'd love to add this customizer back.

qualidafial:
@arteam @stevenschlansker Would one of you mind reviewing this?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/SqlQuery.java
```diff
@@ -58,7 +61,15 @@ public Handler buildHandler(Class<?> sqlObjectType, Method method) {
 
         @Override
         void configureReturner(Query q, SqlObjectStatementConfiguration cfg) {
-            cfg.setReturner(() -> magic.map(q, q.getContext()));
+            cfg.setReturner(() -> {
+                StatementContext ctx = q.getContext();
+                Type elementType = magic.elementType(ctx);
+                UseRowMapper useRowMapper = getMethod().getAnnotation(UseRowMapper.class);
+                ResultIterable<?> iterable = useRowMapper == null
```
arteam: It's a pity we can't use the `Optional` idiom here because of the type mismatch. I often find very appealing to replace a ternary operator with a series of `Optional` transformations, because they look very clear and elegant. But for such a simple check it's probably not worth it and the C-style code looks better.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestGetGeneratedKeysPostgres.java
```diff
@@ -53,11 +61,11 @@ public void tearDown() throws Exception {
 
     public interface DAO {
         @SqlUpdate("insert into something (name, id) values (:name, nextval('id_sequence'))")
-        @GetGeneratedKeys(columnName = "id")
+        @GetGeneratedKeys("id")
         long insert(@Bind("name") String name);
 
         @SqlBatch("insert into something (name) values (:name)")
-        @GetGeneratedKeys(columnName = "id")
+        @GetGeneratedKeys("id")
```
arteam: Very nice

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestGetGeneratedKeysPostgres.java
```diff
@@ -83,4 +91,103 @@ public void testBatch() throws Exception {
             assertThat(dao.findNameById(ids[1])).isEqualTo("Macklin");
         });
     }
+
+    @Test
+    public void testUseRowMapperUpdate() throws Exception {
+        dbRule.getJdbi().useHandle(handle -> {
+            handle.execute("drop table something");
```
arteam: We could move the code for recreating the `something` table to the `setUp` block and avoid duplication.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestGetGeneratedKeysPostgres.java
```diff
@@ -83,4 +91,103 @@ public void testBatch() throws Exception {
...
+
+            UseRowMapperDao dao = handle.attach(UseRowMapperDao.class);
+
+            List<IdCreateTime> results = dao.insertBatch("foo", "bar");
+            assertThat(results).extracting(ic -> ic.id).containsExactly(1L, 2L);
+            assertThat(results).extracting(ic -> ic.createdOn).hasSize(2).doesNotContainNull();
+        });
+    }
+
+    public interface UseRowMapperDao {
```
arteam: Just a nit, but I think it would be better to move the `UserRowMapperDao` class after the tests to `RegisterRowMapperDao`.  I think it's better from the point of maintaining a consistent order of the items in the class.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/SqlQuery.java
qualidafial: That's not a bad idea--I can always use `Optional.ofNullable`

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestGetGeneratedKeysPostgres.java
qualidafial: No problem.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/SqlQuery.java
qualidafial: Ugh, the compiler system won't tolerate the `ResultIterable<?>` return type on the lambda. I give up.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestGetGeneratedKeysPostgres.java
qualidafial: The tests all have slightly different setup. Instead of duplicating the SQL I moved the table create to methods in each DAO. I actually like it better this way--you can tell how each setup fits into each DAO's methods.

Issue #692, opened 2017-01-21
Refactor SqlStatementCustomizer like we did Configurer?


qualidafial:
Recently we had `ConfigurerFactory` and `Configurer`, and we were producing a configurer factory only to turn right around and apply the configurer to the config registry. We refactored it so there was only one interface and it would eagerly configure the registry.

Could/should we do the same thing to `SqlStatementCustomizerFactory` / `SqlStatementCustomizer`?

Current API:

```java
public interface SqlStatementCustomizerFactory {
    default SqlStatementCustomizer createForType(Annotation annotation,
                                                 Class<?> sqlObjectType) throws SQLException { ... }

    default SqlStatementCustomizer createForMethod(Annotation annotation,
                                                   Class<?> sqlObjectType,
                                                   Method method) throws SQLException { ... }

    default SqlStatementCustomizer createForParameter(Annotation annotation,
                                                      Class<?> sqlObjectType,
                                                      Method method,
                                                      Parameter param,
                                                      int index,
                                                      Object arg) throws SQLException { ... }
}

public interface SqlStatementCustomizer {
    void apply(SqlStatement<?> q) throws SQLException;
}
```

Proposed replacement:

```java
public interface SqlStatementCustomizerFactory {
    default void customizeForType(SqlStatement<?> statement,
                                  Annotation annotation,
                                  Class<?> sqlObjectType) { ... }

    default void customizeForMethod(SqlStatement<?> statement,
                                    Annotation annotation,
                                    Class<?> sqlObjectType,
                                    Method method) { ... }

    default void customizeForParameter(SqlStatement<?> statement,
                                       Annotation annotation,
                                       Class<?> sqlObjectType,
                                       Method method,
                                       Parameter param,
                                       int index,
                                       Object arg) { ... }
}
```

qualidafial:
To clarify, I'm just raising the question so we can have a discussion--not advocating we actually do it.

One possible reason _not_ to do this, is that you have you can't reflect once and cache the computation for types and methods like you can now. There's really no benefit to caching on the parameter.

I think we should look at the customizers we have, see how they're being used, and evaluate the possible impact before proceeding.

qualidafial:
Starting a spike on this

qualidafial:
Actually, I'm putting this off until the other stuff is done--this will touch a lot of files so I don't want to tie up other PRs

stevenschlansker:
[...]

PR #709, opened 2017-01-22
Spike: Refactor SqlStatementCustomizer like we did Configurer.

qualidafial:
Refactors SqlStatementCustomizer per #692.

Best viewed with `?w=1` in the URL to cut out whitespace noise.

This will conflict with PR #708--I'd rather merge that one first.

What I like about this PR is that it removes a class and makes implementing a customizing annotation a little simpler.

What I don't like is the performance implications. This loses the ability to precompute customizers at the class and method levels, which means that we have to recompute those at call time. Without a proper performance benchmark it's difficult to know whether this will make a difference.

At the parameter level, we already recompute every time at call time, so it's really no worse performance-wise than before.

So this has me wondering whether we ought to be going in the opposite direction, namely that we should pursue @stevenschlansker's idea of a `SqlStatementParameterCustomizer` at the parameter level. This way we can precompute some things at the parameter declaration level. I believe (but cannot confirm) this would have an important performance impact for batches.

So unless the team convinces me that this PR is a good idea, I'm voting -1 on this one. I just wanted to show what it ended up looking like, and give my reasoning for being against it.


qualidafial:
Opened #710

qualidafial:
With #711 opened, I'm even more convinced this is the wrong approach. Closing.

PR #714, opened 2017-01-23
Don't connect sql objects on hashCode(), equals(), or toString()

qualidafial:
Fixes bug identified in #707


src/test/java/org/skife/jdbi/v2/sqlobject/TestObjectMethodBehavior.java
```diff
@@ -21,13 +21,15 @@
...
 /**
  * Sometimes the GC will call {@link #finalize()} on a SqlObject from
- * extremely sensitive places from within the GC machinery.  JDBI should not
- * open a {@link Connection} just to satisfy a (no-op) finalizer.
+ * extremely sensitive places from within the GC machinery.
  * <a href="https://github.com/brianm/jdbi/issues/82">Issue #82</a>.
+ * Other methods also do not use a {@link Connection} and can cause issues acquiring one.
  */
-public class TestFinalizeBehavior
+public class TestObjectMethodBehavior
```
arteam: Maybe it's a better idea to write a new test instead changing the finalize test?

src/main/java/org/skife/jdbi/v2/sqlobject/SqlObject.java
```diff
@@ -187,9 +187,7 @@ else if (mixinHandlers.containsKey(raw_method)) {
         // this is an implicit mixin, not an explicit one, so we need to *always* add it
         handlers.putAll(CloseInternalDoNotUseThisClass.Helper.handlers());
 
-        handlers.putAll(EqualsHandler.handler());
         handlers.putAll(ToStringHandler.handler(sqlObjectType.getName()));
```
arteam: I'm wondering why we don't remove `ToStringHandler`. Do we expect that someone could override it?

qualidafial: Without the override, CGLIB gives it a pretty ugly toString(). There are also some tests that expect a specific toString(), so I opted not to mess with it.

src/test/java/org/skife/jdbi/v2/sqlobject/TestObjectMethodBehavior.java
qualidafial: I didn't mind this rename--it just expanded the same requirement (invocation does not connect) from finalize to the other object methods. If you feel strongly about it though, I'm OK with breaking it out.

q3aiml: FWIW I took the naming more or less from jdbi3: https://github.com/jdbi/jdbi/blob/jdbi3/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestSqlObjectMethodBehavior.java

arteam:
Looks good to me. Let's wait for a review @stevenschlansker, then ship it.

q3aiml:
Thank you everyone for being so responsive. I appreciate all your time and it shows in how pleasant jdbi is to use

PR #743, opened 2017-01-26
Fix NPE in @SqlBatch when batch arguments are empty

### LINKED ISSUE #741: NPE in SqlBatch.BatchHandler if the batch returns no results

qualidafial:
Fixes #741 

Best viewed with `?w=1` in URL


qualidafial:
Expediting this fix since I depend on it for private work.

qualidafial:
Originally that's what I did, but had second thoughts about adding a public API method with only one caller at this stage. If it comes up again that's probably grounds to factor it out into a utility method.

PR #753, opened 2017-02-07
Allow for a plugin point to supply a custom SqlObjectStatementCustomizer

aharin:
Allow to specify an instance of  SqlStatementCustomizerFactory  that will be used to bind parameters to SQL statements when parameters are not annotated. 

When custom SqlStatementCustomizerFactory is not specified the default Bind.Factory is used. 

This is required for Kotlin support as it needs a different parameter customizer. See Issue #282


sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/CustomizingStatementHandler.java
```diff
@@ -72,28 +70,40 @@
...
+        return Stream.of(defaultParameterCustomizer(type, method, parameter, i));
+    }
+
+    /**
+     * Default parameter customizer for parameters with no annotations.
+     */
+    private static BoundCustomizer defaultParameterCustomizer(Class<?> type, Method method, Parameter parameter, Integer i) {
+        return (stmt, args) -> getDefaultParameterCustomizerFactory(stmt)
+                .createForParameter(null, type, method, parameter, i)
+                .apply(stmt, args[i]);
```
qualidafial: I'd rather we introduce a separate factory interface for the default customizer, similar to what #754 is doing with default handlers. It feels weird to pass a `null` for the annotation, when the use case is specifically for when there _is_ no annotation. e.g.

```java
interface DefaultParameterCustomizerFactory {
    SqlStatementParameterCustomizer createForParameter(Class<?> sqlObjectType,
                                                       Method method,
                                                       Parameter param,
                                                       int index);
}
```

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/SqlObjectStatementCustomizerConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+    public SqlObjectStatementCustomizerConfiguration createCopy() {
+        return new SqlObjectStatementCustomizerConfiguration(this);
+    }
+
+    /**
+     * Returns the configured {@link SqlStatementCustomizerFactory} used to bind sql statement parameters
+     * when parameter is not explicitly annotated. By default it is configured as an instance of {@link Bind.Factory}.
+     *
+     * @return the configured {@link SqlStatementCustomizerFactory}.
+     */
```
qualidafial: Good docs.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/config/TestSqlObjectStatementCustomizer.java
```diff
@@ -0,0 +1,104 @@
...
+
+        sqlStatementParameterCustomizer = new SqlStatementCustomizerFactory()
+        {
+            SqlStatementCustomizerFactory defaultFactory = new Bind.Factory();
+
+            @Override
+            public SqlStatementParameterCustomizer createForParameter(Annotation annotation, Class<?> sqlObjectType, Method method, Parameter param, int index)
+            {
+                invocationCounter.incrementAndGet();
+                return defaultFactory.createForParameter(annotation, sqlObjectType, method, param, index);
```
qualidafial: I think it would be a better exercise of this feature if the tests demonstrated an alternative binding strategy, instead of just wrapping the default strategy. Even something simple like `(stmt, arg) -> stmt.bind("mybind" + index, arg)` would do the trick, and give a more complete demonstration of what can be accomplished with this feature.

aharin:
I've made the changes as requested. 

The standard implementation of DefaultParameterCustomizerFactory delegates to Bind.Factory so it has to pass null for annotation. I hope this is ok, I did not want to start changing Bind.Factory to extract method that does not take annotation as a parameter. At least now it is a detail of the implementation while the interface itself is clean.

aharin:
Ok, renamed!

qualidafial:
Thanks @aharin for contributing!

PR #754, opened 2017-02-09
Allow for a plugin point to supply a custom DefaultHandlerFactory

aharin:
Allow to overwrite default mechanism for providing dynamic proxy handler for default methods.  
This pull request introduces DefaultHandlerFactory interface that can be customised using new configuration mechanism. The default implementation, DefaultMethodHandlerFactory builds existing DefaultMethodHandler for java default methods  

This is required for Kotlin support as kotlin default methods do not compile Java 8 default methods. See Issue #282


sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -94,7 +94,7 @@ public boolean accepts(Class<?> extensionType) {
                         invocationHandler));
     }
 
-    private Map<Method, Handler> methodHandlersFor(Class<?> sqlObjectType) {
+    private Map<Method, Handler> methodHandlersFor(Class<?> sqlObjectType, ConfigRegistry instanceConfig) {
```
qualidafial: The handlers generated and cached in this method are class-level--they have no instance-specific data and can thus be safely reused by all instances of a given SQL object type.

Passing `instanceConfig` to this method mixes class- and instance-level scope--with the effect that config is locked in the first time a SQL object type is attached.

Moreover, it subverts a design intention in v3, that any changes made to a config class should be honored on a forward basis by new SQL object instances.

A couple options:

1. Leave it the way it is in this PR. Document that the default method handler config property must be set at the `Jdbi` level before the first SQL object is created.
2. Defer creation of the handler for a given method until the first time it is invoked. I rather dislike this approach, as it subverts our ability to sanity-check interfaces on `attach`. Also it adds some kind of bastard "meta-handler-factory" type which is kinda gross.
3. Make the default handler a property of `SqlObjectFactory` itself. This is a departure from the usual approach of just adding another config property, but given the pervasive effect of this setting I think it's justifiable. Apps (and plugins) could alter the default behavior as follows:

```java
jdbi.findFactory(SqlObjectFactory.java).ifPresent(f -> f.setDefaultMethodHandler(myMethodHandler));
```

@jdbi/contributors I'm leaving toward options 1 or 3. Opinions?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -135,7 +135,11 @@ private Handler buildBaseHandler(Class<?> sqlObjectType, Method method) {
                             sqlMethodAnnotations));
         }
 
-        if (method.isDefault()) {
+        DefaultHandlerFactory defaultHandlerFactory = instanceConfig
+                .get(DefaultHandlerFactoryConfiguration.class)
+                .getDefaultHandlerFactory();
```
qualidafial: Could we pull this up to the `attach` method, and pass down the default handler factory instead of the config registry?

sqlobject/src/test/java/org/jdbi/v3/sqlobject/config/TestDefaultHandlerFactoryConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+            public Handler buildHandler(Class<?> sqlObjectType, Method method) {
+                invocationCounter.incrementAndGet();
+                return delegate.buildHandler(sqlObjectType, method);
+            }
+
+            @Override
+            public boolean accepts(Class<?> sqlObjectType, Method method) {
+                return delegate.accepts(sqlObjectType, method);
+            }
+        };
```
qualidafial: As in #753, I'd like this test to demonstrate the feature in action--not just wrapping the default handler.

Maybe demonstrate that non-default, non-annotated methods could be given default behaviors, e.g. executing queries or updates according to method name?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
qualidafial: I've had a few days to think about this, and I think option 1 is the most practical approach: go ahead and use a config property, and explicitly document that the setting is locked in for a sql object type the first time it is instantiated (or invoked, for on-demand).

aharin:
Great,

I will be back on Saturday. Hopefully I will be able to make the changes
over the weekend.

qualidafial:
Please also take a look at #755. Your PR got me thinking about another solution that might be cleaner in the long term, although it require breaking backward compatibility with 3.0.0-beta0.. :grimacing:

aharin:
Ok, all requested changes done.

qualidafial:
@aharin Have a look at #756, I think this might be a more flexible approach--allows you to preempt the default handler, rather than having to substitute it.

qualidafial:
Closing this PR in favor of #756 and #758. Thanks for contributing and for following up with tests!

PR #756, opened 2017-02-19
Refactor HandlerFactory to allow for alternative method handler mappi…

qualidafial:
…ng strategies.

I'm submitting this as an alternative to PR #754 (which would add a config property for a "default" method handler as an alternative to existing default method handler).

However I believe the approach in this commit is a better choice long-term, as it allows clients to override the out-of-the-box behavior for any method,
rather than only being able to swap out the "fall-back" method handler.

The public API changes are modest:

* [BREAKING CHANGE] Change HandlerFactory to return `Optional<Handler>` instead of `Handler`
* [BREAKING CHANGE] Change semantics of `@SqlMethodAnnotation`: `value()` returns a handler type instead of handler factory type.
  * The handler type must have a public no-arg, (Method), or (Class, Method) constructor.
* Add Handlers registry with last-registered-wins semantics like our other registries
* Refactor SqlObjectFactory toi use the Handlers registry.
  * Existing logic was moved out to SqlMethodHandlerFactory and DefaultMethodHandlerFactory, which are registered out of the box.

To do
* [x] Needs tests to exercise custom `HandlerFactory`s that can override the default behavior

CC @aharin

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/CreateSqlObject.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/CreateSqlObject.java
@@ -20,6 +20,7 @@ import java.lang.annotation.Target;
 import java.lang.reflect.Method;
 
 import org.jdbi.v3.core.extension.Extensions;
+import org.jdbi.v3.core.extension.HandleSupplier;
 
 /**
  * Use this annotation on a sql object method to create a new sql object with the same underlying handle as the sql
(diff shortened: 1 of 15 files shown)


qualidafial:
The build error is a compile failure, and it makes no sense to me..

stevenschlansker:
Wow, that is a bizarre build failure.  Could be a compiler difference in type resolution between IntelliJ and javac?  Ran into that a couple of times with Eclipse.

aharin:
Agreed, this is a better solution then #754 . I have submitted a pull request #758 for this branch that adds a test that verifies that a custom HandlerFactory could be added. Hopefully this addresses the TODO

Issue #763, opened 2017-02-27
Support Map and Multimap return types for SQL Object methods


qualidafial:
Example usage:

```java
// simple key/value mapping for two columns
@SqlQuery("select key, value from config")
@KeyColumn("key")
@ValueColumn("value")
Map<String,String> getConfigs()

// index by id
@SqlQuery("select * from users")
@KeyColumn("id")
Map<Long, User> getUsersById()

// master/detail mapping
@SqlQuery("select a.id a_id, a.name a_name, i.id i_id, i.name i_name, ...etc "+
    "from accounts a left join invoices i on a.id = i.account_id")
@RegisterBeanMapper(value = {Account.class,Invoice.class}, prefix= {"a", "i"})
Multimap<Account,Invoice> getAccountsAndInvoices()
```

Proposed design:

* Introduce a built-in CollectorFactory for Map<K,V>, where elementType() returns Entry<K,V>
* Introduce a CollectorFactory in Guava for Multimap<K,V>, where elementType() returns Entry<K,V>
* Introduce a built-in mapper for Entry<K,V> which delegates to the JoinRow mapper using the key and value types from the Entry generic parameters.
* Introduce annotations `@KeyColumn` and `@ValueColumn` to give hints when the map key and/or value is in a single column. If a key/value annotation is present, get the mapper from ColumnMappers. Otherwise get the mapper from RowMappers.
  * Might need to introduce a config class to hold the key/value column name configuration, in order to convey that info from SQL Object to the mapper in Core.

qualidafial:
@stevenschlansker @arteam Any feedback before I start work on this?

arteam:
To be honest, I'm not a big fan of putting a lot of logic in annotations. We provide a `SqlObject` mixin to get a direct access to the current `Handle`, so it possible to do advanced transformation in default methods of SqlObjects. Unless the annotation way drastically reduce the amount of code, I would try to avoid additional complexity.  Having said that, I think it would be great to have some prototype implementation of it. I don't think this would be a wasteful work in any case.

qualidafial:
There's really not a lot of logic on the SQL object side. Just the two annotations that will each set up a config property. Everything else will live in core, and a little in guava.

stevenschlansker:
One of my worries is that there is a hidden cost to constructing the key values so many times.  It is a classic case of object / relational mismatch, but adding this syntax makes it really easy to construct unexpectedly expensive queries.

[...]

Issue #733, opened 2017-01-25
Developer guide: Guava plugin


qualidafial:
* What types are supported by the guava plugin
* How to use it (`Jdbi.installPlugin` vs `.installPlugins()`)

qualidafial:
Tracked in #869

PR #769, opened 2017-03-06
Add collectors for Map, Multimap types

qualidafial:
Addresses #763 

* Add `MapEntryMapper`, which maps rows to `Map.Entry<K,V>`. This mapper is registered out of the box.
  * By default, `MapEntryMapper` maps keys and value using `RowMapper`s registered for the respective key and value types.
  * `MapEntryMapper.Config.setKeyColumn` overrides the default behavior, and will load the key from the specified column, using the `ColumnMapper` registered for the key's type.
  * `MapEntryMapper.Config.setValueColumn` overrides the default behavior, and will load the value from the specified column, using the `ColumnMapper` registered for the value's type.
* Add a built-in collector factory, which collects various JDK `Map<K,V>` types, with an element type of `Map.Entry<K,V>`
* Enhance `GuavaCollectors` to support Guava-specific `Map<K,V>` types, as well as Guava `Multimap<K,V>` types, with an element type of `Map.Entry<K,V>`
* Add `@KeyColumn` and `@ValueColumn` SQL Object annotations, which call `MapEntryMapper.Config.setKeyColumn` and `.setValueColumn`, respectively.

TODO
* [x] Throw exception in map collectors for duplicate entries.
* ~Add asciidocs demonstrating how to return maps from SQL Object methods.~ #774 
* ~Add asciidocs documenting how to use GuavaPlugin, and which types it adds support for--including Map and Multimap types.~ #733


core/src/test/java/org/jdbi/v3/core/mapper/MapEntryMapperTest.java
```diff
@@ -0,0 +1,181 @@
...
+        h.execute("create table config (key varchar, value varchar)");
+        h.prepareBatch("insert into config (key, value) values (?, ?)")
+                .add("foo", "123")
+                .add("bar", "xyz")
+                .execute();
+
+        Map<String, String> map = h.createQuery("select key, value from config")
+                .configure(MapEntryMapper.Config.class, cfg -> cfg.setKeyColumn("key").setValueColumn("value"))
+                .mapTo(new GenericType<Map.Entry<String, String>>() {})
+                .collect(intoContainer(new GenericType<Map<String, String>>() {}));
```
qualidafial: @stevenschlansker Remember when I pushed back against adding `Query.collectInto(Class<T> | GenericType<T> | Type)`?

I think this test demonstrates that these would probably be desirable, and safe to add.

```java
public interface ResultBearing {
  <R> R collectInto(Class<R> containerType);
  <R> R collectInto(GenericType<R> containerType);
  Object collectInto(Type containerType);
}
```

qualidafial: Oh hell yes I am adding that.

qualidafial: Hell yes I added that

core/src/main/java/org/jdbi/v3/core/collector/BuiltInCollectorFactories.java
```diff
@@ -30,6 +30,14 @@
 
 /**
  * CollectorFactories for built in JDK and jdbi types.
+ * <p>Supported Collection types:</p>
+ * <ul>
+ *     <li>java.util.Optional&lt;T&gt; (throws an exception if more than one row in result)</li>
+ *     <li>java.util.List&lt;T&gt;</li>
+ *     <li>java.util.Set&lt;T&gt;</li>
+ *     <li>java.util.SortedSet&lt;T&gt;</li>
+ *     <li>java.util.Map&lt;K, V&gt; - for row type Map.Entry&lt;K, V&gt;</li>
```
qualidafial: Oops, lots more types supported in MapCollectorFactory that I forgot to add here

core/src/main/java/org/jdbi/v3/core/collector/MapCollectorFactory.java
```diff
@@ -0,0 +1,99 @@
...
+    private static final TypeVariable<Class<Map>> KEY_PARAM;
+    private static final TypeVariable<Class<Map>> VALUE_PARAM;
+
+    static {
+        TypeVariable<Class<Map>>[] mapParams = Map.class.getTypeParameters();
+        KEY_PARAM = mapParams[0];
+        VALUE_PARAM = mapParams[1];
+    }
+
+    private static final Map<Class<?>, Collector<?,?,?>> collectors = ImmutableMap.<Class<?>, Collector<?,?,?>>builder()
```
arteam: I guess it makes sense to declare this variable in the upper-case as a constant, because it's effectively immutable.

core/src/main/java/org/jdbi/v3/core/collector/MapCollectorFactory.java
```diff
@@ -0,0 +1,99 @@
...
+
+    static <K, V, M extends Map<K, V>> Collector<Map.Entry<K,V>, ?, M> toMap(Supplier<M> supplier) {
+        return Collector.of(
+                supplier,
+                MapCollectorFactory::putEntry,
+                MapCollectorFactory::combine);
+    }
+
+    private static <K, V, M extends Map<K, V>> void putEntry(M map, Map.Entry<K, V> entry) {
+        // TODO throw exception if duplicate entry?
```
arteam: I think it would be better to throw an exception on duplicates rather than silently swallowing them. In that case the user will be encouraged to switch from `Map` to `Multimap`, if several values for one key are possible.

core/src/main/java/org/jdbi/v3/core/mapper/MapEntryMapper.java
```diff
@@ -0,0 +1,164 @@
...
+                    .orElseThrow(() -> new NoSuchMapperException("No row mapper registered for map key " + keyType));
+        } else {
+            return config.get(ColumnMappers.class)
+                    .findFor(keyType)
+                    .map(mapper -> new SingleColumnMapper<>(mapper, column))
+                    .orElseThrow(() -> new NoSuchMapperException("No column mapper registered for map key " + keyType + " in column " + column));
+        }
+    }
+
+    private static RowMapper<?> getValueMapper(Type keyType, ConfigRegistry config) {
```
arteam: I think it would be better to name the variable `valueType`, because we are looking for a mapper for the value column.

core/src/main/java/org/jdbi/v3/core/mapper/MapEntryMapper.java
```diff
@@ -0,0 +1,164 @@
...
+    @Override
+    public RowMapper<Map.Entry<K, V>> specialize(ResultSet rs, StatementContext ctx) throws SQLException {
+        return new MapEntryMapper<>(keyMapper.specialize(rs, ctx), valueMapper.specialize(rs, ctx));
+    }
+
+    /**
+     * Configuration class for MapEntryMapper.
+     */
+    public static class Config implements JdbiConfig<Config> {
+        public Config() {
```
arteam: Do we need the default constructor? I can't see where it's used.

core/src/test/java/org/jdbi/v3/core/mapper/MapEntryMapperTest.java
```diff
@@ -0,0 +1,178 @@
...
+                .collectInto(new GenericType<Map<User, Phone>>() {});
+
+        assertThat(map).containsOnly(
+                entry(new User(1, "alice"), new Phone(10, "555-0001")),
+                entry(new User(2, "bob"),   new Phone(20, "555-0002")),
+                entry(new User(3, "cathy"), new Phone(30, "555-0003")));
+    }
+
+    @SuppressWarnings("unchecked")
+    private <K, V, T> Collector<Map.Entry<K, V>, ?, T> intoContainer(GenericType<T> containerType) {
```
arteam: It looks like that method is unused.

guava/src/main/java/org/jdbi/v3/guava/GuavaCollectors.java
```diff
@@ -77,26 +125,158 @@ public void set(T value) {
     public static class Factory implements CollectorFactory {
 
         private static final Map<Class<?>, Collector<?, ?, ?>> collectors =
```
arteam: I think we could declare this variable in the upper case, because the map is immutable. The same for `mapCollectors` and `multimapCollectors`

guava/src/test/java/org/jdbi/v3/guava/TestGuavaCollectors.java
```diff
@@ -107,4 +126,58 @@ public void optionalMultiple() {
...
+
+        assertThat(registry.findElementTypeFor(genericType.getType()))
+                .contains(new GenericType<Map.Entry<Long,String>>(){}.getType());
+
+        Collector<Map.Entry<Long,String>, ?, M> collector = (Collector<Map.Entry<Long, String>, ?, M>) registry
+                .findFor(genericType.getType())
+                .orElseThrow(() -> new IllegalStateException("Missing collector for " + genericType));
+
+        M map = Stream.of(entry(1L, "foo"), entry(2L, "bar")).collect(collector);
+        Assertions.assertThat(map)
```
arteam: We could use just `assertThat`

postgres/src/test/java/org/jdbi/v3/postgres/TestHStore.java
```diff
@@ -133,6 +134,7 @@ public void testWritesWithBindMap() {
     public interface CampaignDao {
 
         @SqlQuery("select caps from campaigns where id=:id")
+        @SingleValue
```
arteam: I don't know whether this good thing or not... Probably good, if we know consider maps as containers.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestReturningMap.java
```diff
@@ -0,0 +1,252 @@
...
+
+        dao.insertUsers(
+                new User(1, "alice"),
+                new User(2, "bob"),
+                new User(3, "cathy"));
+        dao.insertPhone(1, new Phone(10, "555-0001"), new Phone(11, "555-0021"));
+        dao.insertPhone(2, new Phone(20, "555-0002"), new Phone(21, "555-0022"));
+        dao.insertPhone(3, new Phone(30, "555-0003"), new Phone(31, "555-0023"));
+
+        assertThat(dao.getMultimap()).hasSameEntriesAs(
```
arteam: Maybe it's easier just to call `isEqualTo`?

core/src/main/java/org/jdbi/v3/core/collector/MapCollectorFactory.java
qualidafial: I'm actually thinking I'll make them non-static instead.

core/src/main/java/org/jdbi/v3/core/collector/MapCollectorFactory.java
qualidafial: Agreed--this is one of my todos.

core/src/main/java/org/jdbi/v3/core/mapper/MapEntryMapper.java
qualidafial: This is required by ConfigRegistry--all JdbiConfig instances must have a public no-arg constructor, so the registry can create a root config object on demand.

core/src/test/java/org/jdbi/v3/core/mapper/MapEntryMapperTest.java
qualidafial: Good catch. This was used previously, but isn't needed now we have `ResultBearing.collectInfo(GenericType)`

postgres/src/test/java/org/jdbi/v3/postgres/TestHStore.java
qualidafial: Yep. FWIW we already have to do this when selecting a single row of an array column.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestReturningMap.java
qualidafial: Can't do this one. Multimap.equals() contract is basically: this.asMap().equals(that.asMap(). So this common fixture wouldn't work across ListMultimap and SetMultimap.

core/src/main/java/org/jdbi/v3/core/collector/ElementTypeNotFoundException.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.collector;
+
+import org.jdbi.v3.core.JdbiException;
+
+@SuppressWarnings("serial")
```
stevenschlansker: I would normally just declare a `private static final long serialVersionUid = 1L;` rather than suppressing the warning.  One line either way.

core/src/main/java/org/jdbi/v3/core/collector/NoSuchCollectorException.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.collector;
+
+import org.jdbi.v3.core.JdbiException;
+
+@SuppressWarnings("serial")
```
stevenschlansker: ditto
(plus javadocs)

core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
```diff
@@ -88,4 +100,32 @@ public static Type resolveType(Type type, Type contextType) {
...
+     * Given a key and value type, returns the map entry type {@code Map.Entry<keyType,valueType>}.
+     * @param keyType the key type
+     * @param valueType the value type
+     * @return the map entry type
+     */
+    public static Type resolveMapEntryType(Type keyType, Type valueType) {
+        return resolveMapEntryType(TypeToken.of(keyType), TypeToken.of(valueType));
+    }
+
+    private static <K, V> Type resolveMapEntryType(TypeToken<K> keyType, TypeToken<V> valueType) {
```
stevenschlansker: 😱 wow

core/src/main/java/org/jdbi/v3/core/mapper/MapEntryMapper.java
```diff
@@ -0,0 +1,164 @@
...
+                    .orElseThrow(() -> new NoSuchMapperException("No row mapper registered for map key " + keyType));
+        } else {
+            return config.get(ColumnMappers.class)
+                    .findFor(keyType)
+                    .map(mapper -> new SingleColumnMapper<>(mapper, column))
+                    .orElseThrow(() -> new NoSuchMapperException("No column mapper registered for map key " + keyType + " in column " + column));
+        }
+    }
+
+    private static RowMapper<?> getValueMapper(Type valueType, ConfigRegistry config) {
```
stevenschlansker: these two helpers sure look like they could be folded together

core/src/main/java/org/jdbi/v3/core/collector/ElementTypeNotFoundException.java
qualidafial: I think I just copy-pasted this from another class. I'll patch the other one too.

core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
qualidafial: Yeah, this TypeToken use case is a little rough around the edges, but it gets the job done.

core/src/main/java/org/jdbi/v3/core/mapper/MapEntryMapper.java
qualidafial: will do

qualidafial:
I've been holding up merge on this for nearly two weeks to get docs done, but I'm feeling a little blocked.

I think we need to revamp the overall narrative of the docs before there will be an obvious place to plug this feature in.

stevenschlansker:
If you think it's ready to merge other than docs, that's fine by me.

PR #772, opened 2017-03-08
Improvements to KotlinMapper

aharin:
Improvements to KotlinMapper:

- Support a mixed model of constructor +writable  properties set after construction.
- Allow to explicitly specify propery/parameter name using ColumnName annotation

### DIFF
--- a/kotlin/src/test/kotlin/org/jdbi/v3/core/kotlin/KotlinMapperTest.kt
+++ b/kotlin/src/test/kotlin/org/jdbi/v3/core/kotlin/KotlinMapperTest.kt
@@ -5,12 +5,14 @@ import com.nhaarman.mockito_kotlin.whenever
 import org.assertj.core.api.Assertions.assertThat
 import org.jdbi.v3.core.HandleAccess
 import org.jdbi.v3.core.statement.StatementContextAccess
+import org.junit.After
 import org.junit.Before
 import org.junit.Test
 import java.sql.ResultSet
 import java.sql.ResultSetMetaData
 
 class KotlinMapperTest {
+
 
     val resultSet = mock<ResultSet>()
     val resultSetMetaData = mock<ResultSetMetaData>()
@@ -21,6 +23,11 @@ class KotlinMapperTest {
     @Before
     fun setUpMocks() {
         whenever(resultSet.metaData).thenReturn(resultSetMetaData)
+    }
+
+    @After
+    fun debug() {
+        org.mockito.internal.debugging.MockitoDebuggerImpl().printInvocations(resultSet)
     }
 
     @Test fun testDataClassWithOnlyPrimaryConstructor() {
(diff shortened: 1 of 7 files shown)


kotlin/src/test/kotlin/org/jdbi/v3/core/kotlin/KotlinMapperTest.kt
```diff
@@ -40,6 +40,22 @@ class KotlinMapperTest {
 
     }
 
+    @Test fun shouldUseSecondOccurrenceOfColumnWhenMultipleColumnsWithSameNameArePresent() {
+
+        val mapper = KotlinMapper(DataClassWithOnlyPrimaryConstructor::class.java)
+        mockColumns("id", "name", "name")
```
qualidafial: I'm curious--how might this scenario come about in practice?

aharin: Say we want to search for occupations by name, matching on word boundaries. In Postgres you could use regex query like this: `l.name  ~* ('\\y' || :searchTerm)`. When we return occupations we want to highlight the search term within the  name using markdown syntax,  plus I want to return all other fields as in `SELECT l.*, regexp_replace(l.name, '(\\y' || :searchTerm || ')', '*\\1*', 'ig') as name` . In this case l.* will also contain `name` column, however I want to make sure that the second occurrence with the search term highlighted is the one that gets mapped.

qualidafial: Could you rework this into a full database test that demonstrates this scenario?

aharin: Sure, will do!

aharin: Database test added!

qualidafial: Now that I look at this, I'm not sure I agree with this use case.

This appears to hardcode an assumption that if a result set has two columns with the same name, we always want to use the latter. That may work in some scenarios but I don't think it's something we should encourage.

I rather think it's better to leave which column will be selected undefined, and recommend that users formulate their queries to be unambiguous.

I'm open to hear any arguments you have in favor of this, but my gut reaction is that this doesn't seem like the kind of decision we should unilaterally make for our users.

qualidafial: Apologies for the slow response--it's been an unexpectedly busy couple of weeks.

aharin: I thought that it would be a good idea to have tests showing 

- Mapper can handle duplicate columns
- The behaviour is deterministic, i.e. columns are not chosen at random
- Explicitly show which column will be used

I think the standard BeanMapper actually behaves in the same way.

If you think that we do not want to make any promises regarding mapping of duplicate columns, I will be happy to remove these tests.

qualidafial: Now that I come back to this I think I agree with you. Merging.

PR #773, opened 2017-03-15
Add configurable 'untyped null' argument

### LINKED ISSUE #299: Unable to set a null value via JDBI and Sybase JConn4
`@SqlCall("{:retVal= exec sp_CountryCredit_Insert :clientID, :country, :user}")
    OutParameters insertClientCountryCredit(@Bind("clientID") long clientID, @Bind("country") String country, @Bind("user") String user);`

when `country` is bind to NULL, then I'm getting 

>  JZ0SL: Unsupported SQL type 0. 

after a closer look, the jdbi is trying to do

>  statement.setNull(position, Types.NULL);

I override it to 

>  statement.setNull(position, Types.VARCHAR);

and then it seems to work. Is there anything that I'm missing? If not, is it possible to make jdbi smarter to accommodate this situation?

stevenschlansker:
Fixes #299 (although customization per database vendor is not included)

Also changes default from `Types.NULL` to `Types.OTHER` because I think that's a more sensible default given the prior discussion.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
@@ -16,6 +16,7 @@ package org.jdbi.v3.core.argument;
 import static org.jdbi.v3.core.internal.JdbiStreams.toStream;
 
 import java.lang.reflect.Type;
+import java.sql.Types;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CopyOnWriteArrayList;
@@ -27,6 +28,7 @@ import org.jdbi.v3.core.config.JdbiConfig;
 public class Arguments implements JdbiConfig<Arguments> {
     private final List<ArgumentFactory> argumentFactories = new CopyOnWriteArrayList<>();
     private ConfigRegistry registry;
+    private Argument untypedNullArgument = new NullArgument(Types.OTHER);
 
     public Arguments() {
         register(BuiltInArgumentFactory.INSTANCE);
@@ -40,6 +42,7 @@ public class Arguments implements JdbiConfig<Arguments> {
 
     private Arguments(Arguments that) {
         argumentFactories.addAll(that.argumentFactories);
+        untypedNullArgument = that.untypedNullArgument;
     }
 
     public Arguments register(ArgumentFactory factory) {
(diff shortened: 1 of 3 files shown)


stevenschlansker:
tag @kristwaa @dlei in case you care to take a look

core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
```diff
@@ -60,6 +63,22 @@ public Arguments register(ArgumentFactory factory) {
                 .findFirst();
     }
 
+    /**
+     * Configure the {@link Argument} to use when binding a null
+     * we don't have a type for.  Intended for vendor plugins to call.
+     * @param untypedNullArgument the argument to bind
+     */
+    public void setUntypedNullArgument(Argument untypedNullArgument) {
+        this.untypedNullArgument = untypedNullArgument;
```
qualidafial: Should probably guard against null here, just in case.

core/src/test/java/org/jdbi/v3/core/argument/TestMapArguments.java
```diff
@@ -55,6 +53,6 @@ public void testNullBinding() throws Exception
         Map<String, Object> args = Collections.singletonMap("foo", null);
         new MapArguments(args, ctx).find("foo").get().apply(3, stmt, null);
 
-        verify(stmt).setNull(3, Types.NULL);
+        verify(stmt).setNull(3, ((NullArgument)new Arguments().getUntypedNullArgument()).getSqlType());
```
qualidafial: Meh.. `Types.OTHER` is probably safe here.

core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
```diff
@@ -60,6 +63,22 @@ public Arguments register(ArgumentFactory factory) {
                 .findFirst();
     }
 
+    /**
+     * Configure the {@link Argument} to use when binding a null
+     * we don't have a type for.  Intended for vendor plugins to call.
```
qualidafial: This latter sentence, the way it's worded, may discourage users from configuring this by hand if needed. e.g. there is no `JdbiPlugin` for their database vendor, or the plugin does not configure this.

PR #775, opened 2017-03-23
SqlObjectFactory.accepts should throw only for incorrectly implemente…

### LINKED ISSUE #766: SqlObjectFactory.accepts should return false instead of throwing exception
Maybe log the exception message as a warning?

qualidafial:
…d SQL objects

Fixed #766


qualidafial:

PR #776, opened 2017-03-23
Allow annotations on SQL Object methods that override superinterface …

### LINKED ISSUE #767: SQL object types with overridden superinterface methods are treated as default
```java
public interface BaseDao<T> {
  void update(T entity);
}

public interface FooDao extends BaseDao<Foo> {
  @Override
  @SqlUpdate("update foo set name = :name where id = :id")
  void update(@BindBean Foo foo);
}
```

At runtime, `Method.isDefault()` actually does return true for `FooDao.update()`, despite it not looking like a default method.

The trick is that this override method is actually a bridge method, which accepts a more specific type (Foo) than the generic parameter on the superclass (T, which is erased to Object), and merely forwards the request to the method with the erased type.

We should update the `isDefault()` check for annotated methods, to let bridge methods through.

qualidafial:
…methods.

Fixes #767

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlMethodHandlerFactory.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlMethodHandlerFactory.java
@@ -13,14 +13,14 @@
  */
 package org.jdbi.v3.sqlobject;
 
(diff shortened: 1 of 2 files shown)


qualidafial:

PR #777, opened 2017-03-23
Add HandlerDecorators registry, so e.g. JDBI-agnostic annotations can…

### LINKED ISSUE #771: HandlerDecorator registry
There are lots of annotations out there from projects that don't depend on JDBI. It'd be nice to be able to attach decorator behavior to those third-party annotations.

Example: We use @codahale's metrics at $WORK to capture various app metrics. We've standardized mostly on the `@Timed` and `@ExceptionMetered` annotations, and there are hooks in JAX-RS that let us attach the desired behavior to those annotations in REST resource classes.

I propose we add a `HandlerDecorators` registry similar to the `Handlers` registry. The existing `HandleDecorator` interface would remain the same:

```java
public interface HandlerDecorator {
    Handler decorateHandler(Handler base, Class<?> sqlObjectType, Method method);
}
```

The current assumption is that a decorator will always return a new handler that wraps the base handler. However it's perfectly valid to simply return the base handler if the decoration is not applicable to the given method.

Thus we could register a `HandlerDecorator` that looks for the `@Timed` or `@ExceptionMetered` annotations. If the annotations are absent, it would return the base handler. Otherwise it would return an appropriate decorator for the annotation.

@stevenschlansker @arteam Thoughts?

qualidafial:
… have decorator behavior attached to them

Fixes #771 

To do:

* [x] Add proof of concept tests demonstrating how third-party annotations that have nothing to do with JDBI can have behavior attached to them.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlMethodAnnotatedHandlerDecorator.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlMethodAnnotatedHandlerDecorator.java
@@ -46,9 +46,7 @@ public class SqlMethodAnnotatedHandlerDecorator implements HandlerDecorator {
                 .map(e -> e.getAnnotation(DecoratorOrder.class))
                 .filter(Objects::nonNull)
                 .findFirst()
-                .ifPresent(order -> {
-                    annotationTypes.sort(createDecoratorComparator(order).reversed());
-                });
+                .ifPresent(order -> annotationTypes.sort(createDecoratorComparator(order).reversed()));
 
         List<HandlerDecorator> decorators = annotationTypes.stream()
                 .map(type -> type.getAnnotation(SqlMethodDecoratingAnnotation.class))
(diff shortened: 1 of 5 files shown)


qualidafial:

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlMethodAnnotatedHandlerDecorator.java
```diff
@@ -0,0 +1,86 @@
...
+            handler = decorator.decorateHandler(handler, sqlObjectType, method);
+        }
+
+        return handler;
+    }
+
+    private Comparator<Class<? extends Annotation>> createDecoratorComparator(DecoratorOrder order) {
+        List<Class<? extends Annotation>> ordering = Arrays.asList(order.value());
+
+        ToIntFunction<Class<? extends Annotation>> indexOf = type -> {
```
arteam: I think the lambda expression could be passed directly to `Comparator.comparingInt`

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlMethodAnnotatedHandlerDecorator.java
```diff
@@ -0,0 +1,86 @@
...
+            return index == -1 ? ordering.size() : index;
+        };
+
+        return Comparator.comparingInt(indexOf::applyAsInt);
+    }
+
+    private static HandlerDecorator buildDecorator(Class<? extends HandlerDecorator> decoratorClass) {
+        HandlerDecorator decorator;
+        try {
+            decorator = decoratorClass.getConstructor().newInstance();
```
arteam: We could the variable declaration and just use `return decoratorClass.getConstructor().newInstance();`

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlMethodAnnotatedHandlerDecorator.java
```diff
@@ -0,0 +1,86 @@
...
+        List<Class<? extends Annotation>> annotationTypes = Stream.of(method.getAnnotations())
+                .map(Annotation::annotationType)
+                .filter(type -> type.isAnnotationPresent(SqlMethodDecoratingAnnotation.class))
+                .collect(toList());
+
+        Stream.of(method, sqlObjectType)
+                .map(e -> e.getAnnotation(DecoratorOrder.class))
+                .filter(Objects::nonNull)
+                .findFirst()
+                .ifPresent(order -> {
```
arteam: I think we could omit the curly braces in the lambda expression

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlMethodAnnotatedHandlerDecorator.java
qualidafial: Yep, I was overthinking it.

Issue #742, opened 2017-01-26
ConstructorMapper should have public constructors


qualidafial:
The `of` methods only return factories, which means that to use in core you have to make two calls instead of one:

```java
handle.createQuery("select * from foo")
    .registerRowMapper(ConstructorMapper.of(Foo.class))
    .mapTo(Foo.class)
    .list();
```

vs. 

```java
handle.createQuery("select * from foo")
    .map(new ConstructorMapper(Foo.class))
    .list();
```

qualidafial:
Alternatively we could rename the existing `of` methods to e.g. `factoryOf`, and add new `of` methods that return a mapper, not a factory.

qualidafial:
But if we did rename it, it would have to be done before we go beta

qualidafial:
Eh, I don't wanna rename it. Moving back to release blocker milestone.

qualidafial:
Tracked in #779

PR #779, opened 2017-03-24
Add public constructors to ConstructorMapper

qualidafial:
Addresses #742 

@arteam @stevenschlansker Could one of you review this?

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java
@@ -52,7 +52,7 @@ public class BeanMapper<T> implements RowMapper<T>
      * @param type the mapped class
      * @return a mapper factory that maps to the given bean class
      */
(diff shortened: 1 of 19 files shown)


qualidafial:
So I searched the code base for `of(...)` methods, and most of them do return the self-type. The three that don't are `BeanMapper`, `FieldMapper`, and `ConstructorMapper`.

After reading your comment I think I agree that renaming those methods is probably the more attractive option, long-term.

I did find that `VendorSupportedArrayType`, `MapEntryMapper`, `GuavaArguments`, and `GuavaCollectors` all have static methods named `factory` which return a factory type. I'll rename the methods in the other three classes to match.

qualidafial:
@stevenschlansker I've renamed the `of` methods to `factory`, added `of` methods that return the self-type in their place, and made the constructors private. Do you care to review the latest?

stevenschlansker:
Great, having `of` vs `factory` is much clearer.  Thanks.

PR #780, opened 2017-03-27
Make BatchHandler extract the iterable/iterator element type and pass it to the customizer factory

### LINKED ISSUE #760: BatchHandler should extract the iterable/iterator element type and pass it to the customizer factory
From a FIXME comment from #759. See https://github.com/jdbi/jdbi/pull/759#discussion_r102803537 for context.

`Bind.Factory` and now `KotlinSqlStatementCustomizerFactory` are having to unwrap the element type depending on whether the SQL statement is a `PreparedBatch`. This responsibility rightfully belongs to `BatchHandler`.

I propose adding another parameter `Type type` to methods `ParameterCustomizerFactory.createForParameter()` and `SqlStatementCustomizerFactory.createForParameter()`. Most handlers would simply send `param.getGenericReturnType()`, but `BatchHandler` could resolve the element type itself, ensuring that `arg` in the parameter customizer is actually an instance of `type`

qualidafial:
Refactor ParameterCustomizerFactory, SqlStatementCustomizerFactory to accept parameter type. Move reponsibility of detecting batch parameters and figuring out the parameter element type to SqlBatch.

Fixes #760

### DIFF
--- a/jpa/src/main/java/org/jdbi/v3/jpa/BindJpa.java
+++ b/jpa/src/main/java/org/jdbi/v3/jpa/BindJpa.java
@@ -22,6 +22,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 
+import java.lang.reflect.Type;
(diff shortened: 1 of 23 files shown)


qualidafial:

PR #782, opened 2017-03-27
Shade core dependencies: antlr, expiringmap, and guava

### LINKED ISSUE #264: Remove dependency on Guava TypeToken
Or shade it in core, so we don't export it as a transitive dependency.

qualidafial:
Fixes #264 

Shading brings up the core JAR size from ~260K to ~1.4M.

Most of this comes from Guava, which weighs in at 2.5M. The shade plugin's tree shaking reduces this to just the 1.1M of classes that are transitively reachable through JDBI.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -196,7 +196,8 @@
                         <configuration>
                             <artifactSet>
                                 <includes>
-                                    <include>org.antlr:antlr-runtime</include>
+                                    <include>antlr:*</include>
+                                    <include>org.antlr:*</include>
                                     <include>com.google.guava:guava</include>
                                     <include>net.jodah:expiringmap</include>
                                 </includes>
@@ -206,7 +207,11 @@
                             <minimizeJar>true</minimizeJar>
                             <relocations>
                                 <relocation>
-                                    <pattern>org.antlr.runtime</pattern>
+                                    <pattern>org.antlr</pattern>
+                                    <shadedPattern>org.jdbi.v3.core.shaded.org.antlr</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>antlr</pattern>
                                     <shadedPattern>org.jdbi.v3.core.shaded.antlr</shadedPattern>
                                 </relocation>
 
@@ -223,7 +228,7 @@
                                     <shadedPattern>org.jdbi.v3.core.shaded.expiringmap</shadedPattern>
                                 </relocation>
                             </relocations>
-                            <shadedArtifactAttached>false</shadedArtifactAttached>
+                            <shadedArtifactAttached>true</shadedArtifactAttached>
                             <shadeSourcesContent>true</shadeSourcesContent>
                         </configuration>
                     </execution>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -46,7 +46,7 @@
     </licenses>
 
     <dependencies>
-        <dependency>
+        <dependency> <!-- shaded -->
             <groupId>org.antlr</groupId>
             <artifactId>antlr-runtime</artifactId>
         </dependency>
@@ -67,12 +67,12 @@
             <scope>test</scope>
         </dependency>
 
-        <dependency>
+        <dependency> <!-- shaded -->
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
 
-        <dependency>
+        <dependency> <!-- shaded -->
             <groupId>net.jodah</groupId>
             <artifactId>expiringmap</artifactId>
         </dependency>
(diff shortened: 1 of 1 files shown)


stevenschlansker:
Hm.  So, I've never been a huge fan of shading.  It causes more problems than it solves in "well-formed" Maven projects since you can no longer patch any issues in the used libraries, although it does help out people who refuse to use good tools ;)

Would it be worth having a build with a different classifier, so that users can choose the version they want?  (Is this a good use for classifier?  If not we could have a different artifact id)

qualidafial:
On the other hand, depending on popular frameworks like Guava or ANTLR can cause problems downstream, e.g. if some unrelated library/framework depends on a different, mutually incompatible version of that dependency.

We experienced this at $WORK this week--Guava removed a method in 19.0 that another framework depended on. So we're all stuck on 18.0 until they upgrade.

So what is the relative risk with JDBI in this case?

- Risks of shading
  - A bug in a shaded dependency cannot be patched in the field. 
    - Guava: they have a strong track record. I consider the probability, and the potential impact, to be low.
    - ANTLR: no idea on probability. Impact could be high, depending on the nature of the bug.
    - expiringmap: no idea on probability. Low impact.
  - The JAR file becomes _giant_.
    - Becomes a tired bullet point against JDBI on competing projects.
    - All docker images for JDBI projects get 1MB bigger just for funsies
- Risks of traditional dependency
  - Different libraries could depend on mutually incompatible versions of the same library
    - Guava: moderate risk: they're serious about deprecation, but they give users a lot of runway to upgrade. Impact would be that upgrade path is constrained if one library lags on upgrades.
    - ANTLR: Low probability--project is pretty stable. Unsure about impact.

I'm going to give this further though.

stevenschlansker:
I'm going to see if I can spike out building both as separate modules.

stevenschlansker:
@qualidafial I pushed a minor change that I think gets us the best of both worlds.  The main jar is still a normal solo jar, but it also builds a `<classifier>shaded</classifier>` jar.  End users can pick the one they like.  It costs us a little during the build but I don't think that's a huge concern.  WDYT?

qualidafial:
I believe an artifact with a classifier is technically viewed by Maven as a distinct artifact.

This becomes a problem when you depend on more than just core:

```xml
<dependency>
  <groupId>org.jdbi</groupId>
  <artifactId>jdbi3</artifactId>
  <classifier>shaded</classifier>
</dependency>
<dependency>
  <groupId>org.jdbi</groupId>
  <artifactId>jdbi3-postgres</artifactId>
</dependency>
```

Unless I'm mistaken, the `jdbi3-postgres` dependency will drag in `jdbi3` (unshaded) as a transitive dependency. Thus if you use the shaded JAR, you have to add an exclusion to every other dependency that includes JDBI core. Sadness.

Let me do a test to verify.

qualidafial:
A few things:

First, I've confirmed that the unshaded and shaded jars are treated as distinct artifacts. So if you use the shaded jar, you'll have to add exclusions on the unshaded JAR for e.g. jdbi3-postgres, jdbi3-guava, etc. Boo.

It might be possible to mitigate this by providing a BOM project that manages the exclusions for you. I'm not certain that those exclusions will actually be honored though when using the bom with import scope.

The other thing I found out is that, if you attach the shaded jar with a classifier, it throws away the dependency-reduced pom. So adding a dependency on the shaded JAR will still drag in the now-unused transitive dependencies, on top of the shaded version of those dependencies. Boo.

I want to try one more thing, which is to add a separate jdbi3-shaded artifact that uses the shade jar to inline jdbi3 and all its dependencies.

qualidafial:
Ok, I've spun up a shaded jar as a separate artifact. It doesn't work--the package relocation only works for first-party code. It won't rewrite references between packages in dependencies. e.g. it doesn't redirect the classes in jdbi3 core jar to the shaded Guava packages.

So at this point I think the best options available to us are:
- Don't use shading
- Shade all core dependencies, and don't provide an unshaded option.

I lean toward just not using shading for now. We can always add it in the future, should the need arise.

stevenschlansker:
Yuck.  I tend to agree, let's not shade -- that should be a policy decided by the end user, not something an intermediate library imposes.  We can add it later if need be.  Sucks that this situation is so difficult though.

qualidafial:
I wish we could get away from using Guava just for type tokens. It's a shame Classmate's APIs don't implement the standard types in `java.lang.reflect`.

stevenschlansker:
If there really is no alternate solution, should we consider breaking out the minimal needed code into another artifact and dropping guava?

qualidafial:
I wouldn't be opposed to that, but I tried that once and it went _way_ down the rabbit hole. `TypeToken` uses a surprising number of other types in Guava.

stevenschlansker:
😞

qualidafial:
Hmm.. taking another look at this. I might be able to pull this off with some aggressive cherry picking

qualidafial:
Bazinga! See #793

Issue #500, opened 2016-09-25
Named and positional parameters don't mix well


qualidafial:
Discovered while working on tests for #494:

```
public interface Dao {
    @SqlUpdate("update something set name=:name where id=:id")
    void updateBogus(@Bind int id, @BindMap(keys = "integerValue") Map<String, Object> map);
}

@Test
public void testBindMap_KeyNotInMapOrKeys() throws Exception {
    handle.insert("insert into something(id, name) values (3, 'Carol')");
    exception.expect(UnableToExecuteStatementException.class);
    dao.updateBogus(3, emptyMap());
}
```

This intent of the above test is to validate we get the right exception when an unbound parameter is referenced in an SQL statement. However when it ran, because `name` is unbound, it falls back to positional parameters and uses `3`, which is not correct.

I think we need to enforce that SQL statement bound parameters are either all positional, or all named--never mixed.

qualidafial:
Tracked in #787

PR #787, opened 2017-03-29
Forbid mixing named and positional parameters in SQL statements.

qualidafial:
Addresses #500.

In a nutshell: if a named parameter didn't match up with e.g. the method parameter name, JDBI would attempt to match on a positional param. This could have very unexpected results if e.g. the method arguments aren't in the same order as the query.

I'm rather surprised how many of our existing tests relied on this. :grimacing: 

This PR adds the following restrictions:
- A SQL statement must use _either_ named parameters, or positional parameters. Using both in the same statement will throw an exception.
- A statement with positional parameters will only used position-bound arguments.
- A statement with named parameters will only use name-bound arguments.

You can still bind both positional and named, but the statement will only use one category of bindings or the other.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/rewriter/InternalRewrittenStatement.java
+++ b/core/src/main/java/org/jdbi/v3/core/rewriter/InternalRewrittenStatement.java
@@ -15,12 +15,10 @@ package org.jdbi.v3.core.rewriter;
 
 import static org.jdbi.v3.core.rewriter.ParsedStatement.POSITIONAL_PARAM;
 
+import com.google.common.base.Joiner;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.util.List;
-
-import com.google.common.base.Joiner;
-import org.jdbi.v3.core.argument.Argument;
 import org.jdbi.v3.core.statement.Binding;
 import org.jdbi.v3.core.statement.StatementContext;
 import org.jdbi.v3.core.statement.UnableToCreateStatementException;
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/jdbi/v3/core/rewriter/InternalRewrittenStatement.java
```diff
@@ -36,49 +37,55 @@
...
-                    a.apply(i + 1, statement, this.context);
-                } catch (SQLException e) {
-                    throw new UnableToCreateStatementException(String.format("Exception while binding '%s'",
-                            named_param), e, context);
-                }
-                i++;
+            try {
+                a.apply(i + 1, statement, this.context);
+            } catch (SQLException e) {
+                throw new UnableToCreateStatementException(String.format("Exception while binding '%s'",
```
arteam: Nitpick: I think we need to decide with with quotes we wrap the parameter in error messages (single or double quotes).

core/src/main/java/org/jdbi/v3/core/rewriter/InternalRewrittenStatement.java
```diff
@@ -36,49 +37,55 @@
...
+        for (int i = 0; i < paramCount; i++) {
+            int index = i;
+            Argument a = params.findForPosition(i)
+                    .orElseThrow(() -> {
+                        String msg = String.format("Unable to execute, no positional parameter bound at position %s",
+                                index);
+                        return new UnableToExecuteStatementException(msg, context);
+                    });
+            try {
+                a.apply(i + 1, statement, this.context);
```
arteam: `this.` seems unnecessary.

arteam: I think `this.` superfluous, we could use just `context`.

core/src/main/java/org/jdbi/v3/core/rewriter/InternalRewrittenStatement.java
```diff
@@ -36,49 +37,55 @@
...
-                                            "the JDBC 'start at 1' scheme) has been set.",
-                                    named_param, index, index + 1);
-                            return new UnableToExecuteStatementException(msg, context);
-                        });
+            bindNamed(params, statement);
+        }
+    }
+
+    private void bindPositional(Binding params, PreparedStatement statement) {
+        int paramCount = stmt.getParams().size();
```
arteam: I guess it doesn't make much sense to extract `stmt.getParams().size();` to a variable. It's not used outside of the for loop condition.

core/src/main/java/org/jdbi/v3/core/rewriter/InternalRewrittenStatement.java
```diff
@@ -36,49 +37,55 @@
...
+        for (int i = 0; i < paramCount; i++) {
+            int index = i;
+            Argument a = params.findForPosition(i)
```
arteam: I think we could avoid declaring the argument variable and just invoke `apply` after `orElseThrow`.

brianm:
Is this lined up for 2.x or 3.x We MUST NOT do this in 2.x

qualidafial:
3.0 only

core/src/main/java/org/jdbi/v3/core/rewriter/InternalRewrittenStatement.java
qualidafial: I like single quotes better, because then I can be lazy and not escape them

PR #791, opened 2017-04-02
Generate javadocs for kotlin classes, release 3.0.0-beta1

qualidafial:
The last hurdle in the way of releasing another beta: the Sonatype OSS repository wants us to generate a javadoc jar for the kotlin artifacts.

Long term we should produce a kdoc which will give kotlin users a more Kotlin-idiomatic picture of the API. This PR just gets us over the release hurdle imposed by Sonatype.


qualidafial:
Note that 3.0.0-beta1 has already been released to Central

PR #793, opened 2017-04-02
Copy TypeToken implementation from Guava, remove Guava as core dependency

qualidafial:
So I tried doing a class-by-class copy, but that quickly spiraled out of control.

In this PR, I've only copied the portions of each class that are directly used by `GenericTypes`. This increased the jar size by 44KB, as opposed to 1.4MB larger when we were shading the entire Guava JAR.

:tada: 

On the downside.. now we have to keep an eye on Guava and follow any bugfixes they apply.


core/src/findbugs/findbugs-exclude.xml
```diff
@@ -15,4 +15,8 @@
     <Match>
         <Package name="~org\.jdbi\.v3\.core\.internal\.lexer\.*" />
     </Match>
+    <Match>
+        <!-- lifted from Google Guava -->
+        <Package name="~org\.jdbi\.v3\.core\.generic\.internal\.*" />
```
stevenschlansker: I'd usually prefer to use the `@SuppressFBWarnings` annotation over excluding an entire package, but I'm guessing you did this so that we can have fewer local modifications?

core/src/main/java/org/jdbi/v3/core/generic/internal/Preconditions.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+package org.jdbi.v3.core.generic.internal;
+
+import javax.annotation.Nullable;
+
+public final class Preconditions {
+  private Preconditions() {}
+
+  public static void checkArgument(boolean expression) {
```
stevenschlansker: Do we actually use this zero-argument variety?  Seems like we should always have a message and that using this is almost certainly wrong.

core/src/main/java/org/jdbi/v3/core/generic/internal/Preconditions.java
```diff
@@ -0,0 +1,90 @@
...
+      @Nullable String errorMessageTemplate,
+      @Nullable Object... errorMessageArgs) {
+    if (!expression) {
+      throw new IllegalArgumentException(format(errorMessageTemplate, errorMessageArgs));
+    }
+  }
+
+  public static <T> T checkNotNull(T reference) {
+    if (reference == null) {
+      throw new NullPointerException();
```
stevenschlansker: ditto here, nothing more annoying than a `NullPointerException` that doesn't tell you what you forgot to fill in

core/src/main/java/org/jdbi/v3/core/generic/internal/Preconditions.java
```diff
@@ -0,0 +1,90 @@
...
+  public static void checkState(
+          boolean expression,
+          @Nullable String errorMessageTemplate,
+          @Nullable Object... errorMessageArgs) {
+    if (!expression) {
+      throw new IllegalStateException(format(errorMessageTemplate, errorMessageArgs));
+    }
+  }
+
+  static String format(String template, @Nullable Object... args) {
```
stevenschlansker: Is this so that arguments get appended to the end if you forget a format specifier?  The implementation looks really strange to me, it is called `format` like `String.format` but if you try to use *any* documented format specifier other than `%s` it will silently behave wrong.  I'm not sure that surprise is worth the nicety of having forgotten arguments appended.

core/src/main/java/org/jdbi/v3/core/generic/internal/TypeResolver.java
```diff
@@ -0,0 +1,427 @@
...
+    Map<TypeVariableKey, Type> mappings = new HashMap<>();
+    populateTypeMappings(mappings, checkNotNull(formal), checkNotNull(actual));
+    return where(mappings);
+  }
+
+  TypeResolver where(Map<TypeVariableKey, ? extends Type> mappings) {
+    return new TypeResolver(typeTable.where(mappings));
+  }
+
+  private static void populateTypeMappings(
```
stevenschlansker: Yikes. 😱

core/src/main/java/org/jdbi/v3/core/mapper/MapEntryMapper.java
```diff
@@ -99,7 +100,7 @@ private MapEntryMapper(RowMapper<K> keyMapper, RowMapper<V> valueMapper) {
 
     @Override
     public Map.Entry<K, V> map(ResultSet rs, StatementContext ctx) throws SQLException {
-        return immutableEntry(keyMapper.map(rs, ctx), valueMapper.map(rs, ctx));
+        return ImmutableEntry.of(keyMapper.map(rs, ctx), valueMapper.map(rs, ctx));
```
stevenschlansker: You could consider using `AbstractMap.SimpleEntry` which might remove the need for the `ImmutableEntry` class.

stevenschlansker:
Also, I stumbled across this, which might be interesting to look at:
https://github.com/jhalterman/typetools

I'm not at all sure they have a better implementation than this, it's just something I saw that might be relevant.

core/src/findbugs/findbugs-exclude.xml
qualidafial: There were only a handful of errors, I don't have a problem adding annotations.

core/src/main/java/org/jdbi/v3/core/generic/internal/Preconditions.java
qualidafial: This is just copy-pasted from Guava. Looking at the usages, none of them pass extra args.

core/src/main/java/org/jdbi/v3/core/generic/internal/TypeResolver.java
qualidafial: ¯\\\_(ツ)\_/¯

core/src/main/java/org/jdbi/v3/core/mapper/MapEntryMapper.java
qualidafial: I was not aware of that implementation. Good to know.

Issue #798, opened 2017-04-16
Do we need all of Handle.execute, Handle.insert, and Handle.update?


qualidafial:
@stevenschlansker @arteam 

Both `execute` and `insert` just delegate to `update`, and return the update count.

What would you think of just making `execute()` return int, and eliminating the other two?

arteam:
I think it's a good idea. This should reduce noise and standardize the API.
I think the users will not have a hard time to update.

qualidafial:
Tracked in #800

PR #800, opened 2017-04-16
Remove Handle.update, Handle.insert in favor of Handle.execute

qualidafial:
Addresses #798


qualidafial:

Issue #799, opened 2017-04-16
OracleReturning: switch to zero-based parameter indices


qualidafial:
Not sure how I missed it during prior refactoring, but `OracleReturning.returnParameters().register(position, sqlType)` is expecting a 1-based index.

We should switch this to zero-based and do the translation behind the scenes, to stay consistent with the rest of Jdbi.

qualidafial:
Tracked in #801

PR #801, opened 2017-04-16
Use 0-based indices in OracleReturning.returnParameters()

qualidafial:
Addresses #799


qualidafial:

arteam:
Looks good, thank you!

stevenschlansker:
Is there any good way to catch code that silently breaks because of this?  If you port forward most jdbi code from v2, you can find the compiler errors and fix them ... this one is silent.

PR #807, opened 2017-05-03
hiding implementation in internal packages

### LINKED ISSUE #691: Can we exclude inner impl/factory classes from javadocs?
Inner factory classes like `BindBean.Factory` are only public so that JDBI can instantiate them to satisfy an annotation, but they shouldn't be part of the published API. They are pointless and noisy in the javadocs.

Find out if we can exclude these methods from generated javadocs. A few possibilities I found on [Stack Overflow](http://stackoverflow.com/a/26641492)

* Annotate them as `@Deprecated` and tell javadoc to exclude deprecated types. This one feels like an abuse of semantics.
* Generate javadocs with [Doxygen](http://www.stack.nl/~dimitri/doxygen/index.html). This one feels heavy-handed.

Maybe there's another approach--investigate and see if it's doable, and if it's worth it when we see what that solution looks like.

stevenschlansker:
fixes #691

do you like this approach?  it's more files, but it makes the javadoc a lot nicer.  I can finish the job if you think it's an improvement as well.


qualidafial:
I think putting them in an internal package is a reasonable workaround, I just wish we didn't have to. JDK9 introduces the `@hidden` javadoc tag--i wonder how hard it would be to use JDK9 only for docs?

stevenschlansker:
@qualidafial actually I was imagining that jdk9 would only push us further down this path -- they are public but are explicitly not part of the public API, and the most natural way to "hide" these in a module manifest will be if they are in a `internal` package anyway.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -36,7 +29,16 @@
 import java.util.function.BiConsumer;
 import java.util.stream.Stream;
 
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.extension.ExtensionFactory;
+import org.jdbi.v3.core.extension.ExtensionMethod;
+import org.jdbi.v3.core.extension.HandleSupplier;
+import org.jdbi.v3.sqlobject.config.Configurer;
+import org.jdbi.v3.sqlobject.config.ConfiguringAnnotation;
```
qualidafial: Lol, we gotta get synced up on the proper import order. I don't much care which order wins, but it's silly to have this churn all the time.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlQueryHandler.java
```diff
@@ -0,0 +1,50 @@
...
+import java.lang.reflect.Type;
+
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.result.ResultIterable;
+import org.jdbi.v3.core.statement.Query;
+import org.jdbi.v3.core.statement.StatementContext;
+import org.jdbi.v3.sqlobject.statement.UseRowMapper;
+
+public class SqlQueryHandler extends CustomizingStatementHandler<Query> {
+    private final ResultReturner magic;
```
qualidafial: this field name still makes me chuckle

qualidafial:
@stevenschlansker Were there any other classes you wanted to migrate as part of this PR before we merge?

stevenschlansker:
Yep, I only sorted through the top few packages.  I'll spruce this up a bit more shortly.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
```diff
@@ -113,6 +74,6 @@ public SqlStatementParameterCustomizer createForParameter(Annotation annotation,
          */
         THROW;
 
-        static final String valueNotHandledMessage = "EmptyHandling type on BindList not handled. Please report this to the jdbi developers.";
+        public static final String valueNotHandledMessage = "EmptyHandling type on BindList not handled. Please report this to the jdbi developers.";
```
qualidafial: This message isn't really part of the API. I think I'd rather move it to an internal package than expose it.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/transaction/TransactionalCallback.java
```diff
@@ -13,6 +13,13 @@
  */
 package org.jdbi.v3.sqlobject.transaction;
 
+/**
+ * Callback that expects an open transation and produces a result.
```
qualidafial: typo: transation -> transaction

qualidafial:
@stevenschlansker It would be better to resolve this PR before I start work on making `@RegisterXyz` annotations repeatable. Will you have time to wrap this up soon? Should I go ahead and address the PR comments myself?

stevenschlansker:
I'll pick it back up, sorry for the delay!

sqlobject/src/main/java/org/jdbi/v3/sqlobject/config/internal/ValueColumnImpl.java
```diff
@@ -0,0 +1,30 @@
...
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.mapper.MapEntryMapper;
+import org.jdbi.v3.sqlobject.config.Configurer;
+import org.jdbi.v3.sqlobject.config.ValueColumn;
+
+public class ValueColumnImpl implements Configurer {
+    @Override
+    public void configureForMethod(ConfigRegistry registry, Annotation annotation, Class<?> sqlObjectType, Method method) {
+        ValueColumn valueColumn = (ValueColumn) annotation;
+        registry.get(MapEntryMapper.Config.class).setValueColumn(valueColumn.value());
```
stevenschlansker: How about `MapEntryRowMapper` at least?  That way it at least follows conventions

qualidafial: Let's name the config class `MapEntryMappers`. Up to you whether you want to do it in this PR or track it in a separate issue.

stevenschlansker:
@qualidafial sorry for the delay, will leave open for a bit in case you want to review the new changes, but I think it's good to merge.  Not 100% complete but I got the majority of internal types out of javadoc.

core/src/main/java/org/jdbi/v3/core/h2/H2DatabasePlugin.java
```diff
@@ -11,13 +11,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jdbi.v3.core.vendor;
+package org.jdbi.v3.core.h2;
```
qualidafial: @stevenschlansker Would you be opposed to moving this to `org.jdbi.v3.h2`?

core/src/main/java/org/jdbi/v3/core/mapper/MapEntryRowMapper.java
```diff
@@ -36,7 +36,7 @@
  * @param <K> entry key type
  * @param <V> entry value type
  */
-public class MapEntryMapper<K, V> implements RowMapper<Map.Entry<K, V>> {
+public class MapEntryRowMapper<K, V> implements RowMapper<Map.Entry<K, V>> {
```
qualidafial: Not sure I agree with this rename. With the exception of `JoinRowMapper` (where `Row` belongs to `Join` more than `Mapper` IMO), all the `RowMapper` implementations omit `Row` from their names.

qualidafial:
Found some more inner classes to move. Stand by..

qualidafial:
@stevenschlansker I generated the javadoc on my machine with that last commit, and I don't see any more implementation inner classes listed.

I just have one last disagreement on the `MapEntry[Row]Mapper` name, which I outlined above. This is probably bikeshedding at this point, so feel free to merge like it is if you feel strongly about the rename.

qualidafial:
@stevenschlansker One last thing: we should add release notes for the backward breaking API changes (moving JdbiPlugin to core package, renamed MapEntryMapper)

stevenschlansker:
Backed off the naming change.  We can bikeshed about it separately so we can 🚢

stevenschlansker:
🎆

PR #833, opened 2017-06-26
Use test scope for assertj for sqlobject artifact

electrum:


qualidafial:
Good catch! Thanks for the PR

PR #837, opened 2017-06-27
Add convenience handle factory for ConnectionFactory

electrum:
This adds symmetry with the other create/open methods and allows
replacing `open().useHandle()` with try-with-resources.


qualidafial:
@electrum Thanks David!

PR #838, opened 2017-06-27
SQL Arrays: if no static element type is available (e.g. List<?>), try guessing rather than just failing.

stevenschlansker:
I'm not 100% sure of this change, but it's pretty frustrating to have `query.bind("foo", Collections.singleton("hi"))` fail even if you register all your array argument factories.

cc @strykerd

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgumentFactory.java
@@ -47,8 +47,12 @@ public class SqlArrayArgumentFactory implements ArgumentFactory {
                     .map(arrayType -> new SqlArrayArgument<>(arrayType, value));
         }
 
-        return GenericTypes.findGenericParameter(type, Collection.class)
-                .flatMap(lookup)
+        final Collection<?> collection = (Collection<?>) value;
+        // If there's no static type, guess based on the first element
+        final Type eltType = GenericTypes.findGenericParameter(type, Collection.class)
+            .orElse(collection.isEmpty() ? String.class : collection.iterator().next().getClass());
+
+        return lookup.apply(eltType)
                 .map(arrayType -> new SqlArrayArgument<>(arrayType, (Collection<?>) value));
     }
 }
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -7,6 +7,7 @@
   - [breaking] Moved JoinRowMapper.JoinRow class to top-level class
   - [breaking] Modified @Register* annotations to be repeatable, instead of using
     array attributes.
+  - array binder now guesses type based on first element of untyped collections, rather than passing
 
 3.0.0-beta1
   - [breaking] Refactored SqlStatementCustomizerFactory.createForParameter(...)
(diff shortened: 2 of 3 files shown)


qualidafial:
What about `bindByType`?

```java
query.bindByType("foo", Collections.singleton("hi"), new GenericType<Set<String>>(){})
```

On the surface I don't object to auto-discovery by observing the element types, but I have reservations about `String` when the collection is empty.

core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgumentFactory.java
```diff
@@ -47,8 +47,12 @@
                     .map(arrayType -> new SqlArrayArgument<>(arrayType, value));
         }
 
-        return GenericTypes.findGenericParameter(type, Collection.class)
-                .flatMap(lookup)
+        final Collection<?> collection = (Collection<?>) value;
+        // If there's no static type, guess based on the first element
+        final Type eltType = GenericTypes.findGenericParameter(type, Collection.class)
+            .orElse(collection.isEmpty() ? String.class : collection.iterator().next().getClass());
```
qualidafial: A couple things here:

* Let's use `orElseGet` so we don't produce a guess when we can compute the element type.
* We should go by the first _non-null_ element, otherwise we risk an NPE:

```java
Type eltType = GenericTypes.findGenericParameter(type, Collection.class)
    .orElseGet(() ->
        collection.stream()
            .filter(Objects::nonNull)
            .map(Object::getClass)
            .findFirst()
            .orElse(null));
```

* I think assuming `String` as the element type when the collection is empty or contains only nulls is a bad idea. If the collection might by empty, then users should call `bindByType`, and give JDBI a generic signature to go off of. This is consistent with the requirement that they call `bindByType` for any null parameter of a custom data type. If this is a concern, I wouldn't be opposed to a set of `bindCollection` methods:

```java
public final <T> This bindCollection(int position,
                                     Collection<T> collection,
                                     Type<T> elementType) { ... }

public final <T> This bindCollection(int position,
                                     Collection<T> collection,
                                     GenericType<T> elementType) { ... }

public final <T> This bindCollection(String name,
                                     Collection<T> collection,
                                     Type<T> elementType) { ... }

public final <T> This bindCollection(String name,
                                     Collection<T> collection,
                                     GenericType<T> elementType) { ... }
```

stevenschlansker: I'm OK with requiring explicit type binding like you suggest above, the convenience methods make it a lot nicer.  I don't think the `String` default is particularly harmful since it is only used if there are no elements, in which case no elements are bound, but I agree it looks weird.

stevenschlansker:
I think it's worth considering that when we do require `bindByType` or `bindCollection` we should probably figure out how to improve the error message if you do the wrong thing -- right now the error is not obvious at all and required me to step through with a debugger.

qualidafial:
I'm +1 to throwing an exception in SqlArrayArgumentFactory when the argument is a Collection and we can't figure out the element type.

qualidafial:
Also! We should consider adding an exception when the element type _is_ known, but there's no SQL array type registered for that element type.

PR #845, opened 2017-06-30
Fix maven coordinates for postgres plugin in docs

zikani03:
Small fix for maven coordinates for postgres plugin in docs


qualidafial:
Good catch! Thanks for the PR

PR #850, opened 2017-07-06
MapMapper: use LinkedHashMap to preserve column ordering

### LINKED ISSUE #848: DefaultMapper should extend LinkedHashMap
This would preserve the order of columns in the result set in the map. If not fixed in jdbiv2 

https://github.com/jdbi/jdbi/blob/jdbi2/src/main/java/org/skife/jdbi/v2/DefaultMapper.java#L69

it should be fixed in jdbi 3

https://github.com/jdbi/jdbi/blob/master/core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java#L75

stevenschlansker:
Fixes #848

### DIFF
--- a/core/src/test/java/org/jdbi/v3/core/statement/TestQueries.java
+++ b/core/src/test/java/org/jdbi/v3/core/statement/TestQueries.java
@@ -20,12 +20,14 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.Assertions.entry;
 import static org.jdbi.v3.core.locator.ClasspathSqlLocator.findSqlOnClasspath;
 
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/jdbi/v3/core/statement/TestQueries.java
```diff
@@ -380,4 +382,19 @@ public void testQueriesWithNullResultSets() throws Exception
...
     }
+
+    @Test
+    public void testMapMapperOrdering() throws Exception
+    {
+        h.execute("insert into something (id, name) values (?, ?)", 1, "hello");
+        h.execute("insert into something (id, name) values (?, ?)", 2, "world");
+
+        List<Map<String, Object>> rs = h.createQuery("select id, name from something")
+                              .setMaxRows(1)
```
qualidafial: This max rows disagrees with the assertion, which is why the test is failing.

stevenschlansker: Whoops, totally didn't mean to commit that.  Copy + paste error 😨

Issue #847, opened 2017-06-30
Bad Error Message on No ArgumentFactory


toadzky:
When using `@BindBean` without a registered argument factory for one of the properties, JDBI throws an error saying it can't find the property. The error should probably say that it's missing an argument factory instead. Here's a pretty printed copy of the error it currently gives:

```
! org.jdbi.v3.core.statement.UnableToExecuteStatementException: 
    Unable to execute, no named parameter matches 'contact'. 
    [statement:"INSERT INTO subscriptions (subscriber, target, expiration, contact, dialog) VALUES (:subscriber, :target, :expiration, :contact, :dialog) ON CONFLICT DO UPDATE SET expiration = :expiration, dialog = :dialog",
     rewritten:"INSERT INTO subscriptions (subscriber, target, expiration, contact, dialog) VALUES (?, ?, ?, ?, ?) ON CONFLICT DO UPDATE SET expiration = ?, dialog = ?", 
     arguments:{ 
        positional:{}, 
        named:{}, 
        finder:[{lazy bean property arguments "Subscription(
            id=0, 
            subscriber=CAMID-1cQX2fdjnSP3hoNQ, 
            target=2000, 
            expiration=2017-06-30T19:32:28.552Z, 
            contact=sip:CAMID-1cQX2fdjnSP3hoNQ@10.4.3.19:5065,
            type=Basic,
            dialog=DialogInfo(
                callId=a24040d1c005714ce12f1c142f881eb2, 
                localTag=BYHynZJ4M2HHVfJm,
                remoteTag=D9A11B48-B25C3255,
                format=application/xpidf+xml,
                cseq=2
            )
        )"]}]
```

qualidafial:
Tracked in #852

PR #852, opened 2017-07-13
Improve error messages when a bound bean property has no registered argument factory

qualidafial:
Addresses #847.

@arteam or @stevenschlansker, would you please review?


core/src/main/java/org/jdbi/v3/core/argument/BeanPropertyArguments.java
```diff
@@ -64,7 +65,16 @@ public BeanPropertyArguments(String prefix, Object bean)
 
                     try
                     {
-                        return ctx.findArgumentFor(getter.getGenericReturnType(), getter.invoke(bean));
+                        Type propertyType = getter.getGenericReturnType();
+                        Object propertyValue = getter.invoke(bean);
+                        Argument argument = ctx.findArgumentFor(propertyType, propertyValue)
```
arteam: I think an explicit check that the argument is present is better than extracting an optional value to present value and then back to `Optional`.
```java
Optional<Argument> argument = ctx.findArgumentFor(propertyType, propertyValue);
if (argument.isPresent()) {
    throw new UnableToCreateStatementException(
       String.format("No argument factory registered for type [%s] for bean property [%s] on [%s]",
       propertyType, propertyName, bean), ctx);
}
return argument;
```

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindBean.java
```diff
@@ -68,4 +77,59 @@ public void testBindBeanPrefix() {
...
     }
+
+    @Test
+    public void testNoArgumentFactoryRegisteredForProperty() {
+        handle.execute("create table beans (id integer, value_type varchar)");
+
+        assertThatThrownBy(() -> handle.attach(BeanDao.class).insert(new Bean(1, ValueType.valueOf("foo"))))
+                .hasMessageContaining("No argument factory registered");
+
+        handle.registerArgument(new ValueTypeArgumentFactory());
```
arteam: I think it's better to split this test to two: one for checking the error message and the other one for the successful execution with the registered mapper.

Issue #843, opened 2017-06-29
SqlObject + Positional Arguments + No -parameters compile flag


toadzky:
When I define this method in my SqlObject:

```java
@SqlQuery("SELECT state FROM devices WHERE owner = ?")
List<DeviceState> getDeviceStatesForUser(String userId);
```

and try to use it, I get the following error:

`A parameter was not given a name, and parameter name data is not present in the class file, for: public abstract java.util.List com.alianza.softswitch.presence.database.Devices.getDeviceStatesForUser(java.lang.String) :: java.lang.String arg0`

I do not have the `-parameters` flag turned on in my project.

qualidafial:
I've started work on this.

I don't want to spin up a maven artifact just to test this. So my plan is to disable the `-parameters` compiler flag on one of our existing artifacts, and add the test to that artifact. Blech..

qualidafial:
Tracked in #854

PR #854, opened 2017-07-14
Fix SQL Object + Positional parameter + No -parameters compile flag

qualidafial:
Addresses #843.

@arteam or @stevenschlansker: Could you please review?

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -218,16 +218,6 @@
                 </executions>
             </plugin>
             -->
-
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <configuration>
-                    <compilerArgs>
-                        <arg>-parameters</arg>
-                    </compilerArgs>
-                </configuration>
-            </plugin>
         </plugins>
     </build>
 </project>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -218,6 +218,16 @@
                 </executions>
             </plugin>
             -->
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <compilerArgs>
+                        <arg>-parameters</arg>
+                    </compilerArgs>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 </project>
--- a/noparameters/pom.xml
+++ b/noparameters/pom.xml
@@ -31,6 +31,7 @@
 
     <properties>
         <maven.deploy.skip>true</maven.deploy.skip>
+        <no_parameter_names>true</no_parameter_names>
     </properties>
 
     <dependencies>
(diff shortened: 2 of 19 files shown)


noparameters/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+~   distributed under the License is distributed on an "AS IS" BASIS,
+~   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+~   See the License for the specific language governing permissions and
+~   limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
```
arteam: It's really pity there is no "noparameters" flag for the java compiler, so we are forced to copypast the maven compiler plugin configuration in all modules.

noparameters/src/test/java/org/jdbi/v3/noparameters/TestSqlObjectNoParameterNames.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Test
+    public void namedParameterWithNoParameterNamesCompiledOrSupplied() throws Exception {
+        assertThatThrownBy(() -> h.attach(Dao.class).getByNamedId(1))
+                .isInstanceOf(UnableToExecuteStatementException.class)
+                .hasMessageContaining("no named parameter matches 'id'");
+    }
+
+    @RegisterBeanMapper(Something.class)
+    public interface Dao {
```
arteam: I think we could add a test for the error message of a missed name for the `@BindList` annotation, too.

core/pom.xml
```diff
@@ -218,6 +218,16 @@
                 </executions>
             </plugin>
             -->
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
stevenschlansker: Could we avoid the copypasta by activating this in the parent with a profile?  The profile could be active by default and suppressed by a `.javac-no-parameters` flag file or the like.

qualidafial: I'll try it.

qualidafial: That worked. I used a property that's only defined in the noparameters pom. Thanks

noparameters/src/test/java/org/jdbi/v3/noparameters/TestSqlObjectNoParameterNames.java
qualidafial: Agreed. Probably likewise for `@Define` and `@DefineList`.

core/pom.xml
qualidafial: Spoke too soon. From what I can tell, the set of activated profiles is locked in at the parent pom--it doesn't look like you can change it in individual artifacts. Reinstating copypasta. _sad panda_

Issue #774, opened 2017-03-23
Document how to use Map collectors (core and sqlobject)


qualidafial:
PR #769 added support for `Map` return types, where the key and value types may map individual columns or some subset of columns from the result row.

Add a section to docs that demonstrates:
* The three return type use cases from #763 
* What the equivalent code would be in the core API

qualidafial:
Starting work on this

qualidafial:
Tracked in #862

PR #862, opened 2017-07-28
Document how to use Map<K,V> collectors

qualidafial:
Addresses #774 

Adds documentation for `Map<K,V>` result types for both Core and SQL Object.

Also added two convenience methods to `Configurable`: `setMapKeyColumn(String)` and `setMapValueColumn(String)`.

@stevenschlansker or @arteam, could one of you please review?

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/config/Configurable.java
+++ b/core/src/main/java/org/jdbi/v3/core/config/Configurable.java
@@ -15,9 +15,6 @@ package org.jdbi.v3.core.config;
 
 import java.lang.reflect.Type;
 import java.util.function.Consumer;
-
-import org.jdbi.v3.core.statement.SqlParser;
-import org.jdbi.v3.core.statement.TimingCollector;
 import org.jdbi.v3.core.argument.ArgumentFactory;
 import org.jdbi.v3.core.argument.Arguments;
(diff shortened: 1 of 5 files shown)


qualidafial:
@stevenschlansker @arteam Ping

qualidafial:
@stevenschlansker It sounds like you're advocating making the key and value column names properties of a config object. This actually _is_ how it works: `MapEntryMapper` looks up the key / value column names from `MapEntryMappers` config class, and responds accordingly. This was also the simplest way to make `@KeyColumn` and `@ValueColumn` annotations to work.

qualidafial:
And if you meant it the other way around, there's a few reasons I did it that way:

* The `Map.Entry` mapper factory is installed out of the box.
  * It has to be a factory in order to be able to map arbitrary key and value types
  * You don't want to be setting one-off properties on a shared factory
* This way, `@KeyColumn` and `@ValueColumn` are simple configuring annotations. Were they not, we'd have to have special one-off code inside sqlobject to interpret them depending on whether neither, one, the other, or both are present.

I actually think we should endorse using config properties as a best practice, for when an object is registered globally (e.g. mappers) but can be configured locally.

This is similar in spirit to using `Provider<T>` in dependency injection, e.g. when a request-scoped object is used by a singleton.

PR #866, opened 2017-08-01
Fix bogus type reference in javadoc

### LINKED ISSUE #863: Javadoc comments are incorrect on the TransactionHander interface
The comment reads `The default implementation, <code>ConnectionTransactionHandler</code>...` but no such class exists.  It should probably also use `@link` instead of code.

qualidafial:
Fixed #863 

@stevenschlansker or @arteam, please review


arteam:
LGTM

PR #868, opened 2017-08-04
jdbi3 replicator: creates anything you want

### LINKED ISSUE #841: Automatic generation of result classes
SQL Objects are really nice to use until you need to return a tuple. Then you have to create a custom class and mapper (or use `BeanMapper` and sacrifice immutability).

I'm envisioning something where you just write an interface and Jdbi generates the implementation. We'd want automatic mapping from result set names to getters, but allow overrides with annotations. Additionally, we need support for type conversion. For example, how to turn a `varbinary` into a `UUID`.

I started prototyping this many years ago (predates SQL Objects) but never got very far: https://github.com/electrum/dbmagic

Jackson has a similar idea: https://github.com/FasterXML/jackson-modules-base/tree/master/mrbean

stevenschlansker:
fixes #841

cc @electrum

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/replicator/ReplicatedTypes.java
+++ b/core/src/main/java/org/jdbi/v3/core/replicator/ReplicatedTypes.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.replicator;
+
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import org.jdbi.v3.core.config.JdbiConfig;
+import org.jdbi.v3.core.generic.GenericTypes;
+import org.jdbi.v3.core.mapper.RowMapper;
+
+/**
+ * Configuration object for the replicator.
+ */
+public class ReplicatedTypes implements JdbiConfig<ReplicatedTypes> {
+
+    private final Map<Type, Optional<RowMapper<?>>> replicatedTypes = new HashMap<>();
+
+    public ReplicatedTypes() { }
+    public ReplicatedTypes(ReplicatedTypes other) {
+        replicatedTypes.putAll(other.replicatedTypes);
+    }
+
+    @Override
+    public ReplicatedTypes createCopy() {
+        return new ReplicatedTypes(this);
+    }
+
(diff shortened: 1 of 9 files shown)


electrum:
Very nice work!

One big drawback of this approach is they will be very expensive in terms of access and memory usage. For example, if stored for later usage, perhaps as part of another data structure. Can we generate concrete classes for these that use proper types (primitives as appropriate)?

core/src/main/java/org/jdbi/v3/core/replicator/ReplicatedTypes.java
```diff
@@ -0,0 +1,52 @@
...
+    public ReplicatedTypes(ReplicatedTypes other) {
+        replicatedTypes.putAll(other.replicatedTypes);
+    }
+
+    @Override
+    public ReplicatedTypes createCopy() {
+        return new ReplicatedTypes(this);
+    }
+
+    Optional<RowMapper<?>> mapperFor(Type type) {
```
qualidafial: It would be good to be able to register a replicated type as an alternative to an annotation on the interface. That way a model class can be agnostic to jdbi and use e.g. Google AutoValue or Immutables.io annotations.

qualidafial:
@stevenschlansker Been thinking about it--should the replicator stuff be in its own artifact?

My design philosophy for Jdbi has been "no magic," or more specifically, "visible magic only." I loathe the Spring approach where just adding something to the classpath magics up your shit.

My rule of thumb: there must be at least one line of code in your app, that explains why Jdbi behaves the way it does.

In that sense, `Jdbi.installPlugins()` is a compromise--rolling up all plugins into a single method call. But the app developer is making the call.

I'm now on the fence about my earlier suggestion that ReplicatorPlugin could be installed by default. There _is_ still one line of code to suggest what's happening: the `@Replicated` annotation on the interface. But I have some doubts about whether an annotation is visible enough. A method call to install the plugin would tend to reside in the "JDBI setup" part of a code base, and thus be more discoverable.

For those who choose to use `Jdbi.installPlugins()`, I think that providing the plugin but making it discoverable through `ServiceLoader` would be surprising to the user. Therefore, I suggest we move all this stuff to a new artifact (e.g. `jdbi3-replicator`) and provide a ServiceLoader manifest.

Also: what about a name like "auto mapper?"

arteam:
I'm in the same boat as @qualidafial. I don't think we should provide things which work "automagically". To fight boilerplate and provide convenient "value classes", recently there has been a shift in the Java community to use annotation processors like Google's [AutoValue](https://github.com/google/auto), [Immutables](https://immutables.github.io/) and Ryan Harter's [auto-value-gson](https://github.com/rharter/auto-value-gson). I think we should consider moving in the same direction and use the source class generation approach. In my option, it's more safe, more obvious for the user, and provides better performance instead of using reflection in runtime.

stevenschlansker:
Thanks for the feedback.  I'm shelving this for now as it won't be a quick win and will focus on burning down the release blockers instead.  (I'll be out for the next week but will try to write some on the plane and such)

alwins0n:
just a quick thought: this functionality could be achieved using the Tuple projection of the vavr plugin. Which basically gives you generic but typesafe query result objects

qualidafial:
@stevenschlansker Are you still interested in pursuing this? I'm kind of leaning toward the Immutables integration rather than this approach with proxies.

stevenschlansker:
I agree, proper Immutables support is more interesting than this approach, at least for now.

PR #869, opened 2017-08-05
Refactor CollectionColumnMapper into CollectorColumnMapper

qualidafial:
While looking into #733, I realized we were duplicating work between SQL arrays and collectors.

We have a `CollectionColumnMapper` class which maps a handful of collection types for array column. We also have `GuavaMappers` which supports `ImmutableList<T>` array columns.

I've refactored the above mapper into `CollectorColumnMapper`, which relies on `JdbiCollectors` registry as well as `ColumnMappers` registry, similar to how SQL Object does to resolve method return types.

With this change, registering a `CollectorFactory` for a given container type will add support for both query return types, and SQL array column mapping.

@stevenschlansker or @arteam, would you please review?


qualidafial:
@stevenschlansker @arteam ping

qualidafial:
Last commit addresses #733

Issue #872, opened 2017-08-10
Need Access To More Binding Param Details


TigerC10:
I'm trying to write a StatementRewriter, but I need more access to the `Binding` params.  There are no getters to access the param data unless you _already_ know the positional index or the named key.

Would be great to have a few more getters on the `Binding` class.

mattmills1023:
👍

qualidafial:
Is this for v2? `StatementRewriter` no longer exists in v3--it was split into `TemplateEngine` and `SqlParser`.

TigerC10:
Ah, yeah, looks like I was using `2.78`.  Maybe #715 needs to be updated to include the bit about `StatementRewriter`, as it looks to be in use by some projects (like [DropWizard](https://github.com/dropwizard/dropwizard/blob/master/dropwizard-jdbi/src/main/java/io/dropwizard/jdbi/NamePrependingStatementRewriter.java)).

qualidafial:
@TigerC10 good catch. I added a comment on #715 to include that in the migration guide.

PR #873, opened 2017-08-10
#872: Adding getters to Binding class

TigerC10:
Adding getters to the `Binding` class for #872


qualidafial:
If this was intended for v2, then `master` is the wrong git branch.

I don't think `Binding` is the right place for this in v3. It's already being collected into `ParsedSql`--I think we just need to refactor `StatementContext.getParsedSql` to return a `ParsedSql` with the SQL string and all bound parameter data, rather than the `String` it returns now.

TigerC10:
Closing this in favor of #877.

Issue #875, opened 2017-08-14
StatementContext.getParsedSql() should return ParsedSql instead of  String


qualidafial:
Per #872, #873

qualidafial:
tracked in #875

PR #878, opened 2017-08-14
Change StatementContext.getParsedSql() to return ParsedSql instead of String

TigerC10:
Addresses #875


TigerC10:
@qualidafial would this suffice for #875?

qualidafial:
Thanks @TigerC10!

TigerC10:
Glad to help out 👍

PR #882, opened 2017-08-24
fixed incorrect urls in SqlParser and TemplateEngine apidoc links

kimmoeklund:
Fixed apidoc links to SqlParser and TemplateEngine. Tested with local build.


qualidafial:
Thanks @kimmoeklund!

PR #885, opened 2017-09-05
Add a simple Collector example. Fixes #726

### LINKED ISSUE #726: Developer guide: Collectors, collector factories

stevenschlansker:


stevenschlansker:
It looks like Travis is failing due to OOM.  I don't think this change could cause that?  I wonder if something changed on their end, or why it suddenly started failing...

qualidafial:
Travis is failing on master, on a build that previously passed, so it's definitely not caused by this PR.

PR #893, opened 2017-09-27
PreparedBatch: throw an exception if you add() a batch with no bindings at all

stevenschlansker:
reported by @mikebell90


stevenschlansker:
Alternately, we could just silently do nothing instead of throwing.  That'd be more friendly and forgiving at the expense of allowing multiple forms that do the same thing, which I'm usually against...

PR #904, opened 2017-10-05
Make Handle.close() play nice with container managed transactions

### LINKED ISSUE #881: Transaction check in Handle.close() does not play nice with container managed transactions
See https://github.com/HubSpot/guice-transactional

guice-transactional provides a method interceptor around `@javax.transaction.Transactional`-annotated methods, and pins the thread to a single connection with an open transaction for the duration of the method.

However, in v3 we added a sanity check when you close a handle: if a transaction is still open, roll back and throw an exception. This forces developers to explicitly resolve open transactions rather than relying on Jdbi to decide for them.

Investigate whether v3 plays nice with guice-transactional. If not, one possible workaround is to disable the transaction check on close, if the transaction was already started when the handle was constructed.

qualidafial:
Fixes #881

### DIFF
--- a/core/src/test/java/org/jdbi/v3/core/transaction/TestTransactionsAutoCommit.java
+++ b/core/src/test/java/org/jdbi/v3/core/transaction/TestTransactionsAutoCommit.java
@@ -16,6 +16,7 @@ package org.jdbi.v3.core.transaction;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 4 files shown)


qualidafial:
I reached out privately to the [person on the mailing list](https://groups.google.com/forum/?fromgroups#!topic/jdbi/stgm0wsZbww) who reported this problem. He no longer works at that job and cannot confirm this fix.

qualidafial:
@stevenschlansker @arteam Ping. Can we merge this?

stevenschlansker:
OK to merge by me.  I still wish we had a report that it's confirmed to fix a problem for someone.

core/src/test/java/org/jdbi/v3/core/transaction/TestTransactionsAutoCommit.java
```diff
@@ -63,7 +64,7 @@ public void restoreAutoCommitInitialStateOnUnexpectedError() throws Exception
 
         // expected behaviour chain:
         // 1. store initial auto-commit state
-        inOrder.verify(connection).getAutoCommit();
+        inOrder.verify(connection, atLeastOnce()).getAutoCommit();
```
stevenschlansker: mocks :unamused:

PR #924, opened 2017-10-29
Developer guide: row mappers and factories

### LINKED ISSUE #717: Developer reference: Query

qualidafial:
Fixes #717 (in that it removes redundant todos that are already covered elsewhere)

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -541,15 +541,13 @@ LocalDate releaseDate = handle.createQuery(
     .findOnly();
 ----
 
-TODO:
-
-* demonstrate providing a column mapper inline
-* demonstrate mapToBean()
-* demonstrate mapToMap()
-
 === Mappers
 
-Jdbi makes use of mappers to convert result data into Java objects.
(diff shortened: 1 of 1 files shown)


docs/src/adoc/index.adoc
```diff
@@ -659,24 +919,159 @@ Out of the box, Jdbi supports mapping database columns to the following types:
...
+
+[source,java]
+----
+Type embeddedType = GenericTypes.resolveType(Optional.class.getTypeParameters()[0], type);
+----
+
+The `GenericTypes.resolveType` method accepts a generic type parameter to
+resolve, and a context type in which to resolve it.
+
+The expression `Optional.class.getTypeParameters()[0]` gives the type variable
```
stevenschlansker: This is great explanation, but it feels repetitive from the same text around line 723.  Maybe we should just have a section on generic type handling and refer it from the two locations?

qualidafial: Will do

qualidafial:
@stevenschlansker I've moved out the `GenericType` / `GenericTypes` stuff to a separate section. Would you care to review before I merge?

PR #926, opened 2017-10-31
Add the abilty to have SqlObject methods accept a Consumer<T>

stevenschlansker:
This is handy when you want only a side effect.  You could already
do this with Stream or Iterable's forEach, but then you end up
owning a Closeable you don't really care for

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -2820,6 +2820,18 @@ If you like StringTemplate, the <<StringTemplate 4>> module also provides a
 SqlLocator, which can load SQL templates from StringTemplate 4 files on the
 classpath.
 
+=== Consumer Methods
+
+As a special case, you may provide a `Consumer<T>` argument in addition to other bound parameters.
+The provided consumer is executed once for each row in the result set.
+The static type of parameter T determines the row type.
+
+[source,java]
+----
+@SqlQuery("select id, name from users")
+void forEachUser(Consumer<User> consumer);
+----
+
 === Default Methods
 
 Occasionally a use case comes up where SQL Method annotations don't fit. In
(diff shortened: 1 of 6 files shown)


docs/src/adoc/index.adoc
```diff
@@ -1628,6 +1628,16 @@ As in the Core API, neither `@BindMap`, `@BindBean`, `@BindFields`, nor
...
 
+As a special case, you may provide a `Consumer<T>` argument in addition to other bound parameters.
+The provided consumer is executed once for each row in the result set.
+The static type of parameter T determines the row type.
+
+[source,java]
+----
+@SqlQuery("select id, name from users")
+void forEachUser(Consumer<User> consumer);
+----
```
qualidafial: `@SqlQuery` hasn't been properly introduced by this point in the guide. This would probably be a better fit in the next section.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
```diff
@@ -226,4 +246,30 @@ protected Type elementType(StatementContext ctx)
...
     }
+
+    static class ConsumerResultReturner extends ResultReturner
+    {
+        private final int consumerIndex;
+        private final Type elementType;
+
+        ConsumerResultReturner(Method method, int consumerIndex) {
+            this.consumerIndex = consumerIndex;
+            elementType = GenericTypes.findGenericParameter(method.getParameters()[consumerIndex].getParameterizedType(), Consumer.class)
```
qualidafial: Meh: could use `method.getGenericParameterTypes()[consumerIndex]`

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
```diff
@@ -226,4 +246,30 @@ protected Type elementType(StatementContext ctx)
...
+                    .orElseThrow(() -> new IllegalStateException("Couldn't find concrete type for Consumer<T>"));
+        }
+
+        @Override
+        protected Object result(ResultIterable<?> iterable, StatementContext ctx) {
+            @SuppressWarnings("unchecked")
+            Consumer<Object> consumer = (Consumer<Object>)
+                ctx.getConfig(SqlObjectStatementConfiguration.class).getArgs()[consumerIndex];
+            iterable.forEach(consumer);
```
qualidafial: Very slick.

PR #931, opened 2017-11-01
fixup mapped type in docs for .list() call

alwins0n:


docs/src/adoc/index.adoc
```diff
@@ -3499,7 +3499,7 @@ Last but not least we can now project simple queries to Vavr tuples like that:
 // given a 'tuples' table with t1 int, t2 varchar, t3 varchar, ...
 List<Tuple3<Integer, String, String>> tupleProjection = handle
         .createQuery("select t1, t2, t3 from tuples")
-        .mapTo(new GenericType<List<Tuple3<Integer, String, String>>>() {})
+        .mapTo(new GenericType<Tuple3<Integer, String, String>>() {})
         .list();
```
qualidafial: I've merged this.

As long as we're using `GenericType`, one possible alternative is to use the `collectInto()` method for brevity:

```java
List<Tuple3<Integer, String, String>> tupleProjection = handle
        .createQuery("select t1, t2, t3 from tuples")
        .collectInto(new GenericType<List<Tuple3<Integer, String, String>>>() {});
```

That way you just have `collectInto` instead of both `mapTo` and `list`.

alwins0n: you are absolutely correct! i think this is what i was going for in the first place but changed it to show different ways of usage.

However, it makes much more sense to use `collectInto()` here, because as you pointed out, `GenericType` is used here anyway.

feel free to change!

PR #935, opened 2017-11-03
improve transaction check

mseele:
see https://groups.google.com/forum/#!topic/jdbi/nkNAruHZc5I


qualidafial:
This PR fails unit tests. I'm really not comfortable changing core behavior like this for a corner case--not without a more convincing explanation.

I think your scenario warrants writing a custom TransactionHandler for your app instead.

I'd also be ok with adding a new transaction handler, with this behavior.

mseele:
I'm fine with having my own custom TransactionHandler (this is how it's implemented right now). Just made the PR's because Steven Schlansker ask me to do so.
For me it's ok to delete the PR's.

qualidafial:
Ok. Thank you all the same for taking the time to submit PRs!

PR #945, opened 2017-11-05
Update README.md for v3

qualidafial:


qualidafial:
@stevenschlansker or @arteam, would you care to review?

PR #950, opened 2017-11-08
Support binding nested properties with bindBean, bindFields, and bindMethods

### LINKED ISSUE #122: Add support for composite types in @BindBean
It would be nice if composite classes would be supported using the @BindBean annotation.

For example:

```
select * from entity where field = :bean.compositeAttribute.id
```

I would be very convenient if JDBI would be able to traverse the object's relationships and bind the result to the query.

### LINKED ISSUE #395: Is there an effort to implement recursive @Bindbean
Currently in DAOs we have to pass nested objects of db objects using separate parameters. Is there any ongoing effort to make @Bindbean recursive so that we can access nested fields inside DAOs?

qualidafial:
Fixes #395.
Fixes #122.

This required a pretty heavy refactor of `BeanPropertyArguments`, `ObjectFieldArguments`, `ObjectMethodArguments`, and their superclasses.

In addition to the goal of supporting nested bindings, this PR:

* Introduces some caching of class members / property descriptors, to speed up binding and reduce redundant introspection.
* Hides some protected methods that should have been package private

@stevenschlansker or @arteam would you please review?

Putting this in the post v3 release milestone so we don't hold up release. If you think it's ready to go now, I'm fine with merging and including it in v3.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/argument/BeanPropertyArguments.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/BeanPropertyArguments.java
@@ -13,13 +13,22 @@
  */
 package org.jdbi.v3.core.argument;
 
+import static java.util.stream.Collectors.toMap;
+
 import java.beans.BeanInfo;
 import java.beans.IntrospectionException;
 import java.beans.Introspector;
 import java.beans.PropertyDescriptor;
 import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Stream;
 
+import net.jodah.expiringmap.ExpirationPolicy;
+import net.jodah.expiringmap.ExpiringMap;
 import org.jdbi.v3.core.statement.StatementContext;
 import org.jdbi.v3.core.statement.UnableToCreateStatementException;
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/jdbi/v3/core/argument/BeanPropertyArguments.java
```diff
@@ -29,7 +38,24 @@
  */
 public class BeanPropertyArguments extends MethodReturnValueNamedArgumentFinder
 {
-    private BeanInfo info;
+    private static final Map<Class, Map<String, PropertyDescriptor>> CLASS_PROPERTY_DESCRIPTORS = ExpiringMap.builder()
+        .expiration(10, TimeUnit.MINUTES)
+        .expirationPolicy(ExpirationPolicy.ACCESSED)
+        .entryLoader((Class type) -> {
```
stevenschlansker: raw types cause warnings, will `Class<?>` work for cases like these?

core/src/main/java/org/jdbi/v3/core/argument/ObjectFieldArguments.java
```diff
@@ -25,48 +37,52 @@
  */
 public class ObjectFieldArguments extends ObjectPropertyNamedArgumentFinder
 {
+    private static final Map<Class, Map<String, Field>> CLASS_FIELDS = ExpiringMap.builder()
+        .expiration(10, TimeUnit.MINUTES)
+        .expirationPolicy(ExpirationPolicy.ACCESSED)
+        .entryLoader((Class type) ->
```
stevenschlansker: ditto

PR #951, opened 2017-11-09
Support mapping of `@Nested` properties in reflection mappers.

### LINKED ISSUE #949: Embedded entity support in reflective mappers
Jdbi's reflective mappers don't have anything analogous to JPA's `@Embedded` annotation. Currently all properties/fields/constructor arguments ("mapped element" for short) have to be column-mapped.

The canonical example model:

```java
class Employee {
  long id;
  @Embedded EmploymentPeriod period;
}

@Embeddable class EmploymentPeriod {
  LocalDate startDate;
  LocalDate endDate;
}
```

By default, the embedded property reads from the column used by the JPA `@Column` annotation. However, the column names can be overridden using `@AttributeOverride` annotations.

I think we could achieve something similar using a `@Nested` annotation, with an optional column name prefix attribute.

* For each mapped element
  * If the element is annotated `@Nested`
    * Map the element with a row mapper of the same type, applying the column name prefix from the annotation if defined.
  * Else if there is a column mapper for the element type
    * Map the element with that column mapper (i.e. the current behavior)
  * Else skip the element (constructor mapper will throw instead of skip)

Coming back to our earlier example:

```java
class Employee {
  long id;
  @Nested EmploymentPeriod period;
}

class EmploymentPeriod {
  LocalDate startDate;
  LocalDate endDate;
}
```

* `BeanMapper.of(Employee.class)` could map from columns `id`, `start_date`, and `end_date`.
* `BeanMapper.of(Employee.class, "emp")` could map from `emp_id`, `emp_start_date`, and `emp_end_date`.

If we provide a column name prefix on the annotation:

```java
class Employee {
  long id;
  @Nested("period") EmploymentPeriod period;
}
```

* `BeanMapper.of(Employee.class)` could map from `id`, `period_start_date`, and `period_end_date`.
* `BeanMapper.of(Employee.class, "emp")` could map from `emp_id`, `emp_period_start_date`, and `emp_period_end_date`.

Each reflective mapper would map embedded entities with the same mapper type: e.g. `BeanMapper` maps nested entities with another `BeanMapper`.

qualidafial:
Fixes #949 

Introduces the `@Nested` annotation, per #949.

Updates BeanMapper, ConstructorMapper, and FieldMapper to honor the `@Nested` annotation on accessors, constructor parameters, and fields (respectively).

Refactors some common elements of the three mappers out to a package private utility class.

Note: With this feature, the "strict checking" config option of `ReflectionMappers` probably needs to be revisited. We might possibly even want to remove strict check v3 if we can't reconcile its semantics when nested properties are involved.

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,7 +1,9 @@
 3.0.0-rc2
   - Row and column mapper for Optional types
-  - Binding of nested attributes e.g. ":user.address.city" with @BindBean, @BindMethods,
-    and @BindFields.
+  - Binding of nested attributes e.g. ":user.address.city" with bindBean(), bindMethods(),
+    bindFields(), (and by extension, @BindBean, @BindMethods, and @BindFields).
+  - Mapping of nested attributes with BeanMapper, ConstructorMapper, and FieldMapper, using
+    the @Nested annotation.
 
 3.0.0-rc1
   - SQL Object methods may have a Consumer<T> instead of a return type. See
--- a/core/src/main/java/org/jdbi/v3/core/mapper/Nested.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/Nested.java
@@ -1,0 +1,35 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.mapper;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+/**
+ * Signals that the annotated element is a nested mapped type.
+ */
+@Retention(RUNTIME)
+@Target({PARAMETER, FIELD, METHOD})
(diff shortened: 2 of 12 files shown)


qualidafial:
@stevenschlansker or @arteam please review.

Please also note the last paragraph of the PR description--this decision could impact v3 release.

qualidafial:
@stevenschlansker I could see strict checking working with nested objects as follows:

* Reflection mappers exclude columns that do not begin with the specific prefix from the strictness check. We probably want to do this regardless.
* Thus, nested mappers which use a prefix own the strictness check for column that begin with that prefix
* Parent mappers exclude columns from their own strictness check that begin with a nested entity's prefix.

This breaks down completely if the nested entity does not use a prefix:

* The parent mapper will consider all columns it would normally check as belonging to the nested entity
* The nested entity will fail the strict check for columns mapped by the parent entity--since they have the same prefix as the parent.

So I think we should throw an exception if strict check is enabled, and any nested entities do not declare a prefix.

qualidafial:
I've got strict matching on nested beans working on bean mapper--this will work.

I'll update the PR shortly.

qualidafial:
@stevenschlansker I've updated strict column matching. Could you review the latest?

qualidafial:
@stevenschlansker ping

qualidafial:
@stevenschlansker I've modified the strict check again. This latest should provide a more strictly correct strict check--without the requirement that nested entities have a prefix:

* Only the root reflection mapper does strictness checks--not the nested mappers.
* The strictness check excludes columns that do not begin with the root mapper's prefix.
* Nested mappers are specialized with a separate, private method, which accepts a `List<String> unmatchedColumns`.
* The root mapper and nested mappers remove elements from `unmatchedColumns` for every column they match to an attribute.
* After the root mapper and all nested mappers have been specialized, fail the strict check if any unmatched columns remain.

@jdbi/contributors As part of this change, I've also changed the semantics of how column name prefixes work, so I wanted to run it by the whole team:

* Prefixes are now prepended to the java name, instead of subtracted from the column name before column matching.
* `SnakeCaseColumnNameMatcher` now ignores underscores in the column name _and_ the java name.

The first was done so that column names like `contact_addr_street` didn't require underscores in prefixes, e.g. `@BindBean("contact_")` and `@Nested("addr_")`.

However that broke a few tests that were already using underscores in prefixes. The second fixed it so you could provide the prefix if you wanted to, but you don't have to.

PR #956, opened 2017-11-12
SQL Object honor annotations from supertypes.

### LINKED ISSUE #438: Honor sqlobject annotations in supertypes

### LINKED ISSUE #825: Developer guide: implementing a generic DAO interface
```java
package com.app;

public interface CrudDao<T, ID> {
    @SqlUpdate
    @UseClasspathSqlLocator
    void insert(T entity);

    @SqlQuery
    @UseClasspathSqlLocator
    T getById(ID id);

    @SqlQuery
    @UseClasspathSqlLocator
    List<T> list();

    @SqlUpdate
    @UseClasspathSqlLocator
    void update(T entity);

    @SqlUpdate
    @UseClasspathSqlLocator
    void delete(ID id);
}

@RegisterBeanMapper(Contact.class)
public interface ContactDao extends CrudDao<Contact, Long> {}

@RegisterBeanMapper(Account.class)
public interface AccountDao extends CrudDao<Account, UUID> {}
```

In this example, SQL is loaded from files on the classpath, depending on the method called and which DAO subclass is used.

* com/app/ContactDao/insert.sql
* com/app/ContactDao/getById.sql
* com/app/ContactDao/list.sql
* com/app/ContactDao/update.sql
* com/app/ContactDao/delete.sql
* com/app/AccountDao/insert.sql
* com/app/AccountDao/getById.sql
* com/app/AccountDao/list.sql
* com/app/AccountDao/update.sql
* com/app/AccountDao/delete.sql

Note that `@UseClasspathSqlLocator` is repeated on every method instead of once on the `CrudDao` type. This is because right now, we only look for annotations on the SQL object type itself, and not  supertypes (see #438). If a DAO subtype overrides a method from a supertype, we only look at the annotations present in the subtype.

In the future we may change this to look for annotations in the supertype, in which case we can update this example.

qualidafial:
Fixes #438 
Fixes #825 

@stevenschlansker or @arteam: please review.


sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/internal/TimestampedConfig.java
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.sqlobject.customizer.internal;
+
+import java.time.Clock;
+
+import org.jdbi.v3.core.config.JdbiConfig;
+
+public class TimestampedConfig implements JdbiConfig<TimestampedConfig> {
```
arteam: It looks to me the change in the `@Timestamped` annotation should be a part of a separate PR.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/CustomizingStatementHandler.java
```diff
@@ -56,15 +57,21 @@
...
-        statementCustomizers = concat(methodCustomizers, parameterCustomizers(type, method))
-            .collect(Collectors.<BoundCustomizer>toList());
+        final Stream<BoundCustomizer> methodCustomizers =  annotationsFor(method)
+            .map(a -> instantiateFactory(a).createForMethod(a, type, method))
+            .map(BoundCustomizer::of);
+
+        final Stream<BoundCustomizer> parameterCustomizers = parameterCustomizers(type, method);
+
+        statementCustomizers = Stream.of(typeCustomizers, methodCustomizers, parameterCustomizers)
+            .flatMap(Function.identity())
```
arteam: For my personal taste, `.reduce(Stream.empty(), Stream::concat)` looks more natural.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/internal/TimestampedConfig.java
qualidafial: This config class is in the `internal` package, so it's not actually part of the public API.

qualidafial: Although the change to make `@Timestamped` applicable to the type, that part should be separate.

Yeah, I'll revert. _sigh_

PR #959, opened 2017-11-21
 Make bindList work with arbitrary SqlParser implementations

### LINKED ISSUE #958: BindList assumes ColonPrefix naming
[SqlStatement.bindList](https://github.com/jdbi/jdbi/blob/master/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java#L1263) always uses `:` for naming, which breaks with other parsers.

brianm:
Fixes #958


brianm:
Yeah, that thread made me go hmm :-)

Issue #840, opened 2017-06-28
Allow SQL Object interfaces to be private


electrum:
We follow the pattern of using DAO objects as an implementation detail of a public API. Forcing them to be public violates encapsulation and is a departure from v2.

The check was introduced in 473be6c39, but for seemingly no reason. Creating proxies for non-public interfaces works fine, as long as a security manager doesn't explicitly deny it. I see no reason for Jdbi to have such a policy.

qualidafial:
From the [Proxy javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Proxy.html):

> Proxy classes are non-public, final, and not abstract if any of the proxy interfaces is non-public.

> If a proxy class implements a non-public interface, then it will be defined in the same package as that interface.

These two facts combined mean that our `InvocationHandler` does not have access to your package-private interface.

I'd be more than happy to be proven wrong, but we did spend a lot of time trying to work around this, and came up empty.

electrum:
Is this a Java 9 issue? In Java 8, absent a security manager, you can do anything (using reflection or code generation). But I'm likely misunderstanding the issue. Do you have an example of something that doesn't work?

qualidafial:
Sure. Comment out the check in `SqlObjectFactory.accepts()`. Then open `TestRegisterBeanMapper` and make `BlogDao` interface package-private, and run the test.

It _does_ work for non-default methods, e.g. `@SqlQuery`.

But default methods don't work. `DefaultMethodHandler` must have access to the class to invoke the default method. It can't invoke it, because the interface is 1) package-private, and 2) in a different package.

```
java.lang.RuntimeException: java.lang.IllegalAccessException: class is not public: org.jdbi.v3.sqlobject.config.TestRegisterBeanMapper$BlogDao.getArticleWithComments(long)Optional/invokeSpecial, from org.jdbi.v3.sqlobject.config.TestRegisterBeanMapper$BlogDao/2

	at org.jdbi.v3.sqlobject.DefaultMethodHandler.<init>(DefaultMethodHandler.java:65)
	at org.jdbi.v3.sqlobject.DefaultMethodHandlerFactory.buildHandler(DefaultMethodHandlerFactory.java:60)
	at org.jdbi.v3.sqlobject.Handlers.lambda$findFor$0(Handlers.java:55)
...
[...]

PR #967, opened 2017-11-24
First spike at @Generate annotation processor that implements SqlObject

stevenschlansker:
Not ready for merge yet, asking for early feedback.

tag @electrum @brianm @qualidafial @arteam 

Allows use of abstract classes and package-private or protected methods.

See #840

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/extension/Extensions.java
+++ b/core/src/main/java/org/jdbi/v3/core/extension/Extensions.java
@@ -18,6 +18,7 @@ import java.util.Optional;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.jdbi.v3.core.config.JdbiConfig;
+import org.jdbi.v3.meta.Beta;
 
 /**
  * Configuration class for defining {@code Jdbi} extensions via {@link ExtensionFactory}
@@ -95,6 +96,7 @@ public class Extensions implements JdbiConfig<Extensions> {
      * @param allowProxy whether to allow use of Proxy types
      * @return this
      */
+    @Beta
     public Extensions setAllowProxy(boolean allowProxy) {
         this.allowProxy = allowProxy;
         return this;
@@ -103,6 +105,7 @@ public class Extensions implements JdbiConfig<Extensions> {
     /**
      * @return whether Proxy classes are allowed to be used
      */
+    @Beta
     public boolean isAllowProxy() {
         return allowProxy;
     }
@@ -115,6 +118,7 @@ public class Extensions implements JdbiConfig<Extensions> {
     /**
      * Throw if proxy creation is disallowed.
      */
+    @Beta
     public void onCreateProxy() {
         if (!isAllowProxy()) {
             throw new IllegalStateException("Creating onDemand proxy disallowed. Ensure @GenerateSqlObject annotation is being processed by `jdbi3-generator` annotation processor.");
--- a/core/src/main/java/org/jdbi/v3/core/extension/Extensions.java
+++ b/core/src/main/java/org/jdbi/v3/core/extension/Extensions.java
@@ -25,6 +25,7 @@ import org.jdbi.v3.core.config.JdbiConfig;
  */
 public class Extensions implements JdbiConfig<Extensions> {
     private final List<ExtensionFactory> factories = new CopyOnWriteArrayList<>();
+    private boolean allowProxy = true;
 
     /**
      * Create an empty {@link ExtensionFactory} configuration.
@@ -89,8 +90,34 @@ public class Extensions implements JdbiConfig<Extensions> {
                 .findFirst();
     }
 
+    /**
+     * Allow using {@link java.lang.reflect.Proxy} to implement extensions.
+     * @param allowProxy whether to allow use of Proxy types
+     * @return this
+     */
+    public Extensions setAllowProxy(boolean allowProxy) {
+        this.allowProxy = allowProxy;
+        return this;
+    }
+
+    /**
+     * @return whether Proxy classes are allowed to be used
+     */
+    public boolean isAllowProxy() {
+        return allowProxy;
+    }
+
     @Override
     public Extensions createCopy() {
         return new Extensions(this);
+    }
+
+    /**
+     * Throw if proxy creation is disallowed.
+     */
+    public void onCreateProxy() {
+        if (!isAllowProxy()) {
+            throw new IllegalStateException("Creating onDemand proxy disallowed. Ensure @GenerateSqlObject annotation is being processed by `jdbi3-generator` annotation processor.");
+        }
     }
 }
--- a/generator/pom.xml
+++ b/generator/pom.xml
@@ -1,0 +1,124 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+~   Licensed under the Apache License, Version 2.0 (the "License");
+~   you may not use this file except in compliance with the License.
+~   You may obtain a copy of the License at
+~
+~   http://www.apache.org/licenses/LICENSE-2.0
+~
+~   Unless required by applicable law or agreed to in writing, software
+~   distributed under the License is distributed on an "AS IS" BASIS,
+~   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+~   See the License for the specific language governing permissions and
+~   limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>jdbi3-parent</artifactId>
+        <groupId>org.jdbi</groupId>
+        <version>3.10.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jdbi3-generator</artifactId>
+    <name>jdbi3 Generator</name>
+    <description>jdbi Generator creates SqlObject implementations and simple value classes</description>
+    <url>http://jdbi.org/</url>
+
+    <licenses>
+        <license>
+            <name>Apache License 2.0</name>
+            <url>http://www.apache.org/licenses/LICENSE-2.0.html</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+
+    <properties>
+        <moduleName>org.jdbi.v3.generator</moduleName>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-sqlobject</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.squareup</groupId>
+            <artifactId>javapoet</artifactId>
+            <version>1.9.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-core</artifactId>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-compile</id>
+                        <configuration>
+                            <compilerArgs>
+                                <arg>-parameters</arg>
+                                <arg>-proc:none</arg>
+                            </compilerArgs>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>default-testCompile</id>
+                        <configuration>
+                            <compilerArgs>
+                                <arg>-parameters</arg>
+                            </compilerArgs>
+                            <annotationProcessorPaths>
+                                <annotationProcessorPath>
+                                    <groupId>org.jdbi</groupId>
+                                    <artifactId>jdbi3-generator</artifactId>
+                                    <version>${project.version}</version>
+                                </annotationProcessorPath>
+                            </annotationProcessorPaths>
+                            <annotationProcessors>
+                                <annotationProcessor>org.jdbi.v3.generator.GenerateSqlObjectProcessor</annotationProcessor>
+                            </annotationProcessors>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
@@ -15,44 +15,50 @@ package org.jdbi.v3.sqlobject;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
-import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Proxy;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
-import java.util.WeakHashMap;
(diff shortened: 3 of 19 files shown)


arteam:
General feedback:
* It works, i like it! Seems to be a nice tool for users who are migrating from jdbi2.
* It looks like a generated class uses reflection to invoke actual methods as well as the proxy implementation. I don't think reflection is really fast on the JVM and also makes code more hard to reason about. In the Proxy case it's justifiable, but if we generate source code I would expect a "pure" Java implementation. Not sure if it's possible though.
* I still think we should not encourage users to use abstract classes for representing DAO. I know there are some issues with using private methods in JDK9, but it seems to be a better option would be try to fix them and sticking to interfaces as the "official" way of representing SQL objects. It seems to me the only reason users ask for abstract classes is the ability to have private SQL methods, which was advertised as one of major JDK9 language feature improvements.

stevenschlansker:
It's true that some reflection remains, but the generated code is not actually using any directly.  The `Handler` interface is all our code.  It just turns out that handling the invocation as an `Method+Object[]` makes the two implementations look similar.  From there, I briefly looked through the Handler implementations and most of it is just looking up by `Method` or other once-per-query work.  I wouldn't be surprised if the overhead disappeared as soon as you introduce any real work.

There is some overhead in creating each instance of your type -- it still does reflection to create the handler map.  We probably could eliminate this too and cache a lot of the rest.

It's true, this is not a code generator that makes idiomatic fluent jdbi or anything like that.  But since not all users will want to use code gen initially this compromise means the entire thing is quite simple.

I agree, I'm not sure how essential `abstract class` SqlObjects are.  But sometimes the need to separate the intermediate "presentation" logic from the generated logic feels cumbersome.  This also allows you to introduce state if you like, which could be considered evil...

generator/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-sqlobject</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.squareup</groupId>
+            <artifactId>javapoet</artifactId>
+            <version>1.9.0</version>
```
qualidafial: Provided scope?

generator/src/main/java/org/jdbi/v3/generator/GenerateProcessor.java
```diff
@@ -0,0 +1,139 @@
...
+                map(VariableElement::getSimpleName)
+                .collect(Collectors.joining(","));
+        String methodField = "_" + e.getSimpleName() + "Method";
+        typeBuilder.addField(Method.class, methodField, Modifier.PRIVATE, Modifier.FINAL, Modifier.STATIC);
+        typeBuilder.addStaticBlock(CodeBlock.of("$L = $L.class.getMethod($S, $L);\n",
+                methodField,
+                ee.getEnclosingElement().getSimpleName(),
+                e.getSimpleName(),
+                ee.getParameters().stream().map()));
+        processingEnv.getElementUtils().
```
qualidafial: what's happening here?

stevenschlansker: This creates a static field for each method and a static block that fills it with its corresponding `java.lang.reflect.Method` instance, so we can easily dispatch to the existing invocation handler without making changes..

stevenschlansker:
Yes.  I'm not sure exactly what the compelling use case is -- I personally have had basically no issues with the `Proxy` implementation.  But maybe it's helpful to some people that want to save state in their SqlObject 🤷‍♀️ ?

Either way, this isn't a current development focus for me.

agentgt:
> Yes. I'm not sure exactly what the compelling use case is -- I personally have had basically no issues with the Proxy implementation. But maybe it's helpful to some people that want to save state in their SqlObject 🤷‍♀️ ?

The compelling use case is [GraalVM native-image](https://www.graalvm.org/docs/reference-manual/aot-compilation/) as well as cleaner stack traces / debugging. 

Also while the runtime performance is probably negligible compared to actual calls to the database there is a startup performance loss when using proxies. For those interested in doing lambda like function micro services startup time is important.

You can register JDK proxies with GraalVM native-image but it is quite painful. Some folks generate processors just to generate the configuration files.

stevenschlansker:
I had an opportunity to play with Graal recently and was quite impressed.  If anyone wants to make this dream a reality, I think it's quite viable -- the POC here worked well already -- but it'll probably take a bit of elbow grease by a volunteer to get it over the finish line :)
I'll kick it off by giving this guy a fresh rebase...

agentgt:
I'm equally impressed how modular and well designed jdbi is!

I'll try my best to see what I can do to make the current implementation better/complete.

We are planning on doing some graal-native with a few of our projects and seeing that this is possible will make it easier to stick or convert to JDBI.

stevenschlansker:
If you achieve good results and don't mind writing up a blog post or case study, we'd certainly feature it :)

I will try to rewrite the current code to reduce use of reflective `Method`s, it shouldn't be too hard.

stevenschlansker:
I found this issue informative on the subject: https://github.com/oracle/graal/issues/955

generator/pom.xml
stevenschlansker: The generator itself uses it at execution time.  We expect users to include this artifact itself in provided scope.

generator/src/main/java/org/jdbi/v3/generator/GenerateProcessor.java
stevenschlansker: Now we use a `HandlerKey` class instead of the `Method` objects.  That way we don't have any reflective types at all.

stevenschlansker:
I rebased the code (wow, it was old!) and removed the use of `Method` and other reflective types from the generated code per @arteam 's suggestion.  Now the generated code looks like:

```java
package org.jdbi.v3.generator;

public class AbstractClassDaoImpl extends NonpublicSubclassTest.AbstractClassDao implements SqlObject {
  private final Function<Object[], Object> getHandle_0;
  private final Consumer<Object[]> insert_1;
  private final Function<Object[], Object> list0_2;
  private final Function<Object[], Object> list0_3;

  public AbstractClassDaoImpl(HandleSupplier handle, Function<HandlerKey, Handler> handlerFactory) {
    final Handler getHandle_0 = handlerFactory.apply(HandlerKey.of("getHandle", org.jdbi.v3.core.Handle.class, new Class<?>[] {}));
    this.getHandle_0 = Unchecked.function(args -> getHandle_0.invoke(this, args, handle));
    final Handler insert_1 = handlerFactory.apply(HandlerKey.of("insert", void.class, new Class<?>[] {int.class,java.lang.String.class}));
    this.insert_1 = Unchecked.consumer(args -> insert_1.invoke(this, args, handle));
    final Handler list0_2 = handlerFactory.apply(HandlerKey.of("list0", java.util.List.class, new Class<?>[] {}));
    this.list0_2 = Unchecked.function(args -> list0_2.invoke(this, args, handle));
    final Handler list0_3 = handlerFactory.apply(HandlerKey.of("list0", java.util.List.class, new Class<?>[] {int.class}));
    this.list0_3 = Unchecked.function(args -> list0_3.invoke(this, args, handle));
  }

  @Override
  public Handle getHandle() {
    return (org.jdbi.v3.core.Handle) getHandle_0.apply(new Object[] {});
  }

  @Override
  void insert(int id, String name) {
     insert_1.accept(new Object[] {id,name});
  }

  @Override
  List<Something> list0() {
    return (java.util.List<org.jdbi.v3.core.Something>) list0_2.apply(new Object[] {});
  }

  @Override
  List<Something> list0(int id) {
    return (java.util.List<org.jdbi.v3.core.Something>) list0_3.apply(new Object[] {id});
  }
}
```

which is actually pretty reasonable!

stevenschlansker:
I marked this as @Beta and since the vast majority of the new code is in its own artifact, I think this is now approaching good to merge!  Of course it's still an "incubating" feature, but including it in the build means people might see and use it, and prevents it from rotting...

stevenschlansker:
It turns out we cannot get away from handling `Method` objects at all, as our `ExtensionMethod` type must carry it along.  I don't think that's a significant problem (in particular we never `invoke` it, only use as a Map key), just a bit of a shame.

agentgt:
FWIW (I don’t know all the details and it’s friday :) )

In terms of Graal VM native image as long as you procure the method with static strong constants you are fine.

Thus using meta data on java.reflect.method is not a problem (most of which is cached including annotations) so long as you provide a constant string names to access the class and the method.

stevenschlansker:
The current approach is to look them all up using `Class.getMethod` or `Class.getDeclaredMethod` in a `static {}` initializer.  The only weirdness is a `ThreadLocal` handoff of some metadata.

I still need to get configuring annotations to work.

stevenschlansker:
Configuring annotations now work.

stevenschlansker:
And now `default` and `private` `interface` methods do as well 😸

stevenschlansker:
I re-built one of $WORK's core service DAOs with `@GenerateSqlObject` and it worked with no changes!

stevenschlansker:
Aaaaaand now we also generate On-Demand implementations classes too :sunglasses:

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
```diff
@@ -0,0 +1,134 @@
...
+import org.jdbi.v3.core.internal.exceptions.Sneaky;
+import org.jdbi.v3.sqlobject.GenerateSqlObject;
+import org.jdbi.v3.sqlobject.Handler;
+
+public final class SqlObjectInitData {
+    private static final Object[] NO_ARGS = new Object[0];
+    public static final ThreadLocal<SqlObjectInitData> INIT_DATA = new ThreadLocal<>();
+
+    public final boolean concrete;
+    public final Class<?> extensionType;
```
TheRealMarnes: `private`

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
```diff
@@ -0,0 +1,134 @@
...
+    public final UnaryOperator<ConfigRegistry> instanceConfigurer;
+    public final Map<Method, UnaryOperator<ConfigRegistry>> methodConfigurers;
```
TheRealMarnes: `private`

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -83,122 +93,172 @@ private boolean looksLikeSqlObject(Class<?> extensionType) {
...
-        return handlersCache.computeIfAbsent(sqlObjectType, type -> {
-            final Map<Method, Handler> handlers = new HashMap<>();
-
-            handlers.putAll(handlerEntry((t, a, h) ->
-                    sqlObjectType.getName() + '@' + Integer.toHexString(t.hashCode()),
-                Object.class, "toString"));
-            handlers.putAll(handlerEntry((t, a, h) -> t == a[0], Object.class, "equals", Object.class));
-            handlers.putAll(handlerEntry((t, a, h) -> System.identityHashCode(t), Object.class, "hashCode"));
-            handlers.putAll(handlerEntry((t, a, h) -> h.getHandle(), SqlObject.class, "getHandle"));
+        if (data.concrete) {
```
TheRealMarnes: please no public field usage

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public static SqlObjectInitData initData() {
+        final SqlObjectInitData result = INIT_DATA.get();
+        if (result == null) {
+            throw new IllegalStateException("Implemented SqlObject types must be initialized by SqlObjectFactory");
+        }
+        return result;
+    }
+
+    public static Method lookupMethod(String methodName, Class<?>... parameterTypes) {
```
TheRealMarnes: should we add suppressions for `unused` on methods only called from reflective code? Better indicates current state on one hand, but adds maintenance to code changes on the other.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
```diff
@@ -0,0 +1,134 @@
...
+        return lookup0(initData().extensionType, methodName, parameterTypes);
+    }
+
+    private static Method lookup0(Class<?> klass, String methodName, Class<?>... parameterTypes) {
```
TheRealMarnes: not a fan of the `0` suffix for internal delegation, there must be a more meaningful name you can give such methods

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+            throw new IllegalStateException(
+                    String.format("can't find %s#%s%s", klass.getName(), methodName, Arrays.asList(parameterTypes)), e);
+        }
+    }
+
+    // Thanks Holger!
+    // https://stackoverflow.com/questions/35331327/does-java-8-have-cached-support-for-suppliers
+    public Supplier<InContextInvoker> lazyInvoker(Object target, Method method, HandleSupplier handle, ConfigRegistry instanceConfig) {
+        return new Supplier<InContextInvoker>() {
```
TheRealMarnes: I think at this point it's better to just make this a proper class of its own, this is more than just a lambda

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -83,122 +93,172 @@ private boolean looksLikeSqlObject(Class<?> extensionType) {
...
             try {
-                handlers.putAll(handlerEntry((t, a, h) -> null, sqlObjectType, "finalize"));
-            } catch (IllegalStateException expected) {
-                // optional implementation
+                SqlObjectInitData.INIT_DATA.set(data);
+                return extensionType.cast(
+                    Class.forName(extensionType.getPackage().getName() + "." + extensionType.getSimpleName() + "Impl")
```
TheRealMarnes: let's make `"Impl"` a constant on a new class that handles these naming concerns?

core/src/main/java/org/jdbi/v3/core/extension/Extensions.java
```diff
@@ -89,8 +90,34 @@ public boolean hasExtensionFor(Class<?> extensionType) {
                 .findFirst();
     }
 
+    /**
+     * Allow using {@link java.lang.reflect.Proxy} to implement extensions.
+     * @param allowProxy whether to allow use of Proxy types
+     * @return this
+     */
+    public Extensions setAllowProxy(boolean allowProxy) {
```
TheRealMarnes: Not that I have a better idea, but have you thought it through that presenting this feature as a boolean decision "proxy or other" is the way to go? Not something like `setSqlObjectFactory` or something where you can plug in a `ProxyFactory` and a `GeneratedClassFactory`? A distinction like that might also be a good starting point for later refactoring, since as far as I understand the point is to have two mutually exclusive sqlobject generators, so slapping their code together in 1 class sounds like conflation.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
stevenschlansker: It's generated code.  I don't think adding getters will make anything clearer, it's already clear as mud.  The code is tightly coupled here by design.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
stevenschlansker: That's how the JDK does it... I can use `unsafe` instead I guess.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
stevenschlansker: The argument capture is too nice here.  Writing out the constructor and fields and copy statements brings no value.

core/src/main/java/org/jdbi/v3/core/extension/Extensions.java
stevenschlansker: Well, the point of this toggle really is, "inform Jdbi that it is forbidden to make Proxy objects" -- usually because it's unimplemented or not desired.  This is not meant to be a general "switch between SqlObject" configuration toggle.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
stevenschlansker: No need -- it's `public`.

stevenschlansker:
I cleaned up a lot of the variable names as you suggested.  Thanks for the review.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
TheRealMarnes: I hardly think a method being `public` is a free pass to look unused outside API, but ok

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
TheRealMarnes: I'm talking about the `LazyInvoker` part, not the rest of the body yet. The LazyInvoker contraption returned by this method is definitely an independent component of which the complexity can be set apart from all this.

core/src/main/java/org/jdbi/v3/core/extension/Extensions.java
TheRealMarnes: Ok, so more of a general setting to be used by potential other features too, just to indicate whether or not Proxies are a go.

PR #972, opened 2017-12-03
added support for option mapping

alwins0n:
using my plugin in real world development I realized that a critical column mapper was missing - the one for `Option`.

this is the code i use to work around that, "inspired" by the builtin OptionalMapper

### DIFF
--- a/vavr/src/main/java/org/jdbi/v3/vavr/VavrOptionMapper.java
+++ b/vavr/src/main/java/org/jdbi/v3/vavr/VavrOptionMapper.java
@@ -29,21 +29,19 @@ import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
 
 public class VavrOptionMapper<T> implements ColumnMapper<Option<T>> {
 
-    private final Type type;
+    private final Type nestedType;
 
-    private VavrOptionMapper(Type type) {
-        this.type = type;
+    private VavrOptionMapper(Type nestedType) {
+        this.nestedType = nestedType;
     }
 
-    public static ColumnMapper<?> of(Type type) {
-        return new VavrOptionMapper(type);
-    }
-
-    public static ColumnMapperFactory factory() {
+    static ColumnMapperFactory factory() {
         return (type, config) -> {
             Class<?> rawType = getErasedType(type);
             if (rawType == Option.class) {
-                return Optional.of(VavrOptionMapper.of(type));
+                final Type nestedType = GenericTypes.findGenericParameter(type, Option.class)
+                        .orElseThrow(() -> new NoSuchMapperException("No mapper for raw Option type"));
+                return Optional.of(new VavrOptionMapper<>(nestedType));
             }
             return Optional.empty();
         };
@@ -52,11 +50,8 @@ public class VavrOptionMapper<T> implements ColumnMapper<Option<T>> {
     @SuppressWarnings("unchecked")
     @Override
     public Option<T> map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
-        final ColumnMapper<?> mapper = ctx.findColumnMapperFor(
-                GenericTypes.findGenericParameter(type, Option.class)
-                        .orElseThrow(() -> new NoSuchMapperException("No mapper for raw Option type")))
-                .orElseThrow(() -> new NoSuchMapperException("No mapper for type " + type + " nested in Option"));
-
+        final ColumnMapper<?> mapper = ctx.findColumnMapperFor(nestedType)
+                .orElseThrow(() -> new NoSuchMapperException("No mapper for type " + nestedType + " nested in Option"));
         return (Option<T>) Option.of(mapper.map(r, columnNumber, ctx));
     }
 
--- a/vavr/src/main/java/org/jdbi/v3/vavr/VavrOptionMapper.java
+++ b/vavr/src/main/java/org/jdbi/v3/vavr/VavrOptionMapper.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.vavr;
+
+import io.vavr.control.Option;
+import org.jdbi.v3.core.generic.GenericTypes;
+import org.jdbi.v3.core.mapper.ColumnMapper;
+import org.jdbi.v3.core.mapper.ColumnMapperFactory;
+import org.jdbi.v3.core.mapper.NoSuchMapperException;
+import org.jdbi.v3.core.statement.StatementContext;
+
+import java.lang.reflect.Type;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Optional;
+
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
+public class VavrOptionMapper<T> implements ColumnMapper<Option<T>> {
+
+    private final Type type;
+
+    private VavrOptionMapper(Type type) {
+        this.type = type;
+    }
+
+    public static ColumnMapper<?> of(Type type) {
+        return new VavrOptionMapper(type);
+    }
+
+    public static ColumnMapperFactory factory() {
+        return (type, config) -> {
+            Class<?> rawType = getErasedType(type);
+            if (rawType == Option.class) {
+                return Optional.of(VavrOptionMapper.of(type));
+            }
+            return Optional.empty();
+        };
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public Option<T> map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
+        final ColumnMapper<?> mapper = ctx.findColumnMapperFor(
+                GenericTypes.findGenericParameter(type, Option.class)
+                        .orElseThrow(() -> new NoSuchMapperException("No mapper for raw Option type")))
+                .orElseThrow(() -> new NoSuchMapperException("No mapper for type " + type + " nested in Option"));
+
+        return (Option<T>) Option.of(mapper.map(r, columnNumber, ctx));
+    }
+
+}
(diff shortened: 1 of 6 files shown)


vavr/src/test/java/org/jdbi/v3/vavr/TestVavrOptionMapperWithDB.java
```diff
@@ -0,0 +1,110 @@
...
+    @Test
+    public void testOptionWithinObjectIfMissing_shouldBeNone() throws SQLException {
+        final SomethingWithOption result = dbRule.getSharedHandle()
+                .registerRowMapper(ConstructorMapper.factory(SomethingWithOption.class))
+                .createQuery("select id, name from something where id = 2")
+                .mapTo(SomethingWithOption.class)
+                .findOnly();
+
+        assertThat(result.getName()).isInstanceOf(Option.class);
+        assertThat(result.getName().getOrNull()).isNull();
```
qualidafial: Wouldn't hurt to add `assertj-vavr` as a test dependency to `jdbi3-vavr` so you can do fluent assertions on vavr types.

https://github.com/assertj/assertj-vavr

vavr/src/test/java/org/jdbi/v3/vavr/TestVavrOptionMapperWithDB.java
```diff
@@ -0,0 +1,110 @@
...
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+
+            SomethingWithOption that = (SomethingWithOption) o;
+
+            if (id != that.id) return false;
+            return name != null ? name.equals(that.name) : that.name == null;
```
qualidafial: Aside: I'm assuming this was generated. What IDE are you using? Most of them should be able to generated Java 7-style equals that use `java.util.Objects.equals(Object, Object)` to avoid these null checks.

vavr/src/test/java/org/jdbi/v3/vavr/TestVavrOptionMapperWithDB.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        @Override
+        public int hashCode() {
+            int result = id;
+            result = 31 * result + (name != null ? name.hashCode() : 0);
+            return result;
```
qualidafial: Ditto for `java.util.Objects.hash(Object...)`

vavr/src/main/java/org/jdbi/v3/vavr/VavrOptionMapper.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public static ColumnMapper<?> of(Type type) {
+        return new VavrOptionMapper(type);
+    }
+
+    public static ColumnMapperFactory factory() {
+        return (type, config) -> {
+            Class<?> rawType = getErasedType(type);
+            if (rawType == Option.class) {
+                return Optional.of(VavrOptionMapper.of(type));
```
qualidafial: I think this lambda should be where we unwrap the `Option<T>` generic type instead of `map()`, so that `of(type)` receives the element type `T` instead of `Option<T>`.

Maybe rename the `of()` method to `ofElement()` to remove ambiguity?

One side benefit of doing the `findGenericParameter(...).orElseThrow(...)` here instead of in `map()` is that this code executes _before_ the database statement is executed.

vavr/src/main/java/org/jdbi/v3/vavr/VavrOptionMapper.java
```diff
@@ -0,0 +1,63 @@
...
+        };
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public Option<T> map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
+        final ColumnMapper<?> mapper = ctx.findColumnMapperFor(
+                GenericTypes.findGenericParameter(type, Option.class)
+                        .orElseThrow(() -> new NoSuchMapperException("No mapper for raw Option type")))
+                .orElseThrow(() -> new NoSuchMapperException("No mapper for type " + type + " nested in Option"));
```
qualidafial: Probably want tests covering these exceptions.

vavr/src/test/java/org/jdbi/v3/vavr/TestVavrOptionMapperWithDB.java
alwins0n: There is no stable release yet and i cannot find it in any repository.

vavr/src/test/java/org/jdbi/v3/vavr/TestVavrOptionMapperWithDB.java
alwins0n: Yeah can redo that - I rarely generate them via IDE since I use lombok

qualidafial:
Please see #975 (need a CollectorFactory for Option) and add that to this PR as well. See `GuavaCollectors.toOptional()` for an implementation example.

vavr/src/main/java/org/jdbi/v3/vavr/VavrOptionMapper.java
alwins0n: Can do, but is wanted to stay consistent with the `OptionalMapper` implementation. Are you aware that this code is just analogue to the core code?

vavr/src/test/java/org/jdbi/v3/vavr/TestVavrOptionMapperWithDB.java
qualidafial: :disappointed:

vavr/src/main/java/org/jdbi/v3/vavr/VavrOptionMapper.java
qualidafial: Yikes. I did not know that. Thankfully, that class is nonpublic. Sheesh..

qualidafial: Would you mind hoisting that generic parameter check in `OptionalMapper.map()` into `BuiltInMapperFactory.build()` while you're at it?

vavr/src/test/java/org/jdbi/v3/vavr/TestVavrOptionMapperWithDB.java
qualidafial: Fun fact: Lombok's `@AllArgsConstructor` generates `@java.beans.ConstructorProperties` annotations on the constructor, so you can use `ConstructorMapper` on those types.

qualidafial:
The Option collector factory is implemented in #977.

vavr/src/test/java/org/jdbi/v3/vavr/TestVavrOptionMapperWithDB.java
qualidafial: https://github.com/assertj/assertj-vavr/issues/8

pivovarit: @alwins0n is right - it's not done yet. If you're looking for a workaround, you can simply convert Vavr's `Option` to Java 8's `Optional` and leverage dedicated assertions

alwins0n: I already rewrote the test suite for this pull request and decided to equal on the result object `SomethingWithOption` as a whole which makes a lot more sense (since it can verify also the correct record via id)

vavr/src/test/java/org/jdbi/v3/vavr/TestVavrOptionMapperWithDB.java
alwins0n: I know ;) This is the killer feature for me to work with vavr jackson and jdbi for a nice functional rest api

vavr/src/main/java/org/jdbi/v3/vavr/VavrOptionMapper.java
alwins0n: done. furthermore i removed the `of` method anyway. i dont see a usecase where someone would need a public access to create a new `OptionMapper` - same goes for package private visibility of the factory method.

If you agree, same could be said for the `OptionalMapper` class

I renamed the variable to make it clear that the nested type is under consideration

qualidafial: With the `factory()` method hidden package private and the constructor private, does it make sense for `VavrOptionMapper` to be public at all? I think it's okay for now to just expose the feature through the plugin.

qualidafial:
Let's make VavrOptionMapper package private and then we're good to merge.

vavr/src/main/java/org/jdbi/v3/vavr/VavrOptionMapper.java
alwins0n: Thinking about it - I will leave the factory method public and the constructor private. It can make sense to allow public access to the factory if you only want jdbi customized with selected features of the plugin and not use the `VavrPlugin` class. I will check if this holds for all features that are installed via VavrPlugin class

The constructor however will remain private since it makes no sense to me to register mappers of Option<T> with only handpicked T.

You agree with this?

alwins0n:
Can you comment on my thoughts about public access? I will change tomorrow whatever you decide fits the plugin spirit of the project best

vavr/src/main/java/org/jdbi/v3/vavr/VavrOptionMapper.java
qualidafial: I think we're in agreement on visibility:

* Reduce the visibility of `OptionMapper` to package private
* Reduce the visibility of `OptionalMapper.of()` to package private. Since the class is already package private, reducing the method visibility doesn't hurt anything.
* (Optional, if you care to take the time) Refactor `OptionalMapper` so it accepts the nested type instead of `Optional<Something>`--similar to what you did with `VavrOptionMapper`.

alwins0n: Are we in agreement? :) 

I just proposed to leave the visibility of `VavrOptionMapper` public to enable client calls to the factory method. that way the user can register parts of the plugin with JDBI directly instead of using the `VavrPlugin` class.

Then again I am not sure if a JDBI plugin is expected to be partly installable or only as a whole - I don't know if there is any guideline here

qualidafial: There aren't any plugin guidelines yet--we're still learning how people will use this. We'll document best practices as we discover them.

As for the public/private thing, I can see both sides of it. If we kept it public, we'd want to make the factory method also public--right now it's package private.

However I'd rather not introduce permanent public API based on speculation. We don't know yet whether people will want access to the mapper, or will be content with just the plugin.

If we make it private now, we can always make it public later if users ask for it. But if we make it public now and realize later that it was a mistake, we can't make it private without breaking compatibility.

So my preference is to make the mapper private, and see how far we get with just the plugin.

qualidafial:
Looks good. Merging.

qualidafial:
@alwins0n Thanks for the PR!

PR #977, opened 2017-12-05
Add collector factories for Optional[Int|Long|Double], and Vavr Option<T>

### LINKED ISSUE #975: Vavr plugin should provide CollectorFactory for Option
If we only provide mappers then a method like `@SqlQuery Option<Foo> get(long id)` will return null if there are no rows.

In this case, `Option<Foo>` is a really a collection type with zero to one `Foo` elements.

### LINKED ISSUE #976: Add CollectorFactory for OptionalInt, OptionalLong, OptionalDouble
A `@SqlQuery OptionalInt getThing(args)` method would return null if the query returned no rows.

We have column mappers for all the optional primitives, but in the above case, `OptionalInt` is a collection of zero to one `int`s.

qualidafial:
Fixes #975.
Fixes #976.

@stevenschlansker or @arteam could you please review?

/cc @alwins0n

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/collector/OptionalBuilder.java
+++ b/core/src/main/java/org/jdbi/v3/core/collector/OptionalBuilder.java
@@ -13,22 +13,51 @@
  */
 package org.jdbi.v3.core.collector;
 
-import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Supplier;
 
-class OptionalBuilder<T> {
-    private Optional<T> optional = Optional.empty();
+class OptionalBuilder<T, OPT_T> {
+  private final Supplier<OPT_T> empty;
+  private final Function<T, OPT_T> factory;
 
-    public void set(T value) {
-        if (optional.isPresent()) {
-            throw new IllegalStateException(
-                    String.format("Multiple values for Optional type: ['%s','%s',...]",
-                            optional.get(),
-                            value));
-        }
-        optional = Optional.ofNullable(value);
+  boolean set;
+  T value;
+
+  OptionalBuilder(Supplier<OPT_T> empty, Function<T, OPT_T> factory) {
+    this.empty = empty;
+    this.factory = factory;
+  }
+
+  void set(T value) {
+    if (set) {
+      throw tooManyValues(this.value, value);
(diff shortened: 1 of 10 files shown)


qualidafial:
Ping @stevenschlansker or @arteam: could one of you please review?

core/src/main/java/org/jdbi/v3/core/collector/OptionalBuilder.java
```diff
@@ -13,22 +13,51 @@
...
-            throw new IllegalStateException(
-                    String.format("Multiple values for Optional type: ['%s','%s',...]",
-                            optional.get(),
-                            value));
-        }
-        optional = Optional.ofNullable(value);
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+class OptionalBuilder<T, OPT_T> {
```
alwins0n: just my 2 cents here: maybe `OptionalBuilder` is not the best class name since it kind of implies the java8 optional. basically it is a builder for all zero-one collections.
also OPT_T reads a little weird to me. maybe stick with R as in the java8 collector api (where R is the return value type)

qualidafial: The class name thing was on my mind when I was working on this. There are several different terms to choose from just looking at the [Wikipedia page on the topic](https://en.wikipedia.org/wiki/Option_type). Ultimately I decided to leave it as `OptionalBuilder` since `Optional` is the term the JDK uses, and this builder is package private anyway.

As for the generic parameter name, I've actually names like this in several JDK classes, e.g. `PrimitiveIterator<T, T_CONS>`, `AbstractPipeline<E_IN, E_OUT, S>`, along with several other types in the `java.util.stream` package. That said, we can probably get away with renaming this parameter to just `O`.

qualidafial:
@stevenschlansker or @arteam: ping?

qualidafial:
@stevenschlansker or @arteam: Could one of you please review?

stevenschlansker:
Oops, sorry, I had a review but forgot to hit the commit button 😞

qualidafial:
Any action items?

PR #986, opened 2017-12-15
Establish a code of conduct

qualidafial:
This has been pressing on my mind with recent headlines about harassment.

In the interest of keeping Jdbi a friendly, inclusive, and welcoming place, I propose we establish a project code of conduct. I don't like thinking about harassment happening within our own community, but having those rules in place would give us a principled basis on which to respond.

@brianm As part of this, we will need to set up an email address `abuse@jdbi.org`, and designate a team member to receive and respond to reports. I will volunteer myself for that role if nobody else wants it.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,46 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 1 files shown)


qualidafial:
@stevenschlansker This was waiting on the `abuse@jdbi.org` email to get created--do you know if this happened? I discussed it with @brianm a bit but I'm not sure if an account actually got created. If so we should make sure the code of conduct is updated with the correct email.

stevenschlansker:
Whoops, sorry, no I didn't.  I forgot we were holding it (it didn't have the on-hold tag).  I think the doc is still valuable as-is, even if the email isn't set up yet.

Issue #987, opened 2017-12-20
jdbi3-spring4 plugin does not close handle on Transaction end


alwins0n:
Expected: after leaving a method/class with springs `@Transactional` I expect the used handle (retrieved via `JdbiUtil#getHandle`) to be closed (method call to `close`).

This is not the case - I think the `JdbiUtil.Adapter#beforeCompletion()` method should closed the transactional handle.

Am willing to work on this

qualidafial:
Transactions and handles have different lifecycles. A single handle may host any number of transactions over time.

Whatever method is calling `JdbiUtil.getHandle()` should likewise be calling `JdbiUtil.closeIfNeeded()` to clean up.

alwins0n:
I understand, but let me think loudly here: in an environment where there a framework managed transactions these should take precedence over the library managed transactions. imho. I will address this in the other issue.

Anyway, the use of `JbiUtil.getHandle()` and `JdbiUtil.closeIfNeeded()` is not mentioned in the docs - at least the plugin user should not be required to figure all out by him/herself

qualidafial:
I agree, we could do better on the docs here. The main problem is that most of the project maintainers use Dropwizard instead of Spring, so our Spring understanding is somewhat lacking.

hgschmie:
We did lots of changes to the resource handling in Jdbi since 3.35.0. Closing this, please reopen if the problem still exists.

Issue #983, opened 2017-12-12
jdbi3-spring4 plugin needs documentation / (rework?)


alwins0n:
1. From the doc it is not clear that you have to use the JdbiUtil class for the declarative transaction management to work. This becomes only apparent if you look at the test code. You would excpect everything to work automagically after you set up your FactoryBean and inject Jdbi.
2. The FactoryBean is probably only useful if you setup spring via xml (since it offers some fast-config params). Since a lot of ppl nowadays use java-config only the doc maybe should reflect that (even if it is actually trivial)

Other than that i was asking myself if the plugin should maybe offer a `JdbiBean`, or something like that, which has the same interface as jdbi but delegates calls to an underlying jdbi instance and handles transactions itself.
That way you would not need to use JdbiUtil or anything like that but use jdbi the way you are used to while in the background spring handles transactions for you - makes sense?

Am willing to work on this issue if needed

qualidafial:
I don't use Spring at $WORK so I'm only I'm only vaguely aware of Spring conventions.

Would providing a `JdbiBean` class provide any advantage above and beyond e.g. a `@Bean Jdbi jdbi()` method in a `@Configuration` class in the user's app?

qualidafial:
Maybe all this needs is an example in the docs using java config, with a blurb on how to set up Jdbi to work with Spring's container-managed transactions.

qualidafial:
`Jdbi` isn't really intended to be subclassed--that's why it has a private constructor.

I think we'd be better off with a builder-style utility class e.g. `SpringJdbiBuilder` with spring-centric chaining builder methods to set up specific things, like e.g. container managed transactions. If Spring needs a new hook to control some internal behavior in Jdbi, I think we could accommodate that.

alwins0n:
I think we should definitly accommodate that. Imho spring-jdbi should be a dropin replacement for spring-jdbc - it would definitly boost acceptance in spring managed environments.

That being said, maybe its best to provide an `JdbiTemplate` class similar to springs `JdbcTemplate` (only similar in name). That class would have an underlying `Jdbi` instance, and all relevant methods `Jdbi` used for querying (useHandle withHandl... maybe even NO `inTransaction`, etc.). 

If this solution is accepable/preferred it would be nice to have a common (small) interface between `Jdbi` and `JdbiTemplate` which would just consist of `useHandle` and `withHandle` and `open` (which may be used to bypass spring transaction management) which could be called `JdbiActions` or something like that.

[...]

PR #989, opened 2017-12-21
WIP draft of JdbiTemplate based plugin

alwins0n:
This PR addresses #983 but also #987 

The idea is to follow springs conventions as closely as possible and expose a narrow interface that, similar to springs JdbcOperations, solely relies on the framework for transaction management.

If this draft is approved I will add tests and better documentation.

Also it can be considered to make `JdbiUtil` non-public.

Furthermore I was thinking of supporting spring boot which uses a lot more classpath magic to even avoid writing java config

### DIFF
--- a/spring4/src/main/java/org/jdbi/v3/spring4/JdbiOperations.java
+++ b/spring4/src/main/java/org/jdbi/v3/spring4/JdbiOperations.java
@@ -1,0 +1,42 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.spring4;
+
+import org.jdbi.v3.core.HandleCallback;
+import org.jdbi.v3.core.HandleConsumer;
+import org.jdbi.v3.core.Jdbi;
+
+/**
+ * similar to {@link JdbiOperations} this class defines the basic Jdbi operations that should be used in a Spring environment.
+ * <p>
+ * implemented by {@link JdbiTemplate}
+ */
+public interface JdbiOperations {
+
+    /**
+     * like {@link Jdbi#withHandle(HandleCallback)} but respects springs transactions
+     */
+    <R, X extends Exception> R withHandle(HandleCallback<R, X> callback) throws X;
+
+    /**
+     * like {@link Jdbi#useHandle(HandleConsumer)} but respects springs transactions
+     */
+    default <X extends Exception> void useHandle(HandleConsumer<X> callback) throws X {
+        withHandle(h -> {
+            callback.useHandle(h);
+            return null;
+        });
+    }
+
+}
--- a/spring4/src/main/java/org/jdbi/v3/spring4/JdbiTemplate.java
+++ b/spring4/src/main/java/org/jdbi/v3/spring4/JdbiTemplate.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.spring4;
+
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.HandleCallback;
+import org.jdbi.v3.core.Jdbi;
+
+/**
+ * Spring template for querying with Jdbi. This class is transaction aware and respects spring's
+ * {@link org.springframework.transaction.annotation.Transactional}
+ */
+public class JdbiTemplate implements JdbiOperations {
+
+    private final Jdbi jdbi;
+
+    public JdbiTemplate(Jdbi jdbi) {
+        this.jdbi = jdbi;
+    }
+
+    /**
+     * like {@link Jdbi#withHandle(HandleCallback)} but respects springs transactions
+     */
+    @Override
+    public <R, X extends Exception> R withHandle(HandleCallback<R, X> callback) throws X {
+        Handle h = JdbiUtil.getHandle(jdbi);
+        R result = callback.withHandle(h);
+        // close if not in transaction
+        JdbiUtil.closeIfNeeded(h);
+        return result;
+    }
+
+    /**
+     * @return the underlying Jdbi instance
+     */
+    public Jdbi getJdbi() {
+        return jdbi;
+    }
+}
--- a/spring4/src/main/java/org/jdbi/v3/spring4/JdbiTemplate.java
+++ b/spring4/src/main/java/org/jdbi/v3/spring4/JdbiTemplate.java
@@ -35,10 +35,11 @@ public class JdbiTemplate implements JdbiOperations {
     @Override
     public <R, X extends Exception> R withHandle(HandleCallback<R, X> callback) throws X {
         Handle h = JdbiUtil.getHandle(jdbi);
-        R result = callback.withHandle(h);
-        // close if not in transaction
-        JdbiUtil.closeIfNeeded(h);
-        return result;
+        try {
+            return callback.withHandle(h);
+        } finally {
+            JdbiUtil.closeIfNeeded(h);
+        }
     }
 
     /**
--- a/spring4/src/main/java/org/jdbi/v3/spring4/JdbiTransactionObject.java
+++ b/spring4/src/main/java/org/jdbi/v3/spring4/JdbiTransactionObject.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.spring4;
+
+import org.springframework.transaction.CannotCreateTransactionException;
+import org.springframework.transaction.SavepointManager;
+import org.springframework.transaction.TransactionException;
+import org.springframework.transaction.TransactionSystemException;
+
+/**
+ * savepoint capable jdbi transaction wrapper
+ */
+public class JdbiTransactionObject implements SavepointManager {
+
+    private static final String JDBI_SAVEPOINT_BASENAME = "JDBI_SAVEPOINT_";
+
+    private int nextSavePointNumber = 0;
+    private HandleHolder handleHolder;
+
+    public JdbiTransactionObject(HandleHolder handleHolder) {
+        this.handleHolder = handleHolder;
+    }
+
+    @Override
+    public Object createSavepoint() throws TransactionException {
+        String savePointName = JDBI_SAVEPOINT_BASENAME + nextSavePointNumber++;
+        try {
+            handleHolder.getHandle().savepoint(savePointName);
+        } catch (org.jdbi.v3.core.transaction.TransactionException ex) {
+            throw new CannotCreateTransactionException("savepoint could not be created", ex);
+        }
+
+        return savePointName;
+    }
+
+    @Override
+    public void rollbackToSavepoint(Object savepoint) throws TransactionException {
+        if (!(savepoint instanceof String)) {
+            throw new IllegalArgumentException("unhandleable savepoint: " + savepoint);
+        }
+
+        try {
+            handleHolder.getHandle().rollbackToSavepoint((String) savepoint);
+        } catch (org.jdbi.v3.core.transaction.TransactionException ex) {
+            throw new TransactionSystemException("savepoint could be rolled back", ex);
+        }
+    }
+
+    @Override
+    public void releaseSavepoint(Object savepoint) throws TransactionException {
+        if (!(savepoint instanceof String)) {
+            throw new IllegalArgumentException("unhandleable savepoint: " + savepoint);
+        }
+
+        try {
+            handleHolder.getHandle().release((String) savepoint);
+        } catch (org.jdbi.v3.core.transaction.TransactionException ex) {
+            // no-op since {@link org.springframework.transaction.support.AbstractTransactionStatus#rollbackToHeldSavepoint}
+            // releases immediately after rollback -> jdbi throws exception
+            // see org.springframework.jdbc.datasource.JdbcTransactionObjectSupport#releaseSavepoint
+        }
+    }
+
+    public HandleHolder getHandleHolder() {
+        return handleHolder;
+    }
+
+    public void setHandleHolder(HandleHolder handleHolder) {
+        this.handleHolder = handleHolder;
+    }
+}
--- a/spring4/src/main/java/org/jdbi/v3/spring4/JdbiUtil.java
+++ b/spring4/src/main/java/org/jdbi/v3/spring4/JdbiUtil.java
@@ -15,7 +15,6 @@ package org.jdbi.v3.spring4;
 
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.Jdbi;
-import org.springframework.transaction.support.TransactionSynchronizationAdapter;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 
 import java.util.HashSet;
@@ -26,7 +25,7 @@ import java.util.Set;
  */
 public class JdbiUtil
 {
-    private static final Set<Handle> TRANSACTIONAL_HANDLES = new HashSet<>();
+    private static final Set<Handle> UNMANAGED_HANDLES = new HashSet<>();
 
     /**
      * Obtain a Handle instance, either the transactionally bound one if we are in a transaction,
@@ -37,16 +36,15 @@ public class JdbiUtil
      */
     public static Handle getHandle(Jdbi jdbi)
     {
-        Handle bound = (Handle) TransactionSynchronizationManager.getResource(jdbi);
-        if (bound == null) {
-            bound = jdbi.open();
-            if (TransactionSynchronizationManager.isSynchronizationActive()) {
-                TransactionSynchronizationManager.bindResource(jdbi, bound);
-                TransactionSynchronizationManager.registerSynchronization(new Adapter(jdbi, bound));
-                TRANSACTIONAL_HANDLES.add(bound);
-            }
+        Handle handle;
+        if (TransactionSynchronizationManager.isSynchronizationActive()) {
+            HandleHolder boundHolder = (HandleHolder) TransactionSynchronizationManager.getResource(jdbi);
+            handle = boundHolder.getHandle();
+        } else {
+            handle = jdbi.open();
+            UNMANAGED_HANDLES.add(handle);
         }
-        return bound;
+        return handle;
     }
 
     /**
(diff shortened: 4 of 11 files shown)


alwins0n:
jeez what happened? ran the tests fine locally... or so i thought. will check it out later

spring4/src/main/java/org/jdbi/v3/spring4/JdbiOperations.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * like {@link Jdbi#withHandle(HandleCallback)} but respects springs transactions
+     */
+    <R, X extends Exception> R withHandle(HandleCallback<R, X> callback) throws X;
+
+    /**
+     * like {@link Jdbi#useHandle(HandleConsumer)} but respects springs transactions
+     */
+    default <X extends Exception> void useHandle(HandleConsumer<X> callback) throws X {
```
arteam: I think it would be better to implement `useHandle` instead of reusing `withHandle`. Having a callback in a callback will make stack traces in case of errors more confusing.

spring4/src/test/java/org/jdbi/v3/spring4/TestJdbiTemplate.java
```diff
@@ -0,0 +1,64 @@
...
+    @Autowired
+    private JdbiOperations jdbiOps;
+
+    @Test
+    public void testJdbiOps() {
+        assertThat(jdbiOps).isNotNull().isInstanceOf(JdbiTemplate.class);
+    }
+
+    @Configuration
+    @EnableTransactionManagement
```
arteam: I believe to make `@EnableTransactionManagement` work you need to define a transaction manager in the configuration. In our case `DataSourceTransactionManager`:
```java
@Bean
public PlatformTransactionManager transactionManager(DataSource dataSource) {
    return new DataSourceTransactionManager(dataSource);
}
```

spring4/src/main/java/org/jdbi/v3/spring4/JdbiTemplate.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * like {@link Jdbi#withHandle(HandleCallback)} but respects springs transactions
+     */
+    @Override
+    public <R, X extends Exception> R withHandle(HandleCallback<R, X> callback) throws X {
+        Handle h = JdbiUtil.getHandle(jdbi);
+        R result = callback.withHandle(h);
+        // close if not in transaction
+        JdbiUtil.closeIfNeeded(h);
```
arteam: I believe we shouldn't close the handle here. It will be closed in the `afterCompletion` method. I think also it will probably be ignored because the handle is still binded to the active transaction.

spring4/src/test/java/org/jdbi/v3/spring4/TestJdbiTemplate.java
```diff
@@ -0,0 +1,64 @@
...
+        public Jdbi jdbi(DataSource dataSource) {
+            return Jdbi.create(dataSource);
+        }
+
+        @Bean
+        public JdbiOperations jdbiOperations(Jdbi jdbi) {
+            return new JdbiTemplate(jdbi);
+        }
+
+    }
```
arteam: It would be nice to add a test how to use Dao class with Spring transactions and `JdbiTemplate`. Something like:
```java
        @Bean
        public Dao dao(JdbiOperations jdbiOperations) {
            return new Dao(jdbiOperations);
        }

    }

    public static class Dao {

        private JdbiOperations jdbiOperations;

        public Dao(JdbiOperations jdbiOperations) {
            this.jdbiOperations = jdbiOperations;
        }

        @Transactional
        public Integer getAmount() {
            return jdbiOperations.withHandle(h -> h.createQuery("select count(*) from something")
                    .mapTo(Integer.class)
                    .findOnly());
        }

        @Transactional
        public void insert(int id, String name) {
            jdbiOperations.useHandle(h -> h.createUpdate("insert into something (id, name) values (?, ?)")
                    .bind(0, id)
                    .bind(1, name)
                    .execute());
        }

        @Transactional
        public void insertMany(List<Integer> ids, List<String> names) {
            for (int i = 0; i < ids.size(); i++) {
                insert(ids.get(i), names.get(i));
            }
        }
    }

    @Test
    public void testTransactionsWork() {
        dao.insertMany(Arrays.asList(1, 2, 3), Arrays.asList("Test1", "Test2", "Test3"));
        System.out.println(dao.getAmount());
    }
```

spring4/src/main/java/org/jdbi/v3/spring4/JdbiOperations.java
alwins0n: Didn't think about that, but yeah, can do

spring4/src/main/java/org/jdbi/v3/spring4/JdbiTemplate.java
alwins0n: that's point - just because you use JdbiTemplate does not guarantee that you are within a spring transaction (most likely still). 

closeIfNeeded closes the handle if it's NOT transaction bound, so exactly what I want here.

spring4/src/test/java/org/jdbi/v3/spring4/TestJdbiTemplate.java
alwins0n: maybe, will look that up. I am used to spring boot where the TM is defaulty defined as long as as a `PlatformTransactionManager` is on the classpath. 

then again the test suite is far from complete - this is WIP

spring4/src/test/java/org/jdbi/v3/spring4/TestJdbiTemplate.java
alwins0n: I will add checks to verify the behaviour I expect from the plugin. this is WIP

alwins0n:
I should add the following: in this PR the Transaction Adapter for spring was modified to close the handle upon completion (as it should imho). I don't think this plugin was used a lot in the wild, but still - this change would maybe modify the behaviour of existing applications.

Either we should not directly expose JdbiUtil - ppl who upgrad the jdbi3-spring would have to deal with this (and maybe replace by JdbiTemplate) or leave public and deprecate the methods as they should not be used directly anymore (and the user will be notified of the change in behaviour)

spring4/src/main/java/org/jdbi/v3/spring4/JdbiOperations.java
qualidafial: Actually.. `Handle.useHandle()` is already delegating to `withHandle()` :man_shrugging:

alwins0n:
fyi: do not merge this already. found some cases where tests will fail. I am trying to understand spring and jdbi transaction management better. At this point I think a custom `PlatformTransactionManager` for Jdbi that does not operate on the DataSource directly but on Jdbi's Handles might be a clean and correct solution. Will update this PR asap

spring4/src/main/java/org/jdbi/v3/spring4/JdbiOperations.java
alwins0n: true that, will leave it that way

spring4/src/main/java/org/jdbi/v3/spring4/JdbiTemplate.java
alwins0n: became obsolete with latest push

alwins0n:
ough... after re-reading spring docs and digging into the transaction SPI as well as jdbi's interface for transactions - I decided to give a fully blown `JdbiTransactionManager` a go!

I pushed the latest version of this draft and its kind of a big one. I do not yet feel confident about everything but I think this is the way to go integrating spring with jdbi.

I tried to reuse as much convienience (Template classes etc.) from spring as possible.

Some points that need to be addressed:
- I rewrote lots of `JdbiUtil`, it breaks the old tests and implementations depending on it. After re-reading the code I think the present logic of JdbiUtil should be packed into `JdbiTemplate` and leave the old implementation be - and deprecate the api, if you feel that we should.
- Now that we have a jdbi `PlatformTransactionManager` we can invoke framework transactions also programmatically. I will show this in the future docs
- Docs need a full rewrite. I intend keeping part of the old docs because the setup with `JdbiFactoryBean` can be very nice/convenient in xml-config. If you approve these changes/PR I will update the docs accordingly

Merry Christmas btw

qualidafial:
I would definitely prefer deprecating JdbiUtil over making an incompatible change.

alwins0n:
Thanks for your comment. I will take that into account once this change is approved on principal. I would also invite ppl to question my approach. The reason I did this larger change (with respect to the existing implementation) was that I think if springs `DatasourceTransactionManager` operates directly on the Ds and its connections - Jdbi might not know what is going on. That way e.g. `Handle`s would be left unclosed after a transaction (where the underlying `Connection` will be closed) and might fail/behave unexpectedly when operated on them.

But maybe I am missing something and this cannot occur anyways and thus I am overcomplicating things here. Please provide feedback! Should we try to get someone from spring framework to comment here? About this being the "right" or "clean" approach?

spring4/src/main/java/org/jdbi/v3/spring4/JdbiTransactionObject.java
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.transaction.TransactionSystemException;
+
+/**
+ * savepoint capable jdbi transaction wrapper
+ */
+public class JdbiTransactionObject implements SavepointManager {
+
+    private static final String JDBI_SAVEPOINT_BASENAME = "JDBI_SAVEPOINT_";
+
+    private int nextSavePointNumber = 0;
```
stevenschlansker: should this be an `AtomicInteger`?

spring4/src/main/java/org/jdbi/v3/spring4/JdbiTransactionObject.java
```diff
@@ -0,0 +1,82 @@
...
+    private HandleHolder handleHolder;
+
+    public JdbiTransactionObject(HandleHolder handleHolder) {
+        this.handleHolder = handleHolder;
+    }
+
+    @Override
+    public Object createSavepoint() throws TransactionException {
+        String savePointName = JDBI_SAVEPOINT_BASENAME + nextSavePointNumber++;
```
stevenschlansker: because this might be called from multiple concurrent threads, I'd guess

spring4/src/test/java/org/jdbi/v3/spring4/TestJdbiTemplate.java
```diff
@@ -0,0 +1,151 @@
...
+        @Bean
+        public DataSource dataSource() {
+            return new EmbeddedDatabaseBuilder()
+                    .setType(EmbeddedDatabaseType.H2)
+                    .build();
+        }
+
+        @Bean
+        public Jdbi jdbi(DataSource dataSource) {
+            return Jdbi.create(dataSource);
```
stevenschlansker: What if this instead created a `DataSouce` implementation that does the `JdbiTemplate` / `JdbiUtil` logic, and delegates to the `dataSource` parameter?

spring4/src/main/java/org/jdbi/v3/spring4/JdbiUtil.java
```diff
@@ -55,38 +53,9 @@ public static Handle getHandle(Jdbi jdbi)
      */
     public static void closeIfNeeded(Handle handle)
     {
-        if (!TRANSACTIONAL_HANDLES.contains(handle))
+        if (UNMANAGED_HANDLES.contains(handle))
```
stevenschlansker: this looks like a memory leak, should this be `remove`?

spring4/src/main/java/org/jdbi/v3/spring4/JdbiTemplate.java
```diff
@@ -0,0 +1,50 @@
...
+    private final Jdbi jdbi;
+
+    public JdbiTemplate(Jdbi jdbi) {
+        this.jdbi = jdbi;
+    }
+
+    /**
+     * like {@link Jdbi#withHandle(HandleCallback)} but respects springs transactions
+     */
+    @Override
```
stevenschlansker: Ok, I think I am starting to see the intent here.  But I wonder, it looks like the only purpose of `JdbiTemplate` is to modify the behavior of `withHandle`.  Instead of introducing it as a wrapper with an interface, what if we wrap the `ConnectionFactory` that gets called in `Jdbi.open()` to invoke the `JdbiUtil.getHandle`?  And for the close logic, you could hook the `StatementBuilder.close()` logic

Then there is no need for this wrapper class, I think.  Does that sound like it might work?

spring4/src/main/java/org/jdbi/v3/spring4/JdbiTransactionObject.java
alwins0n: hmm. on principal the transaction and any transaction related objects are afaik thread bound (at least in the local context of a spring application). that means no more than one thread will ever increment the number of save points.

I will look that up though and make it safe for general purposes if necessary

spring4/src/main/java/org/jdbi/v3/spring4/JdbiTemplate.java
alwins0n: Interesting, let me think this through loudly:

If the connectionfactory is "spring transaction aware" the callstack of using jdbi within a spring transaction would approximately look like this

`Jdbi.useHandle` -> `JdbiUtil.getHandle` -> doStuffWith handle -> `Handle.close` -> `StatementBuilder.close` and `Connection.close`

everything after the "do Stuff" I find problematic. 
- The handle does not know anything about about the transaction
- `StatementBuilder.close` could do the right thing, thats okay!
- the `java.sql.Connection` is closed (Line 132) - bad!
- the handle is marked as closed (released) which may or may not be correct!
- ...

also:
- what if you used `Jdbi.open directly` - you now must not close it but use `JdbiUtil.closeIfNeeded`. are you supposed to know that?
- what if you create a transaction manually? `Jdbi.useTransaction` and so on
...

I have had these concerns before and thats why I decided 
- to wrap the calls to Jdbi via transaction aware classes - so Jdbi is properly informed about every change in the transactions and the handle state reflects the state in the transaction.
- to discourage the use of transaction methods on `Jdbi` - you can start a transaction programmatically via springs mechanisms
- to expose a spring-like API for convenience

If you see any possibility to go without the API  - I am happy to implement it that way. But I am not really happy with static calls and Jdbi not knowing about springs state and vice versa

spring4/src/test/java/org/jdbi/v3/spring4/TestJdbiTemplate.java
alwins0n: That's okay, but see my comment above. I think in this case spring manipulates stuff and state which Jdbi does not know anything about. also other problems I mentioned arise.

alwins0n:
Can I get a ping if this is still of interest? 

The way I see it you are not happy with introducing another API but agree with the need of having a custom `PlatformTransactionManager` that keeps jdbi and spring in sync. Am I correct?

Regarding the API I investigated whether `JdbiTemplate` could be removed.

- It might be possible to modify the `Jdbi` class in a way that makes a direct usage possible, but it would be heavy modifications. Even if `Jdbi.open()` has a hook for getting the transaciton handle. `Jdbi.withHandle()` closes it after usage and would need a lot of hook code too, etc...
- If you use the plugin you are a Spring developer. I claim here, that Spring developers have no problem using a `xxxTemplate` class to achieve whatever they want in a "spring" way. e.g. have a look at: http://www.buggybread.com/2015/03/spring-framework-list-of-template.html . I would also be willing to remove the `JdbiOperations` interface altough its convention to provide one (for proxying, stubbing, testing)

Btw I noticed that the old version is also not in sync with the spring-style naming conventions for static access utility functions. `JdbiUtil` should be `JdbiUtils` analogue to `DatasourceUtils`, `LdapUtils`, ...

stevenschlansker:
Yes, still of interest!  I'll do more reviewing shortly.

alwins0n:
just wanted to say that I am actively using this PR in a real world project right now. Thus, I would really appreciate feedback. Right now it is running quite well.

I realize that this PR is quite big and the incentive to dig into spring is quite low. I can lay out my design decisions once more to have them challenged properly. Maybe some sequence diagrams might help to understand my reasoning.

Would love to hear how this is going forward.

stevenschlansker:
Never fear, you aren't forgotten.  You're right that Spring support isn't top of our list but we definitely intend to move forward and testing the PR in a real world project is great progress.

alwins0n:
hey guys, I want to move this forward - providing API methods for SQL object (useExtension, onDemand) on the high level approach (JdbiTemplate) as well as tests and updated docs.

Also spring 5 arrived which should be taken into consideration to support primarily. And autoconfigurations for spring boot 2 could be implemented.

should we try to get someone from spring on board to review this? should I publish it standalone? what's the move here?

stevenschlansker:
OK, I don't see any reason to hold it up further - if you want to do another pass at the docs, and I noticed a couple of TODO test cases, then we can merge it!

qualidafial:
@alwins0n I'm curious if `TransactionAwareDataSourceProxy` (whew!) mentioned on #1023 made any difference. You had mentioned you were going to try it out and see if it spares us the complexity of wrapping the transactions ourselves.

alwins0n:
Haven't tried it out yet, but as I said, for the most use cases it will work properly. Again: it might be a problem if `Handle` and `Connection` are not in sync due to this "quickfix" solution. 

Off the top of my head, let me give you an example:
I request a handle from `Jdbi` somewhere within a spring transaction (e.g. annotated method). I do stuff with it and leave the annotated method. Now spring will close the underlying connection of this transaction. `Jdbi`s `Handle` does not know about this. If i plan to use it again (another call to the database) I will go fail (closed unexpected... etc)

On the other hand:
I request (open) a handle from `Jdbi` somewhere within a spring transaction (e.g. annotated method). Now i close it (try finally) since this is what I think I am expected to do. This works, but after control flow leaves the annotated method spring will attempt to close/release the connection -> this (might) fail and results in exceptions (probably).

I did not do thourough tests. Also example 1 might be a corner case (using a handle in one method and passing it out to reuse). But it seemed to me from the beginning that `TransactionAwareDataSourceProxy` and the likes are a hack! it might work most of the time but the proper way to integrate spring and jdbi is to teach spring that `Handle` is the transactional resource to be synced, not the `Connections`.

I know that the code above is scary to non-"springers" (hell, to everyone actually) but I think a library should provide the proper solution, not the hack, even if that means somewhat bloated code.

To wrap up - I will run some tests against the `TransactionAwareDataSourceProxy` to give myself a clear picture. but the mismatch of spring and jdbi as i mentioned will never go away. It just might be that it's not as "bad" as suspected by me

alwins0n:
Can I ask for a decision here? IMHO either:
- document the usage of `TransactionAwareDataSourceProxy` in the plugin docs and make users aware that `Handle` and the underlying connection may not be in sync (e.g. closed). This might not pose a problem if JDBI is purly used in "lambda mode"
- update the plugin based on this PR which is a clean solution but requires the introduction of a Spring-like API ( `JdbiTemplate`)

qualidafial:
Hi @alwins0n, I just realized I never got back to you.

I don't feel like we have enough information to make a call here. On your March 23 comment, you said you were planning to try out `TransactionAwareDataSourceProxy` and see where that gets us. I'm still waiting to hear your insights from that experiment.

Speaking personally, I am a little gun shy about adding Spring integration APIs, because I don't want to screw it up. It's hard for me to judge what's idiomatic in Spring because I don't use it--I use Dropwizard at my day job.

So I'm relying rather heavily on you and other community members who care to weigh in on what the best course of action should be.

If you've already posted your findings and I've lost track of it, please link me to it.

alwins0n:
Well, I used my current test suite to run against a `TransactionAwareDataSourceProxy` wiring and it worked fine. That are my results as of now.

It is probably the best and simplest solution to document that configuration and leave the plugin be, I have to admit.

We should just warn that passing around handles to other methods (which may trigger leaving springs transaction context) or creating transactions on jdbi manually while being in springs transaciton context may lead to unexpected results.

as long as you rely on spring for transaction management and use jdbi's handle purely in "callback" mode (which is 99% of all cases) you should be fine.

Some minor improvements to the plugin could be made anyway as I mentioned on the whole thread here (naming etc.)

alwins0n:
I am, but where? Is it really part of the plugin documentation? It feels more like some blog-post content (how to integrate jdbi with spring....)

stevenschlansker:
I was imagining it'd have a whole section in the main jdbi docs, perhaps near the end with other "addons".
Also, if we are willing to tag the module as `@Beta` or `@Experimental`, that'd go a long way towards easing my worries about keeping a stable api.

spring4/src/main/java/org/jdbi/v3/spring4/JdbiUtil.java
```diff
@@ -13,20 +13,19 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
 /**
  * Utility for working with Jdbi and Spring transaction bound resources
  */
 public class JdbiUtil
 {
-    private static final Set<Handle> TRANSACTIONAL_HANDLES = new HashSet<>();
+    private static final Set<Handle> UNMANAGED_HANDLES = new HashSet<>();
```
osi: This set should be either synchronized or a set made from a concurrent hash map

stevenschlansker:
Just checking in -- does anyone use this code in production / how's testing going?  Do you still intend to continue with the checklist of TODOs above?

stevenschlansker:
Sounds like we're going to close this PR for now, but if anyone is interested in carrying the flag across the finish line we can always re-open it.

PR #995, opened 2017-12-30
Sqlite support for Jdbi

### LINKED ISSUE #994: Exception thrown when binding argument of supported type.
I'm trying to insert a bean into an sqlite database. The bean has a url field of type URL. When I try inserting it into the database i get this error:

````java
Exception in thread "main" org.jdbi.v3.core.statement.UnableToCreateStatementException: Exception while binding named parameter 'url' [statement:"INSERT INTO book (title, url) VALUES (:title, :url);", rewritten:"INSERT INTO book (title, url) VALUES (:title, :url);", parsed:"ParsedSql{sql='INSERT INTO book (title, url) VALUES (?, ?);', parameters=ParsedParameters{positional=false, parameterNames=[title, url]}}", arguments:{ positional:{}, named:{}, finder:[{lazy bean property arguments "jdbc.Book@5f3a4b84"]}]
	at org.jdbi.v3.core.statement.ArgumentBinder.bindNamed(ArgumentBinder.java:63)
	at org.jdbi.v3.core.statement.ArgumentBinder.bind(ArgumentBinder.java:27)
	at org.jdbi.v3.core.statement.SqlStatement.internalExecute(SqlStatement.java:1443)
	at org.jdbi.v3.core.result.ResultProducers.lambda$returningUpdateCount$0(ResultProducers.java:39)
	at org.jdbi.v3.core.statement.Update.execute(Update.java:53)
	at org.jdbi.v3.core.statement.Update.execute(Update.java:41)
	at org.jdbi.v3.sqlobject.statement.internal.SqlUpdateHandler.lambda$new$2(SqlUpdateHandler.java:56)
	at org.jdbi.v3.sqlobject.statement.internal.SqlUpdateHandler.lambda$configureReturner$3(SqlUpdateHandler.java:69)
	at org.jdbi.v3.sqlobject.statement.internal.CustomizingStatementHandler.invoke(CustomizingStatementHandler.java:155)
	at org.jdbi.v3.sqlobject.statement.internal.SqlUpdateHandler.invoke(SqlUpdateHandler.java:29)
	at org.jdbi.v3.sqlobject.SqlObjectFactory.lambda$null$13(SqlObjectFactory.java:162)
	at org.jdbi.v3.core.LazyHandleSupplier.invokeInContext(LazyHandleSupplier.java:73)
	at org.jdbi.v3.sqlobject.SqlObjectFactory.lambda$createInvocationHandler$14(SqlObjectFactory.java:161)
	at com.sun.proxy.$Proxy4.addBook(Unknown Source)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.jdbi.v3.core.OnDemandExtensions.lambda$null$0(OnDemandExtensions.java:60)
	at org.jdbi.v3.core.Jdbi.withExtension(Jdbi.java:464)
	at org.jdbi.v3.core.OnDemandExtensions.lambda$create$1(OnDemandExtensions.java:57)
	at com.sun.proxy.$Proxy4.addBook(Unknown Source)
````
Here are my dao interface and bean class

````java
public interface BookDao {
    @SqlUpdate("CREATE TABLE IF NOT EXISTS book (" +
            "id INTEGER PRIMARY KEY, " +
            "title TEXT, " +
            "url TEXT);")
    void createTableIfNotExits();

    @SqlQuery("SELECT * FROM book;")
    @RegisterConstructorMapper(Book.class)
    List<Book> getAllBooks();

    @SqlUpdate("INSERT INTO book (title, url) VALUES (:title, :url);")
    boolean addBook(@BindBean Book book);

    @SqlUpdate("DELETE FROM book WHERE id=:id;")
    boolean removeBooks(@BindBean Book book);
}
````

````java
public class Book {
    private final int id;
    private String title;
    private URL url;

    public Book(@ColumnName("id") int id, @ColumnName("title") String title, @ColumnName("url") URL url) {
        this.id = id;
        this.title = title;
        this.url = url;
    }

    public int getId() {
        return id;
    }

    public String getTitle() {
        return title;
    }

    public URL getUrl() {
        return url;
    }
}
````

According to the documentation, URL is a supported type so i'm thinking it would automatically be mapped to a string for me. Am I doing something wrong?

### PR DESCRIPTION
**Pull request by edwinnyawoli**
Added plugin to support using sqlite with jdbi. 
Fixes #994

### DIFF
--- a/sqlite/src/main/java/org/jdbi/v3/sqlite3/SQLitePlugin.java
+++ b/sqlite/src/main/java/org/jdbi/v3/sqlite3/SQLitePlugin.java
@@ -19,7 +19,7 @@ import org.jdbi.v3.core.spi.JdbiPlugin;
 import java.net.URL;
 
 /**
- * Jdbc plugin for SQLite
+ * Jdbi plugin for SQLite.
  */
 public class SQLitePlugin implements JdbiPlugin {
     @Override
--- a/sqlite/src/main/java/org/jdbi/v3/sqlite3/SQLitePlugin.java
+++ b/sqlite/src/main/java/org/jdbi/v3/sqlite3/SQLitePlugin.java
@@ -1,0 +1,27 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.sqlite3;
+
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.spi.JdbiPlugin;
+
+/**
+ * Jdbc plugin for SQLite
+ */
+public class SQLitePlugin implements JdbiPlugin {
+    @Override
+    public void customizeJdbi(Jdbi jdbi) {
+        jdbi.registerArgument(new URLArgumentFactory());
+    }
+}
--- a/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLArgumentFactory.java
+++ b/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLArgumentFactory.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jdbi.v3.sqlite3;
+
+import org.jdbi.v3.core.argument.AbstractArgumentFactory;
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.argument.ArgumentFactory;
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+import java.net.URL;
+import java.sql.Types;
+
+/**
+ * Build URL objects as strings.
+ */
+public class URLArgumentFactory extends AbstractArgumentFactory<URL> {
+    /**
+     * Constructs an {@link ArgumentFactory} for type {@code URL}.
+     */
+    public URLArgumentFactory() {
+        super(Types.VARCHAR);
+    }
+
+    @Override
+    protected Argument build(URL url, ConfigRegistry config) {
+        return (position, statement, ctx) -> statement.setString(position, url.toString());
+    }
+}
--- a/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLArgumentFactory.java
+++ b/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLArgumentFactory.java
@@ -25,7 +25,7 @@ import java.sql.Types;
 /**
  * Build URL objects as strings.
  */
-public class URLArgumentFactory extends AbstractArgumentFactory<URL> {
+class URLArgumentFactory extends AbstractArgumentFactory<URL> {
     /**
      * Constructs an {@link ArgumentFactory} for type {@code URL}.
      */
--- a/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
+++ b/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
@@ -22,7 +22,7 @@ import java.net.URL;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 
-public class URLColumnMapper implements ColumnMapper<URL> {
+class URLColumnMapper implements ColumnMapper<URL> {
 
     @Override
     public URL map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
--- a/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
+++ b/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
@@ -13,18 +13,26 @@
  */
 package org.jdbi.v3.sqlite3;
 
-import org.jdbi.v3.core.Jdbi;
-import org.jdbi.v3.core.spi.JdbiPlugin;
+import org.jdbi.v3.core.mapper.ColumnMapper;
+import org.jdbi.v3.core.statement.StatementContext;
 
+import java.net.MalformedURLException;
+import java.net.URI;
 import java.net.URL;
+import java.sql.ResultSet;
+import java.sql.SQLException;
 
-/**
- * Jdbc plugin for SQLite
- */
-public class SQLitePlugin implements JdbiPlugin {
+public class URLColumnMapper implements ColumnMapper<URL> {
+
     @Override
-    public void customizeJdbi(Jdbi jdbi) {
-        jdbi.registerArgument(new URLArgumentFactory());
-        jdbi.registerColumnMapper(new URLColumnMapper());
+    public URL map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
+        URL url = null;
+
+        try {
+            url = URI.create(r.getString(columnNumber)).toURL();
+        } catch (MalformedURLException e) {
+            e.printStackTrace();
+        }
+        return url;
     }
 }
--- a/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
+++ b/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
@@ -29,7 +29,7 @@ class URLColumnMapper implements ColumnMapper<URL> {
         URL url = null;
 
         try {
-            url = URI.create(r.getString(columnNumber)).toURL();
+            url = new URL(r.getString(columnNumber));
         } catch (MalformedURLException e) {
             throw new SQLException(e);
         }
--- a/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
+++ b/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
@@ -31,7 +31,7 @@ public class URLColumnMapper implements ColumnMapper<URL> {
         try {
             url = URI.create(r.getString(columnNumber)).toURL();
         } catch (MalformedURLException e) {
-            e.printStackTrace();
+            throw new SQLException(e);
         }
         return url;
     }
--- a/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
+++ b/sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
@@ -26,13 +26,10 @@ class URLColumnMapper implements ColumnMapper<URL> {
 
     @Override
     public URL map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
-        URL url = null;
-
         try {
-            url = new URL(r.getString(columnNumber));
+            return new URL(r.getString(columnNumber));
         } catch (MalformedURLException e) {
             throw new SQLException(e);
         }
-        return url;
     }
 }
--- a/sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
+++ b/sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
@@ -48,8 +48,7 @@ public class TestUrls {
     @Test
     public void testInsertUrlSuccessful() throws MalformedURLException {
         String goolgeString = "http://www.google.com";
-        URL googleUrl = null;
-        googleUrl = URI.create(goolgeString).toURL();
+        URL googleUrl = new URL(goolgeString);
 
         handle.createUpdate("INSERT INTO foo VALUES (:url)")
                 .bind("url", googleUrl)
--- a/sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
+++ b/sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
@@ -46,8 +46,8 @@ public class TestUrls {
 
     @Test
     public void testInsertUrlSuccessful() throws MalformedURLException {
-        String goolgeString = "http://www.google.com";
-        URL googleUrl = new URL(goolgeString);
+        String googleString = "http://www.google.com";
+        URL googleUrl = new URL(googleString);
 
         handle.createUpdate("INSERT INTO foo VALUES (:url)")
                 .bind("url", googleUrl)
@@ -67,7 +67,7 @@ public class TestUrls {
     }
 
     @Test
-    public void testInserUrlUsingBindByType() throws MalformedURLException {
+    public void testInsertUrlUsingBindByType() throws MalformedURLException {
         URL githubUrl = new URL("http://www.github.com");
 
         handle.createUpdate("INSERT INTO foo VALUES (:url)")
--- a/sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
+++ b/sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
@@ -78,4 +78,13 @@ public class TestUrls {
         Assertions.assertThat(dbUrl).hasToString(githubUrl.toString());
     }
 
+    @Test(expected = ResultSetException.class)
+    public void testInsertNullUrlUsingBindByType() {
+        handle.createUpdate("INSERT INTO foo VALUES (:url)")
+                .bindByType("url", null, URL.class)
+                .execute();
+
+        handle.createQuery("SELECT url FROM foo").mapTo(URL.class).findOnly();
+    }
+
 }
--- a/sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
+++ b/sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
@@ -36,7 +36,6 @@ public class TestUrls {
         jdbi.installPlugin(new SQLitePlugin());
         handle = jdbi.open();
         handle.useTransaction(handle -> {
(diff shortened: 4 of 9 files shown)

### CODE REVIEW

File: sqlite/src/main/java/org/jdbi/v3/sqlite3/URLArgumentFactory.java [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+import org.jdbi.v3.core.argument.ArgumentFactory;
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+import java.net.URL;
+import java.sql.Types;
+
+/**
+ * Build URL objects as strings.
+ */
+public class URLArgumentFactory extends AbstractArgumentFactory<URL> {
```
qualidafial: Let's make this and the mapper class package private so that the plugin is the only public API

sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
```diff
@@ -0,0 +1,38 @@
...
+public class URLColumnMapper implements ColumnMapper<URL> {
+
+    @Override
+    public URL map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
+        URL url = null;
+
+        try {
+            url = URI.create(r.getString(columnNumber)).toURL();
+        } catch (MalformedURLException e) {
+            e.printStackTrace();
```
qualidafial: I'm of two minds on this exception. On one hand it seems like the more correct approach is to raise the exception rather than swallow it. On the other hand, that means if a malformed URL somehow gets into the database, you could never read that row out again without an exception--lame.

Any any rate, this should be logged with SLF4J rather than printed to STDOUT

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Test
+    public void testInsertUrlSuccessful() {
+        String goolgeString = "http://www.google.com";
+        URL googleUrl = null;
+        try {
+            googleUrl = URI.create(goolgeString).toURL();
+        } catch (MalformedURLException e) {
+            e.printStackTrace();
```
qualidafial: Test method should just throw exception rather than swallowing an unexpected exception.

qualidafial:
Can we also get a section in `index.adoc` about this new module?

sqlite/src/main/java/org/jdbi/v3/sqlite3/SQLitePlugin.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.jdbi.v3.sqlite3;
+
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.spi.JdbiPlugin;
+
+import java.net.URL;
+
+/**
+ * Jdbc plugin for SQLite
```
stevenschlansker: This is a `jdbi` plugin :)
Also the javadoc is missing the required `.` termination.

sqlite/src/main/java/org/jdbi/v3/sqlite3/URLArgumentFactory.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Constructs an {@link ArgumentFactory} for type {@code URL}.
+     */
+    public URLArgumentFactory() {
+        super(Types.VARCHAR);
+    }
+
+    @Override
+    protected Argument build(URL url, ConfigRegistry config) {
+        return (position, statement, ctx) -> statement.setString(position, url.toString());
```
stevenschlansker: Should this handle a null url?

sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
stevenschlansker: I think I'd rather it throw.  The data you requested from the database cannot be represented in the requested container type.  `null` is indistinguishable from a valid return of a different value, so that's no good.  So we must fail.

If your data may be a URL or maybe not, you should fetch it as a string and guard the conversion, IMO.  Or we could make this a configurable policy with choices of `THROW` vs `SWALLOW_RETURN_NULL`.  But I think the default has to be to throw the exception.

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
```diff
@@ -0,0 +1,66 @@
...
+
+    Handle handle;
+
+    @Before
+    public void setUp() throws Exception {
+        Jdbi jdbi = Jdbi.create("jdbc:sqlite::memory:");
+        jdbi.installPlugin(new SQLitePlugin());
+        handle = jdbi.open();
+        handle.useTransaction(handle -> {
+            handle.execute("DROP TABLE IF EXISTS foo");
```
stevenschlansker: I don't think `foo` can ever exist here, right, since the above opens a unique in-memory db?

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
```diff
@@ -0,0 +1,66 @@
...
+        String goolgeString = "http://www.google.com";
+        URL googleUrl = null;
+        try {
+            googleUrl = URI.create(goolgeString).toURL();
+        } catch (MalformedURLException e) {
+            e.printStackTrace();
+        }
+
+        handle.createUpdate("INSERT INTO foo VALUES (:url)")
+                .bind("url", googleUrl)
```
stevenschlansker: Also test that e.g. `bindByType("url", null, URL.class)` works

sqlite/src/main/java/org/jdbi/v3/sqlite3/URLArgumentFactory.java
qualidafial: `AbstractArgumentFactory` already handles nulls.

sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
qualidafial: Yes, I think you're right. +1 to throw

sqlite/src/main/java/org/jdbi/v3/sqlite3/SQLitePlugin.java
edwinnyawoli: lol...the names got mixed up in my head.

qualidafial: I don't know why, they're totally different :trollface:

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
edwinnyawoli: yeah...you're right

sqlite/src/main/java/org/jdbi/v3/sqlite3/SQLitePlugin.java
edwinnyawoli: 😅

sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
```diff
@@ -0,0 +1,38 @@
...
+import java.sql.SQLException;
+
+class URLColumnMapper implements ColumnMapper<URL> {
+
+    @Override
+    public URL map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
+        URL url = null;
+
+        try {
+            url = URI.create(r.getString(columnNumber)).toURL();
```
arteam: We could just use `return`, because `MalformedURLException` is wrapped by another exception and we have only one point of return in the method.

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
```diff
@@ -0,0 +1,61 @@
...
+public class TestUrls {
+
+    Handle handle;
+
+    @Before
+    public void setUp() throws Exception {
+        Jdbi jdbi = Jdbi.create("jdbc:sqlite::memory:");
+        jdbi.installPlugin(new SQLitePlugin());
+        handle = jdbi.open();
+        handle.useTransaction(handle -> {
```
arteam: We could use a more simple lambda-expression form without the curly braces:
```java
handle.useTransaction(h -> h.execute("CREATE TABLE foo(url URL);"))
```

arteam: We could use a more simple form of lambda-expressions without the curly braces:
```java
handle.useTransaction(h -> h.execute("CREATE TABLE foo(url URL);"))
```

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
```diff
@@ -0,0 +1,61 @@
...
+import java.net.URI;
+import java.net.URL;
+
+public class TestUrls {
+
+    Handle handle;
+
+    @Before
+    public void setUp() throws Exception {
+        Jdbi jdbi = Jdbi.create("jdbc:sqlite::memory:");
```
arteam: I think it would be nice to add some random  suffix to the database name to avoid test clashes if we will have more SQLite tests in the future.

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
```diff
@@ -0,0 +1,61 @@
...
+    @After
+    public void tearDown() throws Exception {
+        handle.close();
+    }
+
+    @Test
+    public void testInsertUrlSuccessful() throws MalformedURLException {
+        String goolgeString = "http://www.google.com";
+        URL googleUrl = null;
+        googleUrl = URI.create(goolgeString).toURL();
```
arteam: The `URL` class has a constructor which accepts `String`, so we can initialize it without `URI`:
```java
new URL("http://www.google.com")
```

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
```diff
@@ -0,0 +1,61 @@
...
+
+        handle.createUpdate("INSERT INTO foo VALUES (:url)")
+                .bind("url", googleUrl)
+                .execute();
+
+        Assertions.assertThat(handle.createQuery("SELECT * FROM foo").mapTo(URL.class).findOnly()).isEqualTo(googleUrl);
```
arteam: Just to be sure, I would avoid loading all columns from `foo` and specify the `url` column in `SELECT`

arteam: Just FYI. this `equals` check actually makes a DNS request to resolve a Google's IP address and compare URLs by them. Relevant: http://brian.pontarelli.com/2006/12/05/mr-gosling-why-did-you-make-url-equals-suck/

stevenschlansker: Great catch, maybe we should check `hasToString` instead?

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
stevenschlansker: Single statements also are implicitly in a transaction, so it might be entirely redundant.

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
edwinnyawoli: Thanks for the info.

edwinnyawoli: How come it hasn't been changed since? :thinking:

qualidafial: Oh my god.

sqlite/src/main/java/org/jdbi/v3/sqlite3/URLColumnMapper.java
edwinnyawoli: I'm not quite clear on what you're saying. Do you mean it should return the null in the catch block?

arteam: No, I meant you could just directly return `URI.create(r.getString(columnNumber)).toURL();` without declaring the `url` variable.

edwinnyawoli: Oh ok. I get it now

qualidafial:
Looking over the comments, the remaining changes appear to be:

* Test with `bindByType()`
* Test mapping null URLs
* Use a random database name to avoid test conflicts

@edwinnyawoli Thanks very much--looking forward to merging this!

edwinnyawoli:
I'll make the remaining changes and push soon. Thanks as well for the opportunity.

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
edwinnyawoli: Is it possible to specify a name for in memory databases?

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
```diff
@@ -48,14 +46,14 @@ public void tearDown() throws Exception {
     @Test
     public void testInsertUrlSuccessful() throws MalformedURLException {
         String goolgeString = "http://www.google.com";
```
stevenschlansker: typo `goolge`

edwinnyawoli:
Concerning the random database name, are you suggesting using a file with a random name vs the in memory database?

arteam:
I think you can just append a random UUID to the memory database name; something like `"jdbc:sqlite::memory:" + UUID.randomUUID()`.

edwinnyawoli:
An exception is thrown when creating the database with

````java
        Jdbi jdbi = Jdbi.create("jdbc:sqlite::memory:" + UUID.randomUUID().toString());
````

````java
    org.jdbi.v3.core.ConnectionException: java.sql.SQLException: opening db: ':memory:9f93fda2-466c-4a07-981e-f111d785eca9': The filename, directory name, or volume label syntax is incorrect

	at org.jdbi.v3.core.Jdbi.open(Jdbi.java:321)
	at TestUrls.setUp(TestUrls.java:38)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:24)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
Caused by: java.sql.SQLException: opening db: ':memory:9f93fda2-466c-4a07-981e-f111d785eca9': The filename, directory name, or volume label syntax is incorrect
	at org.sqlite.core.CoreConnection.open(CoreConnection.java:203)
	at org.sqlite.core.CoreConnection.<init>(CoreConnection.java:76)
	at org.sqlite.jdbc3.JDBC3Connection.<init>(JDBC3Connection.java:25)
	at org.sqlite.jdbc4.JDBC4Connection.<init>(JDBC4Connection.java:24)
	at org.sqlite.SQLiteConnection.<init>(SQLiteConnection.java:45)
	at org.sqlite.JDBC.createConnection(JDBC.java:114)
	at org.sqlite.JDBC.connect(JDBC.java:88)
	at java.sql.DriverManager.getConnection(DriverManager.java:664)
	at java.sql.DriverManager.getConnection(DriverManager.java:270)
	at org.jdbi.v3.core.Jdbi.lambda$create$0(Jdbi.java:103)
	at org.jdbi.v3.core.Jdbi.open(Jdbi.java:305)
	... 24 more
````

**Comment by arteam** [comment]
It looks like the SQLite driver doesn't support several in-memory databases. Okay, let it slip, then.

File: sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java [hunk 14]
```diff
@@ -45,15 +46,36 @@ public void tearDown() throws Exception {
...
 
-        URL actualUrl = handle.createQuery("SELECT * FROM foo").mapTo(URL.class).findOnly();
+        URL actualUrl = handle.createQuery("SELECT url FROM foo").mapTo(URL.class).findOnly();
         Assertions.assertThat(actualUrl).hasToString(googleUrl.toString());
     }
 
+    @Test(expected = ResultSetException.class)
+    public void testMapNullUrlThrowsException() {
+        handle.createUpdate("INSERT INTO foo VALUES (:url)")
+                .bind("url", ((URL) null))
```
qualidafial: The null case must be tested with `bindByType`. This case won't actually cause the URL argument to be used.

qualidafial:
Thanks @edwinnyawoli for the PR!

PR #1001, opened 2018-01-11
Add config option to disable transaction termination check.

qualidafial:


qualidafial:
Alternatively, we could move the config class to an inner class, similar to `SerializableTransactionRunner.Configuration`

qualidafial:
That's the idea. We want this enabled by default for folks who aren't doing CMT, but this check is a poor fit for those who are--so cut them some slack.

PR #1006, opened 2018-01-18
Add row reducers

### LINKED ISSUE #996: Support one-to-many mappings in sqlobject
As far as I can tell, one-to-many relations can be mapped using either

- http://jdbi.org/#_resultbearing_reducerows (can map to any type; not supported by sqlobject)
- http://jdbi.org/#_resultbearing_reduceresultset (can map to any type; not supported by sqlobject)
- A JoinRowMapper (can only map to JoinRow; supported by sqlobject)

I would like to be able to return custom beans, containing one-to-many relations, from sqlobject.

Using the example at http://jdbi.org/#_joins i would like to be able to do something similar to the following:

```
public interface ContactDao {

    @SqlQuery("select contacts.id c_id, name c_name, "
        + "phones.id p_id, type p_type, phones.phone p_phone "
        + "from contacts left join phones on contacts.id = phones.contact_id "
        + "order by c_name, p_type ")
    List<Contact> getAll();

    @SqlQuery("select contacts.id c_id, name c_name, "
        + "phones.id p_id, type p_type, phones.phone p_phone "
        + "from contacts left join phones on contacts.id = phones.contact_id "
        + "order by c_name, p_type where contacts.id = :id")
    Optional<Contact> getOne(Long id);

    @SqlUpdate("insert into contacts (id, name) values (:id, :name)")
    void addContact(Long id, String name);
}

//More methds here.
```

I can see 2 ways this might be implemented:

* Add a ResultSetMapper type, that can manually accumulate the results by mapping the entire ResultSet
* Add a RowReducer type (analogous to the BiFunction used by reduceRows), that can be registered in sqlobject.

If someone can help me get started, I would be happy to contribute the code myself, but I would need some guidance as to which implementation is preferred.

qualidafial:
Fixes #996.

New API:

* `RowReducer<C, R>` interface.
* `LinkedHashMapRowReducer` abstract implementation to make the most common reducer use case simple.
* `ResultBearing.reduceRows(RowReducer<C,R>)`
* `@UseRowReducer` SQL Object annotation, supported on `@SqlQuery` methods.
* `ResultBearing.collectRows(Collector<RowView,A,R>)` for good measure.

I've also added / refactored tests to demonstrate usage.

TODO:

* [x] Still not sure about `RowReducer` method names--`createAccumulator()` in particular--blah. Edit: Settled on `container(): C`, `accumulate(C, RowView)`, and `stream(C): Stream<R>`. Renamed `A` generic parameter to `C` for "container."
* [x] Decide whether to support `@UseRowReducer` with `@SqlBatch`, `@SqlUpdate`. Edit: decided not to support until we have a concrete use case.
  * If yes, implement it. I have a plausible implementation for both but haven't committed it.
  * If no, throw an exception (the current behavior). Currently leaning toward no since I can't fathom a legitimate use case.
* [x] Document reducers in developer guide.

CC @rbkrabbe

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/result/LinkedHashMapRowReducer.java
+++ b/core/src/main/java/org/jdbi/v3/core/result/LinkedHashMapRowReducer.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.result;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.stream.Stream;
+
+/**
+ * A row reducer that uses {@link LinkedHashMap} (which preserves insertion order) as an
+ * accumulator, and returns {@code map.values().stream()} as the final result.
+ *
+ * <p>Implementors need only implement the {@link #accumulate(Object, RowView)} method.
+ *
+ * @param <K> the map key type--often the primary key type of {@code <V>}.
+ * @param <V> the map value type, and the result element type--often the "master" object in a
+ *     master/detail relation.
+ */
+public abstract class LinkedHashMapRowReducer<K, V> implements RowReducer<Map<K, V>, V> {
+  @Override
+  public Map<K, V> createAccumulator() {
+    return new LinkedHashMap<>();
+  }
+
+  @Override
+  public Stream<V> stream(Map<K, V> accumulator) {
+    return accumulator.values().stream();
+  }
+
+  public static <K, V> RowReducer<Map<K, V>, V> of(BiConsumer<Map<K, V>, RowView> acc) {
+    return new LinkedHashMapRowReducer<K, V>() {
+      @Override
+      public void accumulate(Map<K, V> accumulator, RowView rowView) {
+        acc.accept(accumulator, rowView);
+      }
+    };
+  }
+}
--- a/core/src/main/java/org/jdbi/v3/core/result/LinkedHashMapRowReducer.java
+++ b/core/src/main/java/org/jdbi/v3/core/result/LinkedHashMapRowReducer.java
@@ -15,7 +15,6 @@ package org.jdbi.v3.core.result;
 
 import java.util.LinkedHashMap;
 import java.util.Map;
-import java.util.function.BiConsumer;
 import java.util.stream.Stream;
 
 /**
@@ -26,25 +25,17 @@ import java.util.stream.Stream;
  *
  * @param <K> the map key type--often the primary key type of {@code <V>}.
  * @param <V> the map value type, and the result element type--often the "master" object in a
- *     master/detail relation.
+ *            master/detail relation.
  */
-public abstract class LinkedHashMapRowReducer<K, V> implements RowReducer<Map<K, V>, V> {
-  @Override
-  public Map<K, V> container() {
-    return new LinkedHashMap<>();
-  }
+@FunctionalInterface
+public interface LinkedHashMapRowReducer<K, V> extends RowReducer<Map<K, V>, V> {
+    @Override
+    default Map<K, V> container() {
+        return new LinkedHashMap<>();
+    }
 
-  @Override
-  public Stream<V> stream(Map<K, V> container) {
-    return container.values().stream();
-  }
-
-  public static <K, V> RowReducer<?, V> of(BiConsumer<Map<K, V>, RowView> accumulator) {
-    return new LinkedHashMapRowReducer<K, V>() {
-      @Override
-      public void accumulate(Map<K, V> map, RowView rowView) {
-        accumulator.accept(map, rowView);
-      }
-    };
-  }
+    @Override
+    default Stream<V> stream(Map<K, V> container) {
+        return container.values().stream();
+    }
 }
--- a/core/src/main/java/org/jdbi/v3/core/result/LinkedHashMapRowReducer.java
+++ b/core/src/main/java/org/jdbi/v3/core/result/LinkedHashMapRowReducer.java
@@ -19,8 +19,8 @@ import java.util.function.BiConsumer;
 import java.util.stream.Stream;
 
 /**
- * A row reducer that uses {@link LinkedHashMap} (which preserves insertion order) as an
- * accumulator, and returns {@code map.values().stream()} as the final result.
+ * A row reducer that uses {@link LinkedHashMap} (which preserves insertion order) as a
+ * result container, and returns {@code map.values().stream()} as the final result.
  *
  * <p>Implementors need only implement the {@link #accumulate(Object, RowView)} method.
  *
@@ -30,20 +30,20 @@ import java.util.stream.Stream;
  */
 public abstract class LinkedHashMapRowReducer<K, V> implements RowReducer<Map<K, V>, V> {
   @Override
-  public Map<K, V> createAccumulator() {
+  public Map<K, V> container() {
     return new LinkedHashMap<>();
   }
 
   @Override
-  public Stream<V> stream(Map<K, V> accumulator) {
-    return accumulator.values().stream();
+  public Stream<V> stream(Map<K, V> container) {
+    return container.values().stream();
   }
 
-  public static <K, V> RowReducer<Map<K, V>, V> of(BiConsumer<Map<K, V>, RowView> acc) {
+  public static <K, V> RowReducer<?, V> of(BiConsumer<Map<K, V>, RowView> accumulator) {
     return new LinkedHashMapRowReducer<K, V>() {
       @Override
-      public void accumulate(Map<K, V> accumulator, RowView rowView) {
-        acc.accept(accumulator, rowView);
+      public void accumulate(Map<K, V> map, RowView rowView) {
+        accumulator.accept(map, rowView);
       }
     };
   }
--- a/core/src/main/java/org/jdbi/v3/core/result/RowReducer.java
+++ b/core/src/main/java/org/jdbi/v3/core/result/RowReducer.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.result;
+
+import java.util.stream.Stream;
+
+/**
+ * Reduces row data from a {@link java.sql.ResultSet} into a stream of result
+ * elements. Useful for collapsing one-to-many joins.
+ *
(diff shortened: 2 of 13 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
```diff
@@ -67,17 +68,20 @@ static ResultReturner forMethod(Class<?> extensionType, Method method)
...
-        else if (Stream.class.isAssignableFrom(returnClass)) {
+        else if (Stream.class.equals(returnClass)) {
             return new StreamReturner(returnType);
         }
-        else if (Iterator.class.isAssignableFrom(returnClass)) {
-            return new IteratorResultReturner(returnType);
+        else if (ResultIterator.class.equals(returnClass)) {
+            return new ResultIteratorReturner(returnType);
+        }
+        else if (Iterator.class.equals(returnClass)) {
```
qualidafial: I changed these from `isAssignableFrom()` to `equals()`, since the former would match any subtype of e.g. `Iterator` as the return type--which would cause a class cast exception.

qualidafial:
@jdbi/contributors This is ready for code review.

qualidafial:
@stevenschlansker or @arteam, would you mind reviewing?

stevenschlansker:
Taking a look now.

core/src/main/java/org/jdbi/v3/core/result/LinkedHashMapRowReducer.java
```diff
@@ -0,0 +1,50 @@
...
+ * A row reducer that uses {@link LinkedHashMap} (which preserves insertion order) as a
+ * result container, and returns {@code map.values().stream()} as the final result.
+ *
+ * <p>Implementors need only implement the {@link #accumulate(Object, RowView)} method.
+ *
+ * @param <K> the map key type--often the primary key type of {@code <V>}.
+ * @param <V> the map value type, and the result element type--often the "master" object in a
+ *     master/detail relation.
+ */
+public abstract class LinkedHashMapRowReducer<K, V> implements RowReducer<Map<K, V>, V> {
```
stevenschlansker: could this be an `@FunctionalInterface` with default methods provided for `container` and `stream`?  is that nicer?

core/src/main/java/org/jdbi/v3/core/result/LinkedHashMapRowReducer.java
```diff
@@ -0,0 +1,50 @@
...
+    return new LinkedHashMap<>();
+  }
+
+  @Override
+  public Stream<V> stream(Map<K, V> container) {
+    return container.values().stream();
+  }
+
+  public static <K, V> RowReducer<?, V> of(BiConsumer<Map<K, V>, RowView> accumulator) {
+    return new LinkedHashMapRowReducer<K, V>() {
```
stevenschlansker: it might get rid of this thunk

core/src/main/java/org/jdbi/v3/core/result/RowReducer.java
```diff
@@ -0,0 +1,54 @@
...
+   */
+  void accumulate(C container, RowView rowView);
+
+  /**
+   * Returns a stream of result elements from the result container.
+   *
+   * @param container the result container
+   * @return stream of result elements.
+   */
+  Stream<R> stream(C container);
```
stevenschlansker: Is returning a Stream a good API here, rather than just using the container?  (which the caller may then stream, or just use as-is)

There is no possibility to actually stream the results here, due to the context closing before the stream is presented to the caller.

core/src/test/java/org/jdbi/v3/core/result/TestReducing.java
```diff
@@ -64,8 +67,44 @@ public void testLeftJoinRowView() throws Exception
...
+            .createQuery("SELECT something.id, name, location FROM something NATURAL JOIN something_location")
+            .collectRows(Collector.<RowView, Map<Integer, SomethingWithLocations>, Iterable<SomethingWithLocations>>of(
+                    LinkedHashMap::new,
+                    (Map<Integer, SomethingWithLocations> map, RowView rv) ->
+                        map.computeIfAbsent(rv.getColumn("id", Integer.class),
+                                            id -> new SomethingWithLocations(rv.getRow(Something.class)))
+                           .locations
+                           .add(rv.getColumn("location", String.class)),
+                    (a, b) -> {
+                        throw new UnsupportedOperationException("shouldn't use combiner");
```
stevenschlansker: we could make this a public constant somewhere

core/src/test/java/org/jdbi/v3/core/result/TestReducing.java
```diff
@@ -64,8 +67,44 @@ public void testLeftJoinRowView() throws Exception
...
+    @Test
+    public void testReduceRows() {
+        List<SomethingWithLocations> result = dbRule.getSharedHandle()
+            .createQuery("SELECT something.id, name, location FROM something NATURAL JOIN something_location")
+            .reduceRows(LinkedHashMapRowReducer.<Integer, SomethingWithLocations>of((map, rv) ->
+                map.computeIfAbsent(rv.getColumn("id", Integer.class),
+                                    id -> new SomethingWithLocations(rv.getRow(Something.class)))
+                   .locations
+                   .add(rv.getColumn("location", String.class))))
+            .collect(toList());
```
stevenschlansker: as an example of the point above about streams, this natural invocation does a possibly unnecessary copy (especially if you use `toSet`)

arteam:
Looks very nice! One gap between the SqlObject and Fluent APIs down.

core/src/main/java/org/jdbi/v3/core/result/RowReducer.java
qualidafial: I chose stream here specifically so we could take advantage of the existing `JdbiCollectors` registry to collect the reduced elements into any supported container.

If the reducer returns the final container (e.g. `List<T>`, `Optional<T>`) instead of an intermediate stream of elements, then you have to write a reduce for every container type.

This way you can use the same reducer for all container types, provided the container is supported / registered into `JdbiCollectors`.

core/src/main/java/org/jdbi/v3/core/result/LinkedHashMapRowReducer.java
qualidafial: My instinct is that it wouldn't buy us anything, but I'll try it out and check.

The issue is that `ResultBearing.reduceRows()` accepts `RowReducer<C, R>`, not this subtype. So the compiler probably wouldn't know what to do with that lambda without casting to `LinkedHashMapRowReducer`.

Alternatively, we could add a `reduceRows(BiConsumer<Map<K, V>, RowView)` overload that used that reducer under the hood.

qualidafial: Probably wouldn't hurt anything though.

core/src/main/java/org/jdbi/v3/core/result/RowReducer.java
qualidafial: I considered pushing this down into sqlobject, but couldn't make the generics work in the annotation:

```java
interface RowReducer<C, R> {
  C container();
  void accumulate(C, RowView);
  R result(C container);
}

@interface UseRowReducer {
  Class<? extends RowReducer<?, Stream<?>>> value();
}

class MyRowReducer implements RowReducer<Anything, Stream<Result>> { ... }

interface MyDao {
  @SqlQuery("...")
  @UseRowReducer(MyRowReducer.class)
  void doTheThing();
}
```

This caused a compiler error on `@UseRowReducer`, because `RowReducer<Anything,Stream<Result>` does not extend `RowReducer<?, Stream<?>>`. Generic wildcards just break down when you try to use nested parameters.

stevenschlansker: You might need to write `Class<? extends RowReducer<?, ? extends Stream<?>>>` or something like that.

qualidafial: > If the reducer returns the final container (e.g. List<T>, Optional<T>) instead of an intermediate stream of elements, then you have to write a reducer for every container type.

This is the primary reason I prefer the reducer returning a stream. The generics issues in SQL Object are secondary.

core/src/main/java/org/jdbi/v3/core/result/LinkedHashMapRowReducer.java
qualidafial: Usage after making it a functional interface, and adding another convenience method in `ResultBearing`:

```java
List<SomethingWithLocations> result = dbRule.getSharedHandle()
    .createQuery("SELECT something.id, name, location FROM something NATURAL JOIN something_location")
    .<Integer, SomethingWithLocations>reduceRows((map, rv) ->
        map.computeIfAbsent(rv.getColumn("id", Integer.class),
                            id -> new SomethingWithLocations(rv.getRow(Something.class)))
           .locations
           .add(rv.getColumn("location", String.class)))
    .collect(toList());
```

core/src/test/java/org/jdbi/v3/core/result/TestReducing.java
qualidafial: Eh.. feels like overkill for something out of a test.

core/src/main/java/org/jdbi/v3/core/result/RowReducer.java
qualidafial: So I've confirmed that the annotation would work as you suggested using `? extends Stream<?>`, however I still think we're better off with `RowReducer` producing a stream.

In my first prototype, the reducer returned the full container object instead of the stream:

```java
interface RowReducer<C,R> {
  C container();
  void accumulate(C container, RowView rowView);
  R result(C container);
}
```

Unfortunately, this meant that if I wanted to reduce into different containers (e.g. `List` vs `Optional`), I had to implement essentially the same reducer repeatedly, once for each container type:

```java
class ListOfContactPhoneReducer
    implements RowReducer<Map<Long, Contact>, List<Contact>> {
  Map<Long, Contact> container() {
    return new LinkedHashMap<>();
  }

  void accumulate(Map<Long, Contact> container, RowView rowView) {
    ...
  }

  List<Contact> result(Map<Long, Contact> container) {
    return new ArrayList<>(container.values());
  }
}
```

```java
class OptionalContactPhoneReducer
    implements RowReducer<Map<Long, Contact>, Optional<Contact>> {
  Map<Long, Contact> container() {
    return new LinkedHashMap<>();
  }

  void accumulate(Map<Long, Contact> container, RowView rowView) {
    ...
  }

  Optional<Contact> result(Map<Long, Contact> container) {
    return container.values().stream().findFirst();
  }
}
```

This felt wasteful to me, especially considering we already have `JdbiCollectors` in place to facilitate collecting elements into arbitrary container types.

I considered putting the `Stream<T>` restriction only in SQL Object, but that would have left us in a situation where you have to implement a `RowReducer` differently in core than in sqlobject.

Ultimately I concluded that it was better to have it return a stream, for consistency and possibility for reuse between core and sqlobject.

This is why I added `ResultBearing.reduceRows(Collector<RowView, ?, R>)`, as a compromise so users could have it either way.

@stevenschlansker Does this sound reasonable?

stevenschlansker: Yeah, that's a real bummer.  I agree with your reasoning.  I'm still sad that the example already shows an example of the problem though -- this pattern essentially guarantees unnecessary copies.

core/src/test/java/org/jdbi/v3/core/result/TestReducing.java
stevenschlansker: Well, it's not just tests -- most of our collectors have similar mergers.

qualidafial: Actually I looked at most of the collectors, and all of them (except this one) do a best-effort merge now.

I think the reasoning was we're providing these collector factory methods for convenience, but there's nothing to stop folks from using them outside of Jdbi. So try to combine containers and play nice.

qualidafial:
Awesome, thanks for the reviews guys. Merging!

arg20:
@qualidafial is a RowReducer created per mapping job? (Can I store state in it and trust that each reducing job gets its own instance of the reducer?)

qualidafial:
The row reducer should put all state in the `container()` object. Each reduced query uses a distinct container instance, which can be any type you want.

Issue #1010, opened 2018-02-01
BindList can't find my named parameter in query and throws exception


johanneszink:
I got a SQL Query with a named parameter `:featureIds`

`const val selectGeoCoordinates = "SELECT gg.feature_id, gg.latitude, gg.longitude, gg.order, gg.group FROM geo.geo_coordinates as gg WHERE gg.feature_id IN (:featureIds);"`

When I try to bind a list of Ints to the named parameter 

```
val values: List<Int> = listOf(4138, 1752, 1161, 371, 1860, 5498)
val geoCoordinatesList = geo.withHandle<List<GeoCoordinates>, Exception> { handle ->
                handle.createQuery(selectGeoCoordinates).bindList("featureIds", values)
                        .mapTo<GeoCoordinates>().list()
            }
```

I get the following exception on execution:

```
    Caused by: org.jdbi.v3.core.statement.UnableToExecuteStatementException: Unable to execute, no named parameter matches 'featureIds'. [statement:"SELECT gg.feature_id, gg.latitude, gg.longitude, gg.order, gg.group FROM geo.geo_coordinates as gg WHERE gg.feature_id IN :featureIds ;", rewritten:"SELECT gg.feature_id, gg.latitude, gg.longitude, gg.order, gg.group FROM geo.geo_coordinates as gg WHERE gg.feature_id IN :featureIds ;", parsed:"ParsedSql{sql='SELECT gg.feature_id, gg.latitude, gg.longitude, gg.order, gg.group FROM geo.geo_coordinates as gg WHERE gg.feature_id IN ? ;', parameters=ParsedParameters{positional=false, parameterNames=[featureIds]}}", arguments:{ positional:{}, named:{__featureIds_0:4138,__featureIds_4:1860,__featureIds_3:371,__featureIds_2:1161,__featureIds_1:1752,__featureIds_5:5498}, finder:[]}]
	at org.jdbi.v3.core.statement.ArgumentBinder.lambda$bindNamed$0(ArgumentBinder.java:58)
	at java.util.Optional.orElseThrow(Optional.java:290)
	at org.jdbi.v3.core.statement.ArgumentBinder.bindNamed(ArgumentBinder.java:57)
	at org.jdbi.v3.core.statement.ArgumentBinder.bind(ArgumentBinder.java:27)
	at org.jdbi.v3.core.statement.SqlStatement.internalExecute(SqlStatement.java:1443)
[...]

PR #1013, opened 2018-02-05
Add usage documentation for @BindList

pikzen:
Explain how @BindList is used, as the need to use `<binding>` might be slightly confusing at first.

Edit: Related to https://github.com/jdbi/jdbi/issues/1010

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -336,6 +336,26 @@ handle.createUpdate("insert into contacts (id, name) values (:id, :name)")
       .execute();
 ----
 
+You can bind multiple values at once, from either a `List<T>` or a list of arguments. Note that this requires your binding to be written using the `<boundValue>` form (as opposed to ``:boundValue` for most of the other bindings.)
+
+[source,java]
+----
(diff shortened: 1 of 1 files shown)


alwins0n:
I think the documentation is missing for the fluent API as well, iirc.

pikzen:
I have never had to use BindList with the fluent API :D. I'll be testing things out later in the day and completing the docs to make sure I'm not making any errors in the docs, although I would assume the usage is the same.

alwins0n:
okay, i was just pointing out that it should be mentioned in the "Core API" section (3.), since SQLObject is just an extension of that. (but could be mentioned there as well, like you did)

pikzen:
Alright, included it into the Core API section, and rebased it onto a single commit.

alwins0n:
you should check the file again. seems like you have some git change artifacts in there

pikzen:
Indeed, I seem to have failed at git for a second here :D

All fixed, thanks for the warning.

pikzen:
Done!

qualidafial:
Thanks for the PR!

PR #1014, opened 2018-02-05
Upgrade Guava to 24.0-jre

arteam:
Upgrade the dependency to the latest version of Guava, so we can make sure that the integration works with it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -255,7 +255,7 @@
             <dependency>
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
-                <version>21.0</version>
+                <version>24.0-jre</version>
             </dependency>
             <dependency>
                 <groupId>io.vavr</groupId>


stevenschlansker:
I worry that this will force end users forward for no reason -- if we declare a 24.0 here, and a consumer uses 22.0, the maven version verifier may note that 22.0 < 24.0 and break their build.

Is there a particular reason to upgrade here?  Unless we use a new feature, I think it's preferable to not upgrade Guava in an intermediate library.

arteam:
For instance, Guava 22 supports immutable primitive arrays: 
* https://github.com/google/guava/blob/master/guava/src/com/google/common/primitives/ImmutableIntArray.java
* https://github.com/google/guava/blob/master/guava/src/com/google/common/primitives/ImmutableLongArray.java

We could write mappers and argument factories for them. Plus, Guava 24 is friendly for Java 9 - modularized applications, since Guava has got rid off JSR-305 `@Nullable` annotations in favour of the Checker Framework, so users don't need to do arcane module overrides.

stevenschlansker:
Java 9 is compelling, as are those primitive types -- but maybe we should bundle the version update with PRs that either introduce the new types, or provide Java 9 support?  I won't block this, just observe that it brings in compatibility concerns without actually achieving either of the goals you outlined above.

qualidafial:
I agree with Steven on this one--let's upgrade only when we're adding support for new features. Or possibly security updates.

stevenschlansker:
I'm going to close this one, please feel free to reopen if there's further discussion to be had.

PR #1024, opened 2018-02-21
Support named parameters with OracleReturning

### LINKED ISSUE #1022: Support named parameters for OracleReturning
One of the main reasons I use jdbi over plain jdbc is because it supports named paramters which makes the code cleaner and more stable.  However, the OracleReturning construct only supports numbered parameters.  Would it be possible to add named parameter support to OracleReturning consistently with the rest of jdbi?

qualidafial:
Fixes #1022 

* Added method `OracleReturning.ReturnParameters.register(String name, int oracleType)`
* Made `Binding` methods `addPositional`, `addNamed`, and `addNamedArgumentFinder` public. `OracleReturning` does not have access to the `SqlStatement` to bind in the usual way, however we are able to add the binding via `StatementContext.getBinding().addXyz()`.
* Added new lifecycle hook `StatementCustomizer.beforeBinding()` to allow customizers to bind extra parameters before they are applied to the statement.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/BaseStatement.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/BaseStatement.java
@@ -98,6 +98,17 @@ abstract class BaseStatement<This> implements Closeable, Configurable<This>
         return typedThis;
     }
 
+    final void beforeBinding(final PreparedStatement stmt) {
+        for (StatementCustomizer customizer : customizers) {
+            try {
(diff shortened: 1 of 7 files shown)


qualidafial:
@arteam The existing argument infrastructure is already converting the zero-based index into a one-based position when it calls into the argument object.

qualidafial:
@jdbi/contributors Is everybody on board with making those `Binding` methods public?

arteam:
I don't see any issues with that.

PR #1026, opened 2018-02-22
Fix regression in typed enum support

stevenschlansker:
tag @pennello

### DIFF
--- a/postgres/src/main/java/org/jdbi/v3/postgres/TypedEnumArgumentFactory.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/TypedEnumArgumentFactory.java
@@ -13,14 +13,11 @@
  */
 package org.jdbi.v3.postgres;
 
-import java.lang.reflect.Type;
 import java.sql.Types;
-import java.util.Optional;
 
+import org.jdbi.v3.core.argument.AbstractArgumentFactory;
 import org.jdbi.v3.core.argument.Argument;
-import org.jdbi.v3.core.argument.ArgumentFactory;
 import org.jdbi.v3.core.config.ConfigRegistry;
-import org.jdbi.v3.core.generic.GenericTypes;
 
 /**
  * Default {@code jdbi} behavior is to bind {@code Enum} subclasses as
@@ -28,12 +25,13 @@ import org.jdbi.v3.core.generic.GenericTypes;
  * If instead you bind it as {@code java.sql.Types.OTHER}, Postgres will
  * autodetect the enum correctly.
  */
-public class TypedEnumArgumentFactory implements ArgumentFactory {
+public class TypedEnumArgumentFactory extends AbstractArgumentFactory<Enum> {
+    public TypedEnumArgumentFactory() {
+        super(Types.OTHER);
+    }
+
     @Override
-    public Optional<Argument> build(Type type, Object value, ConfigRegistry config) {
-        if (GenericTypes.getErasedType(type).isEnum()) {
-            return Optional.of((p, s, c) -> s.setObject(p, value, Types.OTHER));
-        }
-        return Optional.empty();
+    protected Argument build(Enum value, ConfigRegistry config) {
+        return (p, s, c) -> s.setObject(p, value, Types.OTHER);
     }
 }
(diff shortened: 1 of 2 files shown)


postgres/src/main/java/org/jdbi/v3/postgres/TypedEnumArgumentFactory.java
```diff
@@ -13,25 +13,27 @@
...
-    public TypedEnumArgumentFactory() {
-        super(Types.OTHER);
-    }
-
+public class TypedEnumArgumentFactory implements ArgumentFactory {
     @Override
-    protected Argument build(Enum<?> value, ConfigRegistry config) {
-        return (p, s, c) -> s.setObject(p, value, Types.OTHER);
+    public Optional<Argument> build(Type type, Object value, ConfigRegistry config) {
+        if (GenericTypes.getErasedType(type).isEnum()) {
```
stevenschlansker: This check was actually important, `AbstractArgumentFactory` bases its decisions on the static type on AAF, but we want this runtime check.

stevenschlansker:
Build failure due to a leaked out `spring-boot` dependency through parent.  Ugh.

postgres/src/main/java/org/jdbi/v3/postgres/TypedEnumArgumentFactory.java
```diff
@@ -13,25 +13,27 @@
...
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericTypes;
 
 /**
  * Default {@code jdbi} behavior is to bind {@code Enum} subclasses as
  * a string, which Postgres won't implicitly convert to an enum type.
  * If instead you bind it as {@code java.sql.Types.OTHER}, Postgres will
  * autodetect the enum correctly.
  */
-public class TypedEnumArgumentFactory extends AbstractArgumentFactory<Enum<?>> {
```
arteam: Wouldn't it better to fix the issue just by switching `Enum<?>` to `Enum`? This way we won't compare `ParametrizedType` instances, but will check that an enum can be assigned to the `Enum` class.

arteam: Wouldn't it better to fix the issue just by switching `Enum<?>` to `Enum`? This way we won't compare `ParametrizedType` instances, but will check that an enum can be assigned to the `Enum` class.

stevenschlansker: Oh, I hadn't considered that, I hate raw types :)
Either way this AAF behavior is very strange -- it checks type compatibility if it is not parameterized, and type equality if it is.  I think we should pick one or the other (always exact, or reasonable generic type assignability test, but not the current mixed situation)

qualidafial: The current implementation was a best effort attempt at allowing `AbstractArgumentFactory` to handle subtypes of the declared argument type. We punted on the generics question by only allowing exact type. Artem is right that, under the current implementation, this parameter should have been raw `Enum` instead of `Enum<?>`.

We have a couple options to improve this situation with `AbstractArgumentFactory`:

* Allow the raw type to co-vary, but require all the nested parameters to match exactly, e.g. treat `ArrayList<String>` as a subtype of `List<String>`. This is roughly consistent with the existing `Class.isAssignableFrom()` check for raw classes.
* Allow the raw type _and_ the generic parameters to co-vary, e.g. treat `ArrayList<String>`as a subtype of `List<Object>`. I'm really, _really_ hesitant to go that far. Once you screw up generics, you end up supporting that mistake for years. Just look at Hamcrest--they got the generics wrong on a couple critical matchers, and those matchers are unusable as a result.

I'd be in favor of letting the raw type co-vary, but I'd have to hear a damn good argument to support covariance of the nested parameters.

qualidafial: Submitted issue #1029 for making `AbstractArgumentFactory` more lenient with parameterized types.

qualidafial:
Gah, TravisCI is wedged and I can't merge this :angry:

qualidafial:
All hail whitespace commits, the WD-40 of github integrations

PR #1031, opened 2018-03-01
V3.1.0 release

qualidafial:


stevenschlansker:
:+1: fyi I'm out of town through the 15th, so I won't be responding much.  Thanks for running the release!

Issue #1035, opened 2018-03-14
Extending SqlObject interface does not work


christopheblin:
Extending SqlObject to provide a custom implementation like described in http://jdbi.org/#_default_methods does not work with jdbi3-kotlin-sqlobject

    interface ListUsersDao : SqlObject {
        fun listUsers(): List<User> {
            return handle.createQuery("SELECT * FROM users")
                .mapToBean(User::class.java)
                .list()
        }
    }

    val ds = MysqlDataSource()
    ds.serverName = System.getenv("DB_HOST")
    ds.user = System.getenv("DB_USERNAME")
    ds.setPassword(System.getenv("DB_PASSWORD"))
    ds.databaseName = System.getenv("DB_NAME")

    val jdbi = Jdbi.create(ds)
    jdbi.installPlugin(KotlinSqlObjectPlugin())
    val handle = jdbi.open()
    val dao = handle.attach(ListUsersDao::class.java)
    dao.listUsers()

    java.lang.IllegalStateException: Method ListUsersDao.listUsers must be default or be annotated with a SQL method annotation.

	at org.jdbi.v3.sqlobject.SqlObjectFactory.lambda$buildMethodHandler$11(SqlObjectFactory.java:133)
	at java.util.Optional.orElseThrow(Optional.java:290)
	at org.jdbi.v3.sqlobject.SqlObjectFactory.buildMethodHandler(SqlObjectFactory.java:133)
	at org.jdbi.v3.sqlobject.SqlObjectFactory.lambda$null$9(SqlObjectFactory.java:124)
	at java.util.HashMap.computeIfAbsent(HashMap.java:1127)
	at org.jdbi.v3.sqlobject.SqlObjectFactory.lambda$methodHandlersFor$10(SqlObjectFactory.java:124)
	at java.util.Map.computeIfAbsent(Map.java:957)
	at java.util.Collections$SynchronizedMap.computeIfAbsent(Collections.java:2672)
	at org.jdbi.v3.sqlobject.SqlObjectFactory.methodHandlersFor(SqlObjectFactory.java:110)
	at org.jdbi.v3.sqlobject.SqlObjectFactory.attach(SqlObjectFactory.java:87)
	at org.jdbi.v3.core.extension.Extensions.lambda$findFor$0(Extensions.java:71)
	at java.util.Optional.map(Optional.java:215)
	at org.jdbi.v3.core.extension.Extensions.findFor(Extensions.java:71)
[...]

PR #1036, opened 2018-03-14
Add test for Kotlin SQL object `getHandle`

FredDeschenes:
See #1035 (this doesn't fix anything, but adds a missing test case).


qualidafial:
Thanks @FredDeschenes!

PR #1049, opened 2018-03-25
Vavr plugin update

alwins0n:
PriorityQueue support.
Multimap support.
Updated docs

### DIFF
--- a/vavr/src/main/java/org/jdbi/v3/vavr/VavrCollectorFactory.java
+++ b/vavr/src/main/java/org/jdbi/v3/vavr/VavrCollectorFactory.java
@@ -24,6 +24,7 @@ import io.vavr.collection.LinkedHashMap;
 import io.vavr.collection.LinkedHashSet;
 import io.vavr.collection.List;
 import io.vavr.collection.Map;
+import io.vavr.collection.PriorityQueue;
 import io.vavr.collection.Queue;
 import io.vavr.collection.Seq;
 import io.vavr.collection.Set;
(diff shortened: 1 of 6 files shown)


vavr/src/main/java/org/jdbi/v3/vavr/VavrCollectorFactory.java
```diff
@@ -70,14 +78,19 @@
...
                 Tuple.of(HashSet.class, HashSet.collector()),
                 Tuple.of(LinkedHashSet.class, LinkedHashSet.collector()),
                 Tuple.of(TreeSet.class, TreeSet.collector()),
                 // Maps
                 Tuple.of(HashMap.class, HashMap.collector()),
                 Tuple.of(LinkedHashMap.class, LinkedHashMap.collector()),
-                Tuple.of(TreeMap.class, TreeMap.collector())
+                Tuple.of(TreeMap.class, TreeMap.collector()),
+                // Multimaps
+                Tuple.of(HashMultimap.class, HashMultimap.withSeq().collector()),
```
alwins0n: just wanted to mention: the multimap backed by seq seems like a sensible default. but somehow it might be nice to have that configuraable. currently `CollectorFactory` has no injection of configuration.

I think this is fine, since you can explicitly request collection with Multimap backed by Set or whatever (via `collect` method) if needed just wanted to add that thought here

stevenschlansker: We can start adding configuration if necessary, although usually my preference would simply be that the application registers their own mappers after all the defaults get pulled in - last registration wins

PR #1053, opened 2018-03-27
DefaultMethodHandler: terrible no good very bad hack hack now also matches `public` on lookup

### LINKED ISSUE #1015: JDBI3 interface default method doesn't work with IBM javasdk 8
Hello,
We're using IBM sdk Java. And we can't upgrade to JDBI 3.
We can't invoke method after newInstance(declaringClass, MethodHandles.Lookup.PRIVATE), it's like it is considered private
java.lang.RuntimeException: java.lang.IllegalAccessException: $Dao' no access to: $Dao.method:(Dao,Levels,List,List,List,String,Instant)void/invokeSpecial'

It should be MethodHandles.Lookup.PRIVATE | MethodHandles.Lookup.PUBLIC since interface default method is public.
constructor.newInstance(type, MethodHandles.Lookup.PRIVATE | MethodHandles.Lookup.PUBLIC);

IBM Jvm allows us to call private construcor,
but lookup (inside org.jdbi.v3.sqlobject.DefaultMethodHandler)

MethodHandles.Lookup lookup = lookupFor(declaringClass)
has access mode PRIVATE
and it not enough for checkSpecialAccess in this JVM

it works if we pass MethodHandles.Lookup.PRIVATE | MethodHandles.Lookup.PUBLIC to the private constructor
return constructor.newInstance(type, MethodHandles.Lookup.PRIVATE | MethodHandles.Lookup.PUBLIC );
or method.setAccessible(true); in DefaultMethodHandler constructor to workaround checkSpecialAccess internal method

https://developer.ibm.com/javasdk/downloads/sdk8/
it requires registration to upload unfortunately

stevenschlansker:
Fixes #1015 

Hopefully OP can test this, as I don't really care to register and download IBM's JDK.  But it looks harmless to include (it only widens the lookup) and all the tests pass...

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
+++ b/core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
@@ -13,8 +13,8 @@
  */
 package org.jdbi.v3.core;
 
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.InvocationHandler;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 
@@ -71,11 +71,7 @@ class OnDemandExtensions {
             if (Proxy.isProxyClass(target.getClass())) {
                 return Proxy.getInvocationHandler(target).invoke(target, method, args);
             }
-            try {
-                return method.invoke(target, args);
-            } catch (InvocationTargetException e) {
-                throw e.getTargetException();
-            }
+            return MethodHandles.lookup().unreflect(method).bindTo(target).invokeWithArguments(args);
         } catch (RuntimeException | Error e) {
             throw e;
         } catch (Throwable t) {
--- a/core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
+++ b/core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
@@ -17,6 +17,7 @@ import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+
 import org.jdbi.v3.core.internal.JdbiThreadLocals;
 
 class OnDemandExtensions {
@@ -51,22 +52,34 @@ class OnDemandExtensions {
                 return extensionType + "@" + Integer.toHexString(System.identityHashCode(proxy));
             }
 
-            try {
-                if (threadExtension.get() != null) {
-                    return method.invoke(threadExtension.get(), args);
-                }
-                return db.withExtension(extensionType, extension ->
-                        JdbiThreadLocals.invokeInContext(threadExtension, extension,
-                                () -> method.invoke(extension, args)));
+            if (threadExtension.get() != null) {
+                return invoke(method, threadExtension.get(), args);
             }
-            catch (InvocationTargetException e) {
-                throw e.getTargetException();
-            }
+            return db.withExtension(extensionType, extension ->
+                    JdbiThreadLocals.invokeInContext(threadExtension, extension,
+                            () -> invoke(method, extension, args)));
         };
 
         return extensionType.cast(
                 Proxy.newProxyInstance(
                         extensionType.getClassLoader(),
                         new Class[]{extensionType}, handler));
+    }
+
+    private static Object invoke(Method method, Object target, Object[] args) {
+        try {
+            if (Proxy.isProxyClass(target.getClass())) {
+                return Proxy.getInvocationHandler(target).invoke(target, method, args);
+            }
+            try {
+                return method.invoke(target, args);
+            } catch (InvocationTargetException e) {
+                throw e.getTargetException();
+            }
+        } catch (RuntimeException | Error e) {
+            throw e;
+        } catch (Throwable t) {
+            throw new RuntimeException(t);
+        }
     }
 }
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,3 +1,6 @@
+3.1.1
+  - Improve IBM JDK compatibility with default methods
+
 3.1.0
   - The strict transaction handling check in Handle.close() may be disabled via
     getConfig(Handles.class).setForceEndTransactions(false).
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,6 +1,6 @@
 3.1.1
   - Improve IBM JDK compatibility with default methods
-  - Allow non-public SqlObject types, unless they are used onDemand
+  - Allow non-public SqlObject types!!!
 
 3.1.0
   - The strict transaction handling check in Handle.close() may be disabled via
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,5 +1,6 @@
 3.1.1
   - Improve IBM JDK compatibility with default methods
+  - Allow non-public SqlObject types, unless they are used onDemand
 
 3.1.0
   - The strict transaction handling check in Handle.close() may be disabled via
--- a/core/src/main/java/org/jdbi/v3/core/Jdbi.java
+++ b/core/src/main/java/org/jdbi/v3/core/Jdbi.java
@@ -13,8 +13,6 @@
  */
 package org.jdbi.v3.core;
 
-
-import java.lang.reflect.Modifier;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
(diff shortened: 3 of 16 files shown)


stevenschlansker:
And after reviewing #840, I wonder what happens if we just include all of the modes.  Does this then also fix additional access problems we've had?

stevenschlansker:
Upon further testing, this fix also does the relaxation of the check noted in #840, but then you run into a further problem with on demand methods:

```
java.lang.reflect.UndeclaredThrowableException
	at org.jdbi.v3.sqlobject.$Proxy5.finalize(Unknown Source)
	at org.jdbi.v3.sqlobject.TestOnDemandObjectMethodBehavior.testFinalizeDoesntConnect(TestOnDemandObjectMethodBehavior.java:60)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:86)
	at org.eclipse.jdt.internal.junit.runner.TestExecution.run(TestExecution.java:38)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:538)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.runTests(RemoteTestRunner.java:760)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.run(RemoteTestRunner.java:460)
	at org.eclipse.jdt.internal.junit.runner.RemoteTestRunner.main(RemoteTestRunner.java:206)
Caused by: java.lang.IllegalAccessException: Class org.jdbi.v3.core.OnDemandExtensions can not access a member of class org.jdbi.v3.sqlobject.TestOnDemandObjectMethodBehavior$UselessDao with modifiers "public abstract"
	at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102)
	at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:296)
	at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:288)
	at java.lang.reflect.Method.invoke(Method.java:491)
	at org.jdbi.v3.core.OnDemandExtensions.lambda$null$0(OnDemandExtensions.java:60)
	at org.jdbi.v3.core.internal.JdbiThreadLocals.invokeInContext(JdbiThreadLocals.java:23)
	at org.jdbi.v3.core.OnDemandExtensions.lambda$null$1(OnDemandExtensions.java:59)
	at org.jdbi.v3.core.Jdbi.withExtension(Jdbi.java:463)
	at org.jdbi.v3.core.OnDemandExtensions.lambda$create$2(OnDemandExtensions.java:58)
	... 26 more
```

Since this fix also changes that, we'd end up supporting non-public interfaces unless it's used onDemand.

stevenschlansker:
That means there's some risk this might then regress on jdk9 - in which case we'll have to keep the restriction and only keep the first commit.

qualidafial:
@stevenschlansker Could we remove the restriction in on-demand DAOs by using method handles like SQL Object is doing?

stevenschlansker:
I didn't test that.  I'll try that tonight.

stevenschlansker:
The answer is, no, you can't.  However, I did some more thinking about exactly why we're running into this problem.  When your class calls into your private interface, no problem, our InvocationHandler is called just fine.  But then onDemand reflectively tries to dispatch into the regular SqlObject implementation, which is where we violate access checks.

But we are dealing with a Proxy.  And in fact, we don't need to reflectively invoke a Proxy only to call an InvocationHandler with the same call shape we just received... 😁

core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
```diff
@@ -51,22 +52,30 @@
...
-                    return method.invoke(threadExtension.get(), args);
-                }
-                return db.withExtension(extensionType, extension ->
-                        JdbiThreadLocals.invokeInContext(threadExtension, extension,
-                                () -> method.invoke(extension, args)));
-            }
-            catch (InvocationTargetException e) {
-                throw e.getTargetException();
+            if (threadExtension.get() != null) {
+                return invoke(method, threadExtension.get(), args);
```
qualidafial: Nice simplification hiding that try/catch in the invoke method.

core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
```diff
@@ -51,22 +52,30 @@
...
+                            () -> invoke(method, extension, args)));
         };
 
         return extensionType.cast(
                 Proxy.newProxyInstance(
                         extensionType.getClassLoader(),
                         new Class[]{extensionType}, handler));
     }
+
+    private static Object invoke(Method method, Object target, Object[] args) {
```
qualidafial: \<pedant\>Should this be `target, method, args` instead, to match parameter order elsewhere in the project?\<\/pedant\>

core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
```diff
@@ -51,22 +52,30 @@
...
+        try {
+            if (Proxy.isProxyClass(target.getClass())) {
+                return Proxy.getInvocationHandler(target).invoke(target, method, args);
+            }
+            return MethodHandles.lookup().unreflect(method).bindTo(target).invokeWithArguments(args);
```
qualidafial: Not sure `MethodHandles.lookup()` buys us anything now with this proxy workaround. Should we just use `method.invoke()` until a corner case presents itself?

Does `MethodHandle` throw the original exception instead of wrapping it?

core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
```diff
@@ -51,22 +52,30 @@
...
+        } catch (RuntimeException | Error e) {
+            throw e;
+        } catch (Throwable t) {
+            throw new RuntimeException(t);
+        }
```
qualidafial: `InvocationHandler.invoke()` throws `Throwable`. Seems like we could just slap a `throws Throwable` on this method and let it bubble up.

core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
stevenschlansker: Yes, the only reason I made this change is because it means the exceptions are transparent and it simplifies the catch block, and because we did it in the other place.  I can easily leave the original reflection calls in, at the expense of another nested block for exception unwrapping, if that's preferable.

core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
stevenschlansker: Nope, we use this method in a lambda above, which can't throw `Throwable`.

core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
stevenschlansker: I think I'm going to leave this for now, but I'm happy to do a followup and revert this bit if you think it's too cowboy.

pennello:
Just tested this PR on a Java 9 deployment via a local snapshot dependency.  I was previously suffering the following.

    Caused by: java.lang.IllegalAccessException: access to public member failed: com.opentable.service.feedback.api.ReadDao.read(long,Collection,FeedbackType)List/invokeSpecial, from com.opentable.service.feedback.api.ReadDao/2 (unnamed module @260f2144)

where the `ReadDao` was an interface I was getting an instance of with `onDemand`.

Worked flawlessly.  The problem is no longer present.

core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
qualidafial: checked exceptions are the worst

PR #1060, opened 2018-03-31
Implement @SqlScript

### LINKED ISSUE #131: Multiple SqlBatch annotations
Hi, it would be convenient if we could assign more than one SqlBatch annotations to a method of a DAO class.

Example:

``` java
    @SqlBatch("CREATE TABLE IF NOT EXISTS <table> (`id` int(11), `name` varchar(16))")
    @SqlBatch("ALTER TABLE <table> ADD COLUMN `id` int(11)")
    @SqlBatch("ALTER TABLE <table> ADD COLUMN `name` varchar(16)")
    void createTable(@Define("table") tableName);
```

stevenschlansker:
This allows you to execute multiple statements in a batch

Fixes #131

### DIFF
--- a/sqlobject/src/test/resources/org/jdbi/v3/sqlobject/statement/TestSqlScripts$Scriptacular/externalScript.sql
+++ b/sqlobject/src/test/resources/org/jdbi/v3/sqlobject/statement/TestSqlScripts$Scriptacular/externalScript.sql
@@ -1,0 +1,16 @@
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+-- http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


arteam:
`@SqlScript` with actual SQL doesn't seem very useful. Usually SQL scripts tend to be quite big, sometimes they're generated by a database dump. I think we should provide a way to specify a path to an SQL script from the classpath.

stevenschlansker:
I tend to agree with you, but it was an explicit feature request.  I can see some use cases where 2-3 lines are nicer inline than externalized.  But that said, it should work just as well with a sql locator... I'll add a test showing that!

stevenschlansker:
@arteam now it supports loading from classpath via `@UseClasspathSqlLocator`, is that more compelling?

sqlobject/src/test/resources/org/jdbi/v3/sqlobject/statement/TestSqlScripts$Scriptacular/externalScript.sql
```diff
@@ -0,0 +1,2 @@
+CREATE TABLE temp (pk integer primary key);
```
arteam: I guess the license checker wants too see a license boilerplate in the file header.

arteam:
Looks great now!

PR #1062, opened 2018-03-31
Add a jdbi3-testing artifact with JdbiRule

### LINKED ISSUE #789: Add a testing module with reusable test rules
Would be useful to have a module with testing packages to help those building jdbi extensions/plugins implement tests more quickly. Ex. [`H2DatabaseRule`](jdbi/core/src/test/java/org/jdbi/v3/core/rule/H2DatabaseRule.java).

stevenschlansker:
Fixes #789

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -3375,6 +3375,35 @@ public interface AccountDao extends CrudDao<Account, UUID> {
 <1> Method annotations are not inherited on override, so we must duplicate
     those we want to keep.
 
+== Testing
+
+The `jdbi-testing` artifact provides a link:{jdbidocs}/testing/JdbiRule.html[JdbiRule]
+class which provides helpers for writing JUnit tests integrated with a managed database
+instance.  This makes writing unit tests quick and easy!  You must remember to include
+the database dependency itself, for example to get a pure H2 Java database:
+
+[source,xml]
+----
+<dependency>
+    <groupId>com.h2database</groupId>
+    <artifactId>h2</artifactId>
+    <version>1.4.197</version>
+    <scope>test</scope>
+</dependency>
+----
+
+If you want to test against Postgres, you'd instead include:
+
+[source,xml]
+----
+<dependency>
+    <groupId>com.opentable.components</groupId>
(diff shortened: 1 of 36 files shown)


testing/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-core</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <optional>true</optional>
```
arteam: I don't think JUnit is optional, because all the classes in the module depend on `ExternalResource`. I would say the scope `provided` would be a better option. This would indicate that end-users must have `junit` in the classpath to use `jdbi3-testing`.

stevenschlansker: Yeah, I was imagining that we might e.g. add junit5 equivalents and mark both as optional so you can pick one or the other.  But we can do that at a later point.

docs/src/adoc/index.adoc
```diff
@@ -3375,6 +3375,35 @@ public interface AccountDao extends CrudDao<Account, UUID> {
 <1> Method annotations are not inherited on override, so we must duplicate
     those we want to keep.
 
+== Testing
+
+The `jdbi-testing` artifact provides a link:{jdbidocs}/testing/JdbiRule.html[JdbiRule]
```
qualidafial: Should be `jdbi3-testing`?

qualidafial: Actually, I'm thinking it should probably be be `jdbi3-junit4`. JUnit 5 is out and is very different.

qualidafial:
This is really nice!

docs/src/adoc/index.adoc
stevenschlansker: We could do that.  I'd imagined we just mark each of `junit4` and `junit5` and `testng` (or whatever else people contribute) as optional, and we stash it all in one module, otherwise we end up with 5 different artifacts with 1 class each.

qualidafial: I'm okay with that too.

PR #1064, opened 2018-03-31
ResultProducers: add option to suppress exception if there's no result set

### LINKED ISSUE #1063: Question: execute arbitrary sql and return resultset if one exists
I've made myself a little sql console page in a webapp, a la H2's. I use Jdbi's script parser to separate an incoming sql script into statements, handling each statement individually and returning results for each statement. There's a catch-all block for unrecognized statements: `handle.createQuery(statement).mapToMap().list()`

This works fine when `statement = "call crypt_key('aes', null)"` or anything else result-returning, but if `statement = "commit"` for example, it triggers `NoResultsException: Statement returned no results` at `ResultProducers:62`.

Is there a way I can check if the statement uberhaupt returned any results before attempting to get them and triggering an avoidable exception? It would need to happen in between mapToMap() and list() AFAIK, or I'd need to use a different part of the API altogether, but I can't really piece it together. I'd rather not fall back to just catching and ignoring the exception.

Thanks for any help

stevenschlansker:
Fixes #1063

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/result/ResultProducers.java
+++ b/core/src/main/java/org/jdbi/v3/core/result/ResultProducers.java
@@ -18,13 +18,22 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.function.Supplier;
 
+import org.jdbi.v3.core.config.JdbiConfig;
 import org.jdbi.v3.core.statement.StatementContext;
 
 /**
  * Commonly used ResultProducer implementations.
  */
-public class ResultProducers {
-    private ResultProducers() {
+public class ResultProducers implements JdbiConfig<ResultProducers> {
+
+    private boolean allowNoResults;
+
+    public ResultProducers() {
+        this(false);
+    }
+
+    private ResultProducers(boolean allowNoResults) {
+        this.allowNoResults = allowNoResults;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


TheRealMarnes:
That modifies the handle for all queries then, right?

And there's no existing way at the moment?

stevenschlansker:
It's part of a `JdbiConfig` so you can override it at the Jdbi, Handle, or individual statement level.  The `throw` is currently unconditional, yes.

TheRealMarnes:
It's not bad I guess but it's a little unintuitive. My query is `query.mapToMap().list()`. I'd expect to just get empty collections when there's no result. Was that a conscious design decision?

stevenschlansker:
Well, right now it is checking that it won't accidentally return you an empty list in a context you don't expect it.

For example `@SqlQuery("UPDATE ...")` won't return anything useful (there's no result set, it's probably an error) and this check causes that to fail explosively instead of silently.  This allows you to "opt in" to the behavior you want, while keeping the default as strictly checking that queries return a result set.

TheRealMarnes:
Hmm, understandable but kinda clashy with collection concepts...

Is there maybe a way around it for now, with `query.execute(customResultProducer)` that sprinkles optionality over the existing mapmapper-lister thing? I honestly still get lost in the ResultBearing ResultProducer etc jargon... :/

stevenschlansker:
I agree that the ResultBearing / ResultProducing stuff is confusing, I still have to trace through it in the debugger to figure out what's going on sometimes :/

TheRealMarnes:
Just stepped through the call chain in the debugger, I don't think I can get around the exception... it's too deep down to be able to pass in other impls.

I agree your PR here is the cleanest way to do it without basically rewriting how all of jdbi works again (I remember a timezone discussion)... Is there an ETA for the next update?

stevenschlansker:
needs to get approved, and I'd like to get some of the other open PRs merged, but after that we can cut a release shortly.

TheRealMarnes:
Reading through your pull request, it finally really clicked in my mind how jdbi3's configuration system (which I previously thought was... complicated?) works. Fuck me, that is such a beautiful system.

stevenschlansker:
Yeah, you could plug in your own ResultProducer, but I agree it's a lot easier to just make the existing internal ones configurable.

core/src/main/java/org/jdbi/v3/core/result/ResultProducers.java
```diff
@@ -49,16 +58,30 @@ private ResultProducers() {
...
+        return new ResultProducers(false).returningQueryResults();
+    }
+
+    /**
+     * Result producer that returns a {@link ResultBearing} over the statement result rows.
+     *
+     * @return ResultBearing of result rows.
+     * @see PreparedStatement#getResultSet()
+     */
+    public ResultProducer<ResultBearing> returningQueryResults() {
```
qualidafial: Just seeing this. This method and `getResultSet` could be static by getting the config through the `StatementContext`.

qualidafial: @stevenschlansker ^

stevenschlansker: Fixing it in a followup commit, this means we don't have to change the api at all, thanks!

qualidafial: :+1:

PR #1066, opened 2018-04-02
Support SQL Object default methods in JDK9+

### LINKED ISSUE #497: JDK9 will block setAccessible on modules
We're using `setAccessible(true)` to get reflective access to private API in a couple places. We cannot count on this after JDK9 is released.

In particular, `DefaultMethodHandler` in SQL Object is using `setAccessible` to get access to the `MethodHandle.Lookup` for private members, so that we can invoke the default ("super") method.

See:
- http://jigsaw-dev.1059479.n5.nabble.com/creating-proxies-for-interfaces-with-default-methods-td5711955.html
- http://mlvm-dev.openjdk.java.narkive.com/ocxrCtZV/proxy-an-interface-and-call-a-default-method

qualidafial:
Fixes #497.

@keithbranton can you please confirm this fixes SQL Object default methods in JDK 9?

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultMethodHandler.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultMethodHandler.java
@@ -18,6 +18,7 @@ import static java.util.Collections.synchronizedMap;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Map;
 import java.util.WeakHashMap;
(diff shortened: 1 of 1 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultMethodHandler.java
```diff
@@ -18,29 +18,55 @@
...
             try {
-                // TERRIBLE, HORRIBLE, NO GOOD, VERY BAD HACK
-                // Courtesy of:
-                // https://rmannibucau.wordpress.com/2014/03/27/java-8-default-interface-methods-and-jdk-dynamic-proxies/
+                return (MethodHandles.Lookup) PRIVATE_LOOKUP_IN.invoke(null, clazz, MethodHandles.lookup());
+            } catch (IllegalAccessException | InvocationTargetException e) {
+                String message = String.format(
+                        "Error invoking MethodHandles.privateLookupIn(%s.class, MethodHandles.lookup()) in JDK 9+ runtime",
+                        clazz);
+                throw new RuntimeException(message, e);
```
qualidafial: Should we attempt to fall through to the JDK 8 hack here?

keithbranton: I'm not sure it would help - if privateLookupIn exists then the java 8 hack won't likely work - unless Java 10+ is different...

qualidafial: This is one of those "this should never happen" scenarios. Maybe we should keep the exception so we hear about it if our assumption turns out to be invalid.

qualidafial:
Awesome. Thanks Keith.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultMethodHandler.java
kkrgwbj: the MethodHandler performance is  low

qualidafial: You observed a slowdown in the latest change?

stevenschlansker: Also, 'slow' can mean a lot of different things.  Even if the method invocation took 10us (crazy long) it still pales to the cost of a roundtrip and query on most DBs...

PR #1067, opened 2018-04-04
Fix docs that still refer to now-removed LinkedHashMapRowReducer.of() method

### LINKED ISSUE #1034: Factory method " LinkedHashMapRowReducer.of" does not exist
In the documentation under http://jdbi.org/#_resultbearing_reducerows it says that the factory method "LinkedHashMapRowReducer.of" can be used, but it doesn't seem to exist in jdbi-core v3.1.0.

Am I missing something here?

qualidafial:
Fixes #1034


qualidafial:
@stevenschlansker or @arteam, would you mind reviewing?

Issue #1071, opened 2018-04-09
[bug?] templating makes Lexer trip over query


TheRealMarnes:
I'm trying to execute the statement `backup database to ?` with ? bound to a file path as string (e.g. `'db/backup/'`). This command is an hsqldb feature.

It does not work with jdbi: no matter how I try to execute the statement or in what form I bind the parameter, I always get `java.sql.SQLSyntaxErrorException: unexpected token: ? in statement [backup database to ?]`

Is this a known limitation of the binding code? That it can only handle "standard" sql and not foreign commands like this? Or should I go further into details?

qualidafial:
Are you using an `Update` or something else?

TheRealMarnes:
@qualidafial I've tried handle.execute, handle.createUpdate, handle.createCall, handle.prepareBatch, ...

`backup database to ?` keeps failing with
> java.sql.SQLSyntaxErrorException: unexpected token: ? in statement [backup database to ?]
no matter how I try to execute it. Seems to be a parser limitation.

Worse, if I prepare the statement myself and just pass the final string to jdbi to execute without binding/parsing, it works, but only if I leave out the `as files` clause. `backup database to 'foo' as files` fails no matter what, with
>UnableToCreateStatementException: Error rendering SQL template: 'backup database to 'db\backup\backup_2018-04-09_19-35-50\' as files' [statement:"backup database to 'db\backup\backup_2018-04-09_19-35-50\' as files", rewritten:"null", parsed:"null", arguments:{ positional:{}, named:{}, finder:[]}]
>Caused by: org.antlr.runtime.MismatchedTokenException: null
	at org.antlr.runtime.Lexer.match(Lexer.java:197)

stevenschlansker:
Often non-DDL statements do not take parameters at all -- for example you cannot parameterize `DROP TABLE ?`.

Instead, you have to `DROP TABLE <name>`, `.define("name", tableName)`, and use one of the template engines to substitute it textually.

I'm not positive for this `backup` case, but it could easily be the same problem.

stevenschlansker:
The latter `UnableToCreateStatementException` does look like a parser limitation, we should probably treat separate to the initial `backup database to ?` problem.

TheRealMarnes:
Hmm, I guess using templating is the answer in this case. I didn't realize parameterizing a statement is handled in the database, not jdbi/jdbc.

[...]

PR #1072, opened 2018-04-09
#1071 MessageFormatTemplateEngine

TheRealMarnes:
https://github.com/jdbi/jdbi/issues/1071

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/MessageFormatTemplateEngine.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/MessageFormatTemplateEngine.java
@@ -63,7 +63,7 @@ public enum MessageFormatTemplateEngine implements TemplateEngine {
         if (keySet.size() != expectedCount) {
             throw new IllegalArgumentException("expected " + expectedCount + " keys but got " + keySet.size());
         }
-        
+
         if (keySet.size() == 0) {
             return;
         }
--- a/core/src/main/java/org/jdbi/v3/core/statement/MessageFormatTemplateEngine.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/MessageFormatTemplateEngine.java
@@ -15,17 +15,18 @@ package org.jdbi.v3.core.statement;
 
 import java.text.MessageFormat;
 import java.util.AbstractMap;
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.Set;
 
 /**
- * Uses {@link MessageFormat#format(String, Object...)} as a template engine.
+ * Uses the equivalent of {@link MessageFormat#format(String, Object...)} as a template engine.
  *
- * You must use "0", "1", "2", etc as keys. You may {@link org.jdbi.v3.core.config.Configurable#define(String, Object)} values in any order.
+ * You must use "0", "1", "2", etc as keys: start at 0, increment by 1. Keys must be numerically unique. You must {@link org.jdbi.v3.core.config.Configurable#define(String, Object)} as many key/value pairs as there are placeholders in the pattern string.
  *
- * Start at 0, increment by 1, do not repeat any keys, and do not exceed the maximum array size for your system. Leading zeroes are ignored. Invalid keys will trigger an {@link IllegalArgumentException} (or subclasses such as {@link NumberFormatException}) when {@link #render(String, StatementContext)} is called.
+ * You may {@code define} key/value pairs in any order. Keys may contain leading {@code '0'}s.
  *
- * MessageFormat does not throw exceptions when your pattern string's placeholders don't match the values, and neither does this class. This class only validates your keys in order to reliably reconstruct the values array. It does not validate the pattern string or values.
+ * Any invalid use will trigger an {@link IllegalArgumentException} (or subclasses such as {@link NumberFormatException}) when {@link #render(String, StatementContext)} is called – typically when the statement is about to be executed.
  *
  * Example usage:
  * <pre>{@code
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jdbi/v3/core/statement/MessageFormatTemplateEngine.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Comparator;
+
+/**
+ * Uses {@link MessageFormat#format} as a template engine.
+ *
+ * You need to use "0", "1", "2", etc as keys. You can {@link org.jdbi.v3.core.config.Configurable#define} values in any order.
+ *
+ * Start at 0, increment by 1, do not repeat any keys, and do not exceed the maximum array size for your system. Leading zeroes are ignored.
+ *
+ * Keys are NOT checked for semantic or technical correctness — bad keys lead to undefined behavior that may include {@link RuntimeException}s, being ignored, overwriting other values, and wrong ordering,
```
qualidafial: Would be great to have tests for these caveats.

TheRealMarnes:
Right, unit tests and stuff slipped my mind.

TheRealMarnes:
Have you guys considered adding a .editorconfig to the project?

stevenschlansker:
I'm fine with having one -- I never added one because I don't use it personally.

TheRealMarnes:
Hmm, I just learned that MessageFormat is pretty lax in its parsing. Superfluous or missing argument objects don't trigger any exceptions, they're just ignored or their placeholders left in the string.

I think it's best we don't add validation MessageFormat doesn't have, so that this class is just an extension of MessageFormat with no new defined behavior (aside from of course having to parse the key strings) on top of what MessageFormat defines. The tests would just serve as a reference. Sound good?

Bedtime for me...

core/src/test/java/org/jdbi/v3/core/statement/TestMessageFormatTemplateEngine.java
```diff
@@ -0,0 +1,100 @@
...
+
+        templateEngine.render("foo bar", ctx);
+    }
+
+    @Test
+    public void testWithPlaceholdersButNoValues()
+    {
+        attributes.clear();
+
+        templateEngine.render("{0} bar", ctx);
```
qualidafial: Would be good to assert the return value--not just that it doesn't throw an exception.

TheRealMarnes:
Got a bit more familiar with MessageFormat today. It's actually fairly easy to enforce that the placeholders in the string match the given arguments in number. I think that may be a worthwhile validation to add.

TheRealMarnes:
It just came to mind users might want to pass in custom Formats, but that would be quite an edge case I think. Passed Formats are also order-sensitive and need to match the variables/placeholders, so it's a brittle construction. I don't think I'd want to add it yet, but it can be done in the future - make it a normal stateful class (not an enum) so it can accept Formats, and turn the INSTANCE field into a public static final default instance. Letting the user add custom Formats in the future is non-breaking.

core/src/main/java/org/jdbi/v3/core/statement/MessageFormatTemplateEngine.java
```diff
@@ -0,0 +1,97 @@
...
+ * You must use "0", "1", "2", etc as keys: start at 0, increment by 1. Keys must be numerically unique. You must {@link org.jdbi.v3.core.config.Configurable#define(String, Object)} as many key/value pairs as there are placeholders in the pattern string.
+ *
+ * You may {@code define} key/value pairs in any order. Keys may contain leading {@code '0'}s.
+ *
+ * Any invalid use will trigger an {@link IllegalArgumentException} (or subclasses such as {@link NumberFormatException}) when {@link #render(String, StatementContext)} is called – typically when the statement is about to be executed.
+ *
+ * Example usage:
+ * <pre>{@code
+ *     // select bar from foo where col = 'abc'
+ *     jdbi.useHandle(handle -> handle.createCall("select {1} from {0} where col = ''{2}''")
```
qualidafial: Do the `}` in the message format markers end the `@code` block prematurely?

GitHub code highlighting suggests they do.

core/src/main/java/org/jdbi/v3/core/statement/MessageFormatTemplateEngine.java
```diff
@@ -0,0 +1,97 @@
...
+    @Override
+    public String render(String template, StatementContext ctx) {
+        MessageFormat msgFormat = new MessageFormat(template);
+
+        validateKeys(ctx.getAttributes().keySet(), msgFormat.getFormats().length);
+
+        Object[] args = ctx.getAttributes()
+            .entrySet()
+            .stream()
+            .map(x -> new AbstractMap.SimpleImmutableEntry<>(Integer.valueOf(x.getKey()), x.getValue()))
```
qualidafial: Integer.valueOf treats leading zeros as octal notation instead of decimal if I'm not mistaken.

TheRealMarnes: will double-check later but I thought it delegated to a default radix =10. Will fix and test

qualidafial: You may be right, I might be thinking of parseInt. Double check and let me know

core/src/main/java/org/jdbi/v3/core/statement/MessageFormatTemplateEngine.java
TheRealMarnes: Looks alright to me
image

core/src/main/java/org/jdbi/v3/core/statement/MessageFormatTemplateEngine.java
TheRealMarnes: Both parseInt and valueOf delegate to radix = 10

qualidafial:
Thanks for the PR!

PR #1073, opened 2018-04-10
easier life for contributors: add editorconfig

TheRealMarnes:


TheRealMarnes:
File should be expanded with more rules by people with knowledge of the project conventions. I only wrote a java section because that's all I worked with and I just assumed the rules from existing files and checkstyle errors.

qualidafial:
Thanks for the PR!

Issue #1074, opened 2018-04-10
Add optionality to nested beans


28Smiles:
If you bind a bean like
```
@SqlUpdate("INSERT INTO subject (name, right_id) VALUES (:name, :right.id)")
@GetGeneratedKeys("id")
long insert(@BindBean SubjectJdbi subject);
```
With a Bean like
```
public class SubjectJdbi {
    private long id;
    public String name;

    @Nested
    public RightJdbi right;
}
```
And access the nested bean `right.id` you get a nullpointer exception,
which is correct for a statement in most cases,
but I would appreciate if this wasn't the case sometimes.
I wan't to propose to add an opportunity to mark a nested bean as null able,
so `right.id` would be NULL instead of throwing a NullPointerExeption.
Maybe a attribute could be added to `@Nested` flagging the field as null able or `@Nullable` could be read.

PR #1075, opened 2018-04-11
#1074 Added OptionalNestedArgument Annotation for Nested Beans

28Smiles:
#1074 
Added OptionalNestedArgument Annotation to enable optional nested beans.

### DIFF
--- a/core/src/main/antlr3/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g
+++ b/core/src/main/antlr3/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g
@@ -19,7 +19,7 @@ fragment COLON: {input.LA(2) != ':'}?=> ':';
 fragment DOUBLE_COLON: {input.LA(2) == ':'}?=> '::';
 fragment QUESTION: {input.LA(2) != '?'}?=> '?';
 fragment DOUBLE_QUESTION: {input.LA(2) == '?'}?=> '??';
-fragment NAME: 'a'..'z' | 'A'..'Z' | '0'..'9' | '_' | '.' | '#';
+fragment NAME: 'a'..'z' | 'A'..'Z' | '0'..'9' | '_' | '.' | '?.' | '#';
 
 COMMENT: '/*' .* '*/';
 QUOTED_TEXT: QUOTE (ESCAPE_QUOTE | ~QUOTE)* QUOTE;
--- a/core/src/main/java/org/jdbi/v3/core/argument/ObjectPropertyNamedArgumentFinder.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/ObjectPropertyNamedArgumentFinder.java
@@ -42,7 +42,6 @@ abstract class ObjectPropertyNamedArgumentFinder implements NamedArgumentFinder 
     }
 
     @Override
-    @SuppressWarnings("PMD.AvoidDeeplyNestedIfStmts")
     public final Optional<Argument> find(String name, StatementContext ctx) {
         if (name.startsWith(prefix)) {
             final String actualName = name.substring(prefix.length());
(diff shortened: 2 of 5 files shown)


core/src/main/java/org/jdbi/v3/core/argument/OptionalNestedArgument.java
```diff
@@ -0,0 +1,30 @@
...
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Signals that the annotated element is a optional nested mapped argument type.
+ */
+@Retention(RUNTIME)
+@Target({PARAMETER, FIELD, METHOD})
+public @interface OptionalNestedArgument {
```
stevenschlansker: I think I'd rather we find a way to do this other than a new interface.  I'd prefer an attribute on `@Nested`, or detection of `javax.annotation.Nullable` I think

qualidafial: I was thinking `@Nullable` too

28Smiles: I like both, I will change this to `@Nullable` and add your crazy idea with :outher.?inner

core/src/main/java/org/jdbi/v3/core/argument/NullArgumentFinder.java
```diff
@@ -0,0 +1,29 @@
...
+import java.sql.Types;
+import java.util.Optional;
+
+/**
+ * Finds a Null Argument for everything.
+ */
+public class NullArgumentFinder implements NamedArgumentFinder {
+    @Override
+    public Optional<Argument> find(final String name, final StatementContext ctx) {
+        return Optional.of(new NullArgument(Types.OTHER));
```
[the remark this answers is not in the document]
28Smiles: Thanks, I've searched for that

core/src/main/java/org/jdbi/v3/core/argument/ObjectFieldArguments.java
```diff
@@ -81,7 +77,7 @@ public ObjectFieldArguments(String prefix, Object bean) {
     }
 
     @Override
-    NamedArgumentFinder getNestedArgumentFinder(Object obj) {
+    NamedArgumentFinder getNestedArgumentFinder(String name, Object obj) {
```
qualidafial: I think a more effective approach would be to add a `Class<?>` parameter instead of `name`, and modify the code to tolerate `obj == null`.

When you're resolving the type of the property, you use `obj.getClass()` if not null, otherwise fall back to the provided type.

28Smiles:
So which Syntax do you preffere? `bean.?bean.?id` or `bean.bean?.id?`?

28Smiles:
Then I will change this too ?.

28Smiles:
@qualidafial this should be mergeable or is the anything left to change?

28Smiles:
But wasn't it @stevenschlansker who proposed this kind of syntax. In my opinion it makes sence, since you have the additional not null check, if you silently place a null value it may be due a missimplementation of the Extention, in this way we have a kind of failure safety.
I hope you get what i want to say with this.

core/src/test/java/org/jdbi/v3/core/argument/TestBeanArguments.java
```diff
@@ -122,4 +122,95 @@ public void setBar(String bar) {}
         exception.expect(UnableToCreateStatementException.class);
         new BeanPropertyArguments("foo", bean).find("foo.bar", ctx);
     }
+
+    @Test
+    public void testBindNestedOptionalNull() throws Exception
```
qualidafial: These are good tests to have, but we need to supplement them with tests that actually use the database, to confirm that optional properties work as expected.

28Smiles:
I cleaned up my mess of pushes.

core/src/main/java/org/jdbi/v3/core/argument/ObjectPropertyNamedArgumentFinder.java
```diff
@@ -41,6 +42,7 @@
     }
 
     @Override
+    @SuppressWarnings("PMD.AvoidDeeplyNestedIfStmts")
```
28Smiles: Is this PMD Supression ok, should it be for the line or should i remove the if statements in any way?

core/src/main/antlr3/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g
```diff
@@ -19,7 +19,7 @@ fragment COLON: {input.LA(2) != ':'}?=> ':';
 fragment DOUBLE_COLON: {input.LA(2) == ':'}?=> '::';
 fragment QUESTION: {input.LA(2) != '?'}?=> '?';
 fragment DOUBLE_QUESTION: {input.LA(2) == '?'}?=> '??';
-fragment NAME: 'a'..'z' | 'A'..'Z' | '0'..'9' | '_' | '.' | '#';
+fragment NAME: 'a'..'z' | 'A'..'Z' | '0'..'9' | '_' | '.' | '?.' | '#';
```
28Smiles: This is nessesary, to include the `?` in the binding in this case

28Smiles:
@qualidafial @qualidafial how about this?

qualidafial:
@28Smiles This looks ready to me. Did you have any other changes you wanted to push before we merge?

28Smiles:
@qualidafial Yes I think it is ready to merge too

qualidafial:
Thanks for the PR!

PR #1076, opened 2018-04-12
#841 [WIP] Added Support For Immutable Mapping

28Smiles:
Added Support for using [Immutables ](https://immutables.github.io/).


core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableInfo.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public ImmutableInfo(Class immutable, Class builder) {
+        List<Method> getter = Arrays.asList(immutable.getMethods());
+        List<Method> setter = Arrays.asList(builder.getMethods());
+
+        descriptors = getter.stream().map(g -> {
+            try {
+                return new PropertyDescriptor(g.getName(), g, setter.stream().filter(s -> s.getName().equals(g.getName())).findFirst().get());
+            } catch (IntrospectionException e) {
+                throw new IllegalArgumentException();
```
stevenschlansker: `throw new IllegalArgumentException(e)`

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
```diff
@@ -0,0 +1,223 @@
...
+
+    private ImmutableMapper(Class<T> type, String prefix) {
+
+        this.type = type;
+        this.prefix = prefix;
+        this.implementation = (Class<? extends T>) type.getAnnotation(ImmutableImplementation.class).value();
+
+        try {
+            builder = this.implementation.getMethod("builder").invoke(null);
+        } catch (IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
```
stevenschlansker: Could this use the new `MethodHandles` approach to avoid exception wrapping and other ickiness of the old reflection API?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
```diff
@@ -0,0 +1,223 @@
...
+
+        return (r, c) -> {
+            for (int i = 0; i < mappers.size(); i++) {
+                RowMapper<?> mapper = mappers.get(i);
+                PropertyDescriptor property = properties.get(i);
+
+                Object value = mapper.map(r, ctx);
+
+                try {
+                    property.getWriteMethod().invoke(builder, value);
```
stevenschlansker: prefer MethodHandles again, or at least provide a cause of `e`.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
```diff
@@ -0,0 +1,223 @@
...
+                    throw new SQLException();
+                }
+            }
+
+            try {
+                return (T) builder.getClass().getMethod("build").invoke(builder);
+            } catch (IllegalAccessException | NoSuchMethodException e) {
+                throw new IllegalArgumentException();
+            } catch (InvocationTargetException e) {
+                if (e.getTargetException() instanceof IllegalStateException) {
```
stevenschlansker: This catch block looks even stranger.

core/src/test/java/org/jdbi/v3/core/SampleImmutable.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core;
+
+import org.immutables.value.Value;
+import org.jdbi.v3.core.mapper.reflect.ImmutableImplementation;
+
+@ImmutableImplementation(ImmutableSampleImmutable.class)
```
stevenschlansker: Does the `@ImmutableImplementation` serve only to point at a generated class file?  Is there some way we can discover the class file produced instead of having to write it out in the source file, just by the `@Value.Immutable` annotation?

core/pom.xml
```diff
@@ -81,6 +81,12 @@
             <artifactId>expiringmap</artifactId>
         </dependency>
 
+        <dependency>
```
stevenschlansker: I don't think we want to add this as a strong dependency in `core`.  Either it should be a separate module, or the dependency should be `<optional>true</optional>`.  It should probably then register via a `JdbiPlugin`.

qualidafial: Agree. I favor spinning up a `jdbi3-immutables` module.

core/src/test/java/org/jdbi/v3/core/SampleImmutable.java
28Smiles: working on this, but at least the `@Value.Immutable` does not contain any information about the Implementation Class, but I could search it by it's generic name.

core/pom.xml
28Smiles: I don't use maven very often, but I will try that

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
28Smiles: Not here because I can not know the returntype

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
28Smiles: Did it with `MethodHandles` now, but does not get any better, still prefer `MethodHandles` over normal reflection

docs/src/adoc/index.adoc
```diff
@@ -1302,6 +1302,82 @@ List<User> users = handle
...
+handle.registerRowMapper(JoinRowMapper.forTypes(Contact.class, Phone.class);
+List<JoinRow> contactPhones = handle.select("select " +
+        "c.id cid, c.name cname, " +
+        "p.id pid, p.name pname, p.number pnumber " +
+        "from contacts c left join phones p on c.id = p.contact_id")
+    .mapTo(JoinRow.class)
+    .list();
+----
+
+Nested immutable-mapped types can be mapped using the `@Nested` annotation:
```
qualidafial: Can we get unit tests for this scenario?

sqlobject/pom.xml
```diff
@@ -77,6 +77,14 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.immutables</groupId>
+            <artifactId>value</artifactId>
+            <version>2.5.6</version>
+            <optional>true</optional>
+            <scope>test</scope>
```
qualidafial: Test scope is sufficient here, we don't need to make it optional as well.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableInfo.java
```diff
@@ -0,0 +1,48 @@
...
+
+    private List<PropertyDescriptor> descriptors;
+
+    public ImmutableInfo(Class immutable, Class builder) {
+        List<Method> getter = Arrays.asList(immutable.getMethods());
+        List<Method> setter = Arrays.asList(builder.getMethods());
+
+        descriptors = getter.stream().map(g -> {
+            try {
+                return new PropertyDescriptor(g.getName(), g, setter.stream().filter(s -> s.getName().equals(g.getName())).findFirst().get());
```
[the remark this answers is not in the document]
qualidafial: Also, are we recognizing "wither" setter names in builders?

qualidafial:
@28Smiles Ping. We're still interested in including this in the project, if you can find the time to address the items in the reviews above.

28Smiles:
@qualidafial Pong. #joking 
I'm still planning on finishing this, but currently I'm bound to Work, but I'll look forward to Implement this till the end of the Week.
We are using Jdbi in our Jersey WebApplication and I would like to use some Immutables at some point.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
```diff
@@ -246,4 +245,34 @@ private static String paramName(PropertyDescriptor descriptor)
...
+        }).collect(Collectors.toList());
+    }
+
+    private String getRawName(String name) {
+        if (name.startsWith("get")) {
+            if (Character.isUpperCase(name.charAt(3))) {
+                return Character.toLowerCase(name.charAt(3)) + name.substring(4);
+            }
+        }
+        return name;
```
qualidafial: Should also look for `is` prefix for boolean properties.

core/src/test/java/org/jdbi/v3/core/mapper/reflect/ImmutableTest.java
```diff
@@ -71,93 +71,103 @@ private void mockAllNullsResult() throws SQLException {
...
     @Test
     public void shouldSetValueWithBuilder() throws Exception {
-        int aIntVal = 42;
+        int aIntVal1 = 42;
         long aLongVal = 100L;
         String aStringValue = "Foo";
+        int aIntVal2 = 42;
 
-        mockColumns("id", "name", "valueInt");
+        mockColumns("id", "name", "valueInt", "inner_id");
```
qualidafial: Just a reminder: these tests still need to be changed into integration tests, and using a real database. We need to see that immutable mappers work as expected with the whole of Jdbi--not just in isolation.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
```diff
@@ -194,41 +166,44 @@ public T map(final ResultSet rs, final StatementContext ctx) throws SQLException
...
-                    throw (RuntimeException) e;
+
+            return (T) Proxy.newProxyInstance(this.getClass().getClassLoader(), new Class[]{type}, new InvocationHandler() {
+                private final Map<Method, Object> invocationMap = invocations;
+
+                @Override
+                public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+                    Object returnObj = invocationMap.get(method);
+                    if (returnObj == null) {
+                        if (method.isDefault() && constructor != null) {
```
28Smiles: Should be checked for default Values on the evaluation or on the mapping?

28Smiles:
I pfeffere the Implementation with Proxys, because it is more versatile and doesn't depend on any libary except Java

28Smiles:
Also Integration Tests are Implemented

28Smiles:
Sry, forgot about the other Test, I'll fix this

28Smiles:
I'm sorry, my IntelliJ is normally configured to prohibit star imports -_- forgot at this about reconfiguring this

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
```diff
@@ -194,41 +166,44 @@ public T map(final ResultSet rs, final StatementContext ctx) throws SQLException
...
+                            final Class<?> declaringClass = method.getDeclaringClass();
+                            return constructor.newInstance(declaringClass, MethodHandles.Lookup.PRIVATE)
+                                .unreflectSpecial(method, declaringClass)
+                                .bindTo(proxy)
+                                .invokeWithArguments(args);
```
qualidafial: Are you running on Java 9? I think you'll run into problems with `unreflectSpecial` on default methods in Java 8.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
qualidafial: Definitely on mapping--if you try to wait until the method is invoked, the JDBC result set will be closed.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
```diff
@@ -194,41 +166,44 @@ public T map(final ResultSet rs, final StatementContext ctx) throws SQLException
...
-                try {
-                    MethodHandles.lookup().unreflect(property.getWriteMethod()).bindTo(builder).invokeWithArguments(value);
-                } catch (IllegalAccessException e) {
-                    throw new MappingException("Setter for the builder should be public", e);
-                } catch (RuntimeException e) {
-                    throw e;
-                } catch (Throwable e) {
-                    throw new MappingException("Immutable build failed with exception", e);
-                }
+                invocations.put(property.getReadMethod(), value);
```
qualidafial: It seems like you're trying to roll your own Immutables?

I'd much rather just integrate with what Immutables already does than try to roll our own here--why don't we just spin up a separate `jdbi3-immutables` module?

immutable/pom.xml
```diff
@@ -0,0 +1,15 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>jdbi3-parent</artifactId>
+        <groupId>org.jdbi</groupId>
+        <version>3.2.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>immutable</artifactId>
```
qualidafial: Let's make this `jdbi3-immutables` to agree with the other module names.

immutable/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
```diff
@@ -0,0 +1,308 @@
...
+                        } else {
+                            Class<?> returnType = method.getReturnType();
+                            if (returnType.equals(Map.class)) {
+                                return new HashMap<>(); // TODO IMMUTABLE MAP
+                            }
+                            if (returnType.equals(List.class)) {
+                                return new ArrayList<>(); // TODO IMMUTABLE LIST
+                            }
+                            if (returnType.equals(Set.class)) {
+                                return new HashSet<>(); // TODO IMMUTABLE SET
```
qualidafial: Could use `Collections.emptyMap()`, `emptyList()`, and `emptySet()` here.

immutable/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
```diff
@@ -0,0 +1,308 @@
...
+            throw new MappingException(String.format("Mapping immutable type %s " +
+                "didn't find any matching columns in result set", type));
+        }
+
+        return (r, c) -> {
+            final Map<Method, Object> invocations = new HashMap<>();
+            for (int i = 0; i < mappers.size(); i++) {
+                RowMapper<?> mapper = mappers.get(i);
+                PropertyDescriptor property = properties.get(i);
+                Object value = mapper.map(r, ctx);
```
qualidafial: Would be good to wrap the raw values in unmodifiable wrappers if they are a map, set, or collection.

Later on we could also think about supporting array properties.

immutable/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    private String debugName(PropertyDescriptor descriptor) {
+        return String.format("%s.%s", type.getSimpleName(), descriptor.getName());
+    }
+
+    private List<PropertyDescriptor> getInformation(Class<T> immutable) {
+        final List<Method> getter = Arrays.asList(immutable.getMethods());
+
+        return getter.stream().filter(g -> g.getParameterCount() == 0).map(g -> {
```
qualidafial: May want to exclude the `getClass()` method

immutable/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutableMapper.java
```diff
@@ -0,0 +1,308 @@
...
+        }
+        if (name.startsWith("is")) {
+            if (Character.isUpperCase(name.charAt(2))) {
+                return Character.toLowerCase(name.charAt(2)) + name.substring(3);
+            }
+        }
+        return name;
+    }
+
+    private Object invokeDefault(Method method, Object proxy, Class<?> declaringClass, Object[] args) throws Throwable {
```
qualidafial: We already have similar code in SQL object (DefaultMethodHandler).

Let's extract the relevant code from there into an internal utility class in core, and reuse it from here rather than reimplementing.

immutable/src/main/java/org/jdbi/v3/sqlobject/config/internal/RegisterImmutableMapperImpl.java
```diff
@@ -0,0 +1,45 @@
...
+    @Override
+    public void configureForType(ConfigRegistry registry, Annotation annotation, Class<?> sqlObjectType) {
+        RegisterImmutableMapper registerImmutableMapper = (RegisterImmutableMapper) annotation;
+        Class<?> beanClass = registerImmutableMapper.value();
+        String prefix = registerImmutableMapper.prefix();
+        RowMappers mappers = registry.get(RowMappers.class);
+        if (prefix.isEmpty()) {
+            mappers.register(ImmutableMapper.factory(beanClass));
+        } else {
+            mappers.register(ImmutableMapper.factory(beanClass, prefix));
```
qualidafial: No need to if/else branch--you can just pass in the empty prefix

qualidafial:
Decision point:

I don't think that you can call it "immutables" or advertise compatibility with Immutables and have it implemented in proxies--not without completely reimplementing Immutables in Jdbi.

Proxies will produce a _ton_ of garbage and perform slower relative to the immutable classes generated by Immutables.

If we stick with the proxy approach, then I think we should move the classes back into core, but rename the mapper to e.g. `ProxyMapper` and the annotation to `@RegisterProxyMapper`. It might also permit us to shed some of the baggage from Immutables, such as guaranteeing that collection / map properties are immutable. We could advertise that the proxy mapper is low ceremony but slow, and users would be accepting that tradeoff by using it.

On the other hand, if we want to build something that's strictly compatible with Immutables, we should reflectively invoke the builder interface to map immutable objects.

28Smiles:
I will rethink this, thank you for your input. I will probably have a look into this again in the next few days.

PR #1077, opened 2018-04-12
gitattributes, checkstyle, editorconfig

TheRealMarnes:
The top and right now only line tells git to try to determine if a file is text or binary, and if it's text, convert its line endings to the pc-native type. This is done regardless of user settings.

I dislike windows' special line endings so I've disabled autocrlf on my system to be able to work with LF despite being on windows. I dislike git ever changing the content of my files anyway.
However, this caused jdbi's mvn builds to fail on my system: checkstyle checks that files end with _the native_ line endings. Having autocrlf disabled so my files all had LF instead of CRLF caused checkstyle to go boom on every file.

Having a file like this would prevent such situations and is a good thing for project consistency, where project rules should be enforced over personal style anyway.

I expect this file will need to get some more fine-grained rules by people with more knowledge of the project.

This concept is incompatible with editorconfig's `end_of_line` setting which can only be set to `lf` or `crlf`, not `native`. Since the build depends on files having native line endings, I would just remove this setting from editorconfig.

Maybe some discussion needs to be had so there can be a definitive decision about line endings. It's a tricky subject that's as widely debated as spaces-vs-tabs though. Personally I think forcing native line endings is not a bad idea even though it's not my taste. Forcing lf isn't a bad idea either but I expect some people will dislike it. Files that necessarily must stay the same on all platforms (like flyway scripts, which are hashed) can be added to the gitattributes file.


TheRealMarnes:
Making checkstyle accept cross-platform line endings would work too of course...

I don't know, does the jdbi team have an opinion on everything regarding git, newlines, and crossplatform? I'm conflicted myself and can't seem to find a golden solution.

stevenschlansker:
I personally think anything but `lf` only is crazy, but if enough users use broken operating systems and will complain, we might have to stay with `native`.

If we decide to go for `lf` mandated line endings, we could contribute a Checkstyle plugin to enforce this, that would make sense.

TheRealMarnes:
Tbh as a user of such a shitty OS, I feel anything but lf is craziness too. Changing checkstyle to enforce lf instead of native is a good solution for me. The gitattributes file can help enforce it by setting `*.java text eol=lf` (include other extensions too though).

qualidafial:
I'm not opposed to normalizing on LF line endings. Do we know how much effort it is to reconfigure checkstyle?

TheRealMarnes:
I've modified the editorconfig and gitattributes to force `lf` everywhere. I'm not 100% sure it's safe to do because I don't know some of the affected file types, so someone more involved should double-check and test a build.

For checkstyle, my personal `lf` project uses this check to make it happen:
```xml
<module name="RegexpMultiline">
	<property name="format" value="(?s:(\r\n|\r).*)"/>
	<property name="message" value="Line has CRLF or CR line ending."/>
</module>
```

But I don't know where to put it. There's a checkstyle suppressions file but no checkstyle config file. Checkstyle isn't mentioned anywhere in the project so I guess it's inherited through parent poms, which is where I step back and say "nah"...

TheRealMarnes:
https://stackoverflow.com/questions/997021/how-to-get-rid-of-checkstyle-message-file-does-not-end-with-a-newline#15433548

this is probably a better answer

TheRealMarnes:
I can complete builds without issues now on windows and regardless of git settings, and my checkstyle config seems to be more strict than the previous one, so I'd say goal reached :)

PR #1080, opened 2018-04-13
fix #999

### LINKED ISSUE #999: SerializableTransactionRunner.Configuration.maxRetries is badly named
From what I can tell, the maxRetries property is not the number of **re**tries that will be attempted but the total number of attempts that will be made. Reading the source code and running experiments shows this.

SerializableTransactionRunner should be adjusted to count correctly: 1 base attempt followed by max n (specified) retries.

Nonsensical values like -1 and lower should throw an exception at configuration time. Setting 0 would defeat the purpose of this runner but isn't really nonsense - I'd opt to allow it for the sake of test scenarios.

this code was reviewed by

TheRealMarnes:
#999

### DIFF
--- a/core/src/test/java/org/jdbi/v3/core/transaction/TestSerializableTransactionRunner.java
+++ b/core/src/test/java/org/jdbi/v3/core/transaction/TestSerializableTransactionRunner.java
@@ -13,21 +13,22 @@
  */
 package org.jdbi.v3.core.transaction;
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
-
 import java.sql.SQLException;
 import java.util.concurrent.atomic.AtomicInteger;
-
-import org.jdbi.v3.core.rule.H2DatabaseRule;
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/jdbi/v3/core/transaction/TestSerializableTransactionRunner.java
```diff
@@ -38,38 +40,46 @@ public void setUp() throws Exception
     {
         db = Jdbi.create(dbRule.getConnectionFactory());
         db.setTransactionHandler(new SerializableTransactionRunner());
+        db.configure(SerializableTransactionRunner.Configuration.class, config -> config.setMaxRetries(RETRIES));
```
qualidafial: Tip: `configure()` is intended to allow chaining method calls on the callee (db in this instance).

When you just want to set config and don't need to chain on the callee, `getConfig(Class)` is a little simpler to use:

```java
db.getConfig(SerializableTransactionRunner.Configuration.class).setMaxRetries(RETRIES);
```

PR #1081, opened 2018-04-14
[done] fix #981 logging interface

### LINKED ISSUE #981: jdbi3 SQLLog equivalent?
I've been going through the entire javadoc and site doc but I can't seem to find a replacement mechanism for SQLLog. I can see StatementContext has some get*Sql methods, but there's no replacement for DBI.setSQLLog. TimingCollector receives the statementcontext so it could technically be done in there, but that feels like a workaround given the interface's name.

TheRealMarnes:
#981 

I'd like some input from the team about the overall design and whether I should take testing any further.

I personally think the design is ok as this is basically exactly what I do in my personal project for which I opened the issue.

LoggableArgument is a copy-paste from my project, where I use it in all of my custom ArgumentFactories. IIRC the BuiltInArgumentFactory already provides a usable `toString()`, so if users make sure to use LoggableArgument in their custom ArgumentFactories, every Argument should be loggable. Works fine in my project.

As for testing, I think the current tests cover everything and are sufficient even if some more combinations could be made. Writing more tests feels a bit redundant to me since most of the tested code will just be the same.

I felt like deprecating TimingCollector maybe?


TheRealMarnes:
I don't see what the build problem is?

core/src/main/java/org/jdbi/v3/core/argument/LoggableArgument.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.util.Objects;
+import org.jdbi.v3.core.statement.StatementContext;
+
+/**
+ * A wrapper for {@link Argument}s that enables retrieving (e.g. for logging, see {@link org.jdbi.v3.core.statement.SqlLogger}) of the represented value.
+ */
+public class LoggableArgument implements Argument {
```
stevenschlansker: I'm not sure I understand why this wrapper is necessary.  How is it better than simply going through and ensuring all `Argument` types have an appropriate `toString`?

core/src/main/java/org/jdbi/v3/core/statement/SqlLogger.java
```diff
@@ -0,0 +1,40 @@
...
+        public void logBeforeExecution(StatementContext context) {}
+
+        @Override
+        public void logAfterExecution(StatementContext context, long nanos) {}
+
+        @Override
+        public void logException(StatementContext context, SQLException ex) {}
+    };
+
+    void logBeforeExecution(StatementContext context);
```
stevenschlansker: you could make these default methods with empty implementation, that makes it easier to make partial loggers

core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
```diff
@@ -182,11 +181,15 @@ private ExecutedBatch internalBatchExecute() {
...
             try {
+                getConfig(SqlStatements.class).getSqlLogger().logBeforeExecution(getContext());
+
                 final long start = System.nanoTime();
                 final int[] rs =  stmt.executeBatch();
                 final long elapsedTime = System.nanoTime() - start;
+
                 LOG.trace("Prepared batch of {} parts executed in {}ms", bindings.size(), elapsedTime / 1000000L, parsedSql);
                 getConfig(SqlStatements.class).getTimingCollector().collect(elapsedTime, getContext());
+                getConfig(SqlStatements.class).getSqlLogger().logAfterExecution(getContext(), elapsedTime);
```
stevenschlansker: This pattern shows up a lot.  Would it make sense for SqlLogger to have a

```java
default void timedExecute(Runnable action, StatementContext ctx) throws SQLException {
    final SqlLogger logger = ctx.getConfig(SqlStatements.class).getSqlLogger();
    try {
        beforeExecution(ctx);
        long startNs = System.nanoTime();
        action.run();
        afterExecution(ctx, System.nanoTime() - startNs);
    } catch (SQLException | RuntimeException e) {
        logException(ctx, e);
        throw e;
    }
}
````

File: core/src/main/java/org/jdbi/v3/core/statement/Batch.java [hunk 4]
```diff
@@ -91,17 +91,22 @@ public Batch add(String sql)
...
                 // Null for statement, because for batches, we don't really have a good way to keep the sql around.
                 getConfig(SqlStatements.class).getTimingCollector().collect(elapsedTime, getContext());
-                return rs;
+                getConfig(SqlStatements.class).getSqlLogger().logAfterExecution(getContext(), elapsedTime);
 
+                return rs;
             }
             catch (SQLException e)
             {
+                getConfig(SqlStatements.class).getSqlLogger().logException(getContext(), e);
```
qualidafial: Elapsed time may be desirable for errors too?

core/src/test/java/org/jdbi/v3/core/statement/TestSqlLogger.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    @Test
+    public void testStatement() {
+        h.execute(CREATE);
+        h.createUpdate(INSERT).execute();
+
+        assertThat(logger.getRawSql()).containsExactly(CREATE, CREATE, INSERT, INSERT);
+        assertThat(logger.getTimings()).hasSize(2);
+        assertThat(logger.getTimings()).allMatch(x -> x > 0);
```
qualidafial: Heads up: you can chain allMatch() after hasSize() without starting a new statement.

core/src/test/java/org/jdbi/v3/core/statement/TestSqlLogger.java
```diff
@@ -0,0 +1,184 @@
...
+            assertThat(logger.getRawSql()).containsExactly(CREATE, CREATE, INSERT_PREPARED, INSERT_PREPARED);
+            assertThat(logger.getTimings()).hasSize(1);
+            assertThat(logger.getTimings()).allMatch(x -> x > 0);
+            assertThat(logger.getExceptions()).containsExactly((SQLException) e.getCause());
+        }
+    }
+
+    @Test
+    public void testNotSql() {
+        String query = "herp derp";
```
qualidafial: Seems legit

core/src/main/java/org/jdbi/v3/core/argument/LoggableArgument.java
TheRealMarnes: I can look into it again but something about not having access to the value object and Argument being a FunctionalInterface... It wasn't possible to retrofit a proper tostring into Argument without breaking compatibility for implementers.

TheRealMarnes:
I'm not personally interested in timing errors but I'm all for adding it.

core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
TheRealMarnes: I don't really understand...

TheRealMarnes:
So the remaining issues then are `LoggableArgument` and @stevenschlansker pattern thingie I don't quite get?

TheRealMarnes:
Ok so the reason for LoggableArgument is of course that Argument is nothing but the `void apply(pos, stmt, ctx)` method. Its `toString` is no good, and implementing it is impossible because what's it supposed to return? The value object is only received by `apply`, there's nothing `toString` can log. That's why I made LoggableArgument.

stevenschlansker:
Why not just put an `Instant executionStart` to the `StatementContext`?  This is a more extensible way to add timing points (you can have "handle open" and "execute start" as separate times, for example) and it leaves the question of "do we care to time exceptions" up to whoever is implementing the timing collector.

TheRealMarnes:
Good idea. I tried to restrain myself from going overboard with fancy stuff I'm sure some people would call excessive :P But if you're in favor of slapping on some nice Instants and such, I'll get on it

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -143,6 +145,7 @@ public TimingCollector getTimingCollector() {
      * @param timingCollector the new timing collector
      * @return this
      */
+    @Deprecated
```
qualidafial: Can we get a note added to javadoc as well? e.g. `@deprecated use {@link #setSqlLogger(SqlLogger)} instead` or something like that?

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -143,6 +145,7 @@ public TimingCollector getTimingCollector() {
...
     public SqlStatements setTimingCollector(TimingCollector timingCollector) {
         this.timingCollector = timingCollector == null ? TimingCollector.NOP_TIMING_COLLECTOR : timingCollector;
```
qualidafial: We could get rid of the `timingCollector` field altogether if we wrapped the collector passed into this method in a `SqlLogger` adapter, and passed it to `setSqlLogger()`.

core/src/main/java/org/jdbi/v3/core/argument/LoggableArgument.java
qualidafial: I feel like we'd be better off providing `toString()` implementations in all our argument types.

TheRealMarnes: @qualidafial not possible without breaking compatibility afaik. I'm all for it if that's acceptable

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -150,7 +147,12 @@ public TimingCollector getTimingCollector() {
      */
     @Deprecated
     public SqlStatements setTimingCollector(TimingCollector timingCollector) {
-        this.timingCollector = timingCollector == null ? TimingCollector.NOP_TIMING_COLLECTOR : timingCollector;
+        this.sqlLogger = new SqlLogger() {
```
qualidafial: Probably still want the null check here. `timingCollector == null ? SqlLogger.NOP : etc`

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
TheRealMarnes: Hmm, not a fan of this actually, it's screwing up the test logic. Can we put this back?

qualidafial:
Looks like the tests are failing because we're still invoking the timing collector, and so every statement gets tracked twice.

Probably need to look for references to `getTimingCollector()` and make sure that we're exclusively using the SqlLogger now, instead of to both.

TheRealMarnes:
Yeah but that means no grace period for the migration away from TimingCollector. If you're ok with that, I'll fix it

TheRealMarnes:
Maybe best to just remove TimingCollector altogether then?

core/src/main/java/org/jdbi/v3/core/argument/LoggableArgument.java
qualidafial: I think we can find a compromise here: in `SqlStatement.toArgument(Type, Object)`, why don't we just wrap each argument in a simple `Argument` wrapper that delegates to the original argument, but overrides `toString()`?

core/src/main/java/org/jdbi/v3/core/statement/Batch.java
```diff
@@ -77,16 +78,23 @@ public Batch add(String sql) {
...
-                final long elapsedTime = System.nanoTime() - start;
-                LOG.trace("] executed in {}ms", elapsedTime / 1000000L);
-                // Null for statement, because for batches, we don't really have a good way to keep the sql around.
-                getConfig(SqlStatements.class).getTimingCollector().collect(elapsedTime, getContext());
-                return rs;
 
+                getContext().setCompletionMoment(Instant.now());
+                sqlLogger.logAfterExecution(getContext());
+
+                LOG.trace("] executed in {}ms", getContext().getElapsedTime(ChronoUnit.MILLIS));
```
qualidafial: I think the addition of `SqlLogger` means that these trace statements could go away.

core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
```diff
@@ -167,19 +169,27 @@ private ExecutedBatch internalBatchExecute() {
...
+
                 final int[] rs =  stmt.executeBatch();
-                final long elapsedTime = System.nanoTime() - start;
-                LOG.trace("Prepared batch of {} parts executed in {}ms", bindings.size(), elapsedTime / 1000000L, parsedSql);
-                getConfig(SqlStatements.class).getTimingCollector().collect(elapsedTime, getContext());
+
+                getContext().setCompletionMoment(Instant.now());
+                sqlLogger.logAfterExecution(getContext());
+
+                LOG.trace("Prepared batch of {} parts executed in {}ms", bindings.size(), getContext().getElapsedTime(ChronoUnit.MILLIS), parsedSql);
```
qualidafial: Ditto for this trace statement

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -1358,20 +1360,27 @@ PreparedStatement internalExecute() {
...
-            final long elapsedTime = System.nanoTime() - start;
-            LOG.trace("Execute SQL \"{}\" in {}ms", sql, elapsedTime / 1000000L);
-            getConfig(SqlStatements.class)
-                    .getTimingCollector()
-                    .collect(elapsedTime, getContext());
+
+            getContext().setCompletionMoment(Instant.now());
+            sqlLogger.logAfterExecution(getContext());
+
+            LOG.trace("Execute SQL \"{}\" in {}ms", sql, getContext().getElapsedTime(ChronoUnit.MILLIS));
```
qualidafial: and here

core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
qualidafial: Steven is suggesting adding a function to `SqlLogger` (or just some static helper method somewhere), which encapsulates the logger boilerplate, instead of repeating it in multiple classes:

```java
interface  LogCallback<T, X extends Exception> {
  T invoke() throws X;
}

static <T, X extends Exception> T executeLogged(LogCallback<T, X> callback, StatementContext ctx) throws X {
  SqlLogger logger = ctx.getConfig(SqlStatements.class).getSqlLogger();
  try {
    ctx.setExecutionMoment(Instant.now());
    logger.logBeforeExecution(ctx);

    T result = callback.invoke();

    ctx.setCompletionMoment(Instant.now());
    logger.logAfterExecution(ctx);

    return result;
  } catch (RuntimeException | X e) {
    ctx.setExceptionMoment(Instant.now());
    logger.logException(ctx, e);
    throw e;
  }
}
```

Thus, in PreparedBatch:

```java

try {
    final int[] rs = SqlLoggers.executeLogged(stmt::executeBatch, getContext());

    afterExecution(stmt);

    getContext().setBinding(new Binding());

    return new ExecutedBatch(stmt, rs);
} catch (SQLException e) {
    throw new UnableToExecuteStatementException(Batch.mungeBatchException(e), getContext());
}
```

qualidafial:
@TheRealMarnes I think we're pretty close. Thanks for being patient and putting in the time to make all these changes we've been asking for.

I'm +1 to merge this once we get the timer boilerplate method Steven asked for, and remove the redundant `LOG.trace()` statements.

core/src/test/java/org/jdbi/v3/core/statement/TestSqlLogger.java
```diff
@@ -32,6 +34,7 @@
     private static final String INSERT = "insert into foo(bar) values(1)";
     private static final String INSERT_NULL = "insert into foo(bar) values(null)";
     private static final String INSERT_PREPARED = "insert into foo(bar) values(?)";
+    private static final Predicate<Long> IS_POSITIVE = x -> x >= 0;
```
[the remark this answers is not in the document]
qualidafial: Java 9 shipped a new Clock that was supposed to improve precision.

core/src/main/java/org/jdbi/v3/core/argument/LoggableArgument.java
TheRealMarnes: > we'd ugly up the code turning all our Argument lambda expressions into anonymous inner classes

That would solve it on jdbi's end, but users would still end up implementing Argument, oblivious of needing to implement toString if they want logging to work. So we might as well include LoggableArgument to provide an easy solution for both jdbi internals and users and to make the point more clear. The question is just if we want to break compatibility by changing either the `Argument` class or all methods that take an instance of it or not...

> a simple Argument wrapper that delegates to the original argument, but overrides toString()

Like `LoggableArgument` does?

TheRealMarnes: We could add a `toString` on `Argument` with a default body that returns some helpful message about needing to override it and a pointer to `LoggableArgument` for convenience? Does that allow us to keep it as a `FunctionalInterface`? That method would be clearly visible to every user implementing Argument, would show up when they test logging, and wouldn't break any compatibility...

core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
TheRealMarnes: Ah, now I get it. Hmm... I'll cook something up.

TheRealMarnes: Can't `catch X` though

TheRealMarnes:
I'm not a big fan of SqlLogger.wrap though. Feels strange to delegate control over a database process to a logger. It's like the ticket guy telling the train driver where to take the train...

What do we do about logging? Do we figure it out now and apply the decision before we merge this, or do we handle that separately after securing a foundation?

Personally I suggest we merge this already and then we double-check that all the internal Arguments are loggable. I'll add the informative toString on Argument as a pointer to users, we keep method definitions and the Argument class as they are for compatibility, and users will just have to make a very minor modification to all their ArgumentFactories (wrapping everything in LoggableArgument).

core/src/main/java/org/jdbi/v3/core/argument/LoggableArgument.java
qualidafial: Did you see my [earlier suggestion](https://github.com/jdbi/jdbi/pull/1081#discussion_r184566588)? This would make it unnecessary for everybody to implement `toString()` in their own argument types.

core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
qualidafial: Catch `Exception` and unsafe-cast it to `X` to rethrow.

qualidafial: See `LocalTransactionHandler.java` line 142 for an example.

core/src/main/java/org/jdbi/v3/core/argument/LoggableArgument.java
TheRealMarnes: I see a lot of other types of Argument binding though, like bean binding and such. I'm a little too fuzzy on those implementations to be able to tell just like that if anything/what needs to be done about those...

core/src/main/java/org/jdbi/v3/core/statement/SqlLoggable.java
```diff
@@ -0,0 +1,21 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.statement;
+
+import java.sql.SQLException;
+
+@FunctionalInterface
+public interface SqlLoggable<T> {
+    T invoke() throws SQLException;
```
qualidafial: SQLException is a good compromise. :+1:

TheRealMarnes: well I can leave it as that or make it X with an unsafe cast. I see pros and cons to both, what you?

TheRealMarnes: If I make it X though, SqlLogger will also become generic with an exception type... maybe not desirable

qualidafial: I think `SQLException` is reasonable--let's just stick with that.

TheRealMarnes: Yeah, I just implemented and pushed a refactor to X for demonstration. It works but I don't see it adding any value, and it's a burden on implementers of `logException`. I'd stick to SQLException too.

TheRealMarnes:
I'd like to add an SqliteRule for testing since sqlite doesn't give a crap about data types and constraints. Can I just slap that into this PR since I need it here and merge it along?

core/src/main/java/org/jdbi/v3/core/argument/LoggableArgument.java
qualidafial: Bean binding goes through `NamedParameterFinder` which is a parallel lookup API. All the implementations in core implement `toString()` already.

qualidafial:
If the change is modest that's fine.

TheRealMarnes:
H2DatabaseRule creates a table `something (id identity primary key ...` when instantiated, PgDatabaseRule doesn't. Should that be there? It feels like a presumption too much, creating a table behind a test's back... I don't intend to put it in SqliteRule either.

stevenschlansker:
The `something` table is a relic of v2.  You're correct to not continue propagating it -- tests should explicitly opt in to the schema they need.

TheRealMarnes:
Uh... Is this supposed to happen?

image

I've never had IllegalAccessException before even when mucking around with reflection. A quick google says it's typically an issue in the library side of the code, needing to `setAccessible(true)` and such.

stevenschlansker:
The people who call it a "library issue" do not understand what they are talking about ;)
We cannot call `setAccessible` as it violates the Java security model, and will be disallowed in java 9 as well as under a SecurityManager.

Your test class must be non-private otherwise we cannot discover even public members.

TheRealMarnes:
> Your test class must be non-private

Force of (good) habit :) Learned something useful again tonight

TheRealMarnes:
Some testing shows logging works fine for the built-in Arguments so far, and the fix in SqlStatement succesfully wraps custom types (with custom argumentfactories), but if you `bind(String key, Argument arg)` there's no way to get to the original value, so that Argument needs to implement `toString`.

So LoggableArgument is both useful in jdbi's internal code and to users wishing to bind an Argument instance. I'd say we keep it.

TheRealMarnes:
Well it'll be an anonymous class for us and users then...

TheRealMarnes:
I'm thinking of doing a reflection check for `toString` on found `Argument`s in `SqlStatement#toArgument`, to avoid pointless wrapping. Yay or nay?

TheRealMarnes:
I think I should add a reflection check for existing toString impls on found Arguments in SqlStatement before wrapping it in a new Argument that forces toString to be the toString of the value object, so we don't override desired behavior. Is there a reflection library of choice already in jdbi or do can I go ahead and add [joor](https://github.com/jOOQ/jOOR)?

TheRealMarnes:
done

PR #1086, opened 2018-04-19
[on hold] neutrality plugin to override default config

### LINKED ISSUE #988: Builtin mapper for ZonedDateTime discards time zone information
There exists a built-in mapper for `ZonedDateTime` which looks like SQL's `TIMESTAMP WITH TIME ZONE` is directly supported, but the mapper does not do that.

I think exposing a mapper for zone-aware type that is zone-unaware is not the right thing to do.

TheRealMarnes:
#988

I've had a look through the code. "Fixing" the arguments and mappers is simple enough, but I think some other `Config` classes have the same issue, there are other OOTB impls being registered by default.

I think it's best to separate it all, make a "dumb" no-assumptions core with all the convenient impls provided separately in `Plugin`s, and pre-installed in some of the methods in `Jdbi` to keep compatibility. Separation of church and state or API and impl if you will, so we'd have a `PrimitivesPlugin`, `BuiltInCollectorsPlugin` and the like. They could all be composed into a single `JdbiDefaultsPlugin`.

Thoughts?

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/nothing/SupportNothingPlugin.java
+++ b/core/src/main/java/org/jdbi/v3/core/nothing/SupportNothingPlugin.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.nothing;
+
+import java.util.function.Function;
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.config.Configurable;
+import org.jdbi.v3.core.spi.JdbiPlugin;
+
+public class SupportNothingPlugin extends SupportNothingBase implements JdbiPlugin {
+    public SupportNothingPlugin() {
+        this((clazz) -> new UnsupportedOperationException("You've installed " + clazz.getSimpleName() + " — operation denied"));
+    }
+
+    public SupportNothingPlugin(Function<Class<?>, ? extends RuntimeException> provider) {
+        super(provider);
+    }
+
+    @Override
+    public void customizeJdbi(Jdbi jdbi) {
+        configure(jdbi);
+    }
+
+    @Override
+    public Handle customizeHandle(Handle handle) {
+        return configure(handle);
+    }
+
+    private <T extends Configurable<T>> T configure(T configurable) {
+        configurable.registerArgument(new SupportNothingArgumentFactory(super.provider));
+        configurable.registerRowMapper(new SupportNothingRowMapperFactory(super.provider));
+        configurable.registerColumnMapper(new SupportNothingColumnMapperFactory(super.provider));
+
+        return configurable;
+    }
+}
--- a/core/src/main/java/org/jdbi/v3/core/blank/BlankSlateArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/blank/BlankSlateArgumentFactory.java
@@ -11,22 +11,19 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jdbi.v3.core.nothing;
+package org.jdbi.v3.core.blank;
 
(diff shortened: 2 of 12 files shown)


qualidafial:
With just a cursory look so far: if you want a dumb core, there's a few more registries to blank out: `RowMappers`, `JdbiCollectors`, `SqlArrayTypes`.

I'm trying to keep an open mind, but I'm still not clear on the motivation for this PR.

TheRealMarnes:
Derp. Maybe I'm stupid but why doesn't this basic change work? I moved the Arguments and ColMappers default registrations to the JdbiDefaultsPlugin, which is installed whenever a Jdbi instance is made. Why won't it work?

TheRealMarnes:
Hmm, some of those test constructs are pretty difficult to adapt to this...

TheRealMarnes:
Hmmm... come to think of it, actually...  Couldn't I just write a JdbiPlugin that registers an ArgumentFactory and such that accepts every value but throws a UOE? That would effectively prevent any of the default impls being used. Then you can just add your own converters on top of that.

TheRealMarnes:
You're talking about the DefaultsPlugin idea? I kinda gave that up. It wouldn't work for mappers, only arguments, and doing it from inside Jdbi broke too much stuff in test setups and such that I couldn't figure out.

The clear() methods are the current idea and so far the most feasible and simplest yet very effective...

stevenschlansker:
I was referring to stacking an always-failing factory on top, to "hide" the default configuration.

TheRealMarnes:
Myeah, well that only works for Arguments, so it's not really a total solution...

qualidafial:
If your factory returns `Optional.empty()`, then yes it would fall through to the built-in defaults, but I'm pretty positive that a factory that always throws an exception would do the trick.

TheRealMarnes:
Ooooh I totally missed that there's a `ColumnMapperFactory`! I thought it was only `ArgumentFactory` and then `ColumnMapper`s matched by type. Yeah, that will probably work then, will see tomorrow

TheRealMarnes:
Worth including?

core/src/main/java/org/jdbi/v3/core/nothing/SupportNothingPlugin.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public void customizeJdbi(Jdbi jdbi) {
+        configure(jdbi);
+    }
+
+    @Override
+    public Handle customizeHandle(Handle handle) {
+        return configure(handle);
```
qualidafial: I don't think you want this in both `customizeHandle()` and `customizeJdbi()`. Otherwise anything you register on the `Jdbi` level will get clobbered on the handle.

qualidafial:
At this stage, I'm not certain there's a lot of demand for this from other folks. It would help to hear from others who have issues with the default mappers / arguments. Maybe our core registrations _are_ too opinionated, but I personally don't see it that way.

Some of Jdbi's goals to keep in mind:

* Minimize the work to get from zero to working code with little fuss. The default mappers and arguments serve that goal by providing a bunch of out of the box behavior. Convention over configuration is very much in the project's DNA, and this PR seems to work against that.
* Be flexible and let folks tailor Jdbi to work the way they like. The "last-registered wins" rule serves this goal, letting you clobber any built-in behavior with the way you want it to work.

Some feedback I have for the current state of the PR:

* I'd probably choose a name like `CleanSlatePlugin`, `BlankSlatePlugin`, or something along those lines. But that's just personal taste.
* It feels like introducing a base class for all factories is over-abstracting. Just hard code it in the plugin, e.g. as anonymous inner classes:

```java
public class `BlankSlatePlugin` implements JdbiPlugin {
  public void customizeJdbi(Jdbi jdbi) {
    jdbi.registerArgument((type, value, config) -> {
      throw new UnsupportedOperationException("No argument factory registered for '" + value + "' of type " + type);
    });

    jdbi.registerColumnMapper((ColumnMapperFactory) (type, config) -> {
      throw new NoSuchMapperException("No column mapper registered for type " + type);
    });

    jdbi.registerRowMapper((RowMapperFactory) (type, config) -> {
      throw new NoSuchMapperException("No row mapper registered for type " + type);
    });
  }
}
```

TheRealMarnes:
Regarding the jdbi goals: I would just add this so it's there if people need it and they won't need to figure out the way to do it themselves. It took me long enough, I'm sure others generally won't have an easier time... I wouldn't register it by default or anything, just include it as an optional thing, like the alternative TemplateEngines. Minimize the time it takes to get going, like you said.

The other remarks, yeah, that's all fair enough.

core/src/main/java/org/jdbi/v3/core/nothing/SupportNothingPlugin.java
TheRealMarnes: Isn't that the point? If you install the BlankSlatePlugin on a Jdbi or a Handle, that Jdbi/Handle should get "cleaned".

PR #1088, opened 2018-04-20
Through experimental testing, it's clear that allowing StringTemplate to load templates concurrently is what's causing #1052

### LINKED ISSUE #1052: Random errors with redefinition of template and no such template
Update:
Found a way to reproduce it consistently. See https://github.com/jarlah/jdbi-stringtemplate4-racecondition-bug

Hi

I am trying to use Jdbi with stringtemplate4. I have created two template files that imports another template file relatively. e.g. import "../whatever/WhateverDao.sql.stg".

The problem occurs randomly, and I think is also solved randomly, that the methods in WhateverDao.sql.stg is not found or being overwritten (nulled out) by thread synchronization issues.

I am using:
spring boot 2.0.0.RELEASE
Jdbi 3.1.0

I have tried to create a JdbiFactoryBean and a normal Jdbi creation. 

I get red errors in console when the random error occurs saying that the template method is not found. 
Below I see other red errors saying the methods have been redefined.
I suppose those two errors are linked.

But I also think that there is a major issue with synchronization.

Does jdbi support import statements in template group files? If not then templates are uselees?

stevenschlansker:
Fixes #1052

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -6,7 +6,7 @@
   - add a TemplateEngine based on java.text.MessageFormat
   - allow absence of 'value' on @MaxRows and validate its (lack of) use on SqlObjects
   - fix SqlObject getting confused about result types due to bridge methods
-  - StringTemplate no longer loads templates concurrently to prevent races
+  - StringTemplate no longer shares template groups across threads, to work around concurrency issues
 
 3.1.1
   - Improve IBM JDK compatibility with default methods
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -6,6 +6,7 @@
   - add a TemplateEngine based on java.text.MessageFormat
   - allow absence of 'value' on @MaxRows and validate its (lack of) use on SqlObjects
   - fix SqlObject getting confused about result types due to bridge methods
+  - StringTemplate no longer loads templates concurrently to prevent races
 
 3.1.1
   - Improve IBM JDK compatibility with default methods
--- a/stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateSqlLocator.java
+++ b/stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateSqlLocator.java
@@ -23,6 +23,7 @@ import java.util.concurrent.TimeUnit;
 
 import net.jodah.expiringmap.ExpirationPolicy;
 import net.jodah.expiringmap.ExpiringMap;
+
 import org.stringtemplate.v4.ST;
 import org.stringtemplate.v4.STGroup;
 import org.stringtemplate.v4.STGroupFile;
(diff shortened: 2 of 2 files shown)


stevenschlansker:
@jarlah, can you verify this fix?

stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateSqlLocator.java
```diff
@@ -130,7 +131,13 @@ public static STGroup findStringTemplateGroup(String path) {
      * @return the loaded StringTemplateGroup.
      */
     public static STGroup findStringTemplateGroup(ClassLoader classLoader, String path) {
-        return CACHE.computeIfAbsent(path, p -> readStringTemplateGroup(classLoader, path));
+        final STGroup cached = CACHE.get(path);
+        if (cached != null) {
+            return cached;
+        }
+        synchronized (StringTemplateSqlLocator.class) {
```
qualidafial: Let's synchronize on a private object rather than the class to be safe.

jarlah:
I will try to verify the fix.

jarlah:
@stevenschlansker doesnt help. Ref https://github.com/antlr/stringtemplate4/issues/61 the real problem is the lazy loading of the template methods. The caching doesnt help at all in this case as i see it. Both dao calls loads the template with getInstanceOf("getCar") after loading it from or setting it in cache. I tried to remove caching all together for fun and test passes. Thats of course just to rule out other bugs. I suggest trying to force load all template methods within the group file at the time of caching it. I tried it. With hardcoded getInstanceOf("getCar") and it worked

jarlah:
But really. It's very easy to test this for me and to try out different options. So ill try to find a solution. If i find one I'll post it here.

jarlah:
see https://github.com/jdbi/jdbi/pull/1090

jarlah:
with my fix, the tests passes every time.

stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateSqlLocator.java
jarlah: #1090 i've used a privat static final Object in my PR.

stevenschlansker:
> To be absolutely safe, create one instance of your STGroup for each thread. In Java, you could use ThreadLocal<T> to initialize your group on-demand (by overriding initialValue() to load the group).

I wonder if we should just take this advice.  That issue is ... eye opening, and makes me regret our use of ST a little bit :(

RELEASE_NOTES
```diff
@@ -6,6 +6,7 @@
   - add a TemplateEngine based on java.text.MessageFormat
   - allow absence of 'value' on @MaxRows and validate its (lack of) use on SqlObjects
   - fix SqlObject getting confused about result types due to bridge methods
+  - StringTemplate no longer loads templates concurrently to prevent races
```
qualidafial: With the change I just merged into this branch, this should probably read "StringTemplate no longer shares template group across threads, to avoid concurrency problems withing STGroup" or something like that.

qualidafial:
@stevenschlansker I just merged @jarlah's PR into this branch, which uses `ThreadLocal` as recommended by the ST team.

@jarlah has confirmed this fixes his template loading problem.

Were there any tests you wanted to add?

stevenschlansker:
I think this is fine.  Writing tests that `ThreadLocal` fixes a race behavior is super nasty.  I will squish these commits as right now we have 12 commits changing 3 lines :P

jarlah:
will you release a 3.1.2 version for this? eager to remove a race condition test on my side...

qualidafial:
@stevenschlansker did you mean to close the PR instead of merging?

stevenschlansker:
Yes: https://github.com/jdbi/jdbi/commit/b6627527b4a5b52f88b06dc7b2c5415f5af78ff9

qualidafial:
Roger that.

PR #1091, opened 2018-04-21
Test registered DAO with generic parameter and returned collection

EthanLozano:
This is a test that exposes a reflection and/or generics error in v3.1.1. The head of master seems to have fixed the problem. But, I figure this test could be useful to make sure that generics and polymorphism work in future releases.

Running this test on v3.1.1 on jdk1.8.0_172 results in the following error:
```
  <testcase name="testRegisterGenericRowMapperAnnotationWorks" classname="org.jdbi.v3.sqlobject.TestRegisteredGenericReturnAndParam" time="0.218">
    <error message="No mapper registered for type interface java.util.List" type="org.jdbi.v3.core.mapper.NoSuchMapperException">org.jdbi.v3.core.mapper.NoSuchMapperException: No mapper registered for type interface java.util.List
	at org.jdbi.v3.sqlobject.TestRegisteredGenericReturnAndParam.testFoodToppingRestrictions(TestRegisteredGenericReturnAndParam.java:50)
	at org.jdbi.v3.sqlobject.TestRegisteredGenericReturnAndParam.testRegisterGenericRowMapperAnnotationWorks(TestRegisteredGenericReturnAndParam.java:40)
</error>
```
The problem seems to come up because `SqlObjectFactory.methodHandlersFor` registers two versions of the same method (one with an Object param and the second with a String param):
```
public default java.util.List TestRegisteredGenericReturnAndParam$Gyro.getToppings(int,java.lang.Object)
```
and
```
public abstract java.util.List TestRegisteredGenericReturnAndParam$Gyro.getToppings(int,java.lang.String)
```
Where only the latter has a generic `returnType` in `ResultReturner.forMethod`, so `CollectedResultReturner` just returns the original collection `java.util.List` for the former.

### DIFF
--- a/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestRegisteredGenericReturnAndParam.java
+++ b/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestRegisteredGenericReturnAndParam.java
@@ -1,0 +1,144 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


qualidafial:
Haha, love the food-themed test cases

PR #1092, opened 2018-04-22
DefineStatementLexer: don't get tripped up on partial define '<identifier'

stevenschlansker:


stevenschlansker:
@a1anw2, can you verify that this fixes your issues?  I did not want to check your query into our repo.

stevenschlansker:
Let us know if the problem isn't fixed for you @a1anw2

PR #1093, opened 2018-04-22
Add a test to show jsonb query '?' can be escaped as '??' and work

### LINKED ISSUE #1069: Jdbi cannot handle some postgres json(b) operators in query
I have just noticed that the following statement fails, due to complaints that positional and named parameters cannot be matched.

But in postgres with json it is valid syntax (querying whether the property exists on the object)

`SELECT * FROM foo WHERE some_col ? 'barProp' AND some_other_col = :param`

where `?` is a query operator. see https://www.postgresql.org/docs/10/static/functions-json.html

is there/should there be any way to explicitly disable positional binding on `SqlStatement` or `Binding` or make this `Configurable` for the current statement?

stevenschlansker:
Fixes #1069

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -3678,6 +3678,10 @@ It also configures SQL array type support for `int`, `long`, `float`, `double`,
 See the link:{jdbidocs}/postgres/package-summary.html[javadoc^] for an
 exhaustive list.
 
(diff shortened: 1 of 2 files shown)


EthanLozano:
The bug appears when including a test with a sql object and binded input: https://github.com/jdbi/jdbi/pull/1094

PR #1094, opened 2018-04-22
TestJsonOperator with SqlObject and binded input

EthanLozano:
Exposes the following bug with sql parsing:
```
org.jdbi.v3.core.statement.UnableToExecuteStatementException: Cannot mix named and positional parameters in a SQL statement: [?, ?, property]
```

### DIFF
--- a/postgres/src/test/java/org/jdbi/v3/postgres/TestJsonOperator.java
+++ b/postgres/src/test/java/org/jdbi/v3/postgres/TestJsonOperator.java
@@ -31,4 +31,14 @@ public class TestJsonOperator {
                 .findOnly())
             .isEqualTo(true);
     }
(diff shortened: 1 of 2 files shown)


qualidafial:
Working on a fix that incorporates these test cases.

qualidafial:
Thank you for the test cases! I've opened #1104 which incorporates your unit tests, and fixes the bug.

Would you please check out that branch and give it a spin?

qualidafial:
#1104 is merged, which includes your tests. Thanks!

EthanLozano:
@qualidafial I'm glad my test cases came in handy!  I gave the this commit a spin and it works against my project's code. Thanks for fixing the issue!

PR #1099, opened 2018-04-24
StringSubstitutor TemplateEngine

TheRealMarnes:
Thought this could be simple and general enough to be worth including. Feel free to call it too niche to be worth bothering with though.

StringSubstitutor could be used to implement the default \<foo\> TemplateEngine too, btw. I mean, this class is basically the generified version of it, with \< and \> set instead of ${ and }.

image

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -83,6 +83,11 @@
 
         <dependency>
             <groupId>org.apache.commons</groupId>
+            <artifactId>commons-text</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.commons</groupId>
             <artifactId>commons-jexl</artifactId>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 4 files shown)


qualidafial:
Ha, love the image.

I'm happy to support more template engines, but I'd prefer they be added as independent modules similar to what we're doing with StringTemplate, instead of adding them as a dependency of core.

TheRealMarnes:
Hm, I wasn't expecting this, hah. I actually deleted this stuff in hindsight, thinking it was just too silly since the whole thing can basically be shortened to `setTemplateEngine((sql, ctx) -> StringSubstitutor.replace(sql, ctx.getAttributes()))`.

But yeah, maybe as a separate artifact, like stringtemplate4 is.

qualidafial:
I've also thought about supporting FreeMarker and/or Mustache templates too, FWIW

stevenschlansker:
For single classes like this, you can also just include the class and mark the dependency as `<optional>`.  I don't love it, but I also really hate one-class-modules.

TheRealMarnes:
Yeah I'm not fond of making a module for every single thing either...

PR #1100, opened 2018-04-24
[done?] checkstyle part 2

TheRealMarnes:
Part 2 of checkstyling


TheRealMarnes:
First commits: remarks from last time

TheRealMarnes:
It seems none of checkstyle, pmd, and findbugs have a rule to detect methods that declare checked exceptions that are never thrown :(

TheRealMarnes:
After merge conflicts later today, I think I've done all the major work I can do for now in checkstyle. What remains is a few individual checks for which code will need to be edited (structural changes), which I'd like to do as separate PRs.

Could this be reviewed and merged ASAP to avoid large numbers of conflicts please? :)

TheRealMarnes:
I think I've done all the major work I can do for now in checkstyle. What remains is a few individual checks for which code will need to be edited (structural changes), which I'd like to do as separate PRs.

@stevenschlansker @qualidafial Could this be reviewed and merged ASAP to avoid large numbers of conflicts please? :) I think the rules I've added with this PR are all quite fair, and generally recommended anyway.

TheRealMarnes:
There's a todo left in the code regarding empty catch blocks, must be a plugin version issue or something

http://checkstyle.sourceforge.net/config_whitespace.html#WhitespaceAround

TheRealMarnes:
Ugh, merges... such a pain in the neck

TheRealMarnes:
Ok, I can't fix the checkstyle version problem. That org.basepom parent pom with its several levels of inheritance and bazillion points of complexity just strings me up.

We probably want to enable the WhiteSpaceAround#allowEmptyCatches property, but it's checksty;e 3.0+, and apparently that's not the version configured from above. I can't figure out how to fix it properly... Someone else, please?

qualidafial:
There should be a checkstyle version property declared in basepom, I'll see if I can find it. To upgrade you just override the property in the jdbi parent pom

qualidafial:
@TheRealMarnes See https://github.com/basepom/basepom/blob/master/foundation/pom.xml#L163

TheRealMarnes:
Problem turns out to be some caching or I don't know what... I changed the config like you said (I was already doing the same thing yesterday) and it still didn't work, but then after screwing around for a minute without making changes to any project files, the next exact same build suddenly behaved differently and worked as expected. Maven's builds don't seem very reproducible when changes to plugins are involved.

TheRealMarnes:
Ok, the empty catch block rule is fixed now :) I configured it for now to demand ignored exceptions be named _ignore(d)_ or _expected_ and for a comment to be optional. I think this is better than the other way around, naming it _e_ and demanding a comment, because a variable named _ignored_ or _expected_ is easier found in searches than a catch block containing a comment. Just my 2 cents, feel free to disagree and I'll change it. That goes for all the settings I've put in here btw. I'm just making initial suggestions.

@qualidafial @stevenschlansker I think this is ready to be reviewed and merged. Especially the fix for
```java
void foo() {
}
```
is something I think will be welcomed.

core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -78,8 +77,8 @@ void setConfig(ConfigRegistry config) {
         this.config.set(config);
     }
 
-    void setConfigThreadLocal(ThreadLocal<ConfigRegistry> config) {
-        this.config = config;
+    void setConfigThreadLocal(ThreadLocal<ConfigRegistry> conf) {
+        this.config = conf;
```
stevenschlansker: When you make changes like this, `this.` becomes redundant.  Should we clean those up as well if we forbid name shadowing?

docs/src/test/java/jdbi/doc/CallTest.java
```diff
@@ -24,11 +23,16 @@ public void testCall() {
 
         // tag::invokeProcedure[]
         OutParameters result = handle
-                .createCall("{:sum = call add(:a, :b)}") // <1>
-                .bind("a", 13) // <2>
-                .bind("b", 9) // <2>
-                .registerOutParameter("sum", Types.INTEGER) // <3> <4>
-                .invoke(); // <5>
+            // <1>
```
stevenschlansker: Does making these changes break the asciidoc?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
```diff
@@ -106,15 +106,34 @@ private static Object checkResult(Object result, Type type) {
         return result;
     }
 
+    static class VoidReturner extends ResultReturner {
```
stevenschlansker: why did this move up suddenly?

src/build/checkstyle-suppress.xml
```diff
@@ -17,9 +17,5 @@
     "http://www.puppycrawl.com/dtds/suppressions_1_1.dtd">
 
 <suppressions>
-    <suppress checks=".*" files="SqlScriptLexer\.java$"/>
-    <suppress checks=".*" files="DefineStatementLexer\.java$"/>
-    <suppress checks=".*" files="ColonStatementLexer\.java$"/>
-    <suppress checks=".*" files="FormatterStatementLexer\.java$"/>
-    <suppress checks=".*" files="HashStatementLexer\.java$"/>
+    <suppress checks="InnerTypeLast|VariableDeclarationUsageDistance|NoFinalizer" files="[A-z]*Test[A-z]*\.java$"/>
```
stevenschlansker: this might miss any tests with `0-9` in the name

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
TheRealMarnes: I re-ordered them to be in the same order as their usage in the methods higher up. I think that was a thing last time, some classes were in order of usage and got reordered to be out-of-order, which wasn't appreciated...

TheRealMarnes:
I thought individual commits per rule was what you wanted, so you could see the changes each rule brings :s

TheRealMarnes:
Local build fails for me atm because of some problem with the kotlin module, so I can't build the doc module either. I looked around online but I can't find anything about that `<1>` related to asciidoc(tor) either... I'm not sure what to look for right now and can't do much until later today. Could you check if the comment above-lining broke the doc? You probably know exactly what to look for and where, unlike me...

stevenschlansker:
Oh, sorry, I didn't even notice that they were commit per rule!  That's fine, I just only saw the mega-diff.  I should have looked more closely!

qualidafial:
I'm not a fan of the rule requiring parameter names to be distinct from field names. Parameter naming is an API design decision first and foremost, and style is a distant second or third consideration here.

If we're worried about bugs, then we should ensure the pmd self-assignment check is turned on.

TheRealMarnes:
@qualidafial sounds fair

TheRealMarnes:
Now that you've mentioned using pmd for it instead, shall I go ahead and slap on a pmd ruleset? I have one ready to go from a private project with every rule enabled, minus the ones that just aren't reasonable or very useful.

TheRealMarnes:
I'll do the naive javatime plugin and maybe a rummage through the poms next btw...

freemarker/src/test/java/org/jdbi/v3/freemarker/BindBeanListTest.java
```diff
@@ -40,11 +39,11 @@
     private static List<Something> expectedSomethings;
 
     @ClassRule
-    public static final H2DatabaseRule dbRule = new H2DatabaseRule();
+    public static final H2DatabaseRule DB_RULE = new H2DatabaseRule();
```
qualidafial: This would probably be better to change to an instance rule instead of a class rule.

freemarker/src/test/java/org/jdbi/v3/freemarker/BindListTest.java
```diff
@@ -42,11 +41,11 @@
     private static List<Something> expectedSomethings;
 
     @ClassRule
-    public static final H2DatabaseRule dbRule = new H2DatabaseRule();
+    public static final H2DatabaseRule DB_RULE = new H2DatabaseRule();
```
qualidafial: ditto, etc

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBeanMapper.java
```diff
@@ -225,22 +224,22 @@ public String toString() {
...
-        @SqlQuery("select f.id f_id, f.name f_name, " +
-                "d.id d_id, d.name d_name, d.contents d_contents " +
-                "from folders f left join documents d " +
-                "on f.id = d.folder_id " +
-                "where f.id = :folderId")
+        @SqlQuery("select f.id f_id, f.name f_name, "
+            + "d.id d_id, d.name d_name, d.contents d_contents "
+            + "from folders f left join documents d "
+            + "on f.id = d.folder_id "
+            + "where f.id = :folderId")
```
qualidafial: Inconsistent indentation with the next method

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestDocumentation.java
```diff
@@ -77,10 +75,11 @@ public void testFiveMinuteSqlObjectExample() throws Exception {
         });
     }
 
-
     @Test
     public void testObtainHandleViaOpen() throws Exception {
-        try (Handle handle = dbRule.getJdbi().open()) {}
+        try (Handle handle = dbRule.getJdbi().open()) {
+            // TODO?
```
qualidafial: the test just verifies that the handle can be opened, so no TODO

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestRegisteredMappersWork.java
```diff
@@ -90,7 +89,8 @@ public void setColor(String color) {
         default Optional<Bean> findByNameDefaultMethod(String name) {
             return getHandle().createQuery("select name, color from beans where name = :name")
                     .bind("name", name)
-                    .mapTo(Bean.class) // uses annotation-registered mapper
+                // uses annotation-registered mapper
+                    .mapTo(Bean.class)
```
qualidafial: inconsistent indentation

src/build/checkstyle.xml
```diff
@@ -1,105 +1,136 @@
...
+        <module name="NoClone"/>
+        <module name="NoFinalizer"/>
+        <module name="OneStatementPerLine"/>
+        <module name="OneTopLevelClass"/>
+        <module name="OuterTypeFilename"/>
+        <module name="ParameterAssignment"/>
+        <module name="StringLiteralEquality"/>
+        <module name="SuperClone"/>
+        <module name="SuperFinalize"/>
+        <module name="TrailingComment"/>
```
qualidafial: I'm on the fence with this one, as it disrupts the call-out markers in the documentation code.

core/src/test/java/org/jdbi/v3/core/TestTooManyCursors.java
```diff
@@ -120,11 +119,11 @@ public static Statement newInstance(Statement stmt, ConnectionInvocationHandler
             }
 
             return (Statement) Proxy.newProxyInstance(stmt.getClass().getClassLoader(),
-                                                      interfaces.toArray(new Class[interfaces.size()]),
+                                                      interfaces.toArray(new Class[0]),
```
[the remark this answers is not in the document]
TheRealMarnes: this was a recommendation from intellij based on some josh bloch-esque jvm knowledge involving performance and redundancy. I think it was that the created array is not the array that's actually used, so giving it any size besides 0 is pointless.

docs/src/test/java/jdbi/doc/CallTest.java
TheRealMarnes: will undo

TheRealMarnes:
Do you want me to close this PR and make a bunch of smaller PRs for just a handful of changes each time instead? Going back now to undo the unwanted changes, or cherrypicking the good changes onto the current master is going to be merge hell, I think. I can probably just do it over again with regard for the feedback in a quarter the time.

src/build/checkstyle.xml
stevenschlansker: This was also one of my least favorites

stevenschlansker:
Nah, at this point, let's not do it all over.  Let's just address the comments and feedback on this PR and in the future try to split it up further.

TheRealMarnes:
I'm not sure how/why/when those method headers got reformatted to multiple lines tbh... might be my intellij style settings

core/src/test/java/org/jdbi/v3/core/TestTooManyCursors.java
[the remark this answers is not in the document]
TheRealMarnes: image

core/src/test/java/org/jdbi/v3/core/statement/TestCallable.java
```diff
@@ -97,19 +88,19 @@ public void testWithNullReturn() throws Exception {
...
         assertThat(out).isNull();
     }
 
-    public static void testProcedure(String in, String[] out) {
-        out = new String[1];
-        out[0] = in;
-    }
+    // used by the db in this test
+    @SuppressWarnings("unused")
+    public static void testProcedure(String in, String[] out) {}
```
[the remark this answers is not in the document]
TheRealMarnes: I don't see how they possibly can... The method assigns a new array to its parameter, and puts a string in this new array.  It doesn't actually change the array object it was given. It's completely moot, unless I'm missing some magic :/

TheRealMarnes:
whoosh...

TheRealMarnes:
Whew, ok, done. I made lots of small commits for clarity, I can squash them down if you wanna keep repo history pollution a little lower.

TheRealMarnes:
Oh yeah, one last thing, I was gonna sort the checkstyle file alphabetically for easier review/comparison

TheRealMarnes:
Once a proper pmd ruleset is in place, some of the checkstyle rules that actually handle code content can be removed, like forbidding parameter assignment...

core/src/test/java/org/jdbi/v3/core/TestTooManyCursors.java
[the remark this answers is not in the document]
qualidafial: Ok.

core/src/test/java/org/jdbi/v3/core/statement/TestCallable.java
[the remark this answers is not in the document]
qualidafial: You're right--I had misread the old code.

TheRealMarnes:
Guys, can we please get this one merged? I tackled the stuff you asked. The longer it stays open, the more conflicts we'll keep getting with master. I've got pmd ready to go, I'll make smaller PRs for that one as asked. I don't expect pmd to give as much trouble because checkstyle already forced a lot of cleanup. I'll remove non-style checks from checkstyle too so each tool focuses on its essence.

stevenschlansker:
@qualidafial you've got the blocking review, anything else to address before merge?

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestDocumentation.java
```diff
@@ -77,10 +76,9 @@ public void testFiveMinuteSqlObjectExample() throws Exception {
         });
     }
 
-
     @Test
     public void testObtainHandleViaOpen() throws Exception {
-        try (Handle handle = dbRule.getJdbi().open()) {}
+        assertThatCode(dbRule.getJdbi()::open).doesNotThrowAnyException();
```
qualidafial: Leaks connection.

TheRealMarnes: True, but doesn't the process end and kill the db anyway? Writing a fix anyway, just interested in the mechanics behind memdbs in tests

TheRealMarnes: Isn't the sharedHandle in H2DatabaseRule et al also a leak then btw?

qualidafial: The test rule closes the connection in its `after()` method.

TheRealMarnes: Well it closes the connection it gets from the handle. Closing the handle itself isn't important then?

qualidafial:
You're right that it's probably not super critical to fix a connection leak in a test. However lots of people (I know a few) use the tests as a model to understand how the library works, so I think it's important to set a good model for how the library _should_ be used, even in tests.

Thank you for sticking it out and being patient with all these changes we've asked for. Great work!

Will merge once the build completes successfully.

qualidafial:
LEEROOOOOOOY JENKIIIIIIIIIINS

TheRealMarnes:
Oh I'm not saying it's not important to write tests "right" at all :) I've seen bad habits sneak into test code and then from there to production code often enough... If your tests are written badly they're not really worth anything as evaluators of other code either.

I'm a little itchy about those databaserules now though. I kinda realized the test was technically a leak after my change, but looking at DatabaseRules, I thought it would be harmless since those keep an open Handle too and only close the underlying connection. Didn't realize opening a handle also immediately opens a connection that needs to be closed though, I thought it was lazy-opened.

How about we change databaserules to close the handle entirely instead of only the connection inside it? It removes the dubiousness of leaving an open handle lying around.

PR #1102, opened 2018-04-26
[done] sortpom

TheRealMarnes:
Automatically tidy, ordered files > allowing arbitrary file structure

Had to add a maven module, and so I noticed...

### DIFF
--- a/guava/pom.xml
+++ b/guava/pom.xml
@@ -29,16 +29,16 @@
             <groupId>org.jdbi</groupId>
             <artifactId>jdbi3-core</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.jdbi</groupId>
             <artifactId>jdbi3-core</artifactId>
             <classifier>tests</classifier>
             <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
         </dependency>
 
         <dependency>
--- a/guava/pom.xml
+++ b/guava/pom.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <!--
 ~   Licensed under the Apache License, Version 2.0 (the "License");
 ~   you may not use this file except in compliance with the License.
@@ -16,8 +16,8 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-        <groupId>org.jdbi</groupId>
         <artifactId>jdbi3-parent</artifactId>
+        <groupId>org.jdbi</groupId>
         <version>3.2.0-SNAPSHOT</version>
     </parent>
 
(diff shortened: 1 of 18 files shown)


stevenschlansker:
Hm, this makes an awful lot of changes.  And some of the dependencies were grouped together intentionally.

I'm not a fan of how much more whitespace there is.  Can we do 2 spaces for xml?

TheRealMarnes:
> some of the dependencies were grouped together intentionally

I understand, but a lot also get added haphazardly or chronologically, which is of no value... You can't really have half order, it's either ordered or it isn't...

I can remove the sortDependencies property but then we live with having dependencies just thrown together in whichever order. Your and @qualidafial 's choice to make, I guess.

.editorconfig
```diff
@@ -13,3 +13,6 @@ trim_trailing_whitespace = false
 
 [*.yml]
 indent_size = 2
+
+[pom.xml]
+indent_size = 2
```
qualidafial: I'm not sure what is the benefit of changing this? Most of our POMs use 4-space tabs.

guava/pom.xml
```diff
@@ -13,65 +13,65 @@
...
-            <artifactId>guava</artifactId>
-        </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jdbi</groupId>
+      <artifactId>jdbi3-core</artifactId>
+    </dependency>
```
qualidafial: Putting Guava before Jdbi? This feels very naive to me.

The order of dependencies in Maven is a factor in dependency resolution. Reordering them can change the outcome of that resolution.

TheRealMarnes: ask @stevenschlansker ¯\\\_(ツ)\_/¯ https://github.com/jdbi/jdbi/pull/1102#issuecomment-385084836

.editorconfig
TheRealMarnes: ask @stevenschlansker ¯\\\_(ツ)\_/¯ #1102 (comment)

guava/pom.xml
TheRealMarnes: Hm, did not know that. Well I've disabled sortDependencies anyway as @stevenschlansker asked, so everybody happy?

qualidafial: At the least, dependencies need to go back in the order they were in before.

TheRealMarnes:
Ok, so, current config is indent=4 and deps are not sorted

stevenschlansker:
we can leave it as 4, that's fine... I thought I saw one going to 8 and reacted negatively :)

qualidafial:
@stevenschlansker Any other feedback or concerns before merging?

TheRealMarnes:
Ok, so, I took a more attentive look and changed the sortpom config a fair bit:

- removed rules that were set to defaults
- removed the "dangerous" rules that can affect builds

It's pretty minimal now, and the changes to existing poms are also quite minimal. Indentation, some header elements, that's basically it. I think it's good to go and risk-free now.

TheRealMarnes:
Maybe we can sort dependencies by scope at least though? It's always nice to have `test` dependencies at the bottom.

TheRealMarnes:
done

qualidafial:
Thanks!

qualidafial:
Hmm, GitHub is saying there are still merge conflicts. Possibly with your other PR that just merged?

TheRealMarnes:
merged

Issue #1101, opened 2018-04-25
Discourage and deprecate StringTemplate4 support


alexcase52:
I propose to review support for this framework.

It doesn't seem to be normally supported. No fix releases since 2014 and growing number of issues in StringTemplate itself and issues and confusion around its integration with JDBI. E.g. issues with loading template files in multi-threaded envs or env with non-trivial-classloading. No responses on issues, even no response on the [issue](https://github.com/antlr/stringtemplate4/issues/196) about project activity since Nov 2017.

The minimum, I think, is to put a notice in Docs.

Thanks

qualidafial:
Let's start with a notice in the docs that the StringTemplate project is inactive.

I'll let the other project members weigh in on the question of deprecating Jdbi's ST integration. Personally I think we should keep supporting it for at least a little while. We only just started supporting ST4 in the first place.

jarlah:
I think we need an alternative. Not having a way to create complex templates for sql is a no go. I need to be able to share sql bits and parts between different templates, and to be able to factor out common stuff into template methods. For me, removing stringtemplate is making me look for another alternative framework. If not I would be forced to make sql files with a lot of duplication.

stevenschlansker:
Let's add a `NOTICE` now.  We can deprecate it once we have a viable replacement.

Does anyone have favorite templating engines that would fit nicely into jdbi?  I've used `mustache` a bit and it was fine, but I'm far from an expert...

qualidafial:
Thanks @jarlah, that was kind of what I expected.

It's unfortunate StringTemplate is being neglected upstream, but as long as people are using it and it works, there's no reason for us to withdraw support.

We've been pretty happy with FreeMarker where I work. Mustache is a good option too.

TheRealMarnes:
How would additional templateEngines be bundled in jdbi? They all bring in external dependencies we don't want in `core`, and neither making a 1-class module nor setting an <optional> mvn dep is desirable... 

IIRC stringtemplate got its own mvn module though, so maybe we should just keep doing that? It sounds like less of a hacky hassle for users than optional dependencies anyway.

I'd be interested in contributing a StringSubstitutor engine (if I haven't already), and Mustache seems pretty fun to do too.

[...]

PR #1105, opened 2018-04-27
[done] StringSubstitutor integration

TheRealMarnes:
#1101 

So I've got a TemplateEngine largely worked out.

I came across `SqlParser`. It deals with string rendering so I'm wondering if I can't do something with that too.

Is an SqlParser allowed to use the `StatementContext` to replace variables in the string based on the defined attributes (`define("key, "value")`) or is it supposed to work independently of the context?


stringsubstitutor/src/main/java/org/jdbi/v3/stringsubstitutor/StringSubstitutorTemplateEngine.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.apache.commons.text.StringSubstitutor;
+import org.jdbi.v3.core.statement.StatementContext;
+import org.jdbi.v3.core.statement.TemplateEngine;
+
+@FunctionalInterface
+public interface StringSubstitutorTemplateEngine extends TemplateEngine {
+    StringSubstitutorTemplateEngine DEFAULTS = substitutor -> {};
+
+    void customize(StringSubstitutor substitutor);
```
qualidafial: What do you anticipate is the use case for this method?

I feel like this would be clearer if this interface were a class instead, and the customize function was an optional `Consumer<StringSubstitutor>` constructor parameter.

stringsubstitutor/src/test/java/org/jdbi/v3/stringsubstitutor/TestStringSubstitutorSqlParser.java
```diff
@@ -0,0 +1,42 @@
...
+import org.jdbi.v3.core.statement.StatementContext;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(MockitoJUnitRunner.class)
```
qualidafial: Prefer `@Rule MockitoRule mockitoRule = MockitoJUnit.rule()` over the test runner.

stringsubstitutor/src/main/java/org/jdbi/v3/stringsubstitutor/StringSubstitutorSqlParser.java
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.stringsubstitutor;
+
+import org.jdbi.v3.core.statement.ParsedSql;
+import org.jdbi.v3.core.statement.SqlParser;
+import org.jdbi.v3.core.statement.StatementContext;
+
+public class StringSubstitutorSqlParser implements SqlParser {
```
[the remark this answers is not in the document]
TheRealMarnes: Well that's what I'm gonna find out :) Just needed to understand SqlParser clearly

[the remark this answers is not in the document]
qualidafial: Sure. Did my earlier answers clear things up? Any other questions?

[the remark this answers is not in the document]
TheRealMarnes: Perfectly clear. Thanks for other remarks

stringsubstitutor/src/test/java/org/jdbi/v3/stringsubstitutor/TestStringSubstitutorSqlParser.java
TheRealMarnes: Cool, I didn't know you could do that. Makes tests nicely extensible, lets you combine mockito with other stuff that usually needs a Runner. Thanks for telling

qualidafial:
Don't forget to add new module to bom

TheRealMarnes:
I can't really seem to find any additional uses for Commons Text. Let's add the module as-is, maybe in the future we'll find something else, and if not, it still makes a nice TemplateEngine.

stringsubstitutor/src/main/java/org/jdbi/v3/stringsubstitutor/StringSubstitutorTemplateEngine.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.apache.commons.text.StringSubstitutor;
+import org.jdbi.v3.core.statement.StatementContext;
+import org.jdbi.v3.core.statement.TemplateEngine;
+
+/**
+ * Register an instance of this interface ({@link org.jdbi.v3.core.statement.SqlStatements#setTemplateEngine}) to use an Apache Commons Text {@link StringSubstitutor} as a {@link TemplateEngine}. This lets you use any pair of strings as variable delimiters, enabling the use of syntax like <pre>select * from ${foo}</pre>, <pre>select * from &lt;foo&gt;</pre>, <pre>select * from %foo%</pre>, etc.
+ */
+@FunctionalInterface
+public interface StringSubstitutorTemplateEngine extends TemplateEngine {
```
qualidafial: I'm still hung up on this being an interface. @stevenschlansker am I being too picky?

stringsubstitutor/src/main/java/org/jdbi/v3/stringsubstitutor/StringSubstitutorTemplateEngineFactory.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+package org.jdbi.v3.stringsubstitutor;
+
+/**
+ * Convenient semantic factory for common {@link StringSubstitutorTemplateEngine} instances.
+ */
+public class StringSubstitutorTemplateEngineFactory {
+    private StringSubstitutorTemplateEngineFactory() {}
+
+    public static StringSubstitutorTemplateEngine between(char prefix, char suffix) {
```
qualidafial: Let's fold these factory methods into `StringSubstitutorTemplateEngine`, so the usage looks like:

```java
jdbi.setTemplateEngine(StringSubstitutorTemplateEngine.between('<','>'));
```

TheRealMarnes:
You sure? Feels like interface pollution to me tbh...

stringsubstitutor/src/main/java/org/jdbi/v3/stringsubstitutor/StringSubstitutorTemplateEngineFactory.java
TheRealMarnes: You sure? Feels like interface pollution to me tbh...

TheRealMarnes:
merged

stringsubstitutor/src/main/java/org/jdbi/v3/stringsubstitutor/StringSubstitutorTemplateEngine.java
stevenschlansker: Yeah, this doesn't feel like an interface to me either.  How about taking a `Consumer<StringSubstitutor>` as a constructor argument instead?  Then you decouple the desire for lambdas from the choice of `class` vs `interface` for this piece of api.

stringsubstitutor/src/main/java/org/jdbi/v3/stringsubstitutor/StringSubstitutorTemplateEngineFactory.java
stevenschlansker: I tend to agree -- if the factory had any state or configurability, then a separate class is fine.  But as is these are simple static factory methods that should probably go on the class itself.

```java
private StringSubTemplateEngine(Consumer<StringSubstitutor> configurer) { ... }
public static SSTE between(prefix, suffix) {
    return new SSTE(sub -> {
        sub.setVar...
    });
}
```

stringsubstitutor/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>jdbi3-parent</artifactId>
+        <groupId>org.jdbi</groupId>
+        <version>3.2.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>jdbi3-stringsubstitutor</artifactId>
```
stevenschlansker: should this be `jdbi3-commons-text` or is there a reason to name it `stringsubstitutor` instead?  The former makes me think Apache, the latter I'd never heard until I looked at the API of Commons Text.

TheRealMarnes: Oh yeah, you're right, commons-text makes more sense

qualidafial:
Looks great! I'll merge this now to avoid any further merge conflicts, but we'll need a follow up PR if you have the time: recent PRs have added a handful of new Maven modules, and we need to make sure that each one is included in the bom, and as a dependency in the docs module (so they're included in javadoc generated for the web site).

TheRealMarnes:
@qualidafial funnily enough I actually wanted to do submit a pom cleanup PR next

qualidafial:
FYI: I want to release 3.2.0 tonight, so I'm focusing on pre-release cleanup and polish for the next couple hours.

TheRealMarnes:
Can't help there then, it's bedtime for me...

PR #1108, opened 2018-04-30
Bind `LocalDate` parameters using `setDate` instead of `setTimestamp`

pazustep:
Converting `LocalDate` arguments to `Timestamp` might have adverse effects, to the point of setting the wrong date in the database if local and server timezones are far enough apart.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
@@ -106,7 +106,7 @@ public class BuiltInArgumentFactory implements ArgumentFactory {
         register(map, UUID.class, Types.VARCHAR, PreparedStatement::setObject);
 
         register(map, Instant.class, Types.TIMESTAMP, (p, i, v) -> p.setTimestamp(i, Timestamp.from(v)));
-        register(map, LocalDate.class, Types.TIMESTAMP, (p, i, v) -> p.setTimestamp(i, Timestamp.valueOf(v.atStartOfDay())));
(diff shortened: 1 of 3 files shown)


jarlah:
I was actually going to suggest add an sqlobject test to check that it's stored correctly. But it's like 50+ lines of test for a simple blatant fix.

mbrito-yieldstreet:
@jarlah tests are not too much; I'll try to get it done.

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
```diff
@@ -106,7 +106,7 @@
         register(map, UUID.class, Types.VARCHAR, PreparedStatement::setObject);
 
         register(map, Instant.class, Types.TIMESTAMP, (p, i, v) -> p.setTimestamp(i, Timestamp.from(v)));
-        register(map, LocalDate.class, Types.TIMESTAMP, (p, i, v) -> p.setTimestamp(i, Timestamp.valueOf(v.atStartOfDay())));
```
qualidafial: Ouch, how did that `TIMESTAMP` sneak through. Good catch.

pazustep:
@qualidafial I've added a couple of tests. I couldn't actually check the fix – that would involve multiple VMs or launching external databases – but these tests at least verify that the `LocalDate` binding change shouldn't break any code that depended on the previous behavior.

qualidafial:
Looks great, thanks!

stevenschlansker:
If we wish to test databases in non-standard TZ configurations, I believe the EmbeddedPostgres allows you to override the database time zone as a server flag.

PR #1110, opened 2018-05-02
Add methods to get attributes from Jdbi - closes #1109

### LINKED ISSUE #1109: Retrieving `defined` attributes
Currently, Jdbi allows us to `define` values that can be interpolated in queries, like so:
```java
Jdbi.define("foo", "fooColumn");
Jdbi.open().createQuery("SELECT <foo> FROM my_foos");
```
But afterwards we are not able to retrieve the value of `foo` directly from the Jdbi class or instance, it would be nice to have `jdbi.getAttribute("foo")` .
The `StatementContext#getAttribute` method does just this but it's only accessible from within a StatementCustomizer

zikani03:
Added a couple of convenience methods to get defined attributes directly from the Jdbi instance. Basically the same way the StatementContext does it.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Jdbi.java
+++ b/core/src/main/java/org/jdbi/v3/core/Jdbi.java
@@ -16,6 +16,7 @@ package org.jdbi.v3.core;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
 import java.util.ServiceLoader;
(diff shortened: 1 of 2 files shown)


qualidafial:
We've tried to limit the config helper methods in `Configurable` (which `Jdbi` implements) to just those that are commonly needed at that usage phase.

That's why you see register methods in `Configurable` but not in `StatementContext`, and you only see methods like `findArgumentFor` in `StatementContext` but not `Confifurable`.

As you've seen it's already accessible through the config class. What is your use case for pulling this up to `Jdbi`?

zikani03:
Thanks for the response. My particular use case was to fetch attributes defined on the context:

For example:

I have a query defined as follows: 
```java
private static final String sql = "SELECT <pkey>,<field> FROM <table> ORDER BY <pkey> LIMIT <fetchLimit> OFFSET :offset";
```
Note how fetch limit is _defined_ and not used as a bound parameter - that's because somewhere in the code it's set like this:

`jdbi.define("fetchLimit", configuration.getFetchLimit());`

Somewhere else we'd like to get the value of `fetchLimit` for logging or to update the offset etc.. which is where the `getAttribute` method would come in handy rather than having to implement a `JdbiConfig` class with a single field...

However, since I eventually figured out how to get the attributes it's more of a convenience than anything at this point :)

PR #1112, opened 2018-05-05
Add support for binding primitive optionals

electrum:
This adds a built-in argument factory for
OptionalInt, OptionalLong and OptionalDouble.


qualidafial:
Thanks!

(can't believe I overlooked this!)

PR #1113, opened 2018-05-08
[asap!] fix test with random failures

TheRealMarnes:
Sorry, stupid little defect in this test that I never realized would be possible


TheRealMarnes:
Sorry, stupid little defect in this test that I never realized would be possible

PR #1118, opened 2018-05-12
Use FreeMarker's built-in classpath template loading instead of rolli…

qualidafial:
…ng it ourselves.

Confirmed this change fixes an error seen while trying to use FreeMarker SQL locator:

```
! java.lang.IllegalArgumentException: URI is not hierarchical
! at java.io.File.<init>(File.java:418)
! at org.jdbi.v3.freemarker.FreemarkerSqlLocator.findTemplateDirectory(FreemarkerSqlLocator.java:47)
! at org.jdbi.v3.freemarker.FreemarkerSqlLocator.findTemplate(FreemarkerSqlLocator.java:54)
! at org.jdbi.v3.freemarker.internal.UseFreemarkerSqlLocatorImpl.lambda$configureForType$1(UseFreemarkerSqlLocatorImpl.java:39)
! at org.jdbi.v3.sqlobject.statement.internal.CustomizingStatementHandler.locateSql(CustomizingStatementHandler.java:172)
```

We were calling URL, URI, File APIs to load files on the classpath, but FreeMarker already provides facilities to load templates in a much easier way.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -18,7 +18,7 @@
     <parent>
         <artifactId>jdbi3-parent</artifactId>
         <groupId>org.jdbi</groupId>
-        <version>3.2.1</version>
+        <version>3.2.2-SNAPSHOT</version>
     </parent>
 
     <groupId>org.jdbi</groupId>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -18,7 +18,7 @@
     <parent>
         <artifactId>jdbi3-parent</artifactId>
         <groupId>org.jdbi</groupId>
-        <version>3.2.1-SNAPSHOT</version>
+        <version>3.2.1</version>
     </parent>
 
     <groupId>org.jdbi</groupId>
--- a/commons-text/pom.xml
+++ b/commons-text/pom.xml
@@ -17,7 +17,7 @@
     <parent>
         <artifactId>jdbi3-parent</artifactId>
         <groupId>org.jdbi</groupId>
-        <version>3.2.1-SNAPSHOT</version>
+        <version>3.2.1</version>
     </parent>
     <artifactId>jdbi3-commons-text</artifactId>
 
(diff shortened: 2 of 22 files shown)


qualidafial:
I'm planning to roll a maintenance release once this is approved and merged.

qualidafial:
Just for context, the exception I pasted above manifested in an app build, but only during integration tests. Unit tests ran fine, but the integration tests run against the fat-jarred application. So it appears that the existing code will only work for template files on the filesystem--not those embedded in JARs.

On another note, I noticed this in the FreeMarker docs: https://freemarker.apache.org/docs/pgui_quickstart_createconfiguration.html:

> Do not needlessly re-create Configuration instances; it's expensive, among others because you lose the template cache. Configuration instances meant to be application-level singletons.

We're creating `Configuration` instances for every template :grimacing:, _and_ we're redundantly keeping a cache when FreeMarker already does this for us.

@jarlah I've modified `FreemarkerSqlLocator` on my machine to create a single `Configuration` in the static initializer, using the context classloader for the classpath template loader. This is working well on my machine, however it breaks one test when it tries to process an `#include` directive:

```ftl
<#include "org/jdbi/v3/freemarker/util.ftl">
```

If I add a leading slash to make it an absolute path, the test passes:

```ftl
<#include "/org/jdbi/v3/freemarker/util.ftl">
```

I feel like this is a reasonable breakage in exchange for using `Configuration` the way it's designed to be used--especially since 3.2.0 doesn't really work if you're running from a JAR.

With this change, the only thing `FreemarkerSqlLocator` does (beyond initializing a configuration for you), is translates a class and template name into a path, and fetches the template for that path from the configuration.

I'm going to do a maintenance release with just this change.

If this setup turns out to be too inflexible, we can add a `FreemarkerSqlLocator.setConfiguration()` method in the future.

qualidafial:
This is released in 3.2.1

jarlah:
Going the implement this at work on monday. We strongly relies on StringTemplate4 at the moment, and its pretty unstable.

Issue #1123, opened 2018-05-15
@BindMethods Bug with respect to generically defined methods


isaki-x:
Greetings,

I've isolated a bug when utilizing `@BindMethods` and I have a fix that works, but I don't love it. I'm looking to start a discussion around this issue before I submit my PR.

The scenario I encountered was specific to an enum I am using that implements an interface. For argument's sake, let's define these as follows:

```Java
public interface JdbiValue<T> {
    T asColumn();
}
```

Then, we have the following, silly example:

```Java
public enum FooType implements JdbiValue<Integer> {
    A(12),
    B(14);

    private Integer val;

    private FooType(final int val) {
        this.val = val;
    }

    @Override
    public Integer asColumn() {
        return this.val;
    }
}
```

Then we have the following SqlObject definition:

```Java
public class FooObject {
    private final FooType type;

    public FooObject(final FooType type) {
        super();
        this.type = type;
    }

    public FooType getType() {
        return this.type;
    }
}
```

```Java
public interface FooDAO {

    @SqlUpdate("UPDATE table SET value = :o.asColumn WHERE id = :id")
    int update(@Bind("id") int id, @BindMethods("o") FooObject o);
}
```

When this is used, an IllegalStateException is thrown out of the `Collectors` class as the default merge operator is in use:

```Java
    /**
     * Returns a merge function, suitable for use in
     * {@link Map#merge(Object, Object, BiFunction) Map.merge()} or
     * {@link #toMap(Function, Function, BinaryOperator) toMap()}, which always
     * throws {@code IllegalStateException}.  This can be used to enforce the
     * assumption that the elements being collected are distinct.
     *
     * @param <T> the type of input arguments to the merge function
     * @return a merge function which always throw {@code IllegalStateException}
     */
    private static <T> BinaryOperator<T> throwingMerger() {
        return (u,v) -> { throw new IllegalStateException(String.format("Duplicate key %s", u)); };
    }
```

So, I created a simple `BinaryOperator<Method>` to capture what was going on (and to return the first parameter to see if that worked, which it did). This was done in `ObjectMethodArguments`.

```Java
    private static final Map<Class<?>, Map<String, Method>> CLASS_METHODS = ExpiringMap.builder()
        .expiration(10, TimeUnit.MINUTES)
        .expirationPolicy(ExpirationPolicy.ACCESSED)
        .entryLoader((Class<?> type) ->
            Stream.of(type.getMethods())
[...]

PR #1124, opened 2018-05-15
Addressed duplicate methods from reflective lookup

isaki-x:
When implementing interfaces with generic types, in Java 8, the
Class.getMethods() invocation will return methods with matching
signatures but different return types. This fix ensures the most
specific return type is the version chosen.

This fixes the issue brought up in #1123.

### DIFF
--- a/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindMethods.java
+++ b/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindMethods.java
@@ -219,7 +219,6 @@ public class TestBindMethods {
         public String toString() {
             return Objects.toString(this.value);
         }
-
     }
 
     public static final class BigIntNumberArgumentFactory extends AbstractArgumentFactory<Number> {
@@ -232,6 +231,5 @@ public class TestBindMethods {
         protected Argument build(final Number value, final ConfigRegistry config) {
             return new BigIntNumberArgument(value);
         }
-
     }
 }
--- a/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindMethods.java
+++ b/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindMethods.java
@@ -1,3 +1,16 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package org.jdbi.v3.sqlobject;
 
 import static org.assertj.core.api.Assertions.assertThat;
--- a/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindMethods.java
+++ b/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindMethods.java
@@ -1,0 +1,240 @@
+package org.jdbi.v3.sqlobject;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Type;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Types;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.argument.ArgumentFactory;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.rule.H2DatabaseRule;
+import org.jdbi.v3.core.statement.StatementContext;
+import org.jdbi.v3.sqlobject.config.RegisterArgumentFactory;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.customizer.BindMethods;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import com.google.common.reflect.TypeToken;
+
+public class TestBindMethods {
+
+    @Rule
+    public H2DatabaseRule dbRule = new H2DatabaseRule().withPlugin(new SqlObjectPlugin());
+
+    @Before
+    public void setUp() {
+        final Handle h = this.dbRule.getSharedHandle();
+        h.execute("CREATE TABLE bind_methods (id IDENTITY PRIMARY KEY, value BIGINT)");
+    }
+
+    @Test
+    public void testBindMethodsDirect() {
+        final PairRowDAO dao = this.dbRule.getJdbi().onDemand(PairRowDAO.class);
+        final long testValue = 709L;
+        final int testId = 5;
+
+        final DatabaseValue<?> dbVal = new DirectDatabaseLongValue(testValue);
+        final PairRow row = new PairRow(testId, dbVal);
+
+        assertThat(dao.insert(row)).isEqualTo(1);
+        assertThat(dao.getById(testId)).isEqualTo(testValue);
+    }
+
+    @Test
+    public void testBindMethodsImplicitOverride() {
+        final PairRowDAO dao = this.dbRule.getJdbi().onDemand(PairRowDAO.class);
+        final long testValue = 708L;
+        final int testId = 6;
+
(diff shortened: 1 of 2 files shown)


qualidafial:
Thank you for tracking down and reporting this.

We dealt with a similar issue recently with SQL Object interface types that inherit from generic interfaces. See #1087. 

If my suspicion is correct, this is actually caused by bridge methods (which were introduced way back in Java 5):

```java
interface Foo<T, R> {
  R foo(T input);
}

class Bar implements Foo<String, Integer> {
  Integer foo(String input) {
    // do stuff
  }
}
```

Any time a class/interface extends/implements a type with generic parameters, and provides concrete parameters, you actually end up with two actual methods for each generic method declared:

```java
class Bar implements Foo<String, Integer> {
  // bridge method synthetically created by the Java compiler
  Object foo(Object input) {
    // casts parameters to the expected type and delegates to the other method
    return foo((String) input);
  }

  // generic method
  Integer foo(String input) {
    // do stuff
  }
}
```

Whether the generic method or the bridge method gets called depends on the calling context:

```java
Bar bar = new Bar();
bar.foo("hello"); // calls generic method

Foo<?, ?> foo = bar;
foo.foo("world"); // calls bridge method
```

I think I'd prefer that we fix this by reusing the logic we already have in `BridgeMethodHandlerFactory`. We could introduce a utility method e.g. `GenericTypes.findBridgedMethod(Method bridgeMethod, Class type): Optional<Method>`, and extract the "find bridged method" logic from the factory. Then use that method in both `BridgeMethodHandlerFactory` and `ObjectMethodArguments`.

Finally, we need some tests that reproduce the bug to prove that this change actually fixes the problem, and to prevent future regressions.

isaki-x:
Thank you for the feedback! I wrote a quick test mock up to see if that solution and my solution give the same answer (and they did, sort of... they have very different goals; my solution is designed for zero arg methods where the desired method is not known as you are pre-building a cache; the #1087 solution validates method signature matches when the method you want is known and a bridge method and you need the concrete, non-bridged method).

```Java
public class FooTest {

    @Test
    public void test1123() {

        final Map<String, Method> result = Arrays.stream(BlackWidow.class.getMethods())
            .filter(m -> m.getParameterCount() == 0)
            .collect(Collectors.toMap(Method::getName, Function.identity(), FooTest::boundMethodMerge));

        final Method chosen = result.get("getValue");
        assertNotNull(chosen);

        final Class<?> retType = chosen.getReturnType();
        assertEquals(String.class, retType);
    }

    @Test
    public void test1087() {

        // This wants bridge methods for method passed in
        Method tmp = null;
        final Method[] allMethods = BlackWidow.class.getMethods();
        for (int i = 0; i < allMethods.length; ++i) {
            if (allMethods[i].isBridge() && "getValue".equals(allMethods[i].getName())) {
                tmp = allMethods[i];
                break;
            }
        }

        assertNotNull(tmp);

        // Play nice with lambda
        final Method method = tmp;

        final Map<String, Method> result = Arrays.stream(allMethods)
            .filter(candidate -> !candidate.isBridge()
                && Objects.equals(candidate.getName(), method.getName())
                && candidate.getParameterCount() == method.getParameterCount())
            .filter(candidate -> {
                Class<?>[] candidateParamTypes = candidate.getParameterTypes();
                Class<?>[] methodParamTypes = method.getParameterTypes();
                return IntStream.range(0, method.getParameterCount())
                    .allMatch(i -> methodParamTypes[i].isAssignableFrom(candidateParamTypes[i]));
            })
            .collect(Collectors.toMap(Method::getName, Function.identity()));

        final Method chosen = result.get("getValue");
        assertNotNull(chosen);

        final Class<?> retType = chosen.getReturnType();
        assertEquals(String.class, retType);
    }

    private static Method boundMethodMerge(Method a, Method b) {
        final Class<?> aClazz = a.getReturnType();
        final Class<?> bClazz = b.getReturnType();

        final Method ret;
        if (aClazz.isAssignableFrom(bClazz)) {
            ret = b;
        } else if (bClazz.isAssignableFrom(aClazz)) {
            ret = a;
        } else {
            throw new IllegalStateException("Method return conflict");
        }

        return ret;
    }

    private interface Spider<T> {

        T getValue();
    }

    private static final class BlackWidow implements Spider<String> {

        private final String value;

        @SuppressWarnings("unused")
        BlackWidow(final String value) {
            super();
            this.value = value;
        }

        @Override
        public String getValue() {
            return this.value;
        }
    }

}
```

So while these two PRs share a common root cause, the problem they aim to solve is different (as is the information available when tackling the problem).

However, the #1087 solution has given me an idea on how I might do this better. I will write some test cases then come back with an answer as to whether or not I stand by my approach or if I want to change it.

qualidafial:
After I left my comment last night, it occurred to me that this might be as simple as making the merge function prefer the non-bridge method over the bridge method.

isaki-x:
That is what I'm currently testing.

isaki-x:
I found another nuanced issue when developing comprehensive tests. If you have an interface A<T>, a class B<T> that implements A's method that returns T, and then a class C that extends B<T> with a specific type, the specific implementation does not get created.

For example:

```Java
    public interface DatabaseValue<T> {

        T getColumnValue();
    }

    public class DatabaseNumberValue<T extends Number> implements DatabaseValue<T> {

        private final T value;

        DatabaseNumberValue(final T value) {
            super();
            this.value = value;
        }

        @Override
        public T getColumnValue() {
            return this.value;
        }
    }

    public final class DatabaseLongValue extends DatabaseNumberValue<Long> {

        DatabaseLongValue(final long value) {
            super(Long.valueOf(value));
        }
    }

    public final class DatabaseLongValueWithOverride extends DatabaseNumberValue<Long> {

        DatabaseLongValueWithOverride(final long value) {
            super(Long.valueOf(value));
        }

        @Override
        public Long getColumnValue() {
            return super.getColumnValue();
        }
    }

    public final class DirectDatabaseLongValue implements DatabaseValue<Long> {

        private final long value;

        DirectDatabaseLongValue(final long value) {
            super();
            this.value = value;
        }

        @Override
        public Long getColumnValue() {
            return Long.valueOf(this.value);
        }
    }
```

When we dump the methods named 'getColumnValue', we see the following:

`[public java.lang.Number org.jdbi.v3.sqlobject.TestBindMethods$DatabaseNumberValue.getColumnValue(), public java.lang.Object org.jdbi.v3.sqlobject.TestBindMethods$DatabaseNumberValue.getColumnValue()]`

When we explicitly provide an overwrite to the implicitly defined method, you see we get three methods (two of which are bridged):

`[public java.lang.Long org.jdbi.v3.sqlobject.TestBindMethods$DatabaseLongValueWithOverride.getColumnValue(), public java.lang.Number org.jdbi.v3.sqlobject.TestBindMethods$DatabaseLongValueWithOverride.getColumnValue(), public java.lang.Object org.jdbi.v3.sqlobject.TestBindMethods$DatabaseNumberValue.getColumnValue()]`

The reason this is an issue is there is no argument for `Number` so my testing fails.

I'm going to rewrite my tests to use explicit objects that I control so I can verify everything works as desired.

isaki-x:
I was wrong; I added an explicit `ArgumentFactory` for `Number` and still get the following error:

```org.jdbi.v3.core.statement.UnableToCreateStatementException: No argument factory registered for type [T] for element [getColumnValue] on [org.jdbi.v3.sqlobject.TestBindMethods$DatabaseLongValue@75881071] [statement:"INSERT INTO bind_methods (id, value) VALUES(:row.getKey, :row.getValue.getColumnValue)", rewritten:"INSERT INTO bind_methods (id, value) VALUES(:row.getKey, :row.getValue.getColumnValue)", parsed:"ParsedSql{sql='INSERT INTO bind_methods (id, value) VALUES(?, ?)', parameters=ParsedParameters{positional=false, parameterNames=[row.getKey, row.getValue.getColumnValue]}}", arguments:{ positional:{}, named:{}, finder:[{lazy object functions arguments "org.jdbi.v3.sqlobject.TestBindMethods$PairRow@365185bd"]}]
```

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindMethods.java
```diff
@@ -0,0 +1,253 @@
...
+        }
+
+        @Override
+        public String toString() {
+            return Objects.toString(this.value);
+        }
+
+    }
+
+    public static final class BigIntNumberArgumentFactory implements ArgumentFactory {
```
qualidafial: AbstractArgumentFactory simplifies this process by handling nulls for you--just pass the SQL type integer to the superconstructor and implement the method, which is only invoked when the value is non-null.

isaki-x: After thinking about it, there is no harm in letting the argument factory handle all the cases. I've made your suggested changes and pushed an update. Thanks!

isaki-x:
Thanks for both the approval and the compliment!

isaki-x:
@qualidafial Is there anything still outstanding that I missed?

qualidafial:
Looks great! Thanks for the PR

isaki-x:
Sorry to bother you all, but is there an ETA for the next release of JDBI? I'm reaching a deadline for which I need this fix and am trying to manage expectations accordingly. Thanks!

qualidafial:
No problem. I'll check with the project members and see if we can cut a release tonight.

isaki-x:
That is amazing, thank you!

qualidafial:
This is released in 3.3.0. It should be available on Maven Central, although it may not be indexed on search.maven.org for a couple hours.

isaki-x:
Thank you so much!

PR #1127, opened 2018-05-16
[done] fix for allowNoResults with mapping

TheRealMarnes:
Statements a la `handle.createQuery(line).configure(ResultProducers.class, rp -> rp.allowNoResults(true)).mapToMap().list()` throw an exception because of mappers calling `EmptyResultSet.getMetaData`, which I've now patched to return EmptyResultSetMetaData, where `getColumnCount() == 0` is the magic bit. This fixes the use case (see tests) and shouldn't break anything.

There is the philosophical argument of throwing exceptions vs returning neutral values (for the other methods), if anyone wants to go into that, but I think it works fine as-is.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/result/EmptyResultSetMetaData.java
+++ b/core/src/main/java/org/jdbi/v3/core/result/EmptyResultSetMetaData.java
@@ -1,3 +1,16 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jdbi/v3/core/result/EmptyResultSetMetaData.java
```diff
@@ -0,0 +1,137 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.result;
+
+import java.sql.ResultSetMetaData;
+
+public class EmptyResultSetMetaData implements ResultSetMetaData {
```
stevenschlansker: Could this class be made package-private, or moved to `internal`?  I don't see it having much value as public API.

TheRealMarnes:
done

PR #1128, opened 2018-05-16
[alternative] DollarSqlParser

TheRealMarnes:
SqlParser for `$foo` placeholders. Not familiar with antlr beyond editing the existing scripts and figuring out how it works that way, so I'm not entirely sure the lexer makes perfect sense as-is (it's an edit of the Hash lexer).

Found a suspicious test case `TestHashPrefixSqlParser.testCommentQuote`. Doesn't really seem to test anything of value. I suspect it needs to look more like my `TestDollarPrefixSqlParser.testCommentAndQuote` to be meaningful.

### DIFF
--- a/core/src/main/antlr3/org/jdbi/v3/core/internal/lexer/DollarStatementLexer.g
+++ b/core/src/main/antlr3/org/jdbi/v3/core/internal/lexer/DollarStatementLexer.g
@@ -1,0 +1,32 @@
+lexer grammar DollarStatementLexer;
+
+@header {
+    package org.jdbi.v3.core.internal.lexer;
+}
+
+@lexer::members {
+  @Override
+  public void reportError(RecognitionException e) {
+    throw new IllegalArgumentException(e);
+  }
+}
+
+fragment QUOTE: '\'';
+fragment ESCAPE: '\\';
+fragment ESCAPE_QUOTE: ESCAPE QUOTE;
+fragment DOUBLE_QUOTE: '"';
+fragment DOLLAR: '$';
+fragment QUESTION: {input.LA(2) != '?'}?=> '?';
+fragment DOUBLE_QUESTION: {input.LA(2) == '?'}?=> '??';
+fragment NAME: 'a'..'z' | 'A'..'Z' | '0'..'9' | '_' | '.' | ':';
+
+COMMENT: '/*' .* '*/';
+QUOTED_TEXT: QUOTE (ESCAPE_QUOTE | ~QUOTE)* QUOTE;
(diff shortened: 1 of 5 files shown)


TheRealMarnes:
Question: is it advised to always make SqlParsers "sql syntax aware" (i.e. with lexers) or are more "dumb" parsers allowable too if the user understands the risks involved (false positive matches)? I'd e.g. contribute an SqlParser that uses regex matching, with customizable prefix and suffix. That would let users pick whatever placeholder format they feel like using, but of course also means they need to watch out for themselves that they don't accidentally write matching text in their queries. Using _foo_ placeholder for example format could end up turning camelcase_table_name into camelcaseFOOname with table=foo...

stevenschlansker:
What does using `$` over the existing `#` or `:` bring?  does it avoid some particular SQL incompatibility or inconvenience?

TheRealMarnes:
Just stylistic choice. $ is pretty common as a placeholder, especially in spring and linux, I felt it was lacking. Would totally understand if you considered it cruft and rejected it.

stevenschlansker:
Same conclusion here as the other one, let's hold this one for community votes, unless we have a clear case where this parser is superior?  There's nothing inherently wrong with it, but absent anyone else asking for it let's not merge too eagerly.

TheRealMarnes:
works for me

qualidafial:
One note: choosing `$foo` for bound variables would conflict with FreeMarker template engine, which uses `$` prefix for template variables.

I'm fairly confident FreeMarker will be our recommended template engine (if the default is inadequate) in the near future. StringTemplate appears totally abandoned, whereas FreeMarker is regularly maintained, and gives pretty great error messages to boot.

TheRealMarnes:
@qualidafial tbh that's just a matter of which tools you combine. Personally I'm starting to home in on using StringSubstitutor with mustaches and $foo parameters due to familiarity with that syntax in many environments. It's the user's problem to figure out which combination of syntax they want to use ¯\\\_(ツ)\_/¯

Can I ask that we go ahead with this for the simple reasons that it's only a minor modification of existing code (no new risks) and gives users a choice between a healthy 3 options instead of only 2? There aren't many usable symbols left after : # and $ either, so I doubt more of these will show up as PRs. We have a healthy 4-ish TemplateEngines to choose from too, a nice number that's not restrictively low nor burdeningly high.

It does really need a review of the antlr code. I modified it without any antlr experience. I can't see any problem with it but that's developer bias. I've already commented on the other TemplateEngine I offered that I've come to believe it's not a good fit and should be closed without merging, so a decision here will wrap up the whole discussion. :)

PR #1129, opened 2018-05-17
[experiment] regex sqlparser

TheRealMarnes:
Quick implementation of a regex-based non-lexing SqlParser.

Has obvious risks but works great if the user watches his step accordingly. Has the advantage of being configurable, unlike the antlr SqlParsers. You gain some you lose some.

Needs some more tests, javadoc, and maybe some more implementation (I don't know yet what to do about positionals for example).

Is this appealing to be fleshed out and merged or nah?

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/RegexSqlParser.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/RegexSqlParser.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


stevenschlansker:
Do you have a brief use case where this parser is uniquely suited?  (i.e. what's it good for?)

TheRealMarnes:
Its only point is to offer more choices of placeholder format, like ${foo}, over the 2 current fixed formats. ${} in particular being pretty common, I felt it was lacking. Users could go for @foo or whatever else too. Comes at the cost of not lexing the input though. Would again totally understand a cruft stamp and rejection.

stevenschlansker:
It's not so much that I'd discard it as "Cruft", but the bar for inclusion to the library is pretty high, and we don't want to start sprouting a ton of single-use-case components.  My recommendation is that it's useful to you, keep it in your code for now -- and if other users reach out and verify it's useful to them as well we can merge it into main.  Sound fair?

TheRealMarnes:
works for me

TheRealMarnes:
I've come to believe we'd better not use this after all, tbh. It doesn't do any text parsing so it's risky to use. Agree we close this and forget about it?

stevenschlansker:
Glad you came around to that conclusion :) looking forward to your next PR

PR #1133, opened 2018-05-21
Introduced JUnit Rule to provide embedded postgres prepared by flyway

Sch3lp:
Also added static factory method in abstract JdbiRule for consistency.

Unsure about the class name, but the class wraps the `PreparedDBRule` from the otj-pg-embedded lib.

This newly introduced rule provides the convenience of having an in memory postgres db, that gets set up with Flyway migrations according to the provided migrationLocations.

Example usage:
```java
public class SomeTest {
   @Rule
   public JdbiRule db = JdbiRule.preparedEmbeddedPostgres("db/migration").withPlugins();
}
```

### DIFF
--- a/testing/src/main/java/org/jdbi/v3/testing/JdbiRule.java
+++ b/testing/src/main/java/org/jdbi/v3/testing/JdbiRule.java
@@ -43,6 +43,15 @@ public abstract class JdbiRule extends ExternalResource {
     }
 
     /**
+     * Create a JdbiRule with an embedded Postgres instance that will be prepared by Flyway.<br>
+     * Your project must depend on the {@code otj-pg-embedded} artifact.
+     * @param migrationLocations classpath locations to scan recursively for flyway migrations. (default: db/migration)
+     */
+    public static JdbiRule preparedEmbeddedPostgres(final String... migrationLocations) {
+        return new PreparedEmbeddedPostgresJdbiRule(migrationLocations);
+    }
+
+    /**
      * Create a JdbiRule with an in-memory H2 database instance.
      * Your project must depend on the {@code h2} database artifact.
      */
--- a/testing/src/main/java/org/jdbi/v3/testing/PreparedEmbeddedPostgresJdbiRule.java
+++ b/testing/src/main/java/org/jdbi/v3/testing/PreparedEmbeddedPostgresJdbiRule.java
@@ -1,0 +1,27 @@
+package org.jdbi.v3.testing;
+
+import com.opentable.db.postgres.embedded.FlywayPreparer;
+import com.opentable.db.postgres.junit.EmbeddedPostgresRules;
+import com.opentable.db.postgres.junit.PreparedDbRule;
+import org.jdbi.v3.core.Jdbi;
+import org.junit.runner.Description;
+import org.junit.runners.model.Statement;
+
+class PreparedEmbeddedPostgresJdbiRule extends JdbiRule {
+
+    private final PreparedDbRule embeddedPreparedDbRule;
+
+    PreparedEmbeddedPostgresJdbiRule(final String... migrationLocations) {
+        embeddedPreparedDbRule = EmbeddedPostgresRules.preparedDatabase(FlywayPreparer.forClasspathLocation(migrationLocations));
+    }
+
+    @Override
+    protected Jdbi createJdbi() {
+        return Jdbi.create(embeddedPreparedDbRule.getTestDatabase());
+    }
+
+    @Override
+    public Statement apply(final Statement base, final Description description) {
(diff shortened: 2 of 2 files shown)


testing/src/main/java/org/jdbi/v3/testing/JdbiRule.java
```diff
@@ -42,6 +42,15 @@ public static JdbiRule embeddedPostgres() {
         return new EmbeddedPostgresJdbiRule();
     }
 
+    /**
+     * Create a JdbiRule with an embedded Postgres instance that will be prepared by Flyway.<br>
+     * Your project must depend on the {@code otj-pg-embedded} artifact.
+     * @param migrationLocations classpath locations to scan recursively for flyway migrations. (default: db/migration)
+     */
+    public static JdbiRule preparedEmbeddedPostgres(final String... migrationLocations) {
```
qualidafial: Would be preferable to make this a method on `JdbiRule` so this is available for all databases:

```java
@Rule public JdbiRule jdbiRule = JdbiRule.embeddedPostgres().preparedWithFlyway("db/migration");
```

Sch3lp: Makes sense. Let's see what I can cough up. :)

Sch3lp: So there seems to be a bit of an chicken or egg problem with the way the `FlywayPreparer` works.

In its current implementation, the `PreparedEmbeddedPostgresJdbiRule` simply uses a wrapped `PreparedDbRule`, which both creates a `Datasource` and performs Flyway migrations on it.

Later in the runtime, `createJdbi()` gets called, and a `Jdbi` object is created with the `Datasource` from the `PreparedDbRule`.

Rules are then applied in order.

## Alternative: Late FlywayPrepare
_Building_ a `JdbiRule` in the fashion of `.preparedWithFlyway()` implies that I can't reuse the `PreparedDbRule` and have to perform the _prepare_ through custom code.

This custom code would create a `FlywayPreparer` like this 
```java
flywayPreparer = FlywayPreparer.forClasspathLocation(locations);
```
But would then have to be used like this:
```java
if (flywayPreparer != null) {
    flywayPreparer.prepare(datasource);
}
```
The problem is that `JdbiRule` doesn't expose a `Datasource` after you've created a `Jdbi` (or `Handle`) object because the Jdbi API doesn't expose it. It's the other way around, you create a Jdbi or Handle from a Datasource.

That is, unless I'm missing something. I'm not super familiar with Jdbi or `otj-pg-embedded` for that matter.

qualidafial: I think the easiest thing is to circumvent the `DatabasePreparer` thing altogether, and run flyway explicitly. I've pushed a proof of concept to this branch: https://github.com/jdbi/jdbi/tree/jdbirule-flyway-migrations

Sch3lp: Ah. I see!

So, how do we go about this PR?

Do we switch to that branch? Or do we continue on this PR?

I'm new to this thing.

qualidafial:
We don't merge anything to master unless it's ready to release immediately.

I appreciate wanting to address the pain and move on, but what we're proposing is public API (even if it is just for testing) which is a long term compatibility promise. We owe it to our users to take a little extra time to get things right the first time.

@stevenschlansker I have an idea I wanted to run by you. First check the branch I linked above then continue reading this comment.

Because `JdbiRule.createJdbi` is currently abstract, existing subclasses have to override that method. On my branch, `createJdbi()` is a concrete method that delegates to the new `createDataSource()` method.

I can't make `createDataSource` abstract because that would break existing implementations. But I _could_ make it throw `UnsupportedOperationException` by default, and document that subclasses must either override `createDataSource` or `createJdbi`, and that some migrations are only supported when you override `createDataSource`.

Is that a fair compromise?

stevenschlansker:
I think that's a fair compromise.  I think `JdbiRule` is also new enough, if we are convinced the new way is strictly better, we could break compatibility -- it's only been one release and it's testing code.  But the `UnsupportedOperationException` approach is fine as well.

arteam:
I'm also fine with making a small breaking change to `JdbiRule` to make the Flyway implementation not limited to Postgres. I don't think there are many (if any) custom implementations of it out there.

qualidafial:
If you're both ok with breaking compatibility in this case then that's probably easier to support long term. I'll make the change.

qualidafial:
@Sch3lp you alright with closing this PR in favor of #1152?

Sch3lp:
You betcha!

________________________________

Issue #1132, opened 2018-05-19
Kotlin support for RegisterJoinRowMapper


gabzim:
I'm using the kotlin plugin so I can have support for kotlin objects without creating a custom mapper, one thing that isn't clear from the docs is how to use this when doing a `JoinRow` type of query, eg:

```
interface CustomerDao {

    @RegisterJoinRowMapper(Customer::class, CustomerAddress::class, CustomerPhone::class)
    @SqlQuery("""select
  c.id       as c_id,
  c.name     as c_name,
  a.id       as a_id,
  a.address1 as a_address1,
  p.id       as p_id,
  p.phone    as p_phone
from customer c
  left join phones p on c.id = p.customer_id
  left join addresses a on c.id = a.customer_id
where c.id = :id
    """)
    fun selectCustomerDetailed(id: Int): Stream<JoinRow>
}
```

However, since here variable names no longer have the plain name that can be matched, there should exist a way to say: I want to create a bean using the automatic mapper provided by the plugin *HOWEVER* the names will be prefixed by *prefix_*

Has anyone done this yet?

arg20:
Also, shouldn't `KotlinMapper` be `open` so we can subclass? The `prefix` in it doesn't seem to work, given that `specialize0` completely ignores the prefix.

arg20:
I submitted a PR with what I believe to be a possible solution to this issue.

gabzim:
You can register kotlin mappers with a given prefix by using `@RegisterKotlinMapper(class, prefix)` or for multiple ones `@RegisterKotlinMappers(...)`

PR #1134, opened 2018-05-21
Enables easy registration of kotlin mappers

gabzim:
It seems the Kotlin Mapper wasn't working correctly when trying to register the mapper for a type without using the factory:

eg: `jdbi.registerMapper(KotlinMapper(Entity::class.java))`

The reason for this is that the `KotlinMapper` is not parametrized and extends `RowMapper` by doing `: RowMapper<Any>`, the problem with this is that `registerMapper` will try to infer the class from the parametrized type of the mapper, and in this case it's `Any` (or `Object`), which I found out by debugging after wrestling with it for a while.

It also adds an annotation to register `KotlinMapper`s in the same way you'd register `BeanMapper` so that you can override the default mapper with a prefix:

```
    @SqlQuery("SELECT id AS c_id, name AS c_name FROM customer WHERE id=:id")
    @RegisterKotlinMapper(Customer::class, "c_")
    fun selectCustomer(@Bind("id") id: Int): Customer
```

Which in that example above is not useful but it IS useful when you're working with joins. This came up as an attempt to troubleshoot: https://github.com/jdbi/jdbi/issues/1132


kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/RegisterKotlinMapper.kt
```diff
@@ -0,0 +1,30 @@
...
+import kotlin.reflect.KClass
+
+/**
+ * Registers a KotlinMapper for a specific kotlin class
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@ConfiguringAnnotation(RegisterKotlinMapperImpl::class)
+@Target(ElementType.TYPE, ElementType.METHOD)
+@Repeatable
+@java.lang.annotation.Repeatable(RegisterKotlinMappers::class)
```
arg20: Both, kotlin's `@Repeatable` and java's `@Repeatable` are needed for the time being because kotlin doesn't support repeatable annotations yet. Presumably, once it does, we can get rid of the plural annotation and its corresponding `RegisterKotlinMappersImpl`

qualidafial:
I'm open to this PR, but we really need tests validating the new annotations.

kotlin/src/main/kotlin/org/jdbi/v3/core/kotlin/KotlinMapper.kt
```diff
@@ -38,22 +39,49 @@ import kotlin.reflect.jvm.jvmErasure
 
 private val nullValueRowMapper = RowMapper<Any?> { rs, ctx -> null }
 
-class KotlinMapper(clazz: Class<*>, private val prefix: String = "") : RowMapper<Any> {
-    private val kClass: KClass<*> = clazz.kotlin
+class KotlinMapper<T: Any>(clazz: Class<T>, private val prefix: String = "") : RowMapper<T> {
```
stevenschlansker: Is this change considered {source, binary} compatible in the Kotlin world?  I'm not sure what the rules are.  I don't think it would be in Java land, but we probably just have to accept that the Kotlin world is more unstable for now?

gabzim: That's a great question, I'm not 100% sure.

gabzim:
@qualidafial Sorry I haven't been able to add tests yet. I'm swamped with the (fairly recent) birth of my daughter and trying to help out in the house whenever I have free time. I'll get to it as soon as I can.

This PR does address a very annoying problem I wrestled with for a while though. The current implementation is counterintuitive when you want to register mappers since it can't infer typing properly.

stevenschlansker:
Congratulations on the new role ;) we'd love to pick this back up when you have time -- unfortunately I'm not much of a Kotlin guy to pick it up easily.

kotlin/src/main/kotlin/org/jdbi/v3/core/kotlin/KotlinMapper.kt
```diff
@@ -38,22 +39,49 @@ import kotlin.reflect.jvm.jvmErasure
...
+
+    companion object {
+
+        /**
+         * Returns a mapper factory that maps to the given kotlin class
+         *
+         * @param type the mapped class
+         * @return a mapper factory that maps to the given kotlin class
+         */
+        fun factory(type: Class<*>): RowMapperFactory {
```
qualidafial: Do we want `KClass` overloads too?

kotlin/src/main/kotlin/org/jdbi/v3/core/kotlin/KotlinMapper.kt
```diff
@@ -38,22 +39,49 @@ import kotlin.reflect.jvm.jvmErasure
...
+        }
+
+        /**
+         * Returns a mapper factory that maps to the given kotlin class
+         *
+         * @param type the mapped class
+         * @param prefix the column name prefix for each mapped kotlin property
+         * @return a mapper factory that maps to the given kotlin class
+         */
+        fun factory(type: Class<*>, prefix: String): RowMapperFactory {
```
qualidafial: Could have a single method with a default parameter instead:

```kotlin
fun factory(type: Class<*>, prefix:String = "") { ... }
```

kotlin/src/main/kotlin/org/jdbi/v3/core/kotlin/KotlinMapper.kt
```diff
@@ -38,22 +39,49 @@ import kotlin.reflect.jvm.jvmErasure
...
     private val constructor = findConstructor(kClass)
     private val constructorParameters = constructor.parameters
     private val memberProperties = kClass.memberProperties.mapNotNull { it as? KMutableProperty1<*, *>}.filter { property ->
         !constructorParameters.any { parameter -> parameter.paramName() == property.propName() }
     }
 
-    private val nestedMappers = ConcurrentHashMap<KParameter, KotlinMapper>()
-    private val nestedPropertyMappers = ConcurrentHashMap<KMutableProperty1<*, *>, KotlinMapper>()
+    private val nestedMappers = ConcurrentHashMap<KParameter, KotlinMapper<*>>()
+    private val nestedPropertyMappers = ConcurrentHashMap<KMutableProperty1<*, *>, KotlinMapper<*>>()
```
qualidafial: If these lines do not compile without the `<*>` type parameter, then I think it's safe to assume that adding the type parameter is unsafe in Kotlin

qualidafial:
@gabzim Just a reminder that we're still interested in this PR if you can find time to respond to reviews.

gabzim:
@qualidafial I'm going over this now. I'll send a push with the requested changes. I've tested this in my local environment manually but I can't get the tests to run :(

gabzim:
@qualidafial Sorry, I'm not done with this PR just yet but yeah, I don't register mappers as it happens out of the box, but I'd say the main benefit of this PR would be the annotation that lets you override it with a prefix. Nevertheless, it is worth noting that even though not necessary, registering a mapper programatically while using a prefix with the KotlinMapper is broken, you can try it out, it shouldn't work.

qualidafial:
I see this misunderstanding from time to time.

Because of Java generic type erasure, you cannot just pass a mapper instance to `register()` if the class of that mapper uses a type variable. It must be extend `RowMapper` or `ColumnMapper` with a concrete type.

You _can_ however register the mapper, if you pass the mapped type with the mapper:

```kotlin
jdbi.registerRowMapper(Entity::class.java, KotlinMapper(Entity::class, "prefix"))
```

gabzim:
It's funny, I just figured that from debugging, yeah, annoying backward compatibility issues. I just pushed a change that corrects that error in the implementation for the annotation `@RegisterKotlinMapper`. I'll figure out how to add tests (or maybe you can help me?) but I just tested this and it works.

gabzim:
@qualidafial Added test

kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/RegisterKotlinMapperImpl.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    override fun configureForType(registry: ConfigRegistry, annotation: Annotation, sqlObjectType: Class<*>) {
+        val registerKotlinMapper = annotation as RegisterKotlinMapper
+        val kotlinClass = registerKotlinMapper.value
+        val prefix = registerKotlinMapper.prefix
+        val mappers = registry.get(RowMappers::class.java)
+        if (prefix.isEmpty()) {
+            mappers.register(kotlinClass.java, KotlinMapper(kotlinClass.java))
+        } else {
+            mappers.register(kotlinClass.java, KotlinMapper(kotlinClass.java, prefix))
```
qualidafial: This if/else clause is unnecessary. You could just unconditionally register with the prefix--an empty prefix is ignored.

kotlin-sqlobject/src/test/kotlin/org/jdbi/v3/kotlin/KotlinSqlObjectPluginTest.kt
```diff
@@ -57,6 +58,10 @@ class KotlinSqlObjectPluginTest {
         @SqlQuery("select id, name from something where id=:id")
         fun findById(id: Int): Thing
 
+        @SqlQuery("select id as t_id, name as t_name from something where id=:id")
+        @RegisterKotlinMapper(Thing::class, "t_")
```
qualidafial: Fun fact: because Jdbi elides underscores when comparing column names, this would also work with just `prefix = "t"`

kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/RegisterKotlinMapperImpl.kt
gabzim: Great point, I actually followed this approach from other files in the code (if you search for `prefix.isEmpty()` you can find the other occurrences which might be worth fixing as well. Submitting this change now.

gabzim:
@qualidafial I added the changes you requested and added a test.

kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/RegisterKotlinMapperImpl.kt
qualidafial: Good to know. These are likely relics from pre-3.0 release, when we were still figuring out how we wanted prefixes to work. I'll keep my eyes open for them.

qualidafial:
Thanks for the PR!

TheRealMarnes:
So is #1132 fixed now then?

qualidafial:
Looks like it. @gabzim can you confirm from a build on master?

gabzim:
I tested, it works. We can close #1132, however, I do have an issue but it's a separate problem. Namely that `JoinRowMapper` will *invoke every mapper for the row*, and if you're doing a `left join` that could mean that if you have users and addresses, you might have a user without an address, and in this case it will fail because the `KotlinMapper` will attempt to create the `Address` with a bunch of `null`s (which is ok with java but not with kotlin data classes unless you want to make all the fields nullable.

This is a separate issue with a join row mapper and kotlin objects. I should open a new ticket.

qualidafial:
I would recommend using a row reducer for outer joins.

qualidafial:
See http://jdbi.org/#__userowreducer

PR #1138, opened 2018-05-24
Default number format for Freemarker configuration

jarlah:
I need to have Freemarker use computer as default number format. It doesnt make sense to use thousand separators for numbers in sql. At least not implicitly. Its too much work having to always use ?c on all numbers to be used in queries.

### DIFF
--- a/freemarker/src/main/java/org/jdbi/v3/freemarker/FreemarkerSqlLocator.java
+++ b/freemarker/src/main/java/org/jdbi/v3/freemarker/FreemarkerSqlLocator.java
@@ -30,6 +30,7 @@ public class FreemarkerSqlLocator {
     static {
         Configuration c = new Configuration(Configuration.DEFAULT_INCOMPATIBLE_IMPROVEMENTS);
         c.setTemplateLoader(new ClassTemplateLoader(selectClassLoader(), "/"));
+        c.setNumberFormat("computer");
         CONFIGURATION = c;
     }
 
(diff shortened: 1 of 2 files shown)


qualidafial:
I'll merge this, but going forward we need a better test to justify these decisions. The tests we have right now are abusing the limits of what people should be doing with template engines.

In any app where I work, I would insist these values be a bound parameter (with a `?` or `:name` marker), not inlined into the SQL with a defined attribute.

`TemplateEngine` can be a vector for SQL injection attacks, so we should be very careful not to nudge people into using the template engine when a bound parameter is safer and more appropriate. The safest policy is to only use template engines for structural shaping of the SQL statement, e.g. including or excluding a where clause, changing the order by, etc.

jarlah:
Ive thought about using BindList. What we do most is actually to join either a list of strings or numbers with comma for use in select where in statements. Will BindList quote strings automatically?(I can check myself)

jarlah:
@qualidafial BindList is useless. I need to be able to know inside the template that the list is non empty. And BindList will as I can see generate either a,b,c or empty string. So if i have select * from B where B.id in (${ids}) and ids is an empty list, the sql will throw

jarlah:
I could of course check with freemarker that the string is non empty, of course. When I think about its a fairly good option.

jarlah:
?has_content

jarlah:
I used @BindList(onEmpty = BindList.EmptyHandling.NULL) coupled with <#if myList??> etc. Works nice.

stevenschlansker:
Turns out it's worth trying to use it a couple of different ways before discarding it as useless ;)
I recommend figuring out how to bind *all* your parameters.  It's worth the trouble in security peace of mind and early error detection.

jarlah:
Ye never mind the useless comment ;)

PR #1139, opened 2018-05-26
Allow flyway migrations from multiple locations

Sch3lp:


Sch3lp:
I was wondering why `getDataSource()` is not private.

I also noticed IntelliJ suggesting to make the `mutex` field final.

Thoughts?

qualidafial:
`JdbiRule` is intended to be extensible by third parties, to allow it to be used with various database vendors that the Jdbi project might not yet support. The flyway migrations idea uncovered that the `getJdbi()` method was the wrong design, so my branch makes `getDataSource()` the preferred method to override going forward. It must be `protected` in order to be overridden in third party subclasses.

Making the mutex final is probably a good idea.

Sch3lp:
`createDataSource()` is protected and should be overridden, but `getDataSource()` can remain private I think.

Update to this PR incoming!

qualidafial:
Thanks!

Sch3lp:
🙌

PR #1143, opened 2018-05-30
[doc] some extra sqllogger javadoc

TheRealMarnes:
It confused me for a second when I started using this class, and I'm the guy who wrote the whole thing, so I'm assuming it would confuse others too...

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlLogger.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlLogger.java
@@ -27,15 +27,27 @@ public interface SqlLogger {
     SqlLogger NOP_SQL_LOGGER = new SqlLogger() {
     };
 
+    /**
+     * Will be called before a query is executed. Note that no timing information will be available in the {@link StatementContext} yet.
+     */
     default void logBeforeExecution(StatementContext context) {
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jdbi/v3/core/statement/SqlLogger.java
```diff
@@ -27,15 +27,27 @@
...
 
+    /**
+     * Will be called after a query has failed. The {@link StatementContext} will contain a start {@link StatementContext#getExecutionMoment} and failure {@link StatementContext#getExceptionMoment} time.
+     */
     default void logException(StatementContext context, SQLException ex) {
     }
 
+    /**
+     * It's probably best not to override this method. It's only here due to architectural trade-off choices.
+     */
```
qualidafial: I kinda still want to make this method a static utility method instead of a default method.

TheRealMarnes: Where would you put it? Can't go on an interface.

TheRealMarnes: https://github.com/jdbi/jdbi/pull/1144

qualidafial: Perfect. Thanks

PR #1145, opened 2018-05-31
Update assertj-core to version 3.10.0 and assertj-guava to version 3.2.0

PascalSchumacher:


sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestInheritedAnnotations.java
```diff
@@ -68,7 +68,7 @@ public void testCrud() throws Exception {
     assertThat(dao.findById(1)).contains(new Character(1, "Moiraine Sedai", inserted, inserted));
 
     Instant modified = mockClock.advance(10, SECONDS);
-    assertThat(inserted).isLessThan(modified);
+    assertThat(inserted).isBefore(modified);
```
PascalSchumacher: `Instant` specific assertions where added some versions ago, so `Comparable` assertions can not be used for `Instant` anymore.

Issue #1154, opened 2018-06-01
Spring dependencies excluded when using jdbi3-bom


dermot-hardy:
In a simple project with no other dependencies I would expect this:

    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>org.jdbi</groupId>
                <artifactId>jdbi3-bom</artifactId>
                <version>3.2.1</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>

    <dependencies>
        <dependency>
            <groupId>org.jdbi</groupId>
            <artifactId>jdbi3-spring4</artifactId>
        </dependency>
    </dependencies>

to be equivalent to this:

    <dependencies>
        <dependency>
            <groupId>org.jdbi</groupId>
            <artifactId>jdbi3-spring4</artifactId>
            <version>3.2.1</version>
        </dependency>
    </dependencies>

But what I actually see is that when using the `jdbi3-bom` many of the Spring dependencies, like `spring-jdbc` for example, are not included.

You can see this by looking in the .war file produced by this simple project:
[pom.xml](https://github.com/jdbi/jdbi/files/2064351/pom.xml.txt)

Or by running `mvn dependency:list` on the project and observing that the Spring dependencies are marked `optional` when the bom is used but that they are not optional when the version is specified directly.

On a more complex project which uses spring uses this bom file causes spring dependencies to be excluded.

PR #1155, opened 2018-06-01
Drop optional flag from spring dependencies in dependency management

dermot-hardy:
Fix https://github.com/jdbi/jdbi/issues/1154


qualidafial:
Thanks @dermot-hardy!

PR #1157, opened 2018-06-02
[experiment] more categorized arguments/mappers for easier composition and more safety

TheRealMarnes:
So this is a POC I wanted to make, based on what I've done in my private project. It ties in to a discussion we previously had about avoiding presumptions about data type support in databases and saddling users up with undesired conversion behavior they'd need to weed out, as well as the `NaiveJavaTimePlugin` I was going to/asked to make, and the db-specific preconfiguration plugins that were brought up once as a desirable Jdbi feature.

At its core are the `PrimitivesPlugin`, `OptionalPrimitivesPlugin`, and `JdbcbasicsPlugin` (built on pure non-primitive jdbc-provided get/set methods, like for `String` and `java.sql` classes). This represents the very core java/sql conversions.

Beside those are the `LegacyTimePlugin` (`java.util.Date` et al), and the `UnreliableJavaTimePlugin` aka `NaiveJavaTimePlugin`.

Where it gets interesting is the `meta` package. This contains the `EssentialsPlugin` with a `controversial` option, to install the core converters and optionally the unsafe ones that are de facto standards but not always desirable. All other data types are blocked out for safety.

Beside that is the convenient `HsqldbPlugin`, which registers the core plugin and a special plugin that uses `getObject`/`setObject` without any types for a set of classes. Hsqldb has explicit support for doing this for some classes such as `UUID` and `java.time` instances, removing the need for any special mapping logic or fickle byte array messing. Installing this plugin tailors Jdbi to support 1:1 the same data types hsqldb supports, no more or less.
This is something I'm personally very fond of, and I think Jdbi would benefit from too. Other databases could get similar plugins, from people who are properly familiar with them and their conversion support. As you can see from the code, a structure like this makes it very easy to tailor a plugin to a database's specific type support.

It's just a POC so no tests, javadoc, or dependency version restrictions atm... I'll write tests later if this idea takes off, but I already know 99% of it will work because my private project basically contains what the HsqldbPlugin does (in less concentrated form), *with* a ton of integration tests to verify nullity and integrity.

I'm a little unsure how Optional support works in jdbi atm so I just included a quick plugin for it... And for enums too while I'm at it ¯\\\_(ツ)\_/¯ I'm not saying the code is perfect or anything yet though...

### DIFF
--- a/core/src/main/java/org/jdbi/v3/wip/EnumPlugin.java
+++ b/core/src/main/java/org/jdbi/v3/wip/EnumPlugin.java
@@ -1,0 +1,30 @@
+package org.jdbi.v3.wip;
+
+import java.sql.Types;
+import java.util.Optional;
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.spi.JdbiPlugin;
+
+public class EnumPlugin implements JdbiPlugin {
+	@Override
+	public void customizeJdbi(Jdbi jdbi) {
+		jdbi.registerArgument((type, value, config) -> {
+            if (type instanceof Class && Enum.class.isAssignableFrom((Class) type)) {
+                if (value == null) {
+                    return Optional.of((position, statement, ctx) -> statement.setNull(position, Types.VARCHAR));
+                }
+            }
+            return Optional.empty();
+        });
+
+		jdbi.registerColumnMapper((type, config) -> {
+            if (type instanceof Class && Enum.class.isAssignableFrom((Class) type)) {
+                return Optional.of((r, columnNumber, ctx) -> r.getObject(columnNumber) == null
+                    ? null
+                    : Enum.valueOf((Class) type, r.getString(columnNumber)));
+            } else {
+                return Optional.empty();
+            }
+        });
+	}
+}
--- a/core/src/main/java/org/jdbi/v3/wip/OptionalPlugin.java
+++ b/core/src/main/java/org/jdbi/v3/wip/OptionalPlugin.java
@@ -1,0 +1,34 @@
+package org.jdbi.v3.wip;
+
+import java.lang.reflect.Type;
+import java.util.Optional;
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.argument.Arguments;
+import org.jdbi.v3.core.generic.GenericTypes;
+import org.jdbi.v3.core.mapper.ColumnMappers;
+import org.jdbi.v3.core.spi.JdbiPlugin;
+
+public class OptionalPlugin implements JdbiPlugin {
+	@Override
+	public void customizeJdbi(Jdbi jdbi) {
+		jdbi.registerArgument((type, value, config) -> {
+			if (Optional.class.equals(GenericTypes.getErasedType(type))) {
+				Optional<Type> t = GenericTypes.findGenericParameter(type, Optional.class);
+				if (t.isPresent()) {
+					return config.get(Arguments.class).findFor(t.get(), value);
+				}
+			}
+			return Optional.empty();
+		});
+
+		jdbi.registerColumnMapper((type, config) -> {
+			if (Optional.class.equals(GenericTypes.getErasedType(type))) {
+				Optional<Type> t = GenericTypes.findGenericParameter(type, Optional.class);
+				if (t.isPresent()) {
+					return config.get(ColumnMappers.class).findFor(t.get());
+				}
+			}
+			return Optional.empty();
+		});
+	}
+}
--- a/core/src/main/java/org/jdbi/v3/wip/categories/OptionalPrimitivesPlugin.java
+++ b/core/src/main/java/org/jdbi/v3/wip/categories/OptionalPrimitivesPlugin.java
@@ -1,0 +1,80 @@
+package org.jdbi.v3.wip.categories;
+
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.Types;
+import java.util.Optional;
+import java.util.OptionalDouble;
+import java.util.OptionalInt;
+import java.util.OptionalLong;
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.generic.GenericType;
+import org.jdbi.v3.core.spi.JdbiPlugin;
+
+import static org.jdbi.v3.wip.categories.PluginUtils.registerClassArgument;
+import static org.jdbi.v3.wip.categories.PluginUtils.registerClassMapper;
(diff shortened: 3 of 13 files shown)


core/src/main/java/org/jdbi/v3/wip/EnumPlugin.java
```diff
@@ -0,0 +1,30 @@
...
+
+		jdbi.registerColumnMapper((type, config) -> {
+            if (type instanceof Class && Enum.class.isAssignableFrom((Class) type)) {
+                return Optional.of((r, columnNumber, ctx) -> r.getObject(columnNumber) == null
+                    ? null
+                    : Enum.valueOf((Class) type, r.getString(columnNumber)));
+            } else {
+                return Optional.empty();
+            }
+        });
```
qualidafial: See `EnumByNameMapperFactory`

core/src/main/java/org/jdbi/v3/wip/OptionalPlugin.java
```diff
@@ -0,0 +1,34 @@
...
+	public void customizeJdbi(Jdbi jdbi) {
+		jdbi.registerArgument((type, value, config) -> {
+			if (Optional.class.equals(GenericTypes.getErasedType(type))) {
+				Optional<Type> t = GenericTypes.findGenericParameter(type, Optional.class);
+				if (t.isPresent()) {
+					return config.get(Arguments.class).findFor(t.get(), value);
+				}
+			}
+			return Optional.empty();
+		});
```
qualidafial: See `BuiltInArgumentFactory`, line 165

core/src/main/java/org/jdbi/v3/wip/OptionalPlugin.java
```diff
@@ -0,0 +1,34 @@
...
+
+		jdbi.registerColumnMapper((type, config) -> {
+			if (Optional.class.equals(GenericTypes.getErasedType(type))) {
+				Optional<Type> t = GenericTypes.findGenericParameter(type, Optional.class);
+				if (t.isPresent()) {
+					return config.get(ColumnMappers.class).findFor(t.get());
+				}
+			}
+			return Optional.empty();
+		});
```
qualidafial: See `BuiltInMapperFactory`, line 107

core/src/main/java/org/jdbi/v3/wip/categories/OptionalPrimitivesPlugin.java
```diff
@@ -0,0 +1,80 @@
...
+
+import static org.jdbi.v3.wip.categories.PluginUtils.registerClassArgument;
+import static org.jdbi.v3.wip.categories.PluginUtils.registerClassMapper;
+import static org.jdbi.v3.wip.categories.PluginUtils.registerGenericTypeArgument;
+import static org.jdbi.v3.wip.categories.PluginUtils.registerGenericTypeMapper;
+
+public class OptionalPrimitivesPlugin implements JdbiPlugin {
+	@Override
+	public void customizeJdbi(Jdbi jdbi) {
+		registerOptionalArgument(jdbi, new GenericType<Optional<Boolean>>() {}, Types.BOOLEAN, PreparedStatement::setBoolean);
```
qualidafial: The existing `Optional` handling for arguments and column mapping takes the `T` parameter (from `Optional<T>` and just looks up an argument / column mapper for type `T`, and delegates to that argument or mapper.

qualidafial:
I like the basic idea of what you're doing, and I think the organization you've chosen will work.

I want to put a few guard rails around this effort:
- Don't break any existing code
- Everything that works through the built in arguments / mappers should work with the "controversial" plugin installed.
- DRY. If the same mapper / argument is implemented in both the built-ins and the plugin, refactor it to some shared location for reuse. Reused arguments and mappers could live in an internal package to avoid polluting the API.

TheRealMarnes:
Hm, I think we weren't on the same frequency, but I like yours better. It was my original wish, actually. You're saying you'd support this 2-phase plan?

- refactor the current `BuiltIn*` godfactories into smaller factories, and bundle them in a `JdbiOotbPlugin` that's installed by default. This alone changes **nothing** about jdbi's OOTB behavior, but makes the built-in factories easier to work with for devs by being more modular.
- add some competing factories like different primitive-null handling, the get/setObject one, etc, and pave the way for building fully tuned database-specific Plugins a la `HsqlDb_v2_4_1_Plugin`?

TheRealMarnes:
> I'm not sure how we'd divest the built in factories, since they are public types (darn it). We could deprecate them but I'd still like to see code reuse on the mappers / argument factories.

We could keep the classes and just make them empty boxes that delegate to the new factories, if you really want to preserve the BuiltIn*Factories as API points. The actual factory code can just be split up and moved to the new, more granular factories.
I can't imagine why someone would 1) refer to something obviously meant to be internal and 2) rely on it staying as it is though. I think if someone did, that's just a lesson for them, to not depend on library internals, or suffer having to adapt to updates. I don't think breaking the debatable API in this situation is an issue. I can't see anyone writing `jdbi.registerArgument(new BuiltInArgumentFactory())` in their own code.

> One difference with the primitives plugin is that you're throwing an exception when a primitive column is null, whereas the current built-in behavior is to return a zero/false value.

Well that code was just my case example, not the real deal... But yeah, this is why I want to break up the builtin factories in the first place. Both are valid and common approaches, I just want to remove the bias of one being registered by default while the other is DIY. Which approach to take should be a simple choice for the user, by passing a boolean or enum constant into jdbi or the factories, or just registering the other factory plugin. Actually, I think primitive+null handling is such a java-inherent thing, it might as well be a boolean option on the ConfigRegistry with a single Primitives*Factory behind it. It's not like this problem is gonna repeat itself for any other factories, and if it does, that factory can be configurable by itself.

> One other option we could pursue is using configuration to control visibility of certain mappers. So a factory could be registered, but it wouldn't resolve certain mappers / arguments if e.g. a config option was disabled.

Basically a configurable factory. I think that can already be done with the existing ConfigRegistry system, or internal state in the factory (though I think I'd prefer the former in this project). And this too would just be all the easier if we split up the BuiltIn*Factory godclasses.

TheRealMarnes:
On a related note, is a package cleanup (more packages with fewer classes in each, and better grouping of classes) on the distant todo list or is the idea dead in the water due to API compatibility, even for a new minor/major version?

TheRealMarnes:
I'm going to close this and work on a more serious approach next, with the feedback so far. This was a useful discussion :) I've got some question marks regarding breaking the public shadow-API, leaving shims to preserve compatibility vs removing deprecated things to favor clean code, etc, but I guess it's best to see to those problems when they occur...

PR #1158, opened 2018-06-03
PMD #0

TheRealMarnes:
_christopher nolan movie droning rumble sound_

I prepared this a few days ago. Updated the pmd plugin to the latest, added it to the pom to run for all module builds. Added a ruleset that activates all rules and then disabled those that jdbi triggered. Idea is to make additional PRs enabling a few rules at a time with the corresponding code fixes and removing overlaps from checkstyle.


qualidafial:
Thanks!

Issue #1096, opened 2018-04-23
SqlObject plugin and BindBean: is there a way to specify java.sql.Types for fields/properties?


igortochansky:
For example, 
```java
@lombok.Data
class Address {
    // It is java.sql.Types.VARCHAR
    // JDBI uses VARCHAR by default for strings, so no action is needed
    private final String country;

    // It has to be java.sql.Types.NVARCHAR
    // Is there some way to specify it?
    // It seems that now there is no a good option. There are two workarounds:
    // 1) Create a new class NString and provide an argument factory for it.
    // 2) Use two annotations (see jpa plugin as example: @BindJpa and @Column)
    private final String city;
}

igortochansky:
`@SqlType` will solve the problem but perhaps it would be more general if `@RegisterArgumentFactory`  may be attached to bean properties (and parameters of Dao's method).

qualidafial:
The problem with `@RegisterArgumentFactory` is scope. Once you register an argument factory that works on `String`, it will use that argument for all strings--not just the one you annotated.

I'm not sure that `@SqlType` will do the trick either. You have to call e.g. `setNString()` instead of `setString()` for NVARCHAR types.

I've had a similar issue with `MACADDR` data types in Postgres. Our solution was to introduce a `MacAddress` class and register an argument factory and a column mapper for that type.

I'm trying to imagine how we might use a qualifying annotation system (similar to javax.inject qualifying annotations like `@Named("foo")`, but that would require us to enhance the existing factory APIs to recognize those annotations.

Unfortunately, I think spinning up an `NString` class may be your best option at the moment.

igortochansky:
@qualidafial, in fact I can change DB schema now. So in my current case I will just replace all `varchars` with `nvarchars` in DDL of my tables. Then I will override default argument factory for strings.
However, usually it is not an option and people need to invent something like `NString`, `MacAddress` or pair of custom annotations. In JPA plugin it is  `@BindJpa` and `@Column`.  Actually, i see, that in our projects based on jdbi v2 a similar approach is already used. There are two annotations `@BindSqlType` for bean's properties and `@BindEntity` marked with `org.skife.jdbi.v2.sqlobject.BindingAnnotation`.
It works but it does not look good because you need to write some custom code and also should remember about both annotations:
```java
@lombok.Data
class Entity {
    @BindSqlType(NVARCHAR)
    private String comments;
}

interface Dao {
[...]

PR #1160, opened 2018-06-04
Qualified types for arguments and column mappers

qualidafial:
See #1096 for context.

We have some situations where the same Java type should be bound or mapped differently depending on context.

- Most strings can be bound with `VARCHAR`, but sometimes you need to bind with `NVARCHAR`.
- Postgres has a `MACADDR` data type that can be mapped to a Java `String`, but it has to be bound with `setObject(PgObject)`.
- Sometimes `Map<String, String>` refers to N key-value rows, and sometimes it refers to an `HSTORE` column from a single row (e.g. in a Postgres table).

My solution is inspired by JSR-330's `@javax.inject.Qualifier` meta-annotation. `@Qualifier` makes it possible to qualify a type so that the following "qualified" types are distinct:

* `String`
* `@NVarchar String`

Changes in this branch:

- New `core.qualifier` package:
  - `@Qualifier` meta-annotation, for identifying annotation types that are treated as type qualifiers. Inspired by JSR-330's `@Qualifier` annotation.
  - `@NVarchar` qualifying annotation for Strings stored in NVARCHAR columns.
  - `QualifiedType` - represents a `java.lang.reflect.Type` with a `Set<Annotation>` of qualifying annotations.
  - `Qualifiers` - utility class for working with qualifying annotations
    - `nVarchar()` static method provides an `NVarchar` instance. Useful for working with `QualifiedType.of(type, qualifiers...)`, `ResultBearing.mapTo(type, qualifiers)`, etc.
    - `getQualifiers(AnnotatedElement...)` method inspects a set of Java elements and returns all the annotations on those elements whose annotation types are annotated with `@Qualifier`.
- `core.argument` package
  - `Arguments.findFor(QualifiedType type, Object value)`
  - `StatementContext.findArgumentFor(QualifiedType, Object value)`
  - Built-in argument factory for `@NVarchar String` types.
  - `SqlStatement.bindBean()`, `.bindMethods()`, and `.bindFields()` (as well as their SqlObject annotation counterparts) honor qualifying annotations on properties, methods, and fields, respectively.
  - Qualifying annotations on an `ArgumentFactory` class are honored.
- Changes to `core.mapper` package
  - `ColumnMappers.register(QualifiedType, ColumnMapper<?>)`
  - `ColumnMappers.findFor(QualifiedType)`
  - `StatementContext.findColumnMapperFor(QualifiedType)`
  - `Configurable.registerColumnMapper(QualifiedType, ColumnMapper<?>)`
  - Built-in column mapper for `@NVarchar String` types.
  - `BeanMapper`, `ConstructorMapper`, and `FieldMapper` honor qualifying annotations on getters/setters/setter params, constructor parameters, and fields, respectively.
  - Qualifying annotations on a `ColumnMapperFactory` class are honored.
  - Qualifying annotations on an inferred column mapper class are honored.
- `SqlStatement` method additions:
  - `bindNVarchar(String)`
  - `bindByType(position, value, QualifiedType)`
  - `bindByType(name, value, QualifiedType)`
- Changes to `core.result` package
  - `ResultBearing.mapTo(Class<T>, Annotation...)`
  - `ResultBearing.mapTo(GenericType<T>, Annotation...)`
  - `ResultBearing.mapTo(QualifiedType)`
- SQL Object
  - Honors qualifying annotations on SQL Object method (qualifier applies to return type)
  - Honors qualifying annotations on SQL Object method parameters
- Kotlin
  - `SqlStatement.bindKotlin()` extension methods
  - `KotlinMapper` honors qualifying annotations on constructor parameters and properties.
- Kotlin SQL Object
  - Plugin honors qualifying annotations on SQL Object methods (applies to return type) and parameters
  - `@BindKotlin` annotation
- JPA
  - `JpaMapper` and `@BindJpa` honor qualifying annotations on fields, getters, setters, or setter params
-  Postgres
    - `@HStore` qualifying annotation for `Map` values stored in HSTORE columns. Binding is supported for any kind of `Map`, and mapping is supported to `Map<String, String>`.
    - `@MacAddr` qualifying annotation for String values stored in MACADDR columns.
    - `PostgresQualifiers` static utility class with `hStore()` and `macAddr()` methods.


TheRealMarnes:
Before reading any code, the general idea sounds ok to me. A bit of a first-world-problem since wrapper classes can have the same net effect, but if the solution is non-intrusive, then by all means, I welcome it. I have asked myself the question all this in the past too.

You mention classes containing consts matching concrete qualifiers like @NVarchar. This sounds like conflicting ideas: there can be any number of such qualifiers/annotations but only a fixed set of those constants. The bindNVarchar method is in the same boat. Why offer it if there cannot be such methods for all of them (since many will be enduser-made)? I've never been fond of biased convenience methods like that when the generic base method is easy enough to use. Thats just me though. Reading the code in detail might reveal some necessity, or we can just leave it at "we wanna provide the most common needs, kitchen sink (constants) included", like how the built-in data type support in all its forms exists.

From what I can tell, this whole feature is sort of a sideline to the existing features and doesn't get in the way of people who choose to ignore it, like how the sqlobject and fluent apis dont get in each other's way. Yay +1

qualidafial:
> The bindNVarchar method is in the same boat. Why offer it if there cannot be such methods for all of them (since many will be enduser-made)?

Binding NVARCHAR is specifically supported by JDBC--that's pretty much where SqlStatement binding methods draw the line.

qualidafial:
I'm wrestling with a question about where the qualifying annotation should go, when reasoning about container-ish type signatures. This applies to both SQL arrays (SqlArrayTypeFactory) and container types (CollectorFactory).

Consider these annotated type signatures:

* `@MacAddr List<String>`
* `List<@MacAddr String>`

The `@MacAddr` qualifier applies to the element type `String` more than to the container type `List`, so the latter form seems more correct.

Note: we may have to make an exception for types like `@MacAddr String[]` since there's no way to annotate the element type of an array.

The problem is, none of the standard sub-interfaces of `java.lang.reflect.Type` provide API to get at annotations. You can only access it through `AnnotatedType`, which does not extend `Type`. Parameter annotations can only be retrieved through the subinterface `AnnotatedParameterizedType`.

That's a major problem, since all our factories use `java.lang.reflect.Type`. This is looking more and more like a breaking change to support this feature, at least if we want the annotations in the "proper" location.

Moreover, none of this plays nice with what we already have in the GenericTypes utility class.

@jdbi/contributors I'm looking for ideas / feedback. Can you recommend any workarounds to the current problem? Is this ticket worth further pursuit or do we abandon this until it's time for 4.x?

qualidafial:
@stevenschlansker @arteam @brianm any feedback?

stevenschlansker:
Would it be possible to create our own subtype of `Type` that holds the annotation data?  That's the only thing that comes to my mind.

I think that 'misplacing' the `@MacAddr` annotation is an acceptable compromise, if that ends up being helpful in making the implementation work.  But any bigger warts and we might decide this has to wait for 4.0, or better JDK support for reflection on annotated types...

Last thought, would an annotation processor based SqlObject be able to work around these reflection shortcomings?  I'd suspect the `javax.model` stuff is pretty thorough.

qualidafial:
I'm not really sure--I've never worked on an annotation processor. It's an interesting idea though.

qualidafial:
Another possible compromise is just to elide qualified type support on SQL arrays for now. That's where things really get messy.

stevenschlansker:
I'm +1 to punting on qualified types for arrays.  Let's not let perfect be the enemy of good.

qualidafial:
@jdbi/contributors I've managed to clean up the public API surface area dramatically.

The new factories for qualified types (`QualifiedArgumentFactory` and `QualifiedColumnMapperFactory`) were made internal classes. Instead of adding a new public interface, I'm scanning the classes ArgumentFactory, ColumnMapper, and ColumnMapperFactory instances as they are registered, and wrapping them in a qualified factory that only matches types with that set of qualifiers.

I've updated the PR description accordingly.

The remaining tasks are to:
- [ ] Put together a list of test scenarios
- [ ] Implement those tests
- [ ] Document qualified types in the developer guide.

Please take a look at the current changes in the PR and tell me if you spot any problems.

qualidafial:
Test scenarios for qualified types--includes both binding and mapping:

- Core
  - Qualified bean property (`BeanMapper`, `SqlStatement.bindBean()`)
    - [ ] Qualifier on getter
    - [ ] Qualifier on setter
    - [ ] Qualifier on setter param
  - [ ] Qualifier on method (`SqlStatement.bindMethods()`)
  - [ ] Qualifier on field (`FieldMapper`, `SqlStatement.bindFields()`)
  - [ ] Qualifier on constructor parameter (`ConstructorMapper`)
  - Sanity check new public API:
    - [ ] Arguments.findFor(QualifiedType, Object)
    - [ ] ColumnMappers.findFor(QualifiedType)
    - [ ] ColumnMappers.register(GenericType<T>, ColumnMapper<T>) - tangential, noticed this was missing from API
    - [ ] ColumnMappers.register(QualifiedType, ColumnMapper<?>)
    - [ ] Configurable.registerColumnMapper(GenericType<T>, ColumnMapper<T>) - tangential, noticed this was missing from API
    - [ ] Configurable.registerColumnMapper(QualifiedType, ColumnMapper<?>)
    - [ ] Configurable.registerRowMapper(GenericType<T>, RowMapper<T>) - tangential, noticed this was missing from API
    - [ ] Mappers.findFor(QualifiedType)
    - [ ] ResultBearing.mapTo(Class<T>, Annotation...)
    - [ ] ResultBearing.mapTo(GenericType<T>, Annotation...)
    - [ ] ResultBearing.mapTo(QualifiedType)
    - [ ] RowMappers.register(GenericType<T>, RowMapper<T>) - tangential, noticed this was missing from API
    - [ ] SqlStatement.bindByType(int, Object, QualifiedType)
    - [ ] SqlStatement.bindByType(String, Object, QualifiedType)
    - [ ] SqlStatement.bindNVarchar(int, String)
    - [ ] SqlStatement.bindNVarchar(String, String)
    - [ ] StatementContext.findArgumentFor(QualifiedType, Object)
    - [ ] StatementContext.findColumnMapperFor(QualifiedType)
    - [ ] StatementContext.findMapperFor(QualifiedType)
    - [ ] `QualifiedType` class
    - [ ] `Qualifiers` static methods
  - [ ] `@NVarchar String` built-in qualifier
  - Custom qualifying annotation
    - [ ] Qualifier on ColumnMapper class
    - [ ] Qualifier on ColumnMapperFactory class
    - [ ] Mapper registered with qualified type: `registerColumnMapper(QualifiedType, ColumnMapper)`
    - [ ] Qualifier on ArgumentFactory class
    - [ ] Works on SQL Object return type
    - [ ] Works on SQL Object method parameters
  - [ ] Multiple qualifiers - should only match if the set of qualifiers is identical.
- SQL Object
  - Qualified binding parameter
    - [ ] General case qualified bindings: updates, selects
    - [ ] Qualifiers on `@SqlBatch` iterable parameters are applied to element type: `@Foo List<String>` treated as `@Foo String`
  - Qualified return type
    - Multi-element results applies qualifying annotation to element type: `@Foo List<String>` treated as collection type `List` with element type `@Foo String`
      - [ ] Collection types
      - [ ] ResultIterable type
      - [ ] ResultIterator type
      - [ ] Iterator type
      - [ ] Stream type
    - Single-value results - applies qualifiers directly to return type
      - [ ] Intrinsically single-value type
      - [ ] `@SingleValue` for collection-like return types
  - [ ] Qualified `Consumer<T>` parameter: `@Foo Consumer<String>` treated as `@Foo String` element type.
  - `@BindBean`, `@RegisterBeanMapper`
    - [ ] Qualifier on getter
    - [ ] Qualifier on setter
    - [ ] Qualifier on setter parameter
  - [ ] `@BindMethods` with qualified method
  - [ ] `@RegisterConstructorMapper` with qualified constructor parameter
  - [ ] `@BindFields`, `@RegisterFieldMapper` with qualified field
  - [ ] `@MapTo` supports `QualifiedType` arguments
- JPA (`JpaMapper`, `@BindJpa`)
  - [ ] Qualifier on field
  - [ ] Qualifier on getter
  - [ ] Qualifier on setter
  - [ ] Qualifier on setter param
- Kotlin
  - [ ] Qualified on constructor parameter
  - [ ] Qualified lateinit property
  - [ ] Qualified var property
  - [ ] `getQualifiers(vararg KAnnotatedElement?)` helper method
  - [ ] SQL Object support on par
- Postgres
  - [ ] `PostgresQualifiers` static methods
  - [ ] `@MacAddr String`
  - [ ] `@HStore Map<String, String>`
  - [ ] Binding (but not mapping) `@HStore Map<Whatever>`

Things that should be documented in the developer guide:

- [ ] Cannot bind positional parameters with `Handle.execute(String sql, Object... args)`. Have to use `Handle.createUpdte(sql).bindByType(...args).execute()` to bind a qualified parameter type.

qualidafial:
Follow up ideas after this PR:

- Is it possible to support qualifying annotation on reified `ResultBearing.mapTo()` method? e.g. `handle.select("...").mapTo<@MacAddr String>()`
- Jdbi and Handle extension methods in `jdbi3-kotlin-sqlobject` should probably have been in `jdbi3-kotlin`, since they extend only core APIs

qualidafial:
Topics to cover in developer guide (WIP):

- [ ] Core
  - [ ] Explain type qualifiers
    - [ ] Allows Jdbi to distinguish between different database data types that map to the same Java type
    - [ ] Core provides out of the box support for `@NVarchar String`, mapping `NVARCHAR` columns
    - [ ] `Handle.execute(String sql, Object... args)` does not work with qualifying annotations. If you need to qualify a parameter, use `Handle.createUpdate(sql).bindByType(col, value, QualifiedType).execute()`
  - [ ] `bindBean()`, `mapToBean()` and `BeanMapper` honor qualifying annotations on property getters, setters, or setter parameters
    - [ ] Hint: Lombok-generated `@Data` or `@Value` classes should use `@Getter(onMethod_ = @SomeQualifier)`, `@Setter(onMethod_ = @SomeQualifier)`, or `@Setter(onParam_ = @SomeQualifier)` in order for the qualifier to be recognized.
  - [ ] `ConstructorMapper` honors qualifying annotations on constructor parameters
  - [ ] `bindMethods()` honors qualifying annotations on methods
  - [ ] `bindFields()`, `FieldMapper` honor qualifying annotations on fields
  - [ ] SQL Arrays do not yet support type qualifiers
- [ ] SQL Object
  - [ ] Qualifiers on SQL Object methods apply to the _element type_ of the method return type.
  - [ ] Qualifiers on SQL Object method parameter apply to the type of the parameter, or to the element type of the parameter for `@SqlBatch` methods.
  - [ ] Qualifiers on SQL Object method parameters of type `Consumer<T>` apply to the element type `T`.
  - [ ] `@MapTo` methods parameter now accept a `QualifiedType` parameter in addition to `Type` or `GenericType`.
  - [ ] `@BindBean`, `@RegisterBeanMapper` honors qualifying annotations on property getters, setters, or setter parameters
    - [ ] Hint: Lombok-generated `@Data` or `@Value` classes should use `@Getter(onMethod_ = @SomeQualifier)`, `@Setter(onMethod_ = @SomeQualifier)`, or `@Setter(onParam_ = @SomeQualifier)` in order for the qualifier to be recognized.
  - [ ] `@RegisterConstructorMapper` honors qualifying annotations on constructor parameters
  - [ ] `@BindMethods` honors qualifying annotations on methods
  - [ ] `@BindFields`, `@RegisterFieldMapper` honors qualifying annotations on fields
- [ ] JPA
  - [ ] `@BindJpa`, `JpaMapper` honor qualifying annotations on getters and setters
- [ ] Kotlin
  - [ ] `bindKotlin()`, `@BindKotlin`, and `KotlinMapper` honor qualifying annotations on constructor parameters, getters, setters, or setter parameters
- [ ] Postgres
  - [ ] Support for mapping and binding `MACADDR` columsn to `@MacAddr String` Java type
  - [ ] Support for mapping `HSTORE` columns to `@HStore Map<String, String>` Java type.
  - [ ] Support for binding any `@HStore Map` Java type (generic or raw) to `HSTORE` columns.
- [ ] Advanced
  - [ ] Custom type qualifiers
    - [ ] Defining the qualifying annotation (`@Qualifier`, `@Target`, `@Retention(RUNTIME)`)
    - [ ] Annotating a class that `ColumnMapper`, `ColumnMapperFactory`, or `ArgumentFactory` class with a qualifying annotation causes that mapper or factory to only apply to qualified types with that same qualifier or set of qualifiers.

qualidafial:
@jdbi/contributors Could any project members review this? I know it's a massive PR, but it should be reasonably straightforward to review as most of the code changes are new tests.

Also: I was at a conference this week and had the opportunity to meet with @stevenschlansker to discuss Jdbi. Our consensus was that we should probably split out the documentation effort into another issue and tackle that separately, to avoid ongoing merge conflicts in this branch. Is everybody on board with that?

TheRealMarnes:
It's high on my todo list atm as I'd really like to see this merged too, I just need to find the time and be able to clear my mind of other things... Possibly this weekend?

core/src/main/java/org/jdbi/v3/core/argument/QualifiedArgumentFactory.java
```diff
@@ -0,0 +1,37 @@
...
+import org.jdbi.v3.core.qualifier.QualifiedType;
+import org.jdbi.v3.meta.Beta;
+
+@FunctionalInterface
+@Beta
+interface QualifiedArgumentFactory {
+    Optional<Argument> build(QualifiedType type, Object value, ConfigRegistry config);
+
+    static QualifiedArgumentFactory adapt(ArgumentFactory factory) {
+        Set<Annotation> qualifiers = getQualifiers(factory.getClass());
```
stevenschlansker: should this have a check
```java
if (factory instanceof QualifiedArgumentFactory) {
    return factory;
}
```
in case someone registers a `QualifiedArgumentFactory` directly?

core/src/main/java/org/jdbi/v3/meta/Beta.java
```diff
@@ -25,15 +25,23 @@
...
  *
- * @see <a href="https://google.github.io/guava/releases/15.0/api/docs/com/google/common/annotations/Beta.html">Courtesy of Kevin Bourrillion's @Beta at Guava</a>
+ * @see <a
+ *     href="https://google.github.io/guava/releases/15.0/api/docs/com/google/common/annotations/Beta.html">Courtesy
+ *     of Kevin Bourrillion's @Beta at Guava</a>
  */
 @Documented
 @Inherited
 @Target({TYPE, FIELD, METHOD, PARAMETER, CONSTRUCTOR, ANNOTATION_TYPE})
+@Beta
```
stevenschlansker: So meta.
Why should Beta annotate itself?  I'm surprised this is even allowed!

qualidafial: Ha, I tried it out for grins and thought it was funny that it worked.. and then forgot about it. I can take it out if you like.

Or we could say use it to say that `@Beta` is reserved for the library itself?

core/src/main/java/org/jdbi/v3/core/argument/QualifiedArgumentFactory.java
qualidafial: `QualifiedArgumentFactory` is package private, and `Arguments.register(QualifiedArgumentFactory)` method is private, so nobody should be able to implement it, let alone register it--not yet anyway.

core/src/test/java/org/jdbi/v3/core/qualifier/QualifiedConstructorParamThing.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, name);
+    }
+
+    @Override
+    public String toString() {
+        return "QualifiedConstructorParamThing{"
```
TheRealMarnes: ugh, manual json-like tostring building... Can we include a tool like ToStringBuilder.reflectionToString to squash all of this debug tostring building logic?

TheRealMarnes: should probably go in a separate pr

qualidafial: This is all IDE-generated code.

I've actually thought about using lombok to simplify our tests. Wouldn't hurt to have some tests to capture lombok common usages too.

PR #1161, opened 2018-06-04
Additional pom fix for build time pom change

isaki-x:
Recent changes to JDBI have introduced another case where the build itself creates a modified pom. Rather than having to blow the file away after every build so I can keep merging master into my local fork, I have simply committed the post-build POM file.


TheRealMarnes:
Ah, yeah, sorry about that. Should've run sortpom as part of the PR

PR #1153, opened 2018-06-01
Fixed for POM conflict issue

isaki-x:
Every time I build and install the latest JDBI3 snapshot, the build changes `core/pom.xml` so when I go to pull the latest commits, git complains that my changes to `core/pom.xml` will be overwritten. This requires me to have to `git checkout -- core/pom.xml` before every pull.

I have been able to reproduce this on an Ubuntu 16.04.4LTS Sever:

```
Apache Maven 3.3.9
Maven home: /usr/share/maven
Java version: 1.8.0_171, vendor: Oracle Corporation
Java home: /usr/lib/jvm/java-8-openjdk-amd64/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "4.4.0-127-generic", arch: "amd64", family: "unix"
```

As well as my Mac:

```
$ mvn --version
Apache Maven 3.5.3 (3383c37e1f9e9b3bc3df5050c29c8aff9f295297; 2018-02-24T14:49:05-05:00)
Maven home: <redacted>/apache-maven-3.5.3
Java version: 1.8.0_171, vendor: Oracle Corporation
Java home: /Library/Java/JavaVirtualMachines/jdk1.8.0_171.jdk/Contents/Home/jre
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "10.13.4", arch: "x86_64", family: "mac"
```

As such, I have simply committed the change made by the build process so this never happens again.

Thanks.

PR #1162, opened 2018-06-04
[housekeeping] pom checkup

TheRealMarnes:
I'm no maven expert, but I believe it's universally bad form to specify inline versions anywhere but the dependencyManagement section of the parent pom, or the properties section for child poms to refer to for plugin versions.

Also moved the checkstyle plugin to what I think is a more correct location and bound it explicitly, along with pmd and sortpom, to a proper lifecycle.

Like I said though, not a maven pro. Have worked with the thing for over 6 years but some of it still confuses me.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -21,7 +21,6 @@
         <version>3.2.2-SNAPSHOT</version>
     </parent>
 
-    <groupId>org.jdbi</groupId>
     <artifactId>jdbi3-bom</artifactId>
     <packaging>pom</packaging>
     <name>jdbi3 BOM</name>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -114,11 +114,6 @@
         <pluginManagement>
             <plugins>
                 <plugin>
-                    <groupId>org.antlr</groupId>
-                    <artifactId>antlr3-maven-plugin</artifactId>
-                    <version>${dep.antlr.version}</version>
-                </plugin>
-                <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-javadoc-plugin</artifactId>
                     <configuration>
@@ -148,7 +143,6 @@
                 <plugin>
                     <groupId>org.eclipse.m2e</groupId>
                     <artifactId>lifecycle-mapping</artifactId>
-                    <version>1.0.0</version>
                     <configuration>
                         <lifecycleMappingMetadata>
                             <pluginExecutions>
--- a/pom.xml
+++ b/pom.xml
@@ -196,6 +196,11 @@
                 <artifactId>jdbi3-vavr</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.jdbi</groupId>
+                <artifactId>jdbi3-oracle12</artifactId>
+                <version>${project.version}</version>
+            </dependency>
 
             <dependency>
                 <groupId>com.google.guava</groupId>
@@ -374,6 +379,42 @@
                 <groupId>org.xerial</groupId>
                 <artifactId>sqlite-jdbc</artifactId>
                 <version>3.8.11.2</version>
+            </dependency>
+
+            <dependency>
+                <groupId>org.antlr</groupId>
+                <artifactId>stringtemplate</artifactId>
+                <version>4.0.2</version>
+            </dependency>
+
+            <dependency>
+                <groupId>javax.annotation</groupId>
+                <artifactId>javax.annotation-api</artifactId>
+                <version>1.3.1</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.oracle.jdbc</groupId>
+                <artifactId>ojdbc7</artifactId>
+                <version>12.1.0.2</version>
+            </dependency>
+
+            <dependency>
+                <groupId>javax.persistence</groupId>
+                <artifactId>persistence-api</artifactId>
+                <version>1.0.2</version>
+            </dependency>
+
+            <dependency>
+                <groupId>joda-time</groupId>
+                <artifactId>joda-time</artifactId>
+                <version>2.9.3</version>
+            </dependency>
+
+            <dependency>
+                <groupId>org.freemarker</groupId>
+                <artifactId>freemarker</artifactId>
+                <version>2.3.28</version>
             </dependency>
 
             <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -506,13 +506,42 @@
                         <jdkVersion>8</jdkVersion>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.antlr</groupId>
+                    <artifactId>antlr3-maven-plugin</artifactId>
+                    <version>${dep.antlr.version}</version>
+                </plugin>
+                <plugin>
+                    <groupId>org.eclipse.m2e</groupId>
+                    <artifactId>lifecycle-mapping</artifactId>
+                    <version>1.0.0</version>
+                </plugin>
+                <plugin>
+                    <groupId>org.jetbrains.kotlin</groupId>
+                    <artifactId>kotlin-maven-plugin</artifactId>
+                    <version>${dep.kotlin.version}</version>
+                </plugin>
+                <plugin>
+                    <groupId>org.asciidoctor</groupId>
+                    <artifactId>asciidoctor-maven-plugin</artifactId>
+                    <version>1.5.3</version>
+                </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-scm-publish-plugin</artifactId>
+                    <version>1.1</version>
+                </plugin>
+                <plugin>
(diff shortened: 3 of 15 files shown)


bom/pom.xml
```diff
@@ -21,15 +21,10 @@
...
-    <groupId>org.jdbi</groupId>
     <artifactId>jdbi3-bom</artifactId>
     <packaging>pom</packaging>
     <name>jdbi3 BOM</name>
-    <description>Jdbi is designed to provide convenient tabular data access in
-        Java(tm). It uses the Java collections framework for query
-        results, provides a convenient means of externalizing sql
-        statements, and provides named parameter support for any database
-        being used.</description>
+    <description>Jdbi is designed to provide convenient tabular data access in Java(tm). It uses the Java collections framework for query results, provides a convenient means of externalizing sql statements, and provides named parameter support for any database being used.</description>
```
stevenschlansker: is this reformatting helpful?  it's a lot harder to read in diffs now (github is fixed width)

core/pom.xml
```diff
@@ -148,7 +148,7 @@
                 <plugin>
                     <groupId>org.eclipse.m2e</groupId>
                     <artifactId>lifecycle-mapping</artifactId>
-                    <version>1.0.0</version>
+                    <version>${dep.plugin.lifecycle.mapping.version}</version>
```
stevenschlansker: this plugin doesn't even actually exist ;)

TheRealMarnes:
Question for someone more knowledgeable than me: can you safely use the parent pom's pluginManagement section to lock plugin versions for child poms, just as you do with dependencyManagement versions? I have this idea in my head that specifying a plugin in pluginManagement that way also actually causes the plugin to be run, unlike what happens with dependencyManagement. I mean, I put checkstyle in pluginManagement initially and it actually ran checkstyle instead of merely locking the configuration.

bom/pom.xml
TheRealMarnes: I guess not, I just thought the source formatting would mess up formatting in displays elsewhere. Guess I should've asked first

core/pom.xml
TheRealMarnes: it was there before I touched it ¯\\\_(ツ)\_/¯ There's more than 1 red marker in these poms... I'm not sure which are intellij brainfarts and which are legit errors, or what some of these plugins are for

stevenschlansker:
I believe that's exactly what the `management` sections are for -- explicitly inheriting configuration down to child poms *without* actually instantiating the element.
So much like `dependencyManagement` is insufficient to bring in a dep, `pluginManagement` does not actually cause the plugin to run.  That's coming from a "real" plugin declaration: https://github.com/basepom/basepom/blob/master/foundation/pom.xml#L910-L922

TheRealMarnes:
Hold off on merging atm please, I'm fixing some stuff up

TheRealMarnes:
> I believe that's exactly what the management sections are for -- explicitly inheriting configuration down to child poms without actually instantiating the element.

That's what I want to believe but it doesn't seem to work that way, plugins in pluginManagement always seem to actually run. I must be missing something. All I can think of is that you maybe need to specify `<configuration><skip>true</skip></configuration>` in the parent and set a proper binding where you want the plugin to really run. :/

bom/pom.xml
TheRealMarnes: reverted

stevenschlansker:
Did you see where I link to the explicit `<plugin>` that I think is actually causing what you are observing?

TheRealMarnes:
Ah.

Well, my hatred for inheritance, especially in poms, grows stronger.

stevenschlansker:
It can be a bit of a mess, and Maven surely has its peculiarities and weaknesses.  But I argue it's still overall the nicest build system I've worked with, once you invest in setting it up properly.

TheRealMarnes:
With the basepom defining checkstyle in the live section, does it make any difference where I put our tuned checkstyle config, in the management or the live section? I'd think putting a plugin in the live section in jdbi-parent makes it also execute for all jdbi child poms, just as the child poms inherit the dependencies from jdbi-parent. So overriding the basepom's checkstyle config in our management section or the live section seems to me like it would give the same results.

TheRealMarnes:
Hmm, so this also means I should move version declarations on plugins in the child poms to the pluginManagement section of our parent pom, right?

stevenschlansker:
General rule is, configuration always goes in `pluginManagement` and execution binding goes in `plugins` (i.e. anything that can be managed, should be)

TheRealMarnes:
Hmm, that's a rule of thumb that makes sense to me :)

TheRealMarnes:
Do I remove the lifecycle-mapping plugin altogether? I'm reading conflicting advice online, though people seem to say you can just remove it and use non-intrusive alternatives instead

stevenschlansker:
Let's leave it for now -- it's helpful to Eclipse users (me).
I'm happy to remove it if there's a better way but let's do that as a separate PR where I can test my IDE doesn't get unhappy.

stevenschlansker:
Right, that advice applies specifically to the `build-helper` plugin.  Other plugins might also have proper m2e lifecycle mappings now, and similarly should then be removed, but we have to actually verify it doesn't break anything.  So it's something to consider but definitely a separate PR.

TheRealMarnes:
I feel like I've drifted off-topic with my changes :/ All I really wanted to do was consolidate version number specifications to avoid conflicts. Now I've moved plugin configuration around and stuff, all of which *seems* like a fine cleanup to me but might subtly make something work badly...

TheRealMarnes:
For example, I see this and can't help but wonder if this was done on purpose to fix some problem or if it just slipped someone's mind: 
image

core/pom.xml
```diff
@@ -162,7 +156,7 @@
                                         </goals>
                                     </pluginExecutionFilter>
                                     <action>
-                                        <ignore></ignore>
+                                        <ignore />
```
PascalSchumacher: This change reverts https://github.com/jdbi/jdbi/pull/1153

TheRealMarnes: yeah, but by fixing it at the core. Empty xml tags should in general be written as closed tags instead of open tags without content. Sortpom by default expands them, which is against that convention. Along with the change in sortpom's config elsewhere in this PR, this issue will be properly gone.

I'm re-doing all the work in this branch right now though, it's become a mess.

TheRealMarnes:
Alright :) @stevenschlansker I've trimmed my changes down to be more conservative and not take any risks. I think the 6 small commits are self-explanatory. I'm happy leaving the poms as they are now, even though there are still some warnings here and there in intellij (but changing any of those just causes issues on jenkins and whatnot)

@PascalSchumacher the bit that concerned you is gone from this, it moved here: https://github.com/jdbi/jdbi/pull/1163

pom.xml
```diff
@@ -196,6 +197,11 @@
                 <artifactId>jdbi3-vavr</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.jdbi</groupId>
+                <artifactId>jdbi3-oracle12</artifactId>
+                <version>${project.version}</version>
+            </dependency>
```
qualidafial: I would move this to the oracle profile.

pom.xml
```diff
@@ -376,6 +382,42 @@
...
+                <groupId>javax.annotation</groupId>
+                <artifactId>javax.annotation-api</artifactId>
+                <version>1.3.1</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.oracle.jdbc</groupId>
+                <artifactId>ojdbc7</artifactId>
+                <version>12.1.0.2</version>
+            </dependency>
```
qualidafial: This should likewise be moved to the Oracle profile.

TheRealMarnes:
@qualidafial done. Seems so obvious in hindsight how certain things about poms work...

pom.xml
```diff
@@ -465,51 +486,109 @@
...
         </pluginManagement>
         <plugins>
             <plugin>
                 <groupId>com.github.ekryd.sortpom</groupId>
                 <artifactId>sortpom-maven-plugin</artifactId>
-                <version>2.8.0</version>
                 <executions>
                     <execution>
-                        <phase>verify</phase>
+                        <phase>compile</phase>
```
qualidafial: I'm curious why we're moving sortpom to the compile phase, and moving kotlin plugin from compile to process-sources phase?

TheRealMarnes: sortpom-compile seemed like a logical fit to me. Compiling is when creation/modification of files happens... Having files changed by running tests or in the verify phase seems more of a mismatch. I can link it to any other phase if you disagree though. I guess I also specified a phase explicitly because I'm not a big fan of just using implicit defaults and not actually knowing/seeing the real config...

The kotlin thing was a jetbrains recommendation, like the commit message says:
image

I guess in this case it makes no practical difference, the compilation already worked, I guess there are no references to any kotlin files from the java files... But I like to pre-emptively follow general recommendations. Again though, you're the boss of course.

qualidafial:
Thanks!

PR #1165, opened 2018-06-05
PMD #2

TheRealMarnes:
A few collateral fixes included, but nothing overwhelming or obscuring

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -115,7 +115,6 @@ public class Handle implements Closeable, Configurable<Handle> {
      * rolled back.
      */
     @Override
-    @SuppressWarnings("PMD.DoNotThrowExceptionInFinally")
     public void close() {
         if (closed) {
             return;
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -13,12 +13,9 @@
  */
 package org.jdbi.v3.core;
 
-import static java.util.Objects.requireNonNull;
-
 import java.io.Closeable;
 import java.sql.Connection;
 import java.sql.SQLException;
-
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.config.Configurable;
 import org.jdbi.v3.core.extension.ExtensionMethod;
@@ -37,6 +34,8 @@ import org.jdbi.v3.core.transaction.TransactionIsolationLevel;
 import org.jdbi.v3.core.transaction.UnableToManipulateTransactionIsolationLevelException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import static java.util.Objects.requireNonNull;
 
 /**
  * This represents a connection to the database system. It is a wrapper around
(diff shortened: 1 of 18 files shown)


TheRealMarnes:
A few uses of SuppressWarnings included for those cases where pmd technically has a point and the rule is valid and useful, but jdbi's mission in those situations demands otherwise...

TheRealMarnes:
Uh, can we talk about the Preconditions class? What's with all these @Nullable on parameters that are... not really permitted to be null? It's failing the build, but it also really doesn't make sense to me.

qualidafial:
Preconditions (well, the entire generics internal package) was lifted from Guava.

We needed the generics introspection capabilities of TypeResolver but didn't want to bring in all of Guava as a core dependency.

core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -116,36 +115,51 @@ public Handle setStatementBuilder(StatementBuilder builder) {
...
+        if (wasInTransaction) {
+            rollback();
+        }
+
+        Throwable suppressed = null;
+        try {
+            statementBuilder.close(getConnection());
+        } catch (Throwable t) {
+            suppressed = t;
+        } finally {
```
qualidafial: I think putting a `finally` around this code block is actually redundant.

This block will always be reached.

core/src/main/java/org/jdbi/v3/core/collector/OptionalPrimitiveCollectorFactory.java
```diff
@@ -28,7 +28,9 @@
 
 class OptionalPrimitiveCollectorFactory implements CollectorFactory {
     private static final Map<Class<?>, Collector<?, ?, ?>> COLLECTORS = new HashMap<>();
-    private static final Map<Class<?>, Class<?>> ELEMENT_TYPES = new HashMap<>(); {
+    private static final Map<Class<?>, Class<?>> ELEMENT_TYPES = new HashMap<>();
+
+    static {
```
qualidafial: good catch

core/src/main/java/org/jdbi/v3/core/collector/OptionalPrimitiveCollectorFactory.java
```diff
@@ -41,16 +43,16 @@
 
     @Override
     public boolean accepts(Type containerType) {
-        return COLLECTORS.containsKey(containerType);
+        return containerType instanceof Class && COLLECTORS.containsKey(containerType);
```
qualidafial: I'm curious why PMD cares about this. `Map.get()` takes an `Object`, not `T`.

core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
```diff
@@ -70,6 +70,7 @@ public void rollback(Handle handle) {
 
     @Override
     public void savepoint(Handle handle, String name) {
+        @SuppressWarnings("PMD.CloseResource")
```
qualidafial: Would be nice if we could annotate `Handle.getConnection()` so that PMD knows the connection wasn't created for this method.

core/src/main/java/org/jdbi/v3/core/collector/OptionalPrimitiveCollectorFactory.java
TheRealMarnes: It wasn't so much pmd as it was intellij and my spidey senses. I think it's important to shortcircuit lookups like this. It's not a collection of types but of classes, and not every type is a class, so not every type should be looked up. In this case it makes no practical difference but the general practice is important, to prevent confusing code of which the intent and mechanism isn't totally clear.

core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
TheRealMarnes: I don't think such a thing exists. There's `@Tainted` et al but I don't think there's anything for "this is a Resource but it doesn't need to be closed"...

core/src/main/java/org/jdbi/v3/core/collector/OptionalPrimitiveCollectorFactory.java
TheRealMarnes: image

TheRealMarnes:
@qualidafial

core/src/main/java/org/jdbi/v3/core/collector/OptionalPrimitiveCollectorFactory.java
qualidafial: I'm not convinced this actually performs better, but it doesn't hurt anything.

core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -116,37 +115,53 @@ public Handle setStatementBuilder(StatementBuilder builder) {
      * rolled back.
      */
     @Override
+    @SuppressWarnings("PMD.DoNotThrowExceptionInFinally")
```
qualidafial: This suppression is superfluous now.

core/src/main/java/org/jdbi/v3/core/collector/OptionalPrimitiveCollectorFactory.java
TheRealMarnes: it wasn't meant for performance anyway :p

core/src/main/java/org/jdbi/v3/core/Handle.java
TheRealMarnes: oh yeah you're right

TheRealMarnes:
@qualidafial

qualidafial:
Thanks!

PR #1166, opened 2018-06-06
@UseTemplateEngine more expandable + fixes

TheRealMarnes:
I'm working on a MustacheTemplateEngine module. During testing, I realized using `@UseTemplateEngine` was failing for it, as well as for StringSubTE. I revised the impl class behind it to do 2 things:
- use a stream-based lookup sequence to avoid pyramid exception blocks
- try to find a const `INSTANCE` in addition to constructors so TemplateEngines don't necessarily need to allow pointless instantiation and to reserve the constructor for internal use. The same idea could apply to factory methods named e.g. `of` if desired.

### DIFF
--- a/commons-text/pom.xml
+++ b/commons-text/pom.xml
@@ -35,23 +35,23 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.apache.commons</groupId>
-            <artifactId>commons-text</artifactId>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-core</artifactId>
         </dependency>
         <dependency>
             <groupId>org.jdbi</groupId>
-            <artifactId>jdbi3-core</artifactId>
+            <artifactId>jdbi3-sqlobject</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-text</artifactId>
         </dependency>
 
         <dependency>
             <groupId>org.jdbi</groupId>
             <artifactId>jdbi3-core</artifactId>
             <classifier>tests</classifier>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.jdbi</groupId>
-            <artifactId>jdbi3-sqlobject</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/commons-text/pom.xml
+++ b/commons-text/pom.xml
@@ -42,10 +42,21 @@
             <groupId>org.jdbi</groupId>
             <artifactId>jdbi3-core</artifactId>
         </dependency>
+
         <dependency>
             <groupId>org.jdbi</groupId>
             <artifactId>jdbi3-core</artifactId>
             <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-sqlobject</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.xerial</groupId>
+            <artifactId>sqlite-jdbc</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
+++ b/commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
@@ -22,6 +22,8 @@ import org.jdbi.v3.core.statement.TemplateEngine;
  * Register an instance of this class ({@link org.jdbi.v3.core.statement.SqlStatements#setTemplateEngine}) to use an Apache Commons Text {@link StringSubstitutor} as a {@link TemplateEngine}. This lets you use any pair of strings as variable delimiters, enabling the use of syntax like <pre>select * from ${foo}</pre>, <pre>select * from &lt;foo&gt;</pre>, <pre>select * from %foo%</pre>, etc.
  */
 public class StringSubstitutorTemplateEngine implements TemplateEngine {
+    public static final TemplateEngine INSTANCE = defaults();
+
     private final Consumer<StringSubstitutor> customizer;
 
     private StringSubstitutorTemplateEngine(Consumer<StringSubstitutor> customizer) {
--- a/commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
+++ b/commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
@@ -22,6 +22,10 @@ import org.jdbi.v3.core.statement.TemplateEngine;
  * Register an instance of this class ({@link org.jdbi.v3.core.statement.SqlStatements#setTemplateEngine}) to use an Apache Commons Text {@link StringSubstitutor} as a {@link TemplateEngine}. This lets you use any pair of strings as variable delimiters, enabling the use of syntax like <pre>select * from ${foo}</pre>, <pre>select * from &lt;foo&gt;</pre>, <pre>select * from %foo%</pre>, etc.
  */
 public class StringSubstitutorTemplateEngine implements TemplateEngine {
+    static final String DEFAULT_PREFIX = "${";
+    static final String DEFAULT_SUFFIX = "}";
+    static final char DEFAULT_ESCAPE = '\\';
+
     private final Consumer<StringSubstitutor> customizer;
 
     /**
@@ -29,9 +33,9 @@ public class StringSubstitutorTemplateEngine implements TemplateEngine {
      */
     public StringSubstitutorTemplateEngine() {
         this(substitutor -> {
-            substitutor.setVariablePrefix("${");
-            substitutor.setVariableSuffix("}");
-            substitutor.setEscapeChar('\\');
+            substitutor.setVariablePrefix(DEFAULT_PREFIX);
+            substitutor.setVariableSuffix(DEFAULT_SUFFIX);
+            substitutor.setEscapeChar(DEFAULT_ESCAPE);
         });
     }
 
--- a/commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
+++ b/commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
@@ -22,11 +22,23 @@ import org.jdbi.v3.core.statement.TemplateEngine;
  * Register an instance of this class ({@link org.jdbi.v3.core.statement.SqlStatements#setTemplateEngine}) to use an Apache Commons Text {@link StringSubstitutor} as a {@link TemplateEngine}. This lets you use any pair of strings as variable delimiters, enabling the use of syntax like <pre>select * from ${foo}</pre>, <pre>select * from &lt;foo&gt;</pre>, <pre>select * from %foo%</pre>, etc.
  */
 public class StringSubstitutorTemplateEngine implements TemplateEngine {
-    public static final TemplateEngine INSTANCE = defaults();
-
     private final Consumer<StringSubstitutor> customizer;
 
-    private StringSubstitutorTemplateEngine(Consumer<StringSubstitutor> customizer) {
+    /**
+     * Default is <pre>${foo}</pre> syntax.
+     */
+    public StringSubstitutorTemplateEngine() {
+        this(substitutor -> {
+            substitutor.setVariablePrefix("${");
+            substitutor.setVariableSuffix("}");
+            substitutor.setEscapeChar('\\');
+        });
+    }
+
+    /**
+     * Customize the given {@link StringSubstitutor} instance to set your preferred prefix, suffix, escape character, and perhaps other configuration. The instance is created by Jdbi, and is not shared nor re-used. Your customizer function however will be re-used for all instances.
+     */
+    public StringSubstitutorTemplateEngine(Consumer<StringSubstitutor> customizer) {
         this.customizer = customizer;
     }
 
@@ -38,15 +50,17 @@ public class StringSubstitutorTemplateEngine implements TemplateEngine {
     }
 
     /**
-     * At the time of writing, defaults means <pre>${foo}</pre> syntax.
+     * @deprecated use the default constructor instead
      */
+    @Deprecated
     public static StringSubstitutorTemplateEngine defaults() {
-        return new StringSubstitutorTemplateEngine(substitutor -> {});
+        return new StringSubstitutorTemplateEngine();
     }
 
     /**
-     * Customize the given {@link StringSubstitutor} instance to set your preferred prefix, suffix, escape character, and perhaps other configuration. The instance is created by Jdbi, and is not shared nor re-used. Your customizer function however will be re-used for all instances.
+     * @deprecated use the constructor instead
      */
+    @Deprecated
     public static StringSubstitutorTemplateEngine withCustomizer(Consumer<StringSubstitutor> customizer) {
         return new StringSubstitutorTemplateEngine(customizer);
     }
--- a/commons-text/src/main/java/org/jdbi/v3/commonstext/UseStringSubstitutorTemplateEngine.java
+++ b/commons-text/src/main/java/org/jdbi/v3/commonstext/UseStringSubstitutorTemplateEngine.java
@@ -24,7 +24,7 @@ import org.jdbi.v3.sqlobject.config.ConfiguringAnnotation;
 @ConfiguringAnnotation(UseStringSubstitutorTemplateEngineImpl.class)
 @Target({ElementType.TYPE, ElementType.METHOD})
 public @interface UseStringSubstitutorTemplateEngine {
-    String prefix() default "${";
-    String suffix() default "}";
-    char escape() default '\\';
+    String prefix() default StringSubstitutorTemplateEngine.DEFAULT_PREFIX;
+    String suffix() default StringSubstitutorTemplateEngine.DEFAULT_SUFFIX;
+    char escape() default StringSubstitutorTemplateEngine.DEFAULT_ESCAPE;
 }
--- a/commons-text/src/main/java/org/jdbi/v3/commonstext/UseStringSubstitutorTemplateEngine.java
+++ b/commons-text/src/main/java/org/jdbi/v3/commonstext/UseStringSubstitutorTemplateEngine.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.commonstext;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+import org.jdbi.v3.commonstext.internal.UseStringSubstitutorTemplateEngineImpl;
+import org.jdbi.v3.sqlobject.config.ConfiguringAnnotation;
+
+@Retention(RetentionPolicy.RUNTIME)
+@ConfiguringAnnotation(UseStringSubstitutorTemplateEngineImpl.class)
+@Target({ElementType.TYPE, ElementType.METHOD})
+public @interface UseStringSubstitutorTemplateEngine {
+    String prefix() default "${";
+    String suffix() default "}";
+    char escape() default '\\';
+}
--- a/core/src/main/java/org/jdbi/v3/core/statement/TemplateEngine.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/TemplateEngine.java
@@ -16,9 +16,9 @@ package org.jdbi.v3.core.statement;
 /**
  * Renders an SQL statement from a template.
  *
- * @see DefinedAttributeTemplateEngine
+ * Note for implementors: define a suitable public constructor for SqlObject's {@code UseTemplateEngine} annotation, and/or create your own custom annotation in case your {@link TemplateEngine} has configuration parameters! Suitable constructors are the no-arg constructor, one that takes a {@link java.lang.Class}, and one that takes both a {@link java.lang.Class} and a {@link java.lang.reflect.Method}.
  *
- * Note for implementors: define a public default constructor, or one that takes a {@link java.lang.Class} and/or {@link java.lang.reflect.Method}, for SqlObject's {@code UseTemplateEngine} annotation, and/or create your own custom annotation in case your {@link TemplateEngine} has configuration parameters!
+ * @see DefinedAttributeTemplateEngine
  */
 @FunctionalInterface
 public interface TemplateEngine {
--- a/core/src/main/java/org/jdbi/v3/core/statement/TemplateEngine.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/TemplateEngine.java
@@ -17,6 +17,8 @@ package org.jdbi.v3.core.statement;
  * Renders an SQL statement from a template.
  *
  * @see DefinedAttributeTemplateEngine
+ *
+ * Note for implementors: define a public default constructor, or one that takes a {@link java.lang.Class} and/or {@link java.lang.reflect.Method}, for SqlObject's {@code UseTemplateEngine} annotation, and/or create your own custom annotation in case your {@link TemplateEngine} has configuration parameters!
  */
 @FunctionalInterface
 public interface TemplateEngine {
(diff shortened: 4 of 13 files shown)


TheRealMarnes:
On a related note, is there still actually an interest in a Mustache module being developed? It was discussed a while ago after @stevenschlansker mentioned it and such

sqlobject/src/main/java/org/jdbi/v3/sqlobject/config/internal/UseTemplateEngineImpl.java
```diff
@@ -46,24 +48,60 @@ public void configureForType(ConfigRegistry registry, Annotation annotation, Cla
...
-            Constructor<? extends TemplateEngine> noArg = value.getConstructor();
-            return noArg.newInstance();
-        } catch (NoSuchMethodException e) {
+                                       Method m) {
+
+        return Stream.of(tryConstructor(value), tryConstructor(value, sqlObjectType), tryConstructor(value, sqlObjectType, m), tryConstant(value, "INSTANCE"))
+            .map(Supplier::get)
+            .filter(Objects::nonNull)
+            .findFirst()
+            .orElseThrow(() -> new IllegalStateException("Unable to instantiate, no viable constructor or constant for " + value.getName()));
```
qualidafial: From my read, the following would be recognized:

```java
public MyTemplateEngine() {}
public MyTemplateEngine(Class<?> sqlObjectType) {}
public MyTemplateEngine(Class<?> sqlObjectType, Method method) {}
public static final MyTemplateEngine INSTANCE = ...
```

I think the `(Class)` and `(Class, Method)` variants are superfluous. The template engine has access to the SQL Object class and method through `ctx.getExtensionMethod()`.

Aside from that, I feel like if a template engine needs to have options (e.g. the StringSubstitutor engine), it should likewise have its own SqlObject annotation to provide those options.

core/src/main/java/org/jdbi/v3/core/statement/TemplateEngine.java
```diff
@@ -17,6 +17,8 @@
  * Renders an SQL statement from a template.
  *
  * @see DefinedAttributeTemplateEngine
+ *
+ * Note for implementors: define an appropriate constructor or constant for SqlObject's UseTemplateEngine annotation!
```
qualidafial: We should actually define what an appropriate constructor or constant looks like.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestUseTemplateEngine.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @Test
+    public void testUseDefinedAttributes() {
+        String selected = jdbi.withExtension(QueriesForDefinedAttributeTE.class, q -> q.select("foo"));
+
+        assertThat(selected).isEqualTo("foo");
+    }
+
+    public interface QueriesForMessageFormatTE {
+        @UseTemplateEngine(MessageFormatTemplateEngine.class)
```
qualidafial: I can't see a valid reason why `MessageFormatTemplateEngine` should not have a public no-arg constructor. It has no state and nothing about it is configurable.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/config/internal/UseTemplateEngineImpl.java
TheRealMarnes: Yeah, that was on my mind too. I didn't want to make functional changes without discussion though, I just refactored the existing code. If you give the go for changing the functionality, I'll give it a try. Maybe in a separate PR? I'm a little confused by some of the implementations though, like for StringTemplate. Most of them really. A lot of their code feels somehow redundant or needlessly complex, and I can't help but think there must be a reason it was done that way.

core/src/main/java/org/jdbi/v3/core/statement/TemplateEngine.java
TheRealMarnes: Is expecting implementors to give the UseTemplateEngine source code a quick look too much to expect? I'm worried if we document the specifics in this separate class, someone will modify UseTemplateEngine's behavior and forget to keep this documentation in sync

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestUseTemplateEngine.java
TheRealMarnes: Maybe it's a good idea to establish a project convention, constructors vs consts and factory methods? I can live with either but I agree having a mix of both is kind of annoying.

TheRealMarnes: Maybe an answer to a different question will help me decide on how I design my classes for jdbi.

Suppose I write a class, give it a public default constructor for convenience like with most classes in jdbi. In the future, something is changed about the class and having a public no-arg constructor becomes something of an issue.

Which is the greater priority? Keeping classes safe and coherent at the cost of changing the API when something becomes risky, or keeping the API intact at the cost of making classes more fragile than they could be?

TheRealMarnes:
Sorry if I'm maybe a little annoying with my walls of text :) I can get kinda philosophical about code practices. I'm open to new ideas, and am undecided or not strongly partial either way about some. There are also project conventions or lack thereof to consider, and how to unite them with personal preferences and contextual details...

TheRealMarnes:
Quick recap:
- keep documentation integrally in UseTemplateEngine or put a copy in TemplateEngine as a reminder? It may be more convenient but it carries the risk of becoming outdated and forgotten
- remove the constructor(a) and constructor(a, b) feature (no engines use it atm afaik, and future ones won't need it)
- add a factory method lookup as an alternative to the constructor lookup?
- deprecate/remove custom UseFooEngine annotations that are of no added value over UseTemplateEngine (afaik all of them)?
- decide whether we wanna generally go with public constructors that may need to be changed in the future (API breakage) or go with factory methods and consts that may offer some leeway/buffer space to prevent breakage (i.e. do we really care users will need to maybe change a few lines of declarative code when they update jdbi?)

TheRealMarnes:
Sorry for the mail spam if you get mails. I'm a hopeless philosopher. I've deleted all the discussion, I think I get the jist of things and can make some reasonable assumptions.

commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
```diff
@@ -35,20 +49,6 @@ public String render(String template, StatementContext ctx) {
...
-        return new StringSubstitutorTemplateEngine(substitutor -> {});
-    }
-
-    /**
-     * Customize the given {@link StringSubstitutor} instance to set your preferred prefix, suffix, escape character, and perhaps other configuration. The instance is created by Jdbi, and is not shared nor re-used. Your customizer function however will be re-used for all instances.
-     */
-    public static StringSubstitutorTemplateEngine withCustomizer(Consumer<StringSubstitutor> customizer) {
-        return new StringSubstitutorTemplateEngine(customizer);
-    }
-
```
qualidafial: These methods were released as API in 3.2.0

commons-text/pom.xml
```diff
@@ -42,12 +42,22 @@
             <groupId>org.jdbi</groupId>
             <artifactId>jdbi3-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-sqlobject</artifactId>
```
qualidafial: Could be an optional dep?

commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
qualidafial: @jdbi/contributors 

I think this highlights that we need a way to do API experiments and include them in releases, without necessarily committing to that API long term.

What if we had a `@Beta` or `@Experimental` annotation that was included in the javadoc? For big features we might put the entire thing in an `experimental` package until it was promoted to a long-term supported feature.

TheRealMarnes: I have often wished there was an @Experimental counterpart to @Deprecated... On the other hand, why would this have been marked experimental? It was a done piece of work. Catching things like that is what project conventions and reviews are for...

Tbh I dont think it's a big deal. Everyone understands updates bring changes, and this is still so fresh I doubt a lot of people are using it. It's also not something you need to refer to 500 times in 1 project...

freemarker/src/main/java/org/jdbi/v3/freemarker/UseFreemarkerEngine.java
```diff
@@ -25,8 +25,11 @@
 /**
  * Configures a SQL object class or method to rewrite SQL statements using Freemarker. Method parameters annotated
  * with {@link Define @Define} are passed to the Freemarker as template attributes.
+ *
+ * @deprecated simply use @UseTemplateEngine(FreemarkerEngine.class) instead
  */
 @ConfiguringAnnotation(UseFreemarkerEngineImpl.class)
 @Target({ElementType.TYPE, ElementType.METHOD})
 @Retention(RetentionPolicy.RUNTIME)
+@Deprecated
```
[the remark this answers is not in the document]
TheRealMarnes: ok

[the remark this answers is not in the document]
TheRealMarnes: ok

sqlobject/src/main/java/org/jdbi/v3/sqlobject/config/internal/UseTemplateEngineImpl.java
```diff
@@ -14,56 +14,34 @@
...
-                                                                                            Method.class);
-                    return constructor.newInstance(sqlObjectType, m);
-                }
-                throw new IllegalStateException("Unable to instantiate, no viable constructor " + value.getName());
-            }
+            return clazz.getConstructor().newInstance();
+        } catch (NoSuchMethodException ignored) {
+            throw new RuntimeException(clazz.getSimpleName() + " does not have a default constructor and thus cannot be used with @" + UseTemplateEngine.class.getSimpleName());
+        } catch (IllegalAccessException | InstantiationException | InvocationTargetException e) {
+            throw new RuntimeException(e);
```
qualidafial: When I suggested getting rid of the `(Class)` and `(Class, Method)` variants, I did not realize it was existing, documented behavior. This behavior should be reinstated--my bad.

stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/UseStringTemplateEngine.java
```diff
@@ -26,8 +26,11 @@
  * Configures a SQL object class or method to rewrite SQL statements using StringTemplate. Method parameters annotated
  * with {@link Define @Define} are passed to the StringTemplate as template
  * attributes.
+ *
+ * @deprecated simply use @UseTemplateEngine(StringTemplateEngine.class) instead
  */
 @ConfiguringAnnotation(UseStringTemplateEngineImpl.class)
 @Target({ElementType.TYPE, ElementType.METHOD})
 @Retention(RetentionPolicy.RUNTIME)
+@Deprecated
```
qualidafial: Disagree

stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/internal/UseStringTemplateEngineImpl.java
```diff
@@ -20,6 +20,10 @@
 import org.jdbi.v3.sqlobject.config.Configurer;
 import org.jdbi.v3.stringtemplate4.StringTemplateEngine;
 
+/**
+ * @deprecated see {@link org.jdbi.v3.stringtemplate4.UseStringTemplateEngine}
+ */
+@Deprecated
```
qualidafial: ditto

commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
qualidafial: > Tbh I dont think it's a big deal. Everyone understands updates bring changes, and this is still so fresh I doubt a lot of people are using it.

That's an assumption, and not as safe as you think. At Eclipse we had a mantra: every API is used by someone. Which means if we make incompatible changes, somebody is going to get mad.

Think about a large software shop. Teams with too many projects on their plate to keep their code groomed regularly. They may only get a chance to update dependencies once or twice a year. When they update from e.g. 3.2.0 to 3.5.0, a semantically versioned project like Jdbi is expected to be fully backward compatible. The only difference users should notice is that bugs are fixed.

> On the other hand, why would this have been marked experimental?

My suggestion is that we start marking _all_ new APIs experimental and give them a few releases to bake. This way, when we start using them in production code we have a grace period for recognizing and fixing design mistakes.

TheRealMarnes: > incompatible changes
 
There's a spectrum to that tbh. Replacing a factory method by a constructor or vice versa is a superficial change that doesn't entail functional changes on either side. Ditto for things like a changed package or renamed method, which only takes a search&replace. Completely restructuring an entire mechanism, now that would be annoying.

Philosophy aside, what do I do with these? I understand we want to keep classes as simple as possible and stick to simply exposing a default constructor wherever reasonable. I could mark the outdated factory methods and constants as deprecated for the time being instead of removing them? They are forward compatible...

qualidafial: It's a spectrum only if you consider a single app module requiring Jdbi in isolation. It becomes an N-dimensional graph when you factor in third party libraries that integrate with us, e.g. Dropwizard's `dropwizard-jdbi3` artifact.

Take Guava. They have always maintained (and documented) that deprecated APIs will eventually be removed. A year or so ago, they made good on that policy, and removed several deprecated APIs. For some Guava users, this made all hell break loose.

At $JOB, we have some library dependencies that transitively depend on Guava. A few depended on an old version and still used those recently-removed APIs. Others depended on newer API in later versions--added after those other APIs were removed.

This made it impossible to use two such dependencies concurrently in a project, if they depended on incompatible versions of Guava. We had projects stalled for _months_ while we waited for those libraries to update their dependencies to a later version of Guava.

It was an awful mess, and one I will not inflict that on our users.

It is noteworthy that the Guava maintainers, after the outcry over those API removals, vowed never to make binary incompatible changes again.

Please take time to review the Eclipse Foundation's guidelines for API compatibility on their wiki. These were curated from years of trial and error, as well as community feedback:

- https://wiki.eclipse.org/Evolving_Java-based_APIs
- https://wiki.eclipse.org/Evolving_Java-based_APIs_2
- https://wiki.eclipse.org/Evolving_Java-based_APIs_3

qualidafial: > Philosophy aside, what do I do with these? I understand we want to keep classes as simple as possible and stick to simply exposing a default constructor wherever reasonable. I could mark the outdated factory methods and constants as deprecated for the time being instead of removing them? They are forward compatible...

Just deprecate, and leave a `@deprecated` javadoc comment directing users to the recommended API alternative.

stevenschlansker: Thanks for putting that so well.  We should strive to avoid breaking API changes unless necessary.  (I still think we can be pragmatic, e.g. a breaking change to test case @Rule just one release after initial merge can slip by occasionally)

I like the idea of @Beta or @Experimental.  I think that lets us lower the bar for inclusion without backing ourselves into a corner.  In particular, it'd let us merge things like the Spring support a little less cautiously.

TheRealMarnes: > Just deprecate

Well that's what I've done now.  Now that the build has passed, I think I've addressed all concerns and have still done what I set out to do...

I get your point and I understand the importance of compatibility, but I can't say I think "we will never change anything ever" is completely unitable with maintaining a project. Sometimes things just need to be changed in light of discoveries and problem fixes, or general improvements. You could keep old bug-free classes around and implement breaking improvements in separate clones of those classes (a la `/rest/api/v1`, `/rest/api/v2`, etc), but then you get an awfully polluted project after a while. Jdbi3 broke a lot of stuff coming from v2, so clearly we're not working under a "we will never change anything ever" policy. If someone wants to use 2 libraries in their project, one of which depends on jdbi v2 and the other on v3, they're fucked, just like those guava users were. It happens, it can't be avoided, and while I very much support minimizing breakage, I think a project's coherence takes priority over legacy support. Keep supporting what you can as long as you can, but don't carve problems into stone for the sake of legacy.

But that's just my opinion. This is jdbi and I expect nothing less than for my code to be rejected and changes demanded if something isn't up to jdbi's rules. :)

Anyway, I think we can all be happy with the current changes, no? I'll try to better adjust my code to your conventions right off the bat in the future so we don't need to discuss so much. I have a better view on your opinions and policies now than before, so this thread has definitely been helpful there. :)

stevenschlansker: Changing the packages when we incremented the major version is an intentional decision which ensures that jdbi v2 and v3 can co-exist in the same codebase.  "If someone wants to use 2 libraries in their project, one of which depends on jdbi v2 and the other on v3, they're fucked" -> nope, they both can co-exist totally cleanly!  (as far as I know)

TheRealMarnes: Hmm, yeah, I forgot that...

So, can I take away from this then that the rule of thumb is
1) API that comes, stays
2) undesired API must stay supported and may only be marked Deprecated
3) the desired breaking cleanup work can be done at the time of a major version increment, as semver dictates that's when all hell is allowed to break loose

That's basically the net result of all this, right? Maybe we can write this down in the readme for both devs and users to see?

commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
```diff
@@ -22,9 +22,29 @@
  * Register an instance of this class ({@link org.jdbi.v3.core.statement.SqlStatements#setTemplateEngine}) to use an Apache Commons Text {@link StringSubstitutor} as a {@link TemplateEngine}. This lets you use any pair of strings as variable delimiters, enabling the use of syntax like <pre>select * from ${foo}</pre>, <pre>select * from &lt;foo&gt;</pre>, <pre>select * from %foo%</pre>, etc.
  */
 public class StringSubstitutorTemplateEngine implements TemplateEngine {
+    /**
+     * @deprecated use the default constructor instead
+     */
+    @Deprecated
+    public static final TemplateEngine INSTANCE = new StringSubstitutorTemplateEngine();
+
```
qualidafial: This is a new member in this PR, no?

I don't see it in the existing Javadoc: http://jdbi.org/apidocs/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.html

Could just be removed.

core/src/main/java/org/jdbi/v3/core/statement/TemplateEngine.java
```diff
@@ -17,6 +17,8 @@
  * Renders an SQL statement from a template.
  *
  * @see DefinedAttributeTemplateEngine
+ *
+ * Note for implementors: define a public default constructor, or one that takes a {@link java.lang.Class} and/or {@link java.lang.reflect.Method}, for SqlObject's {@code UseTemplateEngine} annotation, and/or create your own custom annotation in case your {@link TemplateEngine} has configuration parameters!
```
qualidafial: :+1

commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
TheRealMarnes: Ah, chicken that I am, I added it in the previous commit for the test when I was still holding onto factories and consts

sqlobject/src/main/java/org/jdbi/v3/sqlobject/config/internal/UseTemplateEngineImpl.java
```diff
@@ -14,56 +14,52 @@
...
-        } catch (NoSuchMethodException e) {
+    private static TemplateEngine instantiate(Class<? extends TemplateEngine> engineClass, Class<?> sqlObjectType, @Nullable Method method) {
+        return Stream.of(tryConstructor(engineClass), tryConstructor(engineClass, sqlObjectType), tryConstructor(engineClass, sqlObjectType, method))
+            .map(Supplier::get)
+            .filter(Objects::nonNull)
+            .findFirst()
+            .orElseThrow(() -> new IllegalStateException("Unable to instantiate, no viable constructor for " + engineClass.getName()));
+    }
+
+    private static <T extends TemplateEngine> Supplier<T> tryConstructor(Class<T> c, Object... args) {
```
qualidafial: :+1: good refactor

commons-text/src/main/java/org/jdbi/v3/commonstext/StringSubstitutorTemplateEngine.java
TheRealMarnes: Thanks for catching that, would've been an addition to the regrets if that had gotten through 😁

TheRealMarnes:
Gonna add a commit to fix a javadoc issue, hold on a sec

qualidafial:
@TheRealMarnes I'm ready to merge this if you've made all the changes you planned on.

> On a related note, is there still actually an interest in a Mustache module being developed? It was discussed a while ago after @stevenschlansker mentioned it and such

Sure, Mustache is a nice template engine. But: do you use Mustache, or is this more of a curiosity? You've submitted a couple different template engines already, and I'd prefer we only maintain templates engines for which we have active users.

TheRealMarnes:
I made it because I was under the impression you guys wanted it as a feature, lol

I think for this one, everything's done. The lookup sequence is a little cleaner, UseTemplateEngine is better supported, and stringsubTE has a config annotation :)

TheRealMarnes:
I was thinking of giving SqlLocators a refactor next. I think I saw a lot of duplicated and not-so-trivial logic involving ClassLoaders and paths, and things that could be made slightly more flexible to avoid having to reinvent wheels just to add different rims...

qualidafial:
> I made it because I was under the impression you guys wanted it as a feature, lol

Every feature carries a maintenance burden. We only want that burden for features that have a chance of being widely used.

If there's a large contingent of users that prefer Mustache to FreeMarker for SQL templating then I think that would justify it. But if just one person wants it and everybody else is fine with FreeMarker, then that burden should rightfully fall on the one person who wants to do it differently--in their own code base.

commons-text/src/main/java/org/jdbi/v3/commonstext/UseStringSubstitutorTemplateEngine.java
```diff
@@ -24,7 +24,7 @@
 @ConfiguringAnnotation(UseStringSubstitutorTemplateEngineImpl.class)
 @Target({ElementType.TYPE, ElementType.METHOD})
 public @interface UseStringSubstitutorTemplateEngine {
-    String prefix() default "${";
-    String suffix() default "}";
-    char escape() default '\\';
+    String prefix() default StringSubstitutorTemplateEngine.DEFAULT_PREFIX;
+    String suffix() default StringSubstitutorTemplateEngine.DEFAULT_SUFFIX;
+    char escape() default StringSubstitutorTemplateEngine.DEFAULT_ESCAPE;
```
qualidafial: This refactor is fine, provided the Javadoc shows the literal values for those defaults.

TheRealMarnes: Is it ok like so?

commons-text/src/main/java/org/jdbi/v3/commonstext/UseStringSubstitutorTemplateEngine.java
```diff
@@ -0,0 +1,41 @@
...
+import org.jdbi.v3.sqlobject.config.ConfiguringAnnotation;
+
+@Retention(RetentionPolicy.RUNTIME)
+@ConfiguringAnnotation(UseStringSubstitutorTemplateEngineImpl.class)
+@Target({ElementType.TYPE, ElementType.METHOD})
+public @interface UseStringSubstitutorTemplateEngine {
+    /**
+     * default: <code>${</code>
+     */
+    String prefix() default StringSubstitutorTemplateEngine.DEFAULT_PREFIX;
```
qualidafial: This isn't what I meant--sorry for the misunderstanding.

What I meant is, when you run the build, does javadoc output the default value for the attribute, even though it is a reference to a constant in another class? I ran a build on your branch, and the answer is yes, it does:

```
Element Detail

    prefix

    public abstract String prefix

    default: ${     <-- your javadoc comment

    Default:        <-- generated out of the box by javadoc
        "${"
```

So we can remove those comments, since Javadoc is automatically documenting the default value for us.

TheRealMarnes: ok, reverted

commons-text/src/main/java/org/jdbi/v3/commonstext/UseStringSubstitutorTemplateEngine.java
qualidafial: Yep.

qualidafial:
Looks good, merging.

qualidafial:
Thanks!

PR #1171, opened 2018-06-11
[done] Timeout

### LINKED ISSUE #1170: Add global timeout to Jdbi
I'm unable to specify a timeout on my jdbi instance. I normally expect this from such a lib.

```kotlin
val db = Jdbi.create(someUrl)
db.setTimeout(10, TimeUnit.SECONDS)
```

TheRealMarnes:
fixes #1170 

I'm not sure this can be tested...

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/Batch.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/Batch.java
@@ -60,7 +60,9 @@ public class Batch extends BaseStatement<Batch> {
         try {
             try {
                 stmt = getHandle().getStatementBuilder().create(getHandle().getConnection(), getContext());
+
                 addCleanable(stmt::close);
+                getConfig(SqlStatements.class).customize(stmt);
             } catch (SQLException e) {
                 throw new UnableToCreateStatementException(e, getContext());
             }
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
@@ -13,11 +13,15 @@
  */
 package org.jdbi.v3.core.statement;
 
+import java.sql.SQLException;
+import java.sql.Statement;
 import java.time.temporal.ChronoUnit;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+import javax.annotation.Nullable;
 import org.jdbi.v3.core.config.JdbiConfig;
+import org.jdbi.v3.meta.Beta;
 
 /**
  * Configuration holder for {@link SqlStatement}s.
@@ -28,12 +32,14 @@ public final class SqlStatements implements JdbiConfig<SqlStatements> {
     private TemplateEngine templateEngine;
     private SqlParser sqlParser;
     private SqlLogger sqlLogger;
+    private Integer queryTimeout;
 
     public SqlStatements() {
         attributes = new ConcurrentHashMap<>();
         templateEngine = new DefinedAttributeTemplateEngine();
         sqlParser = new ColonPrefixSqlParser();
         sqlLogger = SqlLogger.NOP_SQL_LOGGER;
+        queryTimeout = null;
     }
 
     private SqlStatements(SqlStatements that) {
@@ -41,6 +47,7 @@ public final class SqlStatements implements JdbiConfig<SqlStatements> {
         this.templateEngine = that.templateEngine;
         this.sqlParser = that.sqlParser;
         this.sqlLogger = that.sqlLogger;
+        this.queryTimeout = that.queryTimeout;
     }
 
     /**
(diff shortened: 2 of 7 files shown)


isaki-x:
This implementation is not thread safe. Perhaps one of the following paradigms should be used?

- Mark the field as volatile.
- Use an AtomicInteger (would require using something like `-1` instead of `null` for unset).
- Gate access with a `Lock`.

TheRealMarnes:
@isaki-x I dont really see why or how this would need to be threadsafe, especially when all the other JdbiConfigs arent desigbed for it either...

isaki-x:
No worries, just offering an observation for a potential issue. If no thread safety guarantees are made around this sort of stuff, then just disregard my previous comment.

stevenschlansker:
FWIW, the `JdbiConfig` class states:
```
 * Implementors should ensure that implementations are thread-safe for access and caching purposes, but not
 * necessarily for reconfiguration.
```
so this shouldn't be a worry currently.

core/src/main/java/org/jdbi/v3/core/connection/Connections.java
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.connection;
+
+import javax.annotation.Nullable;
+import org.jdbi.v3.core.config.JdbiConfig;
+import org.jdbi.v3.meta.Beta;
+
+@Beta
+public class Connections implements JdbiConfig<Connections> {
```
stevenschlansker: Quibble over the naming, a "query timeout" is really a property of `Statements`, not a property of `Connections`.

core/src/main/java/org/jdbi/v3/core/statement/Batch.java
```diff
@@ -60,7 +61,13 @@ public Batch add(String sql) {
         try {
             try {
                 stmt = getHandle().getStatementBuilder().create(getHandle().getConnection(), getContext());
+
                 addCleanable(stmt::close);
+
+                Integer queryTimeout = getConfig(Connections.class).getQueryTimeout();
```
stevenschlansker: Consider, should `Batch` et al actually understand what's going on here, or should we add a `protected void customize(Statement stmt)` that does the actual work to the config class?

core/src/main/java/org/jdbi/v3/core/connection/Connections.java
TheRealMarnes: Hmm, but the timeout applies to the db connection the resultset/statement keep open for lazy loading, doesn't it?

TheRealMarnes:
I added a test attempt but it's not really working. I see h2's setQueryTimeout() definitely being executed, but it refuses to actually time out when the query takes longer. Could be external java procedures are exempt from timeouts or something...

stevenschlansker:
Maybe embedded postgres is a better database to test this against?  That would at least have a real driver timeout.

TheRealMarnes:
Will try

I see you're on a roll tonight :)

TheRealMarnes:
@stevenschlansker that worked, thanks for the golden tip

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -28,19 +32,22 @@
...
     private SqlParser sqlParser;
     private SqlLogger sqlLogger;
+    private Integer queryTimeout;
 
     public SqlStatements() {
         attributes = new ConcurrentHashMap<>();
         templateEngine = new DefinedAttributeTemplateEngine();
         sqlParser = new ColonPrefixSqlParser();
         sqlLogger = SqlLogger.NOP_SQL_LOGGER;
+        queryTimeout = null;
```
stevenschlansker: this line isn't strictly necessary

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -165,6 +172,31 @@ public SqlStatements setSqlLogger(SqlLogger sqlLogger) {
...
+        return queryTimeout;
+    }
+
+    /**
+     * Jdbi does not implement its own timeout mechanism: it simply calls {@link java.sql.Statement#setQueryTimeout}, leaving timeout handling to your jdbc driver.
+     *
+     * @param seconds the time in seconds to wait for a query to complete; 0 to disable the timeout; null to leave it at defaults (i.e. Jdbi will not call {@code setQueryTimeout(int)})
+     */
+    @Beta
+    public SqlStatements setQueryTimeout(@Nullable Integer seconds) {
```
stevenschlansker: Should we take a `Duration` instead of `Integer`?  Attaching units to values prevents many types of common errors (like, assuming this timeout is in milliseconds!)

stevenschlansker: Yeah, an integer number of seconds is pretty dumb here... I would have picked millis :)

stevenschlansker: I'll give you a few minutes to add that if you'd like, then will merge.  I agree that with seconds resolution only it's not nearly as compelling.

TheRealMarnes: I think adding an overload that takes a Duration would be misleading to users since anything below seconds will be rounded, and throwing an IAE to avoid the rounding will be just as much of a PITA :/

PR #1174, opened 2018-06-12
[poc + discuss] support generic Map<String, V> with MapMapper

TheRealMarnes:
I came across an annoyance in my project today. SqlObject has OOTB support for turning a resultset into a (`List` of) `Map<String, Object>` like so:

```
@SqlQuery("select 1.0 as one, 2.0 as two, 3.0 as three from (values(null))")
List<Map<String, Object>> getMapList();
// will be
// "one" -> (some jdbc driver mapped object, e.g. BigDecimal in case of hsqldb),
// "two" -> ...
```

But I wanted to get my `BigDecimal`s out of that map without having to cast them, because you know, what do we have generics for? 

```
@SqlQuery("select 1.0 as one, 2.0 as two, 3.0 as three from (values(null))")
List<Map<String, BigDecimal>> getMapList();
```

This would be ideal. I know the values will be mappable (even directly castable) to `BigDecimal`, so jdbi should be able to extract them with the registered `BigDecimalColumnMapper`. But this does not work because `MapMapper` only responds to `Map<String, Object>` specifically.

I propose we extend the functionality of `MapMapper` by moving it into a `MapMapperFactory` that resolves the `V` in `Map<String, V>` and uses a registered `ColumnMapper` on the context to neatly map the value objects, instead of returning them as `Object` in the `Map` and forcing the user to do casts.

The code I wrote for this PR is just a quick&dirty POC that needs some discussion, but it shows how very easy it is to achieve. Since `MapMapper` is already Core and what I'm suggesting would be an improvement to its support for generics, not a whole different use case, I'd like to put this into Core.

Obviously, if you agree to this idea, I'll reimplement my changes properly. This is just a rough draft.

As an aside, shouldn't the `foldCase` property on `MapMapper` come in a config class, like we have on `SerializableTransactionRunner`? The way it is now seems useful only for the fluent api.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
+++ b/core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
@@ -82,10 +82,14 @@ public class GenericTypes {
      * @return the parameter on the supertype, if it is concretely defined.
      */
     public static Optional<Type> findGenericParameter(Type type, Class<?> parameterizedSupertype) {
-        Type parameterType = resolveType(parameterizedSupertype.getTypeParameters()[0], type);
+        return findGenericParameter(type, parameterizedSupertype, 0);
+    }
+
+    public static Optional<Type> findGenericParameter(Type type, Class<?> parameterizedSupertype, int i) {
+        Type parameterType = resolveType(parameterizedSupertype.getTypeParameters()[i], type);
         return parameterType instanceof Class || parameterType instanceof ParameterizedType
-                ? Optional.of(parameterType)
-                : Optional.empty();
+            ? Optional.of(parameterType)
+            : Optional.empty();
     }
 
     /**
--- a/core/src/main/java/org/jdbi/v3/core/mapper/MapMapperFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/MapMapperFactory.java
@@ -1,0 +1,26 @@
+package org.jdbi.v3.core.mapper;
+
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.Map;
+import java.util.Optional;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericTypes;
+
+public class MapMapperFactory implements RowMapperFactory {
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
```diff
@@ -82,10 +82,14 @@ private GenericTypes() {
      * @return the parameter on the supertype, if it is concretely defined.
      */
     public static Optional<Type> findGenericParameter(Type type, Class<?> parameterizedSupertype) {
-        Type parameterType = resolveType(parameterizedSupertype.getTypeParameters()[0], type);
+        return findGenericParameter(type, parameterizedSupertype, 0);
+    }
+
+    public static Optional<Type> findGenericParameter(Type type, Class<?> parameterizedSupertype, int i) {
```
qualidafial: I like this utility function, surprised we haven't pursued it before now

core/src/main/java/org/jdbi/v3/core/mapper/MapMapperFactory.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.Map;
+import java.util.Optional;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericTypes;
+
+public class MapMapperFactory implements RowMapperFactory {
+    @Override
+    public Optional<RowMapper<?>> build(Type type, ConfigRegistry config) {
+        if (type instanceof ParameterizedType && GenericTypes.getErasedType(type) == Map.class) {
+            Type valueType = GenericTypes.findGenericParameter(type, Map.class, 1).orElseThrow(IllegalStateException::new);
```
qualidafial: This exception would hide a previously registered mapper factory for raw `Map`

TheRealMarnes:
@qualidafial just moved and amended the javadoc on the utility method, and rewrote the generic conditions to fall back to the mapper chain if anything is outside expectations

TheRealMarnes:
Oh, you posted this on the closed PR :p

Sorry about the mail spam, I'd done it too badly the first time around, the proper PR is #1177

PR #1177, opened 2018-06-12
[poc/wip] Support generics in Map mapping

TheRealMarnes:
Right now Jdbi can only map to `Map<String, Object>` in both the fluent api (`mapToMap()`) and SqlObjects (`@SqlQuery Map<String, Object> getMap();`). This saddles the user up with having to do annoying manual type converions: getting `ColumnMapper`s from the context, or doing plain casts.

With this additional factory, which is basically a refactor and slight expansion of `MapMapper`, users will be able to map to a `Map<String, V>` with full generics support. The factory will get an appropriate `ColumnMapper` for V off the context during mapping.

Fluent api: `.mapTo(new GenericType<Map<String, BigDecimal>>() {})` or `.mapToGenericMap(BigDecimal.class)`
SqlObject: just return a `Map<String, BigDecimal>`

Since mapping to `Map` is a core feature already but just lacks proper generics support, I feel this belongs in Core and not in private code.

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,3 +1,9 @@
+3.4.0
+  - New API
+    - GenericMapMapperFactory enables fluent and SqlObject support for mapping to `Map<String, V>` for any `V` that can be handled by a ColumnMapper.
+    - GenericTypes.findGenericParameter(Type, Class) now also takes an index, e.g. to resolve `V` in `Map<K, V>`
+    - ResultBearing.mapToGenericMap(Type) convenience method to use the GenericMapMapperFactory
+
 3.3.0
   - New API
     - SerializableTransactionRunner.setOnFailure(), setOnSuccess() methods allow callbacks to be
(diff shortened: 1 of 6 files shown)


TheRealMarnes:
@qualidafial this is the up-to-date PR :)

I just moved and amended the javadoc on the utility method, and rewrote the generic conditions to fall back to the mapper chain if anything is outside expectations, like you asked

TheRealMarnes:
I've added an exclusion for `<String, Object>` so that that type will keep going to MapMapper as before, and removed the default registration of GenericMapMapper, for the sake of API compatibility. TestHStore showed that this feature conflicts with the use case of mapping to a `Map<?, ?>` the plain `getObject` way.

I'm a bit conflicted on how this should be integrated. On one hand I think it should be registered OOTB because MapMapper is. On the other, there's backward compatibility.

I could build a golden middle way and add a Config class to MapMapper with a boolean `enableGenericsSupport` default false (and collaterally move the `foldCase` setting to it, since the current design isn't so good), and integrate generics support right into Map. Then all a user needs to do to "enable generics support" in MapMapper is flipping the switch. But then aren't we building roundabouts around the concept of "registering the plugins you want"?

TheRealMarnes:
I've no idea what the heck that build failure is...

TheRealMarnes:
Gonna postpone work on this until after breaking up the BuiltIn factories (which I intend to do next)

PR #1180, opened 2018-06-14
3.3.0 release

qualidafial:
Released to Maven Central. Publishing docs to jdbi.org momentarily..


TheRealMarnes:
I thought we weren't gonna remove the deprecated methods to keep compatibility until a major version release?

qualidafial:
@TheRealMarnes You're right, and this is definitely a case of me cheating on our rules. Basically, what it came down to was that we're already making breaking changes this release, the APIs are very new, and the addition of the `@Beta` annotation gives us a convenient excuse to call a mulligan.

PR #1181, opened 2018-06-14
[done] MapMapper JdbiConfig

TheRealMarnes:
This PR enables MapMapper configuration the proper way so that it works with SqlObject and follows project convention, and adds an uppercase option.

100% backward compatible: if foldCase is true, LOWER is used, otherwise the new option value is used, which defaults to NOP. Users who prefer the new API would have a problem using this option in SqlObject for the time being, since they wouldn't be able to turn off foldCase. I've added an override setting for them. Tests show this.

`@Beta` added where necessary, as well as `@Deprecated` so that a cleanup can be done for jdbi4.

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,3 +1,9 @@
+3.4.0
+  - New API
+    - Added a JdbiConfig class to MapMapper to configure column name case changes, with a setting to entirely ignore the old preference (necessary for SqlObject usage).
+  - Deprecations
+    - Deprecated the boolean property `foldCase` on MapMapper in favor of the JdbiConfig mechanism.
+
 3.3.0
   - New API
     - SerializableTransactionRunner.setOnFailure(), setOnSuccess() methods allow callbacks to be
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,8 +1,6 @@
 3.4.0
   - New API
-    - Added a JdbiConfig class to MapMapper to configure column name case changes, with a setting to entirely ignore the old preference (necessary for SqlObject usage).
-  - Deprecations
-    - Deprecated the boolean property `foldCase` on MapMapper in favor of the JdbiConfig mechanism.
+    - Added the MapMappers JdbiConfig class to configure column name case changes in favor of the old boolean toggle.
 
 3.3.0
   - New API
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -1,7 +1,7 @@
 3.4.0
   - New API
     - StatementException.getShortMessage
-    - Added the MapMappers JdbiConfig class to configure column name case changes in favor of the old boolean toggle.
+    - Added the MapMappers JdbiConfig class to configure column name case changes, preferred over the old boolean toggle.
     - SqlStatements.setQueryTimeout(int) to configure the JDBC Statement queryTimeout.
   - Bug Fixes
     - Bridge methods cause SqlObject exceptions to get wrapped in `InvocationTargetException`
--- a/core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
@@ -20,13 +20,11 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.UnaryOperator;
-import org.jdbi.v3.core.config.JdbiConfig;
+import javax.annotation.Nullable;
 import org.jdbi.v3.core.statement.StatementContext;
-import org.jdbi.v3.meta.Beta;
 
 /**
  * Yo dawg, I heard you like maps, so I made you a mapper that maps rows into {@code Map<String,Object>}. Map
@@ -38,23 +36,21 @@ public class MapMapper implements RowMapper<Map<String, Object>> {
      * @deprecated TODO remove in jdbi4
      */
     @Deprecated
-    private final boolean toLowerCase;
+    private final Boolean toLowerCase;
 
     /**
      * Constructs a new MapMapper, with map keys converted to lowercase.
      */
     public MapMapper() {
-        this(true);
+        this(null);
     }
 
     /**
      * Constructs a new MapMapper
-     * @param toLowerCase if true, column names are converted to lowercase in the mapped {@link Map}.
-     *
-     * @deprecated use the {@link Config} class instead
+     * @param toLowerCase if true, column names are converted to lowercase in the mapped {@link Map}. If null, the new MapNappers config will be used.
      */
-    @Deprecated
-    public MapMapper(boolean toLowerCase) {
+    // TODO deprecate when MapMappers.caseChange is out of beta
+    public MapMapper(@Nullable Boolean toLowerCase) {
         this.toLowerCase = toLowerCase;
     }
 
@@ -65,8 +61,11 @@ public class MapMapper implements RowMapper<Map<String, Object>> {
 
     @Override
     public RowMapper<Map<String, Object>> specialize(ResultSet rs, StatementContext ctx) throws SQLException {
-        Config config = ctx.getConfig(Config.class);
-        List<String> columnNames = getColumnNames(rs, !config.getForceNewApi() && toLowerCase ? Config.CaseChange.LOWER : config.getCaseChange());
+        UnaryOperator<String> caseChange = toLowerCase == null
+            ? ctx.getConfig(MapMappers.class).getCaseChange()
+            : (toLowerCase ? MapMappers.LOCALE_LOWER : MapMappers.NOP);
+
+        List<String> columnNames = getColumnNames(rs, caseChange);
 
         return (r, c) -> {
             Map<String, Object> row = new LinkedHashMap<>(columnNames.size());
@@ -79,7 +78,7 @@ public class MapMapper implements RowMapper<Map<String, Object>> {
         };
     }
 
-    private static List<String> getColumnNames(ResultSet rs, Config.CaseChange caseChange) throws SQLException {
+    private static List<String> getColumnNames(ResultSet rs, UnaryOperator<String> caseChange) throws SQLException {
         // important: ordered and unique
         Set<String> columnNames = new LinkedHashSet<>();
         ResultSetMetaData meta = rs.getMetaData();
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -16,19 +16,29 @@
...
  * keys are column names, while map values are the values in those columns. Map keys are converted to lowercase by
  * default.
  */
 public class MapMapper implements RowMapper<Map<String, Object>> {
-    private final boolean foldCase;
+    /**
+     * @deprecated TODO remove in jdbi4
+     */
+    @Deprecated
+    private final boolean toLowerCase;
```
qualidafial: Simpler to make this a `Boolean`, and have this field take precedence if non-null.

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -52,32 +65,106 @@ public MapMapper(boolean foldCase) {
...
+                throw new RuntimeException("column " + name + " appeared twice in this resultset!");
             }
+        }
 
-            columnNames[i] = alias;
+        return new ArrayList<>(columnNames);
+    }
+
+    @Beta
+    public static class Config implements JdbiConfig<Config> {
```
qualidafial: Our naming convention for configuration classes is to pluralize the configured API: e.g. `SqlStatements`, `Arguments`, `RowMappers`, etc.

This convention was chosen to make inline configuration commands read like sentences:

```java
Map<String, Foo> foosById = handle.createQuery("select * from foo")
    .configure(MapEntryMappers.class, config -> config.setKeyColumn("id"))
    .collectInto(new GenericType<Map<String, Foo>>() {});
```
Let's make this a top-level class named `MapMappers`.

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -39,10 +49,13 @@ public MapMapper() {
...
     /**
      * Constructs a new MapMapper
-     * @param foldCase if true, column names are converted to lowercase in the mapped {@link Map}.
+     * @param toLowerCase if true, column names are converted to lowercase in the mapped {@link Map}.
+     *
+     * @deprecated use the {@link Config} class instead
      */
-    public MapMapper(boolean foldCase) {
-        this.foldCase = foldCase;
+    @Deprecated
```
qualidafial: Let's not deprecate until the new API is out of beta

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -52,32 +65,106 @@ public MapMapper(boolean foldCase) {
...
+        return new ArrayList<>(columnNames);
+    }
+
+    @Beta
+    public static class Config implements JdbiConfig<Config> {
+        private CaseChange caseChange;
+        private boolean forceNewApi;
+
+        public Config() {
+            caseChange = CaseChange.NOP;
```
qualidafial: Current default behavior is to convert to lower case. This default must be preserved or we will break existing code.

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -52,32 +65,106 @@ public MapMapper(boolean foldCase) {
...
+            forceNewApi = that.forceNewApi;
+        }
+
+        @Beta
+        public CaseChange getCaseChange() {
+            return caseChange;
+        }
+
+        @Beta
+        public Config setCaseChange(CaseChange caseChange) {
```
qualidafial: Would be more flexible as a `Function<String, String>` that converts column names to keys. Then we don't restrict users to just a few options. They could e.g. use Guava's `CaseFormat` to convert from e.g. `UPPER_SNAKE_CASE` to `camelCase` if they wanted.

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -52,32 +65,106 @@ public MapMapper(boolean foldCase) {
...
+                row.put(columnNames.get(i), rs.getObject(i + 1));
             }
-            if (foldCase) {
-                alias = alias.toLowerCase(Locale.ROOT);
+
+            return row;
+        };
+    }
+
+    private static List<String> getColumnNames(ResultSet rs, Config.CaseChange caseChange) throws SQLException {
```
qualidafial: Nice refactor on this method

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
TheRealMarnes: it is preserved in the current implementation. I'll make sure it stays that way after changing the code for your remarks

TheRealMarnes: `foldCase` is lowercase vs nop. I think it would be good to change this whole thing to default to nop in the future. I used this feature recently and was surprised to see my column names being modified by default. Law of least surprise...

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
TheRealMarnes: done

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
TheRealMarnes: done

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
TheRealMarnes: done

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
TheRealMarnes: done

core/src/main/java/org/jdbi/v3/core/mapper/MapMappers.java
```diff
@@ -0,0 +1,56 @@
...
+    public static final UnaryOperator<String> NOP = UnaryOperator.identity();
+    @Beta
+    public static final UnaryOperator<String> LOCALE_LOWER = s -> s.toLowerCase(Locale.ROOT);
+    @Beta
+    public static final UnaryOperator<String> LOCALE_UPPER = s -> s.toUpperCase(Locale.ROOT);
+
+    private UnaryOperator<String> caseChange;
+
+    public MapMappers() {
+        // TODO jdbi4 law of least surprise: change to nop
```
qualidafial: If you merge from master, there's a `4.0_TODO` file in the project root you can add this to: https://github.com/jdbi/jdbi/blob/master/4.0_TODO

core/src/main/java/org/jdbi/v3/core/transaction/SerializableTransactionRunner.java
```diff
@@ -115,6 +115,7 @@ protected boolean isSqlState(String expectedSqlState, Throwable throwable) {
     /**
      * Configuration for serializable transaction runner
      */
+    // TODO jdbi4 move to top-level class to comply with convention
```
qualidafial: Good catch--another candidate for the `4.0-TODOS` file

TheRealMarnes:
Is moving the todo markers to that file a good idea though? I mean, _all_ todos should be handled or conceded before a major release, and they're placed right at the thing that needs to be changed. Moving them to a todo file removes them from their context and makes them less obvious, and stops your IDE from managing them and reminding you about them too. What's the use of it?

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -28,21 +32,26 @@
...
 
     /**
      * Constructs a new MapMapper
-     * @param foldCase if true, column names are converted to lowercase in the mapped {@link Map}.
+     * @param toLowerCase if true, column names are converted to lowercase in the mapped {@link Map}. If null, the new MapNappers config will be used.
      */
-    public MapMapper(boolean foldCase) {
-        this.foldCase = foldCase;
+    // TODO deprecate when MapMappers.caseChange is out of beta
+    public MapMapper(@Nullable Boolean toLowerCase) {
```
qualidafial: Changing this from `boolean` to `Boolean` is a binary breaking change.

Let's change this back to `boolean`, and let the other constructor explicitly assign null rather than delegating to this constructor.

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -28,21 +32,26 @@
...
      */
     public MapMapper() {
-        this(true);
+        this(null);
     }
 
     /**
      * Constructs a new MapMapper
-     * @param foldCase if true, column names are converted to lowercase in the mapped {@link Map}.
+     * @param toLowerCase if true, column names are converted to lowercase in the mapped {@link Map}. If null, the new MapNappers config will be used.
```
qualidafial: > If null, the new 

The new API will only be considered "new" for a short time, so let's leave out that qualifier.

> MapNappers

Little typo there :wink:

TheRealMarnes:
compatibility fix pushed... do we really wanna move the todo markers too? It feels like a snooze button on an alarm clock to me.

stevenschlansker:
The point of having them separate in a file is that in the particular case of a major release, the todo's are not actionable until you are in the post-3 pre-4 branch.  So having a list of them in one place, to come by and hit them all in that period, is possibly more useful than having them strewn around the codebase tempting you to help out but leaving you unable to.

qualidafial:
There will need to be a significant benefit to justify breaking compatibility for 4.x.

With 3.x the primary benefit was to take advantage of Java 8. I have a few major changes in mind for 4.x, but I don't want to pull that trigger until there's a serious benefit.

We had a similar to-do list during 2.x to 3.x, and it was one of the primary documents driving 3.x development.

RELEASE_NOTES
```diff
@@ -1,6 +1,7 @@
 3.4.0
   - New API
     - StatementException.getShortMessage
+    - Added the MapMappers JdbiConfig class to configure column name case changes in favor of the old boolean toggle.
```
qualidafial: > in favor of the old boolean toggle

"In favor of" is the thing you're moving toward, not away from.

Maybe something like:

```
Added MapMappers config class with `caseChange` (or whatever we name it) property, as a replacement for the `MapMapper(boolean)` constructor.

TheRealMarnes:
I'm not sure what your plans are regarding the new major version and everything to do with it... I just hope it does actually come and we fix every todo and deprecation for it.

qualidafial:
We will clean up everything that's tracked in 4.x, but it needs to be easy to find so as not to be overlooked. Hence the todos file.

Just as a side note: we got about 10 years of mileage out of Jdbi 2.x. It may be a long time before we pursue 4.x in earnest. That's part of why we're so picky about getting names right on the first try--they live a long, long time.

TheRealMarnes:
> but it needs to be easy to find so as not to be overlooked. 

But that's exactly what todo markers are good for... IDEs pick them up and display them in a single convenient overview while keeping them also in their local context. This gives you the advantage of _both_ local markers _and_ a centralized overview like the todos file. Don't tell me you guys work in a plain text editor and need a centralized file for that reason?

image

> That's part of why we're so picky about getting names right on the first try

Oh don't get me wrong, I understand that and am glad about it. It annoys me endlessly that my colleagues at work are so damn sloppy about everything and brush everything off as "it doesn't really matter" when they have no arguments besides aesthetics and laziness to justify technically inferior solutions over the proper way to go that takes no additional effort at all (and in fact saves time, effort, and frustration in the future).

stevenschlansker:
Please don't get frustrated, just consider that we have slightly different workflows than you do :)  I hope you'll see this "inferior" way as just a small hurdle -- hopefully we make few enough mistakes that we do not have to use this often.  Having the file does help us -- I used the previous edition ( https://github.com/jdbi/jdbi/blob/jdbi2/CHANGES_PLANNED_FOR_3_0 ) to inspire 3

We can consider changing the way we do things, if you'd like to propose a process change that's great, but generally let's consider that as a separate issue rather than an addendum on a PR.

stevenschlansker:
cleaned up the nested ternary to a more functional style https://github.com/jdbi/jdbi/pull/1229/commits/ce308dc4a30049b0c5fa76c230e06bc240e7444c and merging as #1229

stevenschlansker:
I also factored out case strategy into its own enum https://github.com/jdbi/jdbi/pull/1229/commits/e418324d7c1236fb1cc957897a0e5145e55ab007 hopefully you like the change

Issue #1182, opened 2018-06-14
POM Dependency Conflict


isaki-x:
Greetings!

I'm opening an issue instead of a PR to get a consensus on what the fix should be (at which point I will submit a PR).

The issue is that when trying to build DropWizard against 3.3.0, the build fails:

https://github.com/dropwizard/dropwizard/pull/2394

The reason for the failure is that jdbi3-core is pulling in both antlr-runtime 3.3 and 3.4. This issue did not exist in 3.2.1.

This problem was introduced by depending on `stringtemplate` 4.0.2:

https://repo.maven.apache.org/maven2/org/antlr/stringtemplate/4.0.2/stringtemplate-4.0.2.pom

As you can see this explicitly requires antlr-runtime 3.3.

**Proposals**

1. Rollback the dependency update for string template.
2. Exclude antlr runtime from the string template dependency definition.

Please let me know which you prefer and I will make it happen.

NOTE: Regardless of the above, I would like to add dependency convergence enforcement to your build to catch this sort of issue going forward.

EDIT: JDBI3 correctly fails its build with enforcement enabled:

```
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.DependencyConvergence failed with message:
Failed while enforcing releasability the error(s) are [
Dependency convergence error for org.antlr:antlr-runtime:3.4 paths to dependency are:
+-org.jdbi:jdbi3-core:3.3.0
  +-org.antlr:antlr-runtime:3.4
and
+-org.jdbi:jdbi3-core:3.3.0
  +-org.antlr:antlr-runtime:3.4
    +-org.antlr:antlr-runtime:3.3
```

EDIT2: What's interesting, is forcing the inclusion of string template in core results in the following failures:

```
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.DependencyConvergence failed with message:
Failed while enforcing releasability the error(s) are [
Dependency convergence error for org.antlr:stringtemplate:4.0.2 paths to dependency are:
+-org.jdbi:jdbi3-core:3.3.0
  +-org.antlr:antlr-runtime:3.4
    +-org.antlr:stringtemplate:4.0.2
and
+-org.jdbi:jdbi3-core:3.3.0
  +-org.antlr:stringtemplate:4.0.2
and
+-org.jdbi:jdbi3-core:3.3.0
  +-org.antlr:stringtemplate:4.0.2
    +-org.antlr:stringtemplate:3.2.1
, 
[...]

PR #1183, opened 2018-06-14
Dependency Convergence Fix for ANTLR

isaki-x:
This PR does two things:

**Enable Maven Enforcement**

I have enabled two built in rules, the first is for Java version (I've set a minimum of Java 8 as the code base does utilize lambda) and the second for dependency convergence checks.

**Exclude antlr-runtime for stringtemplate in Maven managed dependencies**

This seems to fix the issue where a circular dependency is introduced by requiring 3.4 of ANTLR while requiring 4.0.2 for string template.

Please see my notes here: #1182

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
 
         <dep.antlr.version>3.4</dep.antlr.version>
         <dep.checkstyle.version>8.10</dep.checkstyle.version>
-        <dep.dokka.version>0.9.16</dep.dokka.version>
+        <dep.dokka.version>0.9.17</dep.dokka.version>
         <dep.jetbrainsAnnotations.version>13.0</dep.jetbrainsAnnotations.version>
         <dep.kotlin.version>1.2.31</dep.kotlin.version>
         <dep.plugin.checkstyle.version>3.0.0</dep.plugin.checkstyle.version>
@@ -496,6 +496,11 @@
                     <configuration>
                         <jdkVersion>8</jdkVersion>
                     </configuration>
+                </plugin>
+                <plugin>
+                    <!-- Fixes missing plugin during CI builds -->
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-jar-plugin</artifactId>
                 </plugin>
                 <plugin>
                     <groupId>org.antlr</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -496,6 +496,19 @@
                     </configuration>
                 </plugin>
                 <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-jar-plugin</artifactId>
+                    <executions>
+                        <execution>
+                            <id>default-jar</id>
+                            <phase>package</phase>
+                            <goals>
+                                <goal>jar</goal>
+                            </goals>
+                        </execution>
+                    </executions>
+                </plugin>
+                <plugin>
                     <groupId>org.jetbrains.dokka</groupId>
                     <artifactId>dokka-maven-plugin</artifactId>
                     <version>${dep.dokka.version}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -606,6 +606,28 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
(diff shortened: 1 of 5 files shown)


isaki-x:
This PR has turned into 4 changes:

- Added enforcer plugin support.
- Corrected circular dependency for antlr.
- Upgraded Dakka from 0.9.16 to 09.17.
- Explicitly configured maven JAR plugin.

pom.xml
```diff
@@ -489,6 +495,19 @@
                         </excludes>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-jar-plugin</artifactId>
+                    <executions>
+                        <execution>
```
stevenschlansker: What's the purpose of configuring `default-jar` here?

isaki-x: It isn't really required as I don't think JDBI builds any shaded jars. The reason it is there is I lifted it from another POM I have which sets other settings for the plugin as well and this was required to get everything working for that particular build. I removed most of the custom settings and neglected to remove this one. I will remove it.

isaki-x:
@stevenschlansker Good catch on the excess configuration for the jar plugin; I have committed a corrected version of the POM for your review.

pom.xml
```diff
@@ -601,6 +612,28 @@
...
+                <executions>
+                    <execution>
+                        <id>enforce-jdbi</id>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <rules>
+                                <!-- Ensure consistency -->
+                                <dependencyConvergence />
```
qualidafial: We already have a different plugin for this purpose from `basepom`. The enforcer plugin doesn't provide very helpful messages, unfortunately.

pom.xml
```diff
@@ -392,6 +392,12 @@
                 <groupId>org.antlr</groupId>
                 <artifactId>stringtemplate</artifactId>
                 <version>4.0.2</version>
+                <exclusions>
+                    <exclusion>
+                        <artifactId>antlr-runtime</artifactId>
+                        <groupId>org.antlr</groupId>
+                    </exclusion>
+                </exclusions>
```
qualidafial: I'm still unclear why this exclusion would be necessary, since we're managing the `antlr-runtime` dependency version on line 225.

pom.xml
```diff
@@ -489,6 +495,11 @@
                         </excludes>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <!-- Fixes missing plugin during CI builds -->
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-jar-plugin</artifactId>
+                </plugin>
```
qualidafial: Let's try to keep separate concerns in separate PRs. Something like this is easy to review and approve in isolation. In practice it takes much longer to review and approve PRs when multiple concerns are conflated into the same work.

pom.xml
isaki-x: It is because of a circular dependency. You end up with two versions of the same library being pulled in. The dependency management seems to fall over on a circular dependency on a managed dependency. If you remove this exclusion block (and you are DropWizard, or you have the enforcer plugin enabled yourself, which I have done), the build fails. `dropwizard-jdbi3` does not build on anything newer than 3.2.1 of JDBI as a result of the issue this fixes.

pom.xml
isaki-x: Sure, I can move this. This fixed the fact CI was failing which is why I included it. I can put in a separate branch for the CI fixes.

pom.xml
isaki-x: I'm unfamiliar with `basepom`, however it clearly doesn't handle dependency convergence the same way as the enforcer plugin as it allowed a circular dependency reference to make it through (breaking downstream things, like DropWizard, that rely on dependency convergence to prevent bugs/issues). Perhaps there is a way to configure it so that it plays ball with the standard maven plugins used by projects that depend on JDBI3?

isaki-x:
Please see #1186 for a PR with just the CI fixes. Once that is resolved, those changes will disappear from this PR. I do not wish to remove them from here as that will cause the CI to start failing again.

stevenschlansker:
We merged in that PR, this one can now be rebased

isaki-x:
I will reopen with a clean history. The rebase brought in too much noice.

PR #1184, opened 2018-06-15
[done] #1175 allow unused Bindings + todo bug

### LINKED ISSUE #1175: bindNamed() is too restrictive about requiring named parameters if binding is present.
We are using a conditional statement as follows, that causes and exception within bindNamed() when the condition is false. Why not change bindNamed to just return if parameterNames is empty?

```
@SqlQuery("select * from table_name <if(boolean_variable)> where id = :id <endif>")
@UseStringTemplateEngine
public  abstract List<SomeClass> query(@Define("boolean_variable") Boolean boolean_variable,
        @Bind("id") String id);
```

bindNamed() snippet for reference.

```
private static void bindNamed(List<String> parameterNames, Binding binding, PreparedStatement statement, StatementContext context) {
    if (parameterNames.isEmpty() && !binding.isEmpty()) {
        throw new UnableToExecuteStatementException(
            String.format("Unable to execute. The query doesn't have named parameters, but provided binding '%s'.", binding),
            context);
    }
```

TheRealMarnes:
fixes #1175

Testing showed the current code is inconsistent and doesn't cover many cases. It throws an exception when >0 arguments are provided but the query contains 0 variables. It does not throw when too many arguments are provided as long as the query defines at least 1 variable. @rjnay could work around his problem for now by changing his query to `where :id = :id<if(bool)> and id = :id<endif>`. I think this should be fixed. Shall I make a separate PR for that too when other open things have been wrapped up?

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
@@ -50,7 +50,7 @@ class ArgumentBinder {
 
     private static void bindNamed(List<String> parameterNames, Binding binding, PreparedStatement statement, StatementContext context) {
         // TODO throw on any mismatch, not only empty/non-empty
-        if (!context.getConfig(SqlStatements.class).getAllowUnusedBindings() && parameterNames.isEmpty() && !binding.isEmpty()) {
+        if (parameterNames.isEmpty() && !binding.isEmpty() && !context.getConfig(SqlStatements.class).getAllowUnusedBindings()) {
             throw new UnableToExecuteStatementException(
                     String.format("Unable to execute. The query doesn't have named parameters, but provided binding '%s'.", binding),
                     context);
--- a/core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
@@ -34,6 +34,12 @@ class ArgumentBinder {
     }
 
     private static void bindPositional(ParsedParameters params, Binding binding, PreparedStatement statement, StatementContext context) {
+        // best effort: just try +1 (unless we expose a method to get the full binding count)
+        boolean moreArgumentsProvidedThanDeclared = binding.findForPosition(params.getParameterCount()).isPresent();
+        if (moreArgumentsProvidedThanDeclared && !context.getConfig(SqlStatements.class).getAllowUnusedBindings()) {
+            throw new UnableToCreateStatementException("Superfluous positional param at (0 based) position " + params.getParameterCount(), context);
+        }
+
         for (int i = 0; i < params.getParameterCount(); i++) {
             Optional<Argument> argument = binding.findForPosition(i);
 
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
```diff
@@ -48,11 +49,13 @@ private static void bindPositional(int size, Binding binding, PreparedStatement
     }
 
     private static void bindNamed(List<String> parameterNames, Binding binding, PreparedStatement statement, StatementContext context) {
-        if (parameterNames.isEmpty() && !binding.isEmpty()) {
+        // TODO throw on any mismatch, not only empty/non-empty
+        if (!context.getConfig(SqlStatements.class).getAllowUnusedBindings() && parameterNames.isEmpty() && !binding.isEmpty()) {
```
stevenschlansker: Looking up the configuration in the context should be last in the short-circuit, to avoid repeatedly searching configuration for every single bind.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -165,6 +169,17 @@ public SqlStatements setSqlLogger(SqlLogger sqlLogger) {
         return this;
     }
 
+    @Beta
+    public boolean getAllowUnusedBindings() {
+        return allowUnusedBindings;
+    }
+
+    @Beta
+    public SqlStatements setAllowUnusedBindings(boolean allowUnusedBindings) {
```
stevenschlansker: javadoc describing the feature

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -165,6 +169,17 @@ public SqlStatements setSqlLogger(SqlLogger sqlLogger) {
         return this;
     }
 
+    @Beta
```
stevenschlansker: I don't think this feature needs `@Beta` personally.

core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
```diff
@@ -48,11 +49,13 @@ private static void bindPositional(int size, Binding binding, PreparedStatement
     }
 
     private static void bindNamed(List<String> parameterNames, Binding binding, PreparedStatement statement, StatementContext context) {
-        if (parameterNames.isEmpty() && !binding.isEmpty()) {
+        // TODO throw on any mismatch, not only empty/non-empty
```
qualidafial: Exception to this todo is out-parameters.

TheRealMarnes: Never used those. I assume they can't be distinguished?

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -165,6 +167,20 @@ public SqlStatements setSqlLogger(SqlLogger sqlLogger) {
         return this;
     }
 
+    public boolean getAllowUnusedBindings() {
```
qualidafial: Typically boolean getters get an `is` prefix instead of `get`

core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
qualidafial: Not without parsing intent out of the SQL statement, which is way outside our project charter.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
TheRealMarnes: Personally I hate that convention because it often doesn't make any grammatical sense... not every boolean is an `is` state and singular... What would you name the methods? `areUnusedBindingsAllowed` and `unusedBindingsAreAllowed(bool)`?

core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
qualidafial: This is why missing positional parameters are just ignored in the `bindPositional` method.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
qualidafial: Could make it grammatical e.g. `isUnusedBindingAllowed()`

core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
TheRealMarnes: So can we do any extra validation then or is what's present as good as it can possibly be?

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
TheRealMarnes: setter `unusedBindingIsAllowed(bool)` then?

TheRealMarnes:
Hm, I'm confused by the asymmetry between how nameds and positionals are handled. Named args are meant to work the same way as positionals, just with nicer syntax for the user, right? If there's a check for `allowUnusedBindings` for the nameds, shouldn't there also be for positionals? And shouldn't multiplicity checks be the same for both too?

Positionals are allowed to be absent because

> any missing positional Arguments could be return parameters

but `registerOutParameter` registers an Argument, so how can this be legit?

TheRealMarnes:
Sooo I took the liberty to rewrite ArgumentBinder a bit beyond my original intentions and add a test for it. Binding nameds and positionals now behave the same way. Both throw similar exception messages, check for missing `Argument`s, catch binding exceptions, and do a best-effort check that there aren't any excessive `Argument`s, unless allowed by `SqlStatements.allowUnusedBindings`.

If you run the test class I added without the changes to ArgumentBinder, you'll see its behavior was inconsistent. I'd consider this a bugfix.

Regarding out params, they just work. Out params register an `Argument` so the logic in ArgumentBinder that ignored missing `Argument`s (with a comment to explicitly say so) was wrong, I reckon.

TheRealMarnes:
Dayum, couldn't get anything to work yesterday, and now I've fixed 4 of my PRs in half an evening

TheRealMarnes:
@stevenschlansker 
image

Could be solved with a copy variable inside the loop, but then that isn't much better imho. Which one do you want?

TheRealMarnes:
@stevenschlansker 
image

Is this really much better? ;/

stevenschlansker:
How about:

```java
IntStream.range(0, params.getParameterCount())
    .mapToObj(i -> binding.findForPosition(i).orElseThrow(
        () -> new UnableToCreateStatementException(...)))
    .forEach(a -> a.apply(...))
```

or something like that?  might have to add helper method `JdbiOptionals.stream`, it's coming in jdk10 but not available in 8.

TheRealMarnes:
You need a reference to the index to apply the arg though... Can't be done that way, or any other I've tried.

stevenschlansker:
I ended up cleaning it up like this:
```diff
diff --git a/core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java b/core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
index dad92375..83d10cb6 100644
--- a/core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
-            String name = paramNames.get(i);
-
-            Optional<Argument> argument = binding.findForName(name, context);
+            final String name = paramNames.get(i);
 
             try {
-                argument.orElseThrow(() -> new UnableToCreateStatementException(String.format("Missing named parameter '%s'.", name), context)).apply(i + 1, statement, context);
+                binding.findForName(name, context)
+                    .orElseThrow(() -> new UnableToCreateStatementException(String.format("Missing named parameter '%s'.", name), context))
+                    .apply(i + 1, statement, context);
             } catch (SQLException e) {
                 throw new UnableToCreateStatementException(String.format("Exception while binding named parameter '%s'", name), e, context);
             }
```

Sorry the wordsmithing added such a delay.

PR #1194, opened 2018-06-26
Fix write-only property in BeanMapper

sjoerdtalsma:
If there is only a setter method the BeanMapper can still be useful as a RowMapper for output.
In this case, the `descriptor.getReadMethod()` being `null` will cause NPE when determining the type to be mapped.
This is not necessary, as the type can then determined by parameter of the `descriptor.getWriteMethod()`.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java
@@ -26,6 +26,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


stevenschlansker:
Thanks.  This would be great to have a test case added as well, if you can.

sjoerdtalsma:
@stevenschlansker Not sure I can manage that with time being rather hard to come by lately 😉 

But the bug is easily reproduced: try returning a `List<WriteableBean>` using the `BeanMapper`:
```java
public class WriteableBean {
    private String value;
    public void setValue(String value) {
        this.value = value;
    }
    @Override
    public String toString() {
        return getClass().getSimpleName() + '{' + value + '}';
    }
}
```

PR #1195, opened 2018-06-26
[done] findbugs is dead, long live spotbugs

TheRealMarnes:
https://mailman.cs.umd.edu/pipermail/findbugs-discuss/2016-November/004321.html

TLDR findbugs project has/had 1 person with project admin rights, and he went

> poof

so they forked it to be able to continue work on it. AKA spotbugs.


pom.xml
```diff
@@ -573,6 +576,22 @@
                         <sourceDirectories>${project.build.sourceDirectory},${project.build.testSourceDirectory}</sourceDirectories>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>com.github.spotbugs</groupId>
+                    <artifactId>spotbugs-maven-plugin</artifactId>
+                    <version>3.1.3</version>
```
PascalSchumacher: Most recent version is `3.1.5`.

TheRealMarnes:
Hold on a sec please, I'm derping like crazy... fixing this commit

TheRealMarnes:
@PascalSchumacher there, fixed. Sorry, don't know why I was screwing up so much and not even realizing.

PascalSchumacher:
Thanks! 👍 😃

pom.xml
```diff
@@ -573,6 +574,22 @@
                         <sourceDirectories>${project.build.sourceDirectory},${project.build.testSourceDirectory}</sourceDirectories>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>com.github.spotbugs</groupId>
+                    <artifactId>spotbugs-maven-plugin</artifactId>
+                    <version>${spotbugs.version}</version>
+                    <dependencies>
```
PascalSchumacher: The dependency declaration can be omitted as the spotbugs-maven-plugin and spotbugs version are the same.

TheRealMarnes:
Something's pretty wrong with jdbi on windows lately... the builds on my machine are wildly different from the builds on jenkins with non-reproducible or even inconsistent test failures and build plugin errors...

src/build/spotbugs-include.xml
```diff
@@ -0,0 +1,21 @@
...
+~   distributed under the License is distributed on an "AS IS" BASIS,
+~   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+~   See the License for the specific language governing permissions and
+~   limitations under the License.
+-->
+<FindBugsFilter xmlns="https://github.com/spotbugs/filter/3.0.0"
+                xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                xsi:schemaLocation="https://github.com/spotbugs/filter/3.0.0 https://raw.githubusercontent.com/spotbugs/spotbugs/3.1.5/spotbugs/etc/findbugsfilter.xsd">
+	<Match>
+		<Package name="~.*"/>
```
stevenschlansker: What does this `~` matching do?
I tried to look up the file format https://spotbugs.github.io/spotbugs-maven-plugin/usage.html but it links to a 404

pom.xml
```diff
@@ -90,9 +90,10 @@
         <dep.pmd.version>6.0.1</dep.pmd.version>
         <dep.slf4j.version>1.7.25</dep.slf4j.version>
         <dep.spring.version>4.2.4.RELEASE</dep.spring.version>
-        <findbugs.version>3.0.1</findbugs.version>
+        <findbugs.version>3.0.2</findbugs.version>
```
stevenschlansker: should we just remove this, since we aren't using findbugs anymore?

pom.xml
```diff
@@ -573,6 +574,27 @@
...
+                    <version>${spotbugs.version}</version>
+                    <configuration>
+                        <includeFilterFile>${maven.multiModuleProjectDirectory}/src/build/spotbugs-include.xml</includeFilterFile>
+                        <excludeFilterFile>${maven.multiModuleProjectDirectory}/src/build/spotbugs-exclude.xml</excludeFilterFile>
+                    </configuration>
+                </plugin>
+                <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>findbugs-maven-plugin</artifactId>
+                    <executions>
```
stevenschlansker: do we need to both override the execution *and* skip?  or is one or the other sufficient?

src/build/spotbugs-include.xml
TheRealMarnes: `~` tells it to use regex. I couldn't find out anywhere what exactly will be given to the regex to match against though, but trial and error in my private project showed that `.*` simply matches all your project files (i.e. exactly what we want). But yeah, I don't know what the string is that it's matching either :/ It's a pretty common problem in a lot of options like this one in a lot of these tools. Their docs only mention high-level what will be matched without giving a precise example...

TheRealMarnes: I used this for info btw: http://findbugs.sourceforge.net/manual/filter.html

Spotbugs is backward compatible with findbugs, config files included

pom.xml
TheRealMarnes: It's still being used for the `jsr305` artifact

pom.xml
TheRealMarnes: I thought to play it safe. I don't rightly know what's the proper thing to do.

qualidafial:
Would be good to submit a similar PR to the basepom project

stevenschlansker:
basepom 26 includes spotbugs so hopefully we don't need to do additional upstreaming.

qualidafial:
Thanks @TheRealMarnes

Issue #1193, opened 2018-06-26
NPE when using @Define on null values


kaandok:
I'm currently using null values (albeit without success) while trying to implement conditional sql statements

```java
@SqlQuery("select * from test_table <if(created)> where created >= #created <endif>")
List<TestItem> getTestItems(@Define OffsetDateTime created);
```

If I pass a non-null OffsetDateTime value, no problems.
However when I do, it results in an NPE at https://github.com/jdbi/jdbi/blob/2cb484e3b9357e77ba12de09655632a1270ecb49/core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java#L54

This is because the `attributes` (a ConcurrentHashMap instance) doesn't accept null values.

Is this the expected behavior, if not happy to do a simple PR guarding against trying to put null values into the ConcurrentHashMap, which if I'm not mistaken will fix my issue.

If this is the expected behavior, what's the best way to go about using this kind of conditional SQL statements?

TheRealMarnes:
> what's the best way to go about using this kind of conditional SQL statements?

Defining your condition as a separate proper boolean to switch on, i.e. `@Define boolean created = value != null`

Whether `if(null)` will evaluate to false or explode is up to the TemplateEngine too.

stevenschlansker:
Yeah -- usually, you would bind either `true` or `false` unconditionally, so you would never need to bind a null here.

If there is some other reason to need to do so, or if this isn't good for some reason in your use case, we can consider allowing binding of explicit `null` as defined attributes.

kaandok:
Understood, thanks for both (@stevenschlansker @TheRealMarnes) your comments.

That was my current workaround as well.

The reason I was trying it my way (i.e. binding null) was because this was working on the 2.x line with StringTemplate templates.

I agree that for simple queries depending on one condition, passing it via a proper boolean switch is fine. However we have cases where we have > 5 conditions where we can switch on. For these cases I think it's a bit overkill to create another set of arguments in order to switch on them. In addition, there are template engines that do handle null well (Thymeleaf, https://www.thymeleaf.org/ is also one of them which I'm planning to implement for it's text based templates)

If you agree and/or if letting null values be usable for `@Define` annotations just by checking if the value isn't null before putting it into the concurrent hashmap isn't a terrible idea, then I'll create the PR.

Let me know please.

TheRealMarnes:
+1 for that PR from me

TheRealMarnes:
[...]

PR #1200, opened 2018-07-10
Allow null valued attribute definitions

kaandok:
implements #1193

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
@@ -60,6 +60,8 @@ public final class SqlStatements implements JdbiConfig<SqlStatements> {
     public SqlStatements define(String key, Object value) {
         if (value != null) {
             attributes.put(key, value);
+        } else {
+            attributes.remove(key);
         }
         return this;
     }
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -58,7 +58,9 @@ private SqlStatements(SqlStatements that) {
      * @return this
      */
     public SqlStatements define(String key, Object value) {
-        attributes.put(key, value);
+        if (value != null) {
+            attributes.put(key, value);
+        }
```
qualidafial: Should probably `remove(key)` if `value == null`.

kaandok: good idea, thank you.

qualidafial:
The last commit is triggering a PMD rule violation:

```[INFO] PMD Failure: org.jdbi.v3.core.statement.SqlStatements:61 Rule:ConfusingTernary Priority:3 Avoid if (x != y) ..; else ```

Need to invert the if/else clause so it's:

```java
if (value == null) {
  attributes.remove(key);
} else {
  attributes.put(key, value);
}
```

kaandok:
@qualidafial should be good to go now.

PR #1206, opened 2018-08-08
Increase flexibility of KotlinMapper.

cacsar:
Add support for @JdbiConstructor to KotlinMapper.
Allow writeable fields to not be mapped to columns if they are not
late init since they will have been initialized.


cacsar:
After refactoring some kotlin classes that sometimes get loaded via JDBI I hit a few IllegalArgumentExceptions since I had claimed the primary constructor for something else and had some writable fields I didn't need JDBI to try and initialize. This simply makes the KotlinMapper more permissive by supporting the previously existing @JdbiConstructor annotation and makes it so that only lateinit writable fields have to match.

PR #1209, opened 2018-08-09
Avoid formatting string when not needed

alecalve:
When profiling a project using JDBI 3.4.0, I noticed that up to 25% of the samples would happen in `Preconditions#checkNotNull`.

This was due to the error message being formatted even if the reference was not null (as it is in the great majority of the calls).

Running this patch, the profiler doesn't even pick up calls to the method anymore.

The more recent versions of Guava (where I assume this file was pulled from) use this way of checking for nullability.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/generic/internal/Preconditions.java
+++ b/core/src/main/java/org/jdbi/v3/core/generic/internal/Preconditions.java
@@ -14,8 +14,7 @@
 
 package org.jdbi.v3.core.generic.internal;
 
-import java.util.Objects;
-
+@SuppressWarnings("PMD.AvoidThrowingNullPointerException")
 public final class Preconditions {
     private Preconditions() {
         throw new UnsupportedOperationException("utility class");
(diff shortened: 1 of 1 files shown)


TheRealMarnes:
I like the code smell-wise fix of not formatting the string before it's needed, but I also agree with PMD I don't like seeing an NPE being thrown. NPEs are thrown by the jvm itself in the overwhelming majority of cases, and in the context of validation I don't think an argument illegally being null is so different from all other validation failures that it deserves its own exception. Would personally rather see IAE used here, in accordance with PMD recommendation.

alecalve:
I'm fine with making it an IAE, but wouldn't that break things down the line?

`Objects#requireNonNull` is basically identical to this patch's version of `checkNotNull` as it also throws an NPE.

TheRealMarnes:
It's not an exception (type + use case) that's supposed to be caught for any reason, but you're right, it might break some tests or at the very least _technically_ be an API change. I might have to concede this one.

PR #1213, opened 2018-08-14
Allow @Nested objects to be optionally included in result sets

### LINKED ISSUE #1212: Nested bean requires at least one column in all result sets
JDBI version: 3.4.0

Currently, if you add the ```@Nested``` annotation for an 'inner' bean in the 'outer' bean class, then issue a query against the outer bean that doesn't include any columns for the inner bean in the result set, the BeanMapper complains that there are no matching columns in the result set.  This seems to be because the 'columnNames' passed into the specialize0 method are expected to be empty if there are no mappers found, but the column names from the query are passed down to the nested bean mapper, even if none of them apply.

It seems like if the nested bean has no columns on the result set to map to it, the nested property should be ignored instead of having an exception thrown.  Perhaps the specialize0 method should exit early if no mappers are found for a nested bean (by providing some additional context to the method when called from an outer bean mapper)?

Example Bean:

```
public class Child {

    private Long id;
    private String comment;
    private Parent parent;

    @ColumnName("ID")
    public Long getId() {
        return id;
    }

    public void setId(Long id) {
        this.id = id;
    }

    @ColumnName("COMMENT")
    public String getComment() {
        return comment;
    }

    public void setComment(String comment) {
        this.comment = comment;
    }

    @Nested("p")
    public Parent getParent() {
        return parent;
    }

    public void setParent(Parent order) {
        this.parent = order;
    }
}
```

```
public class Parent {

    private Long id;
    private String name;
    private Set<Child> children = Sets.newLinkedHashSet();

    @ColumnName("ID")
    public Long getId() {
        return id;
    }

    public void setId(Long id) {
        this.id = id;
    }

    @ColumnName("NAME")
    public String getName() {
        return name;
    }

    public void setName(String name) {
        this.name = name;
    }

    public Set<Child> getChildren() {
        return children;
    }

    public void setChildren(Set<Child> children) {
        this.children = children;
    }
}
```

Query:

```SELECT ID, COMMENT FROM CHILD WHERE ID = :id```

This query throws:

```
java.lang.IllegalArgumentException: Mapping bean type class com.rw.arcus.example.model.Parent didn't find any matching columns in result set
! at org.jdbi.v3.core.mapper.reflect.BeanMapper.specialize0(BeanMapper.java:182)
! at org.jdbi.v3.core.mapper.reflect.BeanMapper.specialize0(BeanMapper.java:174)
! at org.jdbi.v3.core.mapper.reflect.BeanMapper.specialize(BeanMapper.java:125)
! at org.jdbi.v3.core.result.ResultSetResultIterator.<init>(ResultSetResultIterator.java:38)
! at org.jdbi.v3.core.result.ResultIterable.lambda$of$0(ResultIterable.java:53)
! at org.jdbi.v3.core.result.ResultIterable.findFirst(ResultIterable.java:116)
! at org.jdbi.v3.sqlobject.statement.internal.ResultReturner$CollectedResultReturner.mappedResult(ResultReturner.java:276)
! at org.jdbi.v3.sqlobject.statement.internal.SqlQueryHandler.lambda$configureReturner$0(SqlQueryHandler.java:54)
! at org.jdbi.v3.sqlobject.statement.internal.CustomizingStatementHandler.invoke(CustomizingStatementHandler.java:155)
! at org.jdbi.v3.sqlobject.statement.internal.SqlQueryHandler.invoke(SqlQueryHandler.java:26)
! at org.jdbi.v3.sqlobject.SqlObjectFactory.lambda$null$13(SqlObjectFactory.java:158)
! at org.jdbi.v3.core.ConstantHandleSupplier.invokeInContext(ConstantHandleSupplier.java:52)
! at org.jdbi.v3.sqlobject.SqlObjectFactory.lambda$createInvocationHandler$14(SqlObjectFactory.java:157)
! at com.sun.proxy.$Proxy150.find(Unknown Source)
```

The workaround for this seems to be that I have to specify the foreign key to the Parent class in the result set somewhere, using the prefix.  So, this query works where the other one failed:

```SELECT ID, COMMENT, PARENT as p_ID FROM CHILD WHERE ID = :id```

But this would require including all nested beans in the result set somewhere, even for larger, more complex business models, for example when a one-to-one or many-to-one relationship is defined at deeper than 1 level (e.g. child -> parent -> grandparent -> great-grandparent, etc.).  At that point, every query against Child would have to JOIN the other tables or else return dummy values as placeholders.

### PR DESCRIPTION
**Pull request by strmer15**
Fixes #1212 

The ConstructorMapper, FieldMapper, and BeanMapper now all provide extra context during the specialize0() method to know whether or not a nested call is being made, and if so, to not throw an exception if no columns were mapped but rather create a row mapper that returns null, as per @qualidafial 's suggestion.

The ConstructorMapper was tricky because it throws an exception immediately upon finding an unmatched parameter; I moved the exception handling to after all the parameter mapping so that it could be ignored by nested calls if all the parameters of the nested object were unmapped.  That way, it would catch when a nested object is partially retrieved, as well as when a nested object has other nested objects.

I've also added in test cases to the the mapper tests that hopefully illustrate and cover the fix.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
@@ -21,7 +21,6 @@ import java.lang.reflect.Type;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.OptionalInt;
@@ -157,7 +156,7 @@ public class ConstructorMapper<T> implements RowMapper<T> {
             ctx.getConfig(ReflectionMappers.class).getColumnNameMatchers();
         final List<String> unmatchedColumns = new ArrayList<>(columnNames);
 
-        RowMapper<T> mapper = specialize0(rs, ctx, columnNames, columnNameMatchers, unmatchedColumns, false);
+        RowMapper<T> mapper = specialize0(rs, ctx, columnNames, columnNameMatchers, unmatchedColumns);
 
         if (ctx.getConfig(ReflectionMappers.class).isStrictMatching()
             && unmatchedColumns.stream().anyMatch(col -> col.startsWith(prefix))) {
@@ -175,16 +174,16 @@ public class ConstructorMapper<T> implements RowMapper<T> {
                                      StatementContext ctx,
                                      List<String> columnNames,
                                      List<ColumnNameMatcher> columnNameMatchers,
-                                     List<String> unmatchedColumns,
-                                     boolean nested) throws SQLException {
+                                     List<String> unmatchedColumns) throws SQLException {
 
-        final int count = constructor.getParameterCount();
+        final int parameterCount = constructor.getParameterCount();
         final Parameter[] parameters = constructor.getParameters();
 
-        final List<String> unmatchedParams = new ArrayList<>();
+        int matchedParameterCount = 0;
+        final List<String> unmatchedParameters = new ArrayList<>();
 
-        final List<RowMapper<?>> mappers = new ArrayList<>(count);
-        for (int i = 0; i < count; i++) {
+        final List<RowMapper<?>> mappers = new ArrayList<>(parameterCount);
+        for (int i = 0; i < parameterCount; i++) {
             final Parameter parameter = parameters[i];
 
             Nested anno = parameter.getAnnotation(Nested.class);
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java [hunk 1]
```diff
@@ -171,52 +175,76 @@ public T map(ResultSet rs, StatementContext ctx) throws SQLException {
...
+                mappers.add(mapper);
+            }
+        }
+
+        if (!unmatchedParams.isEmpty()) {
+
+            // nested objects missing all of their 'normal' params should be set to null
+            if (nested && unmatchedParams.size() == Arrays.stream(parameters)
+                .filter(p -> !p.isAnnotationPresent(Nested.class))
+                .count()) {
```
qualidafial: This logic is a bit difficult to reason through, which makes me nervous.

From what I can see, there are three scenarios for nested mappers:

* No constructor parameters matched with any columns - map to null
* All constructor parameters matched to columns - map to object
  * Exception: `@Nested` parameters may map to null and still satisfy this scenario.
* Some but not all parameters matched to columns - throw exception

So here's the algorithm as I see it:
* Introduce a local variable `int matchedParameters = 0;`
* Iterate constructor parameters looking for column matches
  * When a `@Nested` parameter is encountered
    * Clone `columnNames` and pass that to the `specialize0()` call as the `unmatchedColumns` parameter.
    * Upon return from `specialize0()`:
      * Call `unmatchedColumns.retainAll(columnNamesClone)`
      * Increment `matchedParameters` if `columnNamesClone` is not equal to `columnNames`. A simple `size()` lookup is probably sufficient.
  * For non-nested parameters
    * If a column is matched:
      * Increment `matchedParameters`
    * Otherwise
      * Add parameter name to `unmatchedParameters` list.
* If `matchedParameters` is equal to the number of parameters, return the regular mapper.
* If `matchedParameters` is zero, return a null mapper.
* Otherwise, throw an exception

Hmm. I feel like there's a hole in the logic somewhere but I can't put my finger on it.

strmer15: Thanks for the feedback!  The main problem is that, in the @Nested part, if we only increment matchedParameters when ```columnNamesClone.size() != columnNames.size()```, we miss the case where all the nested constructor params were all unmatched, which would mean ```columnNamesClone.size() == columnNames.size()```.  Basically, when I implemented this logic, the ```nestedParametersNotReturned``` and ```nestedPrefixParametersNotReturned``` test cases failed.

What I did to fix it was to always increment matchedParameters counter upon getting the row mapper for the @Nested parameter.  The way I justified it was that, in the recursive call, if the matched parameter count was 0 or all matched, we'd correctly return a row mapper.  If they didn't match correctly, we'd just throw an exception anyways and halt execution.  So, since the recursive call itself is validating the matches, returning from it should always increment the match.  The fact that it returned meant it matched.

strmer15:
@qualidafial Is there anything else you want me to do on this?  Is it possible to merge in for the next release?  Thanks!

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
qualidafial: I'm investigating an alternative approach to this. Will post a link when it's ready to look at.

In short: add a new method `ColumnNameMatcher.columnNameStartsWith(String columnName, String prefix)`, with a default implementation that returns true. The method would be used by reflective mappers, to skip a nested mapping if no columns begin with the prefix in question.

Also: Change `specialize0()` to return an `Optional<RowMapper<T>>` to allow the calling method to decide what to do if no row mapper is available. `specialize()` can throw an exception, whereas `specialize0()` (on nested mappers) can choose to ignore it.

qualidafial: For now I think we should exclude ConstructorMapper from this rework. The sanity checking logic for that one is just crazy hard to reason about.

qualidafial:
@strmer15 thanks for doing a spike on this. I ported several of your unit tests over to the other PR. Just waiting to hear back from other project members on whether they agree to the new behavior.

Issue #1224, opened 2018-08-27
Exceptions thrown by an extension are wrapped in RunTimeException


Porter:
For example (assuming jdbi has the SqlObjectPlugin installed):

```
    interface myDao {
        default void blowUp() throws SQLException { throw new SQLException("boom"); }

        @SqlQuery("SELECT 1")
        int thisMakesMyDaoASqlObject();
    }

    // Fails
    @Test(expected = SQLException.class)
    public void blowUp() throws Exception {
        Jdbi jdbi = getJdbi();
        jdbi.onDemand(myDao.class).blowUp();
    }

    // Passes, but is tedious
    @Test
    public void blowUp2() throws Exception {
        Jdbi jdbi = getJdbi();

        try {
            jdbi.onDemand(myDao.class).blowUp();
            fail("Should have thrown");
        }
        catch (RuntimeException e) {
            assertEquals(SQLException.class, e.getCause().getClass());
        }
    }
```

For unit testing, I would prefer it if the original exception is thrown. Is the current handling of exceptions intentional?

Thank You,
Porter

PR #1225, opened 2018-08-27
Rethrow all Exceptions instead of wrapping them

Porter:
Addresses #1224 

Rethrows exceptions thrown in by a DAO instead of wrapping it in a RunTimeException.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
+++ b/core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
@@ -70,13 +70,13 @@ class OnDemandExtensions {
     }
 
     @SuppressWarnings("PMD.AvoidRethrowingException")
-    private static Object invoke(Object target, Method method, Object[] args) {
+    private static Object invoke(Object target, Method method, Object[] args) throws Exception {
         try {
             if (Proxy.isProxyClass(target.getClass())) {
                 return Proxy.getInvocationHandler(target).invoke(target, method, args);
             }
             return MethodHandles.lookup().unreflect(method).bindTo(target).invokeWithArguments(args);
-        } catch (RuntimeException | Error e) {
+        } catch (Exception | Error e) {
             throw e;
         } catch (Throwable t) {
             throw new RuntimeException(t);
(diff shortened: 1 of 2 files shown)


stevenschlansker:
I like the change, however I am concerned that this might be an API compatibility change.  If someone checked the exact type of thrown exception, this will change behavior.

However, the test case is pretty compelling.  The exception transformation is surprising and non-obvious.  I lean toward making this (slightly-)breaking change.  @jdbi/contributors any other opinions?

core/src/main/java/org/jdbi/v3/core/OnDemandExtensions.java
```diff
@@ -70,13 +70,13 @@ private OnDemandExtensions() {
...
     @SuppressWarnings("PMD.AvoidRethrowingException")
-    private static Object invoke(Object target, Method method, Object[] args) {
+    private static Object invoke(Object target, Method method, Object[] args) throws Exception {
         try {
             if (Proxy.isProxyClass(target.getClass())) {
                 return Proxy.getInvocationHandler(target).invoke(target, method, args);
             }
             return MethodHandles.lookup().unreflect(method).bindTo(target).invokeWithArguments(args);
-        } catch (RuntimeException | Error e) {
+        } catch (Exception | Error e) {
```
qualidafial: Since the method now throws vanilla Exception, is there any downside to just removing the try/catch and letting the error bubble up as is?

Porter: The one downside I can think of is that it makes this change slightly more breaking, as now we stop wrapping all `Throwable`s in a `RunTimeException`. 

Although, in practice, I've never seen any `Throwable` that isn't an `Exception` or `Error`, and allowing errors to bubble up provides maximal exception transparency. 

I think it's best to rip the bandaid off all at once, so I'll make this change

stevenschlansker: Good catch.  Fewer cases to think of is better, let's remove it.

Porter: My bad. It actually turns out that any `Throwable` that isn't `Exception` or `Error` needs to be wrapped in a `RunTimeException`. If not, then we need to say `throws Throwable` in the method signature, which in turn creates the need to change `throws Exception` to `throws Throwable` in many other classes, some of which we can't change such as `java.util.concurrent.Callable#call`. Removing it won't be worth it.

stevenschlansker: Ah, of course.  We should have used `X extends Throwable` instead of `X extends Exception` for our exception bounds, but that error can't be fixed until jdbi 4.

Issue #1231, opened 2018-09-07
JdbiRule should optionally clean up during after()


huhlig:
Unfortunately due to using a couple of PostgreSQL extensions not included with the embedded version I realized I needed to add a cleanup phase to the migration so I tried reworking this and adding Flyway Schema support.

PR #1232, opened 2018-09-07
Adding external PostgreSQL support and improving migration support.

huhlig:
In Support of https://github.com/jdbi/jdbi/issues/1231

### DIFF
--- a/testing/src/main/java/org/jdbi/v3/testing/JdbiRule.java
+++ b/testing/src/main/java/org/jdbi/v3/testing/JdbiRule.java
@@ -14,85 +14,82 @@
 package org.jdbi.v3.testing;
 
 import org.flywaydb.core.Flyway;
-import org.h2.jdbcx.JdbcConnectionPool;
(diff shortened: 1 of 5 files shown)


testing/src/main/java/org/jdbi/v3/testing/JdbiRule.java
```diff
@@ -14,85 +14,82 @@
...
+        return new ExternalPostgresJdbiRule(hostname, port, username, password, database);
     }
 
     /**
-     * Run database migration scripts from {@code db/migration} on the classpath, using Flyway.
-     * @return this
+     * Create a JdbiRule with an in-memory H2 database instance. Your project
+     * must depend on the {@code h2} database artifact.
      */
-    public JdbiRule migrateWithFlyway() {
```
stevenschlansker: I ended up deprecating these methods rather than removing them.

huhlig: Changes seem reasonable to me.

PR #1233, opened 2018-09-07
jdk11, basepom 27

stevenschlansker:


sqlobject/src/test/java/org/jdbi/v3/sqlobject/locator/TestSqlLocator.java
```diff
@@ -32,7 +32,7 @@
     public PgDatabaseRule dbRule = new PgDatabaseRule().withPlugin(new SqlObjectPlugin());
 
     @Test
-    public void testLocateConfigDriven() {
+    public void testLocateConfigDriven() throws Exception {
```
stevenschlansker: type inference changed somewhere along the way, and this no longer compiles in 11??

spring4/pom.xml
```diff
@@ -87,19 +87,4 @@
             </plugin>
         </plugins>
     </build>
-
-    <profiles>
-        <profile>
-            <id>jdk9</id>
-            <activation>
-                <jdk>[1.9,</jdk>
```
stevenschlansker: @qualidafial is this still needed?  It seems to break the jdk11 build (unused dependency `annotation-api`) and removing it doesn't seem to break the jdk9 travis build...

qualidafial: I was kinda throwing stuff at the wall to get the jdk9 build to work. If removing it hurt the build, go for it.

.travis.yml
```diff
@@ -28,6 +28,11 @@ matrix:
       script:
         - src/build/ci-test.sh
       jdk: oraclejdk9
+    - env:
+        - TESTS=Y
```
qualidafial: Couldn't these environment variables be moved into the respective `ci-*.sh` scripts?

kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/Extensions.kt
```diff
@@ -1,4 +1,4 @@
-/*
+/**
```
qualidafial: Any particular reason to mark these up as javadoc instead of comments?

pom.xml
```diff
@@ -428,15 +426,15 @@
             <dependency>
                 <groupId>com.github.spotbugs</groupId>
                 <artifactId>spotbugs-annotations</artifactId>
-                <version>${spotbugs.version}</version>
+                <version>${dep.plugin.spotbugs.version}</version>
```
qualidafial: Looking at Maven Central, it looks like the plugin and annotations jars are only _sort of_ versioned in lockstep.

- [spotbugs-maven-plugin](https://search.maven.org/search?q=g:com.github.spotbugs%20AND%20a:spotbugs-maven-plugin&core=gav)
- [spotbugs-annotations](https://search.maven.org/search?q=g:com.github.spotbugs%20AND%20a:spotbugs-annotations&core=gav)

Not all plugin versions exist for the annotations jar, and vice versa. Additionally, the annotations jar tends to be released days before the plugin--sometimes weeks.

In this instance it's fine, but it may bite us in future basepom upgrades--something to watch out for.

spring4/pom.xml
qualidafial: Also I think this range should be `[1.9,)` (with the trailing `)` paren)--I seem to remember a PR from Dropwizard about this causing a problem.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/locator/TestSqlLocator.java
qualidafial: You're saying JDK 11 breaks type inference on our exception transparency stuff? Well.. shit

kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/Extensions.kt
stevenschlansker: The formatter plugin insists -- if it is not Javadoc style already, it will insert a duplicate license header in Javadoc style.  I'm not sure why this happens for particularly Kotlin source files, maybe we should dig deeper, but this change is necessary for now :/

sqlobject/src/test/java/org/jdbi/v3/sqlobject/locator/TestSqlLocator.java
stevenschlansker: I'm not sure exactly what the break is.  Only this test case required fixing up.
We could submit a question to the jdk mailing lists if we can extract a small reproduction case.  I'm not sure what the impact will be yet, I only just got the build to go green...

spring4/pom.xml
stevenschlansker: In that case I'll just remove it.  We can always add it back in if we remember why it's needed.

.travis.yml
stevenschlansker: I was just copy and pasting what was already here ;)

stevenschlansker:
Ugh.  And of course, jdk11 builds stop working on Travis in the meantime.  I'll revisit this in a day or three.

kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/Extensions.kt
qualidafial: As long as these copyright notices don't show up in the docs, I'm fine with this change.

stevenschlansker:
Looks like Travis is happy again!

pom.xml
mikebell90: It does for 3.1.7, which irpnically is needed for Java 11. The core is out but not the maven plugin

PR #1238, opened 2018-09-13
Make CaseStrategy implement UnaryOperator<String>

stevenschlansker:
Also rename 'locale lower' which I think was confusing because it uses root locale (i.e. *not* locale sensitive)

followup to https://github.com/jdbi/jdbi/commit/e418324d7c1236fb1cc957897a0e5145e55ab007#commitcomment-30452308


core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -51,7 +51,7 @@ public MapMapper() {
      */
     // TODO deprecate when MapMappers.caseChange is out of beta
```
qualidafial: Ok, I could have sworn this wasn't deprecated yet. We should remove the release note that says this constructor was deprecated.

PR #1240, opened 2018-09-15
Reflection -> MethodHandles

TheRealMarnes:
First time working with methodhandles...

Included a little bit of oppurtunistic but very carefully selected refactoring. Tried to run a build locally but kept getting infrastructural errors, I think from postgres... Not exactly a new problem.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/internal/Throwables.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/Throwables.java
@@ -22,17 +22,21 @@ public class Throwables {
     }
 
     /**
-     * returns t directly if it is unchecked, otherwise wraps it in a RuntimeException
+     * unchecked throws t. Returns a bogus exception for compiler satisfaction: always throw the return value even though you'll never get it.
      */
-    public static RuntimeException uncheck(Throwable t) {
+    public static RuntimeException sneakyThrow(Throwable t) {
+        if (t instanceof Error) {
+            throw (Error) t;
+        }
+
         if (t instanceof RuntimeException) {
-            return (RuntimeException) t;
+            throw (RuntimeException) t;
         }
 
         if (t instanceof IOException) {
-            return new UncheckedIOException((IOException) t);
+            throw new UncheckedIOException((IOException) t);
         }
 
-        return new RuntimeException(t);
+        throw new RuntimeException(t);
     }
 }
--- a/core/src/main/java/org/jdbi/v3/core/internal/Throwables.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/Throwables.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.internal;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+
+public class Throwables {
+    private Throwables() {
+        throw new UnsupportedOperationException("utility class");
+    }
+
+    /**
+     * returns t directly if it is unchecked, otherwise wraps it in a RuntimeException
+     */
+    public static RuntimeException uncheck(Throwable t) {
+        if (t instanceof RuntimeException) {
+            return (RuntimeException) t;
+        }
+
+        if (t instanceof IOException) {
+            return new UncheckedIOException((IOException) t);
+        }
+
+        return new RuntimeException(t);
+    }
+}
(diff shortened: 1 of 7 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultMethodHandler.java
```diff
@@ -96,14 +97,11 @@ private static Method privateLookupIn() {
...
-    @SuppressWarnings("PMD.AvoidRethrowingException")
     public Object invoke(Object target, Object[] args, HandleSupplier handle) {
         try {
             return methodHandle.bindTo(target).invokeWithArguments(args);
-        } catch (RuntimeException | Error e) {
-            throw e;
-        } catch (Throwable throwable) {
-            throw new RuntimeException(throwable);
+        } catch (Throwable t) {
+            throw Throwables.uncheck(t);
```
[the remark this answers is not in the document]
TheRealMarnes: Then we'll need to add more catch statements to separate them, which kinda defeats the point of this method... I agree we don't wanna wrap Errors, but catching Error separately so any Throwable left will be an Exception, just adds more boilerplate and a pitfall to remember instead of simplifying things...

I've thought about this before and basically there is no good way to do it, except for the way Unchecked works in jool: instead of transforming exceptions, wrap exception-throwing runnables (and other functionalinterfaces) in a re-usable handler: `Bar bar = Unchecked.function(Foo::toBar).accept(foo).`

Copying Unchecked means stepping into legal argument territory though... apache 2.0 license and whatnot. I personally think a simple Throwable sieve function is so trivial it can't fall under copy protection any more than overriding tostring/hashcode/equals can, but I'm no lawyer...

TheRealMarnes:
How about this change to handle Error? It lets us keep using nice and simple uncheck calls, and doesn't wrap Error. It's just a tad odd from a functional pov, but it works.

core/src/main/java/org/jdbi/v3/core/internal/Throwables.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * throws t if it is an Error, returns t directly if it is unchecked, wraps t in a RuntimeException otherwise
+     */
+    public static RuntimeException uncheck(Throwable t) {
+        if (t instanceof Error) {
+            throw (Error) t;
+        }
+
+        if (t instanceof RuntimeException) {
+            return (RuntimeException) t;
```
stevenschlansker: why not `throw` here (and below) as we do above?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultMethodHandler.java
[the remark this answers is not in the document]
stevenschlansker: It looks like `jool` is Apache 2 same as `jdbi` -- so what licensing issue are you worried about?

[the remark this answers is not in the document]
qualidafial: There is no license conflict. Jdbi is Apache 2.0 license, and so is Jool, but that's beside the point.

It is generally understood in Java that Throwable has "only" two direct subtypes: Error (for unrecoverable problems like out of memory), and Exception for recoverable problems.

It is possible to extend Throwable directly but this is widely frowned upon.

I know there are a few classes in the wild that extend Throwable instead of Exception, but they are the exception (ha) rather than the rule. This is not an edge case I'm willing to accommodate if we have to ugly up our code to do it.

Side note, this is probably why the old Guava `Throwables.propagate()` method threw the exception for you--to avoid the `RuntimeException | Error` disjunct exception type problem.

core/src/main/java/org/jdbi/v3/core/internal/Throwables.java
qualidafial: Concur

stevenschlansker:
Let's get this change in 3.5.0 since we already have a couple of other reflection changes.

core/src/main/java/org/jdbi/v3/core/internal/Throwables.java
TheRealMarnes: I'd like the call site to always do the throw. We can throw it in here as well, but then the throw at the call site is entirely redundant.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/DefaultMethodHandler.java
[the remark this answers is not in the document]
TheRealMarnes: I don't know what you need to do beside copying the code to be permitted to copy it.

As for subtypes: the method should accept Throwable because some APIs declare it that way, like invocations of methodhandles.

core/src/main/java/org/jdbi/v3/core/internal/Throwables.java
qualidafial: That's probably necessary for flow control. It's how Guava's `Throwables.propagate()` worked--the propagate method always threw the exception but had a `RuntimeException` return type so the caller could "throw" a result that was never returned, to satisfy the compiler

TheRealMarnes:
I renamed the method to better describe its function...

image

core/src/main/java/org/jdbi/v3/core/internal/Throwables.java
stevenschlansker: [Guava](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/Throwables.java#L239-L242) methods `throw` unconditionally.  They declare a return of `RuntimeException` or the like but never actually return.

Absent any particular reason to change it, let's copy that style.  No need to do anything dramatic like take the entire class.  I'll make the change unless you're online to do it now.

stevenschlansker: We also need a release note.

stevenschlansker:
oh great, it looks good.

PR #1244, opened 2018-09-17
[wip refactor] reusable high-level classpath builder

TheRealMarnes:
This is a little thing I've had sitting on a branch for a while now. I remember getting annoyed that there are bits and pieces of classpath traversing code everywhere for stuff like loading sql files, and the logic to build the path is reinvented every time. This utility should fit most current and future needs and is holding up well so far in unit tests.

Does it seem an interesting thing to you guys and should I migrate as much as possible ad-hoc pathing to this utility, or nah? I feel it could add some consistency, clarity, and ease to parts of the code, and replace a lot of downcasting-meaningful-objects-to-strings, like in `UseClasspathSqlLocatorImpl` where a nice reflection Method object is turned into a string for manipulation.

I included a touch of portability in the parts involving Paths so that stuff like windows' special path separator won't be a problem. It's basically platform-agnostic, both in its public and private workings.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/locator/ClasspathSqlLocator.java
+++ b/core/src/main/java/org/jdbi/v3/core/locator/ClasspathSqlLocator.java
@@ -21,12 +21,11 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
-
-import org.antlr.runtime.ANTLRInputStream;
-import org.jdbi.v3.core.internal.SqlScriptParser;
-
 import net.jodah.expiringmap.ExpirationPolicy;
 import net.jodah.expiringmap.ExpiringMap;
+import org.antlr.runtime.ANTLRInputStream;
+import org.jdbi.v3.core.internal.SqlScriptParser;
+import org.jdbi.v3.core.locator.internal.ClasspathBuilder;
 
 /**
  * Locates SQL in {@code .sql} files on the classpath.  Given a class and
@@ -35,9 +34,6 @@ import net.jodah.expiringmap.ExpiringMap;
  * The contents are then parsed, cached, and returned for use by a statement.
  */
 public final class ClasspathSqlLocator {
-    private static final char PACKAGE_DELIMITER = '.';
-    private static final char PATH_DELIMITER = '/';
-
     private static final SqlScriptParser SQL_SCRIPT_PARSER = new SqlScriptParser((t, sb) -> sb.append(t.getText()));
 
     @SuppressWarnings("unchecked")
@@ -50,7 +46,7 @@ public final class ClasspathSqlLocator {
             })
             .build();
 
-    private static final String SQL_EXTENSION = ".sql";
+    private static final String SQL_EXTENSION = "sql";
 
     private ClasspathSqlLocator() {}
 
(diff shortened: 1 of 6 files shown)


TheRealMarnes:
On a related note, is there a proper name for this best practice I believe in where you avoid putting all kinds of data into errorprone datatypes like strings wherever possible? Like in spring, instead of injecting external service urls as String, I'd inject them as URLs or URIs, because that's simply what they are. We don't put numbers in Strings either when we have numeric types for it...

PascalSchumacher:
>On a related note, is there a proper name for this best practice I believe in where you avoid errorpronely putting all kinds of data into strings wherever possible?

Not sure if there is a name for this practice but I think `primitive obsession` is the name of the anti-pattern for excessive use of primitive types instead of objects.

TheRealMarnes:
Well I expect to replace a few more existing bits of code with this (difficult to find though), so uses will go up. As for things only used in tests... I wrote them because I expect they will prove useful in general use. We don't wanna write anything expectably needed before the need arises?

TheRealMarnes:
On that note, do you know by heart any other places where classpath work could be delegated to this tool? I don't know the codebase as well as you do.

qualidafial:
Only classes named `*SqlLocator` come to mind. We have Classpath, StringTemplate, and Freemarker SQL locators.

I see the argument for adding convenience API, but considering that `ClasspathBuilder` is in an internal package, any method declared that is not used by the project is just taking up space. It's a different scenario if you're talking about a public API, but I don't think we're ready to cross that bridge yet.

TheRealMarnes:
I've changed as you asked, and slightly optimized(?) the inner workings, avoiding the use of Optional and String reconstructions. The only side-effect is the built classpaths now start with `/`, which is technically different behavior, but I believe it's a good change and it doesn't impact us anyway because we never resolved relative to a class in the first place.

TheRealMarnes:
Something that just came to mind is that it would be very simple to track the total string length as it is built. so the stringbuilder can be scaled to the appropriate size and it will never need to reallocate its array. Is this a worthwhile thing to do? It would add some microclutter for a microgain.

TheRealMarnes:
Hmm, seems there's still a sensitivity to leading slashes... Oh how I hate working with (class)paths in java...

TheRealMarnes:
Alright, all done. I dunno if you agree with the more-performant-less-semantic string handling I've changed it to? It's the classic micro performance gains vs micro maintainability increase trade-off.

qualidafial:
Did you perf test any uses cases before and after your change, to confirm they made a measurable improvement?

Perf profiling is something we've talked about in the past, and that needs to be done. If that's a project that sounds interesting to you, I'd love to see it worked on. The solution could use JMeter or any other measurement tool you like, as long as it works on TravisCI.

TheRealMarnes:
The optimized way is about twice as fast as the natural-language way, but we're talking about speeds in the order of 1–3 microseconds for a typical use case. As for memory usage, there will be some more String garbage in memory, but the method isn't long-running and doesn't create stupidly big numbers of them, so I can't see it becoming a problem. I can't see a situation where thousands of classpaths need to be built at the same time either.

I think it's an excessive optimization, but on the other hand the optimized code isn't difficult to understand or anything, so I'd say let's just leave it in optimized form.

TheRealMarnes:
Aye.

Issue #1211, opened 2018-08-10
JDBI broken with Spring managed transaction


mdeinum:
JDBI doesn't work (anymore?) with Spring with a Spring managed transaction and thus connection. I noticed this while preparing a sample with JDBI and Spring Boot 2.0.  

```
@Override
@Transactional
public Optional<Order> findById(String id) {
    return jdbi.withHandle(handle ->
           handle.createQuery("SELECT id, amount FROM orders WHERE id=:id")
                   .bind("id", id).mapTo(Order.class).findFirst());
}
```

This will eventually fail with an 

```
Caused by: org.springframework.transaction.TransactionSystemException: Could not commit JDBC transaction; nested exception is java.sql.SQLException: Connection is closed
	at org.springframework.jdbc.datasource.DataSourceTransactionManager.doCommit(DataSourceTransactionManager.java:332) ~[spring-jdbc-5.0.8.RELEASE.jar:5.0.8.RELEASE]
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.processCommit(AbstractPlatformTransactionManager.java:746) ~[spring-tx-5.0.8.RELEASE.jar:5.0.8.RELEASE]
	at org.springframework.transaction.support.AbstractPlatformTransactionManager.commit(AbstractPlatformTransactionManager.java:714) ~[spring-tx-5.0.8.RELEASE.jar:5.0.8.RELEASE]
	at org.springframework.transaction.interceptor.TransactionAspectSupport.commitTransactionAfterReturning(TransactionAspectSupport.java:532) ~[spring-tx-5.0.8.RELEASE.jar:5.0.8.RELEASE]
	at org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:304) ~[spring-tx-5.0.8.RELEASE.jar:5.0.8.RELEASE]
	at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:98) ~[spring-tx-5.0.8.RELEASE.jar:5.0.8.RELEASE]
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:185) ~[spring-aop-5.0.8.RELEASE.jar:5.0.8.RELEASE]
[...]

PR #1252, opened 2018-09-30
Spring plugin final

alwins0n:
Hey guys, I decided to create a new pull for the spring plugin rewrite, since #989 is already too outdated and confusing as of now, imho.

I addressed the issues mentioned and lastly suggested by me in #989 and this is my current state.

What has been done:
- complete rewrite of the docs
- moved/renamed the module (makes sense since its works with several major spring versions, current included)
- minor fixes according to #989 
- marked new APIs as Beta. note: so far only the `JdbiTemplate`/Operations class introduces a "real" new API which could break. I saw no point in marking its public methods as Beta since they excatly follow `Jdbi` itself. not sure...
- couple of minor cleanups accoring to the build tools involved.

I wanted to mention that the plugin is more or less in this version in production at my work for 6 months. So far I did not encounter any problems.

What has to be done:
I think I might need a little help to implement the `withExtension` method on `JdbiTemplate`. I thought it would be trivial, but never having used Exentions it got confusing quickly. 

How would a "lazy loaded handle" work within a given transaciton system? Also some of the classes/methods needed for me to at least copy the code are package private etc...

Could you help me out here?


alwins0n:
In my endless creativity I realized that potential there is a very powerful but dangerous integration possible without introducing new APIs! I really want some opinions here as it has potential for confusion

1. Set up a Prototype Proxy for Handle. I.e. every time a `Handle` bean is accessed it returns the transactional handle

```java
@Bean
@Scope(value = "prototype", proxyMode = ScopedProxyMode.TARGET_CLASS)
public Handle handle(Jdbi jdbi) {
    return JdbiUtil.getTransactionalHandle(jdbi);
}
```

where `getTransactionalHandle` is a method that fails if there is no transaction.

2. Inject your Hande (which is a proxied object) into your repository or dao

```java
@Inject
private Handle transactionalHandle;
```

3. Use it. every time a method is called on the handle the proxy delegates it to the trnasactionally bound handle.

**DANGER:**
this fails if you are not in a transaction! seems logical but. if you inject an object into your another one, you might assume that you are able to manipulate it at will. E.g. install a RowMapper in a `@PostConstruct` method.

This may fail unless the whole class is marked as `@Transactional`.

**My Question**

Is this the way to go despite potential drawbacks? It avoids new APIs. And I could probably code together a `FactoryBean` for handles to have the ease of use similar to `JdbiFactoryBean` when setting up the whole thing. (altough I believe the setup of 1. is not that painful)

stevenschlansker:
I don't think I understand why you would want to inject a `Handle` that is potentially in the middle of a unrelated transaction.  For example wouldn't this mean that if your local method fails, it may cause rollback of the previous collaborator that used the Handle?  Who is then responsible for committing your results?

Maybe if you can show a use case for why this is better than the normal `handle.inTransaction(() -> ...)` dance it'll be clearer to me.

qualidafial:
So I want to say up front that we really appreciate your contributions, and I hate giving feedback after the fact on what was clearly a big project. It's important to me that contributors feel like their effort was well spent.

With that said, I have some feedback to give, and I hope you'll take it in stride and be able to incorporate it into this PR, because we really do need better Spring support.

Here goes:

- I _hate_ breaking backward compatibility on our users. Changing artifact IDs and package names qualifies.
- I don't think conflating Spring 4 and 5 into a single `jdbi3-spring` integration artifact is justified, or safe. Spring follows semantic versioning. That means that some things are incompatible between versions, and I don't want to get caught making assumption that "it should be fine" only to find out later that it isn't.
- You're managing the Spring version to Spring 5. Any integration artifact provided by us should advertise this dependency. This is why we chose e.g. `jdbi3-spring4` as the artifact name. Exceptions:
  - `jdbi3-guava` - Because Guava maintainers have stated they won't break backward compatibility any more.
  - `jdbi3-postgres` - Postgres JDBC drivers are not versioned in lockstep with the database version, and they are backward compatible with _very_ old versions of the Postgres database.
- I don't like the wrapper approach to integrating Jdbi with Spring: JdbiTemplate, JdbiOperations, JdbiUtil, HandleHolder. I don't like the way it forces Jdbi users in Spring to always go through those checkpoints rather than the original API. I feel like a better alternative would be to add some hooks into Jdbi, so that Spring can configure Jdbi to play nice with it.
  - If Jdbi enforces restrictions that are invalid in Spring, let's add a config property that can disable those restrictions
  - If Spring needs to be consulted at certain parts of the lifecycle e.g. of a connection or transaction, let's add callback APIs that Spring can use to register listeners.

I don't use Spring in $WORK, so I'm open to the possibility that this is just the way Spring wants to work--if so, I need you to educate me on the what and why.

alwins0n:
don't worry, you are not scaring me away if critique is justified. I myself am not confident about the plugin so far myself. but to address your issues (even though I feel like this PR might be getting out of hand again):

 1. @stevenschlansker 
It does not matter what your reasons are for injecting transactional handles anywhere. The way spring (and java ee) works is. if some method or class is marked as transactional - every db call that is made within that method runs in a transaction. THAT simple.

(simplified) Example on a realworld project: 
Given a `CustomerRepository` and an `EventRepository`, there is a `Activator` service level class that is annotated with `@Transactional`. Its method `activateCustomer` is therefore transactional. Now - on activation of customer X, I want to mark the customer as activated and send a mail (via events). The save of "active=true" and the event should be in one single transaction where the repository MUST NOT know about each other. Mail sending is done later with a batch job which reacts on "CustomerActivatedEvent"s saved in db.

For Reference:
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/annotation/Transactional.html 

the point here is - the `Handle` injected into each repository is the handle participating in that ONE transaction (if configured as propagating). so each Repo uses that handle as if nothing happend, unaware that there is a transaction spanning multiple repositories.

2. @qualidafial 
If you read the docs like the one linked above you will see alot of "@since ..." Spring JDBC is pretty old and not much changes happened over course of its major releases. Heck, I even tested the plugin against spring 3.0 just "for science" (it worked) and it probably would work against spring 2.0 (which is 10 years old or something like that)

IMHO jdbi3-spring5 is AS WRONG as jdbi3-spring4, 3 etc... the spring jdbc api did not change much and I don't foresee it happening in the future (what would be the reason - maybe reactive?).

Secondly, I want to re-state (the most important?) point about spring integration: **you use AOP for transaction management (i.e. `@Transactional` or via config) and expect it to JUST WORK(tm)**.

If jdbi is incapable or unwilling to do that I see a massive problem for its acceptance!
Why do you think the ppl in several issues here recommended to use springs `TransactionAwareDataSourceProxy` as DS for jdbi when integrating the two? **because then spring is managing transactions application wide**, as expected by every spring user ever. (my claim - prove me otherwise)

I am trying to get someone from the spring team to weigh in. I can see that you have difficulties following my reasoning, which is totally understandable if you don't build applications with spring.

Lastly: about the wrapper approach. I am not too happy with that either, but I need to clarify:
- `HandleHolder` is not public API. So not a wrapper that the user deals with. used internally
- `JdbiUtil` is just that, a Util class. You may use it or not.
- The critique about `JdbiTemplate` is reasonable. That's why I suggested the direct injection of `Handle`s. with this approach there is no wrapper (only excpetions if used incorrectly)

alwins0n:
Lol, I just realized that #1211 is excatly such a case. Ppl annotate methods with `@Transactional` and expect everything to automagically work.

We are integrating a low-magic library into a high magic-framework here. Think about the expectations about ease of use by the framework users. Lets be honest here: Jdbi is a plugin to spring, not the other way round.

Maybe this is not the right place for this PR...

qualidafial:
@alwins0n I think your last few comments clarified a lot.

I think a productive next step is to take inventory of all Jdbi default behaviors that don't play nice with Spring. For each of those problems, we need to decide whether the best solution is a configuration option or a lifecycle hook.

Example: we added `getConfig(Handles.class).setForceEndTransactions(false)` config property to disable the out-of-the-box check on `Handle.close()`.

stevenschlansker:
Thanks for explaining further, and your continued pursuit!  To your point about maybe this not being the right place for the PR -- you might be right.  The `jdbi` repository is for the core library itself, and plugins thereof.  Generally Spring likes to *wrap other libraries*, and it sounds like that's what you're doing here.  And you notice all of those projects tend to live under the Spring umbrella.  (witness `spring-jdbc` `hibernate`, etc)

I agree with @qualidafial that we can make any reasonable behavior changes needed (via configs if appropriate).  Do consider whether this really is really the start of a `spring-jdbi` module or standalone project rather than a plugin that should live here.  If you consider this but then still think `jdbi` is a good home for your work we can continue pushing forward.

alwins0n:
Thanks for the discussion, I think we are getting somewhere.

First of all I tried to reach out to spring guys. Turns out, it's not that easy but we will see. Anyhow, your comments got me thinking.

Since there already is a jdbi3-spring4 plugin and since you do not want to break backwards compatibility - maybe this plugin should stay here, but your comments regarding lifecycle - maybe thats the key!

Here is my current idea:
- `Jdbi` gets a hook `setOpenHandleStrategy(Function<Jdbi, Handle> strategy)`
- `Handle` gets a hook `setCloseStrategy(Consumer<Handle> strategy)`

now, the plugin could customize `Jdbi` and `Handle` via these methods and do (along those lines) the following
`Jdbi.setOpenHandleStrategy(JdbiUtil::getHandle)`
`Handle.setCloseStrategy(JdbiUtil::closeIfNeeded)`

this allows to jdbi to be used as if there were no spring - but in fact the handle is always obtained via the Managed transactions of spring. If there is no transaction there is the handle obtained as always (via JdbiUtil) - so a decent fallback!

I did not think this through the end but this sounds very promising to me. What do you think?

stevenschlansker:
That's a much cleaner API integration!  But how about making the strategies part of a `JdbiConfig` class instead of adding additional methods?

````java
jdbi.configure(HandleStrategy.class, s -> {
  s.setOpener(Jdbi::getHandle); // the defaults
  s.setCloser(Handle::close);
});
```

alwins0n:
Just one note to implementation, maybe you can hint me to a nice solution:

As far as I can judge, it won't be able to implement without introducing new API.

`Jdbi#open()` would either use the configured strategy or delegate to the current implementation.
`JdbiUtil#getHandle()` is a strategy that returns the transactional handle OR delegate to the current implementation.

so the latter must be public.

alternatively it probably would be possible to give the strategy a signatur of `Function<Jdbi, Optional<Handle>>` in which case we could delegate to the current `open` implementation when the stragegy return is empty.

another alternative is to have the `OpenStrategy` be a field on Jdbi, which, lich the Transaction Management can be set/get/configured. There would be a default strategy in place, which I could `get`, wrap, and `set` again in the customizeJdbi plugin method

What do you think`?

qualidafial:
@alwins0n Have you had a chance to look at the new `ConnectionFactory.closeConnection(Connection)` method?

ssljivic:
Hi @alwins0n what is the status of this feature? It would be really good to have Spring Tx support in JDBI.

alwins0n:
I wanna be honest here. After working on the integration mentioned on my comment before, I stumbled on some difficulties which would require Jdbi to change its public API a little bit to make it work smoothly.

I figured that the changes probably will result in discussion - which might be not worth it, since springs `TransactionAwareDataSourceProxy` I started to use in my day-job works like a charm for my use cases. And probably for most other spring users.  It should be documented properly, but on the other hand - this should not be JDBIs responsibility.

In the end I gave up development on this, because the transaction plugin would maybe help in some edge cases, but I am even unable (due to time constraints) to prove that.

I can offer to update the code one more time, and let the community decide wether its worth continuing... as of now, I have doubts about that.

stevenschlansker:
Thanks for letting us know.  If you do figure out a good way forward here, please let us know, and we can certainly reopen this or have it live on in a new PR.  For now I'll close it since it sounds like there is not a clear path forward.  Maybe hosting the code in a new repository so that people can test it and prove it out outside of Jdbi's release cycle, and we can talk about integrating it later once there's a bit of a track record.

Thanks for contributing to Jdbi!

ssljivic:
@alwins0n is there a chance that you can share how you have configured the Spring `TransactionAwareDataSourceProxy` ?

alwins0n:
Dead simple. pass Jdbi an instance of the proxy wich you just instatiate in the config.

`Jdbi.create(new TransactionAwareDataSourceProxy(applicationDs))`

I just wonder @stevenschlansker , why you would close this PR but not #989 ? I think this one is a lot more relevant (at least more than the other)

stevenschlansker:
Maybe I read too much into, "In the end I gave up development on this": I thought this mean this PR was unlikely to move forward.  I'm happy to leave it open for now though for further discussion.

alwins0n:
you did not! I actually was suggesting you'd close #989 too :)

PR #1256, opened 2018-10-05
Fix code snippet that extends LinkedHashMapRowReducer instead of implements

juggernaut:
`LinkedHashMapRowReducer` is an interface


qualidafial:
Good catch. That was a leftover from early on in the row reducer development when that interface was an abstract class still.

PR #1257, opened 2018-10-05
Add collector factory for EnumSet

lyind:
* Similar use case as https://github.com/jdbi/jdbi/pull/523
 * Relies on default enum mapping facilities
 * Includes unit test

The PR mentioned above and the old branch "enum_sets" were not merged because of direct ties with PostgreSQL "varbit" SQL type.

This PR avoids these pitfalls.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/collector/EnumSetCollectorFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/collector/EnumSetCollectorFactory.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.collector;
+
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.EnumSet;
+import java.util.Optional;
+import java.util.stream.Collector;
+
+import static org.jdbi.v3.core.generic.GenericTypes.findGenericParameter;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/jdbi/v3/core/collector/EnumSetCollectorFactory.java
```diff
@@ -0,0 +1,66 @@
...
+import static org.jdbi.v3.core.generic.GenericTypes.findGenericParameter;
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
+class EnumSetCollectorFactory implements CollectorFactory {
+
+    @Override
+    public boolean accepts(Type containerType) {
+        // elements of EnumSet must always be enums, no need to check
+        return EnumSet.class.isAssignableFrom(getErasedType(containerType))
+            && containerType instanceof ParameterizedType;
```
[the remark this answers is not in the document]
lyind: Since EnumSet is declared as `abstract class EnumSet<E extends Enum<E>>` the compiler should ensure the element type is `? extends Enum`.

core/src/main/java/org/jdbi/v3/core/collector/EnumSetCollectorFactory.java
```diff
@@ -0,0 +1,63 @@
...
+                () -> EnumSet.noneOf(componentType),
+                EnumSet::add,
+                (left, right) -> {
+                    if (left.size() < right.size()) {
+                        right.addAll(left);
+                        return right;
+                    } else {
+                        left.addAll(right);
+                        return left;
+                    }
```
qualidafial: This optimization is redundant.

The JDK implementations of `EnumSet` already override `addAll(Collection)` to check whether the received set is an `EnumSet`, and do a bitwise OR of their bit fields to quickly combine sets.

- http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/7d1d5f4d019a/src/share/classes/java/util/EnumSet.java
- http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/7d1d5f4d019a/src/share/classes/java/util/RegularEnumSet.java
- http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/7d1d5f4d019a/src/share/classes/java/util/JumboEnumSet.java

lyind: Thank you! Done.

PR #1259, opened 2018-10-07
Fix misworded comment in Vavr Option example

juggernaut:
When name is bound to Option.none() in the query `select * from users
where :name is null or name = :name`, the first clause of the `WHERE`
evaluates to `null is null` which is true, so it matches _any_ row, not
necessarily a row where the name `IS NULL`.

Not sure if this is what you were going for, though. If the intent was
to illustrate an example with the original comment's meaning, the query
should probably be changed to something like `select * from users where
name IS NOT DISTINCT FROM :name`

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -4143,7 +4143,7 @@ Here are some usage examples of the features listed above:
 String query = "select * from users where :name is null or name = :name";
 Option<String> param = Option.of("eric");
 
-// will fetch first user with given name or first user without name (Option.none)
+// will fetch first user with given name or first user with any name (Option.none)
 return handle.createQuery(query)
         .bind("name", param)
         .mapToBean(User.class)


qualidafial:
The idea is to filter query results only if a query parameter is non-null.

This is a good fit for e.g. `GET` requests to a REST resource with optional query parameters.

This pattern can be used with multiple search parameters to mix-and-match parameter combinations:

```sql
"select * from mytable
where (:foo is null or foo = :foo)
and (:bar is null or bar = :bar)
and (:baz is null or baz = :baz)
```

juggernaut:
Fair enough. I read "will fetch first user ... without name" to mean "user where name is null". IMO, it could be a worded a little clearer. How do you feel about something like "fetch user with a given name filter (Option.some), otherwise without a name filter (Option.none)"

qualidafial:
I actually think the wording you've suggested is good. I just wanted to explain the reason we use that pattern in our examples.

Issue #1261, opened 2018-10-09
Can't bind to Postgres enum arrays


juggernaut:
I'm trying to bind enum arrays but getting a `No argument factory registered` error. See code snippet:

```
enum Colors {
        red, green, blue;
}

int i = jdbi.withHandle(h -> {
            h.execute("CREATE TYPE colors as ENUM ('red', 'green', 'blue')");
            h.execute("CREATE TABLE candies (id SERIAL PRIMARY KEY, colors colors[])");

            final Colors[] colors = new Colors[] {Colors.red, Colors.green};
            return h.createUpdate("INSERT INTO candies (colors) VALUES (?)")
                    .bind(0, colors)
                    .execute();
        });
```
I've also tried `.bindBySqlType(0, colors, Types.ARRAY)` and that throws `org.postgresql.util.PSQLException: Cannot cast an instance of [Lcom.ballanceapp.db.PostgresInetTest$Colors; to type Types.ARRAY`

I believe it's possible to work around this using String binding + casting, but is it possible to add support for this natively in JDBI? FWIW, I tested this with plain JDBC using `PreparedStatement.setArray()` and it seems to work.

juggernaut:
@qualidafial thanks, that works! Since it's not immediately obvious, do you think this should be documented someplace? Also, is it worth adding a test case for this scenario? (I can do this)

qualidafial:
We have a section in the developer guide, but it's just a placeholder at the moment. http://jdbi.org/#_sql_arrays

If you could put together a pull request with the basics, that would be much appreciated! I rather have my hands full trying to wrap up a separate PR at the moment.

qualidafial:
If you clone the Jdbi repository, you can find the developer guide source in `docs/src/adoc/index.adoc` and generate the site (from the docs folder) with:

```
mvn org.asciidoctor:asciidoctor-maven-plugin:process-asciidoc
```

qualidafial:
Closing question as answered.

PR #1264, opened 2018-10-14
SQL Array docs

juggernaut:
As promised in https://github.com/jdbi/jdbi/issues/1261 , I've written up some basic docs for the Sql Arrays section based on the TODOs.

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -1593,7 +1593,7 @@ Jdbi can be enhanced to support arbitrary container types. See
 
 === SQL Arrays
 
-Jdbi can bind/map Java arrays and collections to/from SQL arrays:
+Jdbi can bind/map Java arrays to/from SQL arrays:
 
 [source,java]
 ----
@@ -1605,6 +1605,22 @@ handle.createUpdate("insert into groups (id, user_ids) values (:id, :userIds)")
 int[] userIds = handle.createQuery("select user_ids from groups where id = :id")
       .bind("id", 1)
       .mapTo(int[].class)
+      .findOnly();
+----
+
+You can also use Collections in place of arrays, but you'll need to provide
+generic type information if you're using the fluent API:
+
+[source,java]
+----
+handle.createUpdate("insert into groups (id, user_ids) values (:id, :userIds)")
+      .bind("id", 1)
+      .bindByType("userIds", Arrays.asList(10, 5, 70), new GenericType<List<Integer>>() {})
+      .execute();
+
+List<Integer> userIds = handle.createQuery("select user_ids from groups where id = :id")
+      .bind("id", 1)
+      .mapTo(new GenericType<List<Integer>>() {})
       .findOnly();
 ----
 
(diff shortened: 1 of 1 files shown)


qualidafial:
A few details we could add that would make this better:

- SqlArrayType is used for binding arrays. You can map arrays of any type with a registered column mapper.
- You can bind Java arrays or collections to array columns
- Array columns can be mapped to any container type registered with JdbiCollectors. E.g. a `VARCHAR[]` may be mapped into an `ImmutableList<String>` if the guava plugin is installed.

juggernaut:
Thanks for the feedback! I've added an example that shows binding collections instead of arrays, and another section for mapping custom types. Let me know if this looks good.

docs/src/adoc/index.adoc
```diff
@@ -1608,6 +1608,22 @@ int[] userIds = handle.createQuery("select user_ids from groups where id = :id")
...
 ----
 
+You can also use Collections in place of arrays, but you'll need to provide
+generic type information if you're using the fluent API:
+
+[source,java]
+----
+handle.createUpdate("insert into groups (id, user_ids) values (:id, :userIds)")
+      .bind("id", 1)
+      .bindByType("userIds", Arrays.asList(10, 5, 70), new GenericType<List<Integer>>() {})
```
qualidafial: I was thinking we could add a `bindArray(column, collection, elementType)` method to make array binding simpler in core.

(in a separate PR)

docs/src/adoc/index.adoc
```diff
@@ -1680,6 +1696,38 @@ handle.createUpdate("insert into groups (id, user_ids) values (:id, :users)")
...
 
+==== Mapping custom array types
+
+`SqlArrayType` only allows you to bind Java array/collection arguments to their
+SQL counterparts. To map SQL array columns back to Java types, you can register
+a regular `ColumnMapper`:
+
+[source,java]
+----
+public class UserIdListColumnMapper implements ColumnMapper<List<UserId>> {
```
qualidafial: Actually, you only need to register a `ColumnMapper<UserId>` and Jdbi figures out the rest, for any type of collection that's registered with `JdbiCollectors` config class.

docs/src/adoc/index.adoc
```diff
@@ -1680,6 +1696,38 @@ handle.createUpdate("insert into groups (id, user_ids) values (:id, :users)")
...
+                .collect(Collectors.toList());
+    }
+}
+----
+
+[source,java]
+----
+List<UserId> userIds = handle.createQuery("select user_ids from groups where id = :id")
+      .bind("id", 1)
+      .map(new UserIdListColumnMapper())
```
qualidafial: This would change to e.g. `mapTo(new GenericType<List<UserId>>() {})`

docs/src/adoc/index.adoc
juggernaut: thanks, fixed. Much clearer now

docs/src/adoc/index.adoc
juggernaut: Yes, that would indeed be nice - I've created an issue for this https://github.com/jdbi/jdbi/issues/1274

qualidafial:
@juggernaut thanks for contributing!

Issue #1263, opened 2018-10-13
Java 9 Immutable List Nullpointer


28Smiles:
public final This defineList(String key, List<?> values) {
            if (values.isEmpty()) {
                throw new IllegalArgumentException(
                        getClass().getSimpleName() + ".defineList was called with an empty list.");
            }
            if (values.contains(null)) {
                throw new IllegalArgumentException(
                        getClass().getSimpleName() + ".defineList was called with a list containing null values.");
            }
    
            String value = values.stream()
                    .map(Object::toString)
                    .collect(joining(", "));
    
            return define(key, value);
    }

`values.contains(null)` throws a nullpointer exception if list is created with `List.of()`, since it's `contains` method produces a NullpointerException if filled with null:

        @Override
        public boolean contains(Object o) {
            for (E e : elements) {
                if (o.equals(e)) { // implicit nullcheck of o
                    return true;
                }
            }
            return false;
        }

qualidafial:
List.of() API docs say that NPE will be thrown for any null elements: https://docs.oracle.com/javase/9/docs/api/java/util/List.html#of-E-

28Smiles:
@qualidafial yes thats right, but in case of contains, the List will throw a NPE if asked if it contains null

qualidafial:
Do you want Jdbi to stop calling contains() because one implementation doesn't guard against null?

TheRealMarnes:
This very much sounds like a non-jdbi-issue to me, for what my opinion's worth. It's not jdbi's fault the list decides to throw an npe on null items it contains (why can you even get a null into it if it doesn't like nulls?). Jdbi needs to check for nulls, so it does. Your issue is with the list implementation you decided to use.

28Smiles:
@TheRealMarnes @qualidafial I agree it isn't an jdbi issue, but isn't it desireable to at least support all types of list the JCL provides, a fix would be to do it simmilar to the `bindList` implementation, with a stream.

TheRealMarnes:
This List.of that throws an npe on null items (that you somehow got into the list in the first place) and doesn't throw that npe when you stream the items (because logic) is seriously a jdk implementation?

I mean, jdbi kinda has to support it then, but I can't wrap my head around such a thing actually being in the jdk.

28Smiles:
@TheRealMarnes  Yes it is, that is why I opened this Issue.

TheRealMarnes:
[...]

PR #1265, opened 2018-10-14
Fix NPE if SqlStatement::defineList is supplied with an ImmutableList

28Smiles:
See #1263


TheRealMarnes:
I'd like to see a comment added above the line to explain that the unnecessary stream() middle-man is a necessary workaround.

28Smiles:
Oh thx forgot about that

28Smiles:
@TheRealMarnes What do you think about this comment?

TheRealMarnes:
👍

qualidafial:
The spec violates Liskov's Substitution Principle, but I guess that's neither here nor there.

28Smiles:
@qualidafial I do not know if you want to add it later on, but the bugfix is missing in the release notes.

28Smiles:
I could

PR #1268, opened 2018-10-15
Break BuiltInMapperFactory up into functional parts

TheRealMarnes:
This PR makes no functional changes to jdbi, only breaks up and redistributes the BuiltInMapperFactory godclass into more focused, independent, manageable parts. An advantage of this is that it becomes more realistic to start adding configuration to the builtin support, like how the primitives mapping should handle nulls (throw exception or return default: `PrimitivesMapping.nullHandling` or some such). I've added a small dozen such TODOs (potential features or things that might need to be checked) that became more obvious thanks to this refactor.

I intend to do the same to the other 2 BuiltIn classes, after which we can do all kinds of neat things like adding essential config parameters to all the builtin support in a clutter-free way, bundle them into an `OotbPlugin` or something to make them more cleanly installable (instead of being hardcoded into the Context system, see `ColumnMappers`), and I might be able to better implement #1177

I love housekeeping :)

Is this line of work welcome or am I bikeshedding?


TheRealMarnes:
FYI I've done the commits so that if you look at it commit per commit, you can see it's literally just copy-pasting code from one place to another for most of it, to keep the mental complexity down.

core/src/main/java/org/jdbi/v3/core/mapper/BuiltInMapperFactory.java
```diff
@@ -1,216 +0,0 @@
...
-import org.jdbi.v3.core.config.ConfigRegistry;
-import org.jdbi.v3.core.statement.StatementContext;
-
-import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
-
-/**
- * Column mapper factory which knows how to map JDBC-recognized types, along with some other well-known types
- * from the JDK.
- */
-public class BuiltInMapperFactory implements ColumnMapperFactory {
```
stevenschlansker: This is removing what is technically public API.
I doubt anyone uses it - it's installed by default.  But perhaps we should leave it as deprecated and preserve its functionality by moving the list of installed factories from `ColumnMappers` to here?

stevenschlansker: That also leaves this class as the start of your `OotbPlugin`

TheRealMarnes: How do you mean exactly, move the list there?

qualidafial: You could preserve existing behavior by delegating to a list of all the new mapper factories:

```java
private static final List<ColumnMapperFactory> FACTORIES = new ArrayList<>();

static {
        FACTORIES.add(new EnumMapperFactory());
        FACTORIES.add(new OptionalMapperFactory());
        FACTORIES.add(new PrimitiveMapperFactory());
        FACTORIES.add(new BoxedMapperFactory());
        FACTORIES.add(new EssentialsMapperFactory());
        FACTORIES.add(new InternetMapperFactory());
        FACTORIES.add(new SqlTimeMapperFactory());
        FACTORIES.add(new JavaTimeMapperFactory());
        FACTORIES.add(new OptionalMapperFactory());}
}

  return FACTORIES.stream()
    .flatMap(factory -> JdbiStreams.toStream(factory.build(type, config)))
    .findFirst();
}
```

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapperFactory.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Optional;
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
+/**
+ * Column mapper factory which knows how to map Enums.
+ */
+public class EnumMapperFactory implements ColumnMapperFactory {
+    // TODO relation to EnumByNameMapperFactory?
```
qualidafial: I would just use `EnumByNameMapperFactory` rather than introduce a new type--maybe amend that class to use `rawType.asSubclass()` as you've done here

core/src/main/java/org/jdbi/v3/core/mapper/BoxedMapperFactory.java
```diff
@@ -0,0 +1,57 @@
...
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
+/**
+ * Column mapper factory which knows how to map boxed java primitives.
```
qualidafial: I'd like it if each of these new factory classes listed the supported types in javadoc, e.g. in a `<ul>`

core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
```diff
@@ -35,7 +34,16 @@
     private ConfigRegistry registry;
 
     public ColumnMappers() {
-        factories.add(new BuiltInMapperFactory());
+        // TODO BuiltInSupportPlugin?
```
qualidafial: I'm not opposed. `CorePlugin` perhaps?

core/src/main/java/org/jdbi/v3/core/mapper/EssentialsMapperFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
+/**
+ * Column mapper factory which knows how to map high-level essentials like String.
+ */
+public class EssentialsMapperFactory implements ColumnMapperFactory {
+    private static final Map<Class<?>, ColumnMapper<?>> MAPPERS = new HashMap<>();
+
+    // TODO consider other types?
```
qualidafial: Did you have other types in mind?

qualidafial:
Forgot to say: I agree with @stevenschlansker that we can't remove `BuiltInMapperFactory` as this would be binary incompatible, but we can hollow it out and have it just delegate to the new factories.

core/src/main/java/org/jdbi/v3/core/mapper/EssentialsMapperFactory.java
TheRealMarnes: I just figured there may be more essential high-level datatypes than what we have, so we should give it a second of brainstorming, but I had none in mind yet.

BigInteger maybe?

core/src/main/java/org/jdbi/v3/core/mapper/SqlTimeMapperFactory.java
```diff
@@ -0,0 +1,44 @@
...
+
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
+/**
+ * Column mapper factory which knows how to map java.sql timekeeping objects.
+ */
+public class SqlTimeMapperFactory implements ColumnMapperFactory {
+    private static final Map<Class<?>, ColumnMapper<?>> MAPPERS = new HashMap<>();
+
+    // TODO isn't Date support missing here? perhaps other types too? in fact why is Timestamp here if ResultSet has Timestamp support?
```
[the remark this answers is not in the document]
TheRealMarnes: I think the second part is my bad. ResultSet has explicit getDate and getTimestamp support, so I thought Date was left out because the getObjectMapper would get it, but that's only a thing from my private project. That then made me wonder why Timestamp wasn't omitted for the same reason.

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapperFactory.java
TheRealMarnes: I'm not savvy on the logic involved in this function, like the asSubclass() call. Do I replace the logic in EnumMapperByNameFactory with the logic in this class, and then delegate to it? A future improvement would be to make this class configurable to use either ByName or ByOrdinal. I like how this division of responsibilities sounds.

TheRealMarnes:
Can I move as much as possible of this stuff (without breaking compatibility — for now) to internal packages? I think there's way way too much exposed as supposedly API for our own good.

TheRealMarnes:
Are there plans for a major class access restriction move in the future? I feel it's a bit silly trying to minimze the existing technically-but-not-really public API piece by piece if it breaks compatibility anyway and could all be done in one go instead.

TheRealMarnes:
All remarks handled

TheRealMarnes:
Question: should these new factories be public so people can mix-and-match as they like, or do we wanna consider them an implementation detail of the ootbplugin?

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapperFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+import java.lang.reflect.Type;
+import java.util.Optional;
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+/**
+ * Column mapper factory which knows how to map {@link Enum} instances.
+ */
+class EnumMapperFactory implements ColumnMapperFactory {
+    private static final EnumByNameMapperFactory BY_NAME = new EnumByNameMapperFactory();
```
qualidafial: Again--why not just register the existing `EnumByNameMapperFactory` directly instead of adding this factory? That class is already public API.

If the intent is to be able to swap out the default enum binding / mapping strategy, let's save that for a separate PR.

qualidafial: Just noticed this class is now non-public--so this comment doesn't need to block merge.

qualidafial:
To respond to your earlier question: I'm generally ok with making APIs public, provided there is a concrete use case to support that decision.

Otherwise I make things private by default, for a few reasons:
- avoids the effort of hardening an API for mass consumption
- makes it possible to radically refactor without breaking compatibility

qualidafial:
One more thing: I'd be okay with replacing the reference to `BuiltInArgumentFactory` in `ColumnMapper` constructor with registrations of the new factory classes instead. That way `BuiltInArgumentFactory` continues to function like it always did, but isn't used any more by default.

TheRealMarnes:
Well yeah, what I'm asking is do you think there's a reasonable need for users to access the new factories directly, or should everyone roll their own if they decide not to use the ootbplugin?

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapperFactory.java
TheRealMarnes: Well if we just register one or the other, we wouldn't be able to switch the implementation on the fly when the user configures it, we'd need to register the other on top, forever filling up the factories list. I think this should stay as it is for now, it's a necessary abstraction.

TheRealMarnes:
I think it's ready. I'll get started on the similar work on arguments and collectors soon, and then finalize it all with the plugin.

qualidafial:
Tidy work. Thanks!

PR #1275, opened 2018-10-17
Add test for custom array type

juggernaut:
Didn't see a test for this - let me know if this is actually tested someplace else. 
Test binding a custom array type by implementing a `SqlArrayType` and
mapping it back to a Java collection of the element type.


juggernaut:
Fixed; I blame intellij for generating the checkstyle-violating code.

qualidafial:
Happens to me too. :)

Thanks for the PR!

PR #1278, opened 2018-10-17
[done] Break up BuiltInArgumentFactory

TheRealMarnes:
Same deal as before :)


core/src/main/java/org/jdbi/v3/core/argument/BoxedArgumentFactory.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
+class BoxedArgumentFactory implements ArgumentFactory {
+    private static final Map<Class<?>, ArgBuilder<?>> BUILDERS = createInternalBuilders();
```
qualidafial: Might be simpler if these were instance members. Then you could set up argument builders in the constructor instead of a static initializer.

The same could probably apply to the mapper factories from your last PR too.

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
```diff
@@ -13,218 +13,44 @@
...
-                return String.format("{array of %s length %s}", klass.getComponentType(), Array.getLength(value));
-            }
-            return String.valueOf(value);
-        }
-
-        StatementBinder<T> getStatementBinder() {
-            return binder;
-        }
+        return FACTORIES.stream()
+            .map(factory -> factory.build(expectedType, value, config))
```
qualidafial: I prefer flatMap for this. It sucks they left out Optional.stream() in Java 8

```java
return FACTORIES.stream()
    .flatMap(factory -> JdbiStreams.toStream(
        factory.build(expectedType, value, config)))
    .findFirst();
```

core/src/main/java/org/jdbi/v3/core/argument/InternetArgumentFactory.java
```diff
@@ -0,0 +1,68 @@
...
+import java.sql.Types;
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
+// :D
```
qualidafial: 😆

core/src/main/java/org/jdbi/v3/core/argument/JavaTimeArgumentFactory.java
```diff
@@ -0,0 +1,68 @@
...
+        Class<?> expectedClass = getErasedType(expectedType);
+
+        if (value != null && expectedClass == Object.class) {
+            expectedClass = value.getClass();
+        }
+
+        @SuppressWarnings("rawtypes")
+        ArgBuilder v = BUILDERS.get(expectedClass);
+
+        return Optional.ofNullable(v).map(f -> f.build(value));
```
qualidafial: This body gets repeated a lot. We should try to reduce this duplication.

We already have `AbstractArgumentFactory`. Maybe we need a simpler one that doesn't need subclassing, where you pass in a Java type, a SQL type, and a lambda that does the binding. If it's natural enough it might warrant it's own method in `Configurable`.

TheRealMarnes: Yeah, I noticed that... The same thing is happening with the collectors atm. I'll try to find a way.

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
TheRealMarnes: Is that really better though? I find it unintuitive to read with that custom stream tool in the middle.

stevenschlansker: We could consider upping to Java 9 source level, which would give us access to `Optional.stream()`.  Java 8 is unsupported, after all... is it too soon?

core/src/main/java/org/jdbi/v3/core/argument/InternetArgumentFactory.java
stevenschlansker: :upside_down_face:

core/src/main/java/org/jdbi/v3/core/argument/OptionalArgumentFactory.java
```diff
@@ -0,0 +1,99 @@
...
+            }
+        });
+        register(map, OptionalLong.class, Types.BIGINT, (p, i, v) -> {
+            if (v.isPresent()) {
+                p.setLong(i, v.getAsLong());
+            } else {
+                p.setNull(i, Types.BIGINT);
+            }
+        });
+        register(map, OptionalDouble.class, Types.DOUBLE, (p, i, v) -> {
```
stevenschlansker: You might be able to extract this out into a `register(map, OptionalDouble.class, OptionalDouble::getAsDouble, PreparedStatement::setDouble)` saving the need to repeat the lambda body thrice

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
TheRealMarnes: Java 8 is EOL already? I feel like it only just came out and the world is only just starting to use it by default.

core/src/main/java/org/jdbi/v3/core/argument/OptionalArgumentFactory.java
TheRealMarnes: Ah yes, I'd forgotten.

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
TheRealMarnes: image

TheRealMarnes:
image
I've broken the repetitive code down to this, but I can't think of a good place to put this aside from a superclass that would only serve as an implementation holder, not a functional superclass, which is kind of a code smell. Putting it in Configurable feels out of place to me and burdens all these factories with implementing getConfig as well. Any good ideas?

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
qualidafial: AdoptOpenJDK will likely keep maintaining JDK 8 for a while after Oracle stops supporting it

qualidafial:
Could be a static `ArgumentFactory.of(javaType, sqlType, binderLambda)` similar to what we have in `ColumnMapperFactory`, `RowMapperFactory`

core/src/main/java/org/jdbi/v3/core/argument/OptionalArgumentFactory.java
TheRealMarnes: Hmm, now that it's written, I can't say I'm a fan...

image

qualidafial:
Something like this:
```java
@FunctionalInterface
public interface ArgumentFactory {
    Optional<Argument> build(Type type, Object value, ConfigRegistry config);

    static <T> ArgumentFactory of(Class<T> type,
                                  TypedArgumentFactory<T> binder) {
        return of((Type) type, binder);
    }

    static <T> ArgumentFactory of(GenericType<T> type,
                                  TypedArgumentFactory<T> binder) {
        return of(type.getType(), binder);
    }

    static ArgumentFactory of(Type type,
                              TypedArgumentFactory binder) {
        return (t, v, c) -> type.equals(t)
            ? Optional.of(binder.build(v))
            : Optional.empty();
    }

    interface TypedArgumentFactory<T> {
        Argument build(T value);
    }
}
```

thus:

```java
jdbi.registerArgument(UserId.class,
                      value -> (pos, stmt, ctx) -> stmt.setString(value.getId()));
```

TheRealMarnes:
Are the factory methods you just suggested something you want me to add, that you see value in for the API, or would it be best if I put it somewhere internal instead?

TheRealMarnes:
I'm still trying to refactor out the repetition in the factories... Just not really finding any natural way to do it yet, only an implementation-holding parent class.

TheRealMarnes:
Hmm, creating that implementation-parent actually leads to a few cryptical copy-paste-avoidance classes becoming obsolete. Me likes.

TheRealMarnes:
The last refactor shows jdbi has inconsistent ways of binding objects to statements btw. Is each one intentional or would it be a good move for a major release to use a single strategy except where necessary?

stevenschlansker:
You mean for example changing some of the strategies to `bindNull` rather than throwing a NPE?  I think that sort of change is fine to make, it's a small behavior change that only adds functionality (nobody should intentionally run into NPE).  Or do you have something more invasive in mind?

TheRealMarnes:
The one that throws NPE I would indeed like to get rid of right off the bat, it's an anti-feature. But the other strategies should also be consolidated into one unless they have a very specific reason for existing. That would be a behavioral change though.

TheRealMarnes:
@stevenschlansker turns out they're actually null-safe by circumstances... I think the code involved here could use some work later though, if only to make it more obvious how it all works and which things affect each other how.

PR #1281, opened 2018-10-18
Fix typo in ColumnMapper section

juggernaut:
Missed replacing type variable `T` with actual type `Money`


stevenschlansker:
Thanks!

TheRealMarnes:
@stevenschlansker does modifying the doc files have any kind of impact, are there any considerations to when or how the merge happens?

qualidafial:
The doc site is updated by running the script in the docs folder.

Ideally we should do this on every release. In practice, sometimes we forget.

qualidafial:
Correct.

PR #1282, opened 2018-10-18
add basic Immutables integration

stevenschlansker:
this PR attempts to add initial Immutables integration.  it only recognizes the default name styling and method arguments / return types for now.

This also breaks out some of the core BeanMapper reflection logic into a new `PojoProperties` interface, which intends to allow pluggable bean-like conventions.  It's explicitly still internal, though.


core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesTaster.java
```diff
@@ -0,0 +1,185 @@
...
+        registered.add(iface);
+        registered.add(impl);
+    }
+
+    @Override
+    public Optional<PojoProperties<?>> apply(Type t) {
+        final Optional<Class<?>> defn =
+            Arrays.stream(GenericTypes.getErasedType(t).getInterfaces())
+            .filter(
+                i -> i.getAnnotation(Value.Immutable.class) != null
```
stevenschlansker: Note that these checks are totally ineffective right now, none of the annotations have runtime retention...

core/src/test/java/org/jdbi/v3/core/mapper/ImmutablesTest.java
```diff
@@ -0,0 +1,66 @@
...
+                .bindBean(ImmutableSubValue.<String, Integer>builder().t(42).x("foo").build())
+                .execute())
+            .isEqualTo(1);
+
+        assertThat(
+            h.createQuery("select * from immutables")
+                .mapTo(new GenericType<SubValue<String, Integer>>() {})
+                .findOnly())
+            .extracting("t", "x")
+            .isEqualTo(Arrays.asList(42, "foo"));
```
stevenschlansker: :astonished:

stevenschlansker:
Note that I chose to extend the Bean binding and mapping, rather than provide a parallel `bindImmutable` or something like that.  This is mostly because we already have Nested support there -- if we want our "bean-like"s to be mutually nestable they should go through the same code path here.

Taking this further might allow us to for example extend Nested support through FieldMapper down the road, for example.

stevenschlansker:
I added some documentation and polished things up a bit.
I think this is ready for at least initial review.  Do you think that extending the existing bean mapping in this way is the right call?

qualidafial:
Sorry about the merge conflicts--I refactored the reflection mappers a fair amount in the qualified types PR

docs/src/adoc/index.adoc
```diff
@@ -3689,6 +3689,22 @@ instance:
 jdbi.installPlugin(new H2DatabasePlugin());
 ----
 
+== Immutables
+
+link:https://immutables.github.io/[Immutables] is an annotation processor that generates
```
qualidafial: Link needs trailing `^` in link title to open link in new tab.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/locator/TestSqlLocator.java
```diff
@@ -32,7 +33,7 @@
     public PgDatabaseRule dbRule = new PgDatabaseRule().withPlugin(new SqlObjectPlugin());
 
     @Test
-    public void testLocateConfigDriven() {
+    public void testLocateConfigDriven() throws Exception {
```
qualidafial: Why does this throw an exception now?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesTaster.java
```diff
@@ -0,0 +1,286 @@
...
+    static MethodHandle alwaysSet() {
+        return MethodHandles.dropArguments(MethodHandles.constant(boolean.class, true), 0, Object.class);
+    }
+
+    private static <T> T guard(ThrowingSupplier<T> supp) {
+        try {
+            return supp.get();
+        } catch (RuntimeException | Error e) { // NOPMD
+            throw e;
+        } catch (Throwable t) {
```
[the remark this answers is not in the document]
stevenschlansker: The problem is that `MethodHandle.invoke` is declared as `throws Throwable`.
"Classic" reflection avoided this with `UndeclaredThrowableException` but that's gone from the "new" style.

Consider a property method
```java
public int getFoo() throws Throwable {
    throw new Throwable("barf");
}
```

this will be thrown as-is out of the `invoke` call and we have to handle that.  Or am I missing something obvious?

sqlobject/src/test/java/org/jdbi/v3/sqlobject/locator/TestSqlLocator.java
stevenschlansker: I think Travis isn't actually testing against 11 because of toolchains...

stevenschlansker:
Phew.  That was a somewhat gnarly merge.  But I think I resolved all the review comments.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java
```diff
@@ -101,10 +102,17 @@ public static RowMapperFactory factory(Class<?> type, String prefix) {
...
     protected final String prefix;
+    private final PojoProperties<T> beanInfo;
     private final Map<PojoProperty<T>, BeanMapper<?>> nestedMappers = new ConcurrentHashMap<>();
 
+    @SuppressWarnings("unchecked")
     protected BeanMapper(Class<T> type, String prefix) {
+        this(type, (PojoProperties<T>) BeanPropertiesFactory.propertiesFor(type), prefix);
+    }
+
+    protected BeanMapper(Class<T> type, PojoProperties<T> properties, String prefix) {
```
qualidafial: Exposing PojoProperties (an internal class) to API

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesTaster.java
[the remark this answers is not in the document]
stevenschlansker: This is all gone now in favor of `Unchecked`.

stevenschlansker:
OK, this is ready for another pass.  It's far from done but it's stable and usable even in its current infancy.  Everything is `@Beta` so I'm hoping we can ship the current version and incrementally improve over time.

TheRealMarnes:
I'm not sure I can really give an informed content-wise review of this. At most I can check for superficially/obviously fishy stuff, I think

stevenschlansker:
The `immutables` library is well worth learning about for your own benefit, but I understand if you don't want to / have time to do this right now.  I'm mostly just looking for a review to make sure I've not completely buggered it up so that it can be included in the next release; I fully expect to 'own' this feature for a while yet.

TheRealMarnes:
Does this fix #841 ?

TheRealMarnes:
fixes #841

core/src/main/java/org/jdbi/v3/core/internal/JdbiOptionals.java
```diff
@@ -30,4 +30,11 @@ private JdbiOptionals() {
                 .flatMap(supplier -> toStream(supplier.get()))
                 .findFirst();
     }
+
+    public static <T> Stream<T> stream(Optional<T> optional) {
+        if (optional.isPresent()) {
```
electrum: This can be a one-liner if you prefer:
```java
return optional.map(Stream::of).orElse(Stream.of());

File: core/pom.xml [hunk 8]
```diff
@@ -67,6 +67,12 @@
             <artifactId>expiringmap</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.immutables</groupId>
+            <artifactId>value</artifactId>
+            <scope>provided</scope>
```
electrum: This should probably be `<optional>true</optional>`. While they seem to do the same thing, `provided` scope is typically used for things that are required and will be provided at runtime by the container, whereas optional dependencies are, well, optional. From the [Maven documentation](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html): 

> Optional dependencies are used when it's not really possible (for whatever reason) to split a project up into sub-modules.

The above explanation seems to fit this use case. We can think of this Immutables support as a logically separate module.

TheRealMarnes: Oh, yes, isn't it a separate module? I would prefer if it was, with an `ImmutablesPlugin`.

TheRealMarnes: It's debatable though, because this whole feature is built without ever referring to immutables as external code of any kind.

qualidafial: Immutables should be in test scope, as there is no compile-time dependency at all in the main source tree.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
```diff
@@ -0,0 +1,266 @@
...
+                    properties.put(name, createProperty(name, m));
+                }
+            }
+        }
+
+        private boolean isProperty(Method m) {
+            return m.getParameterCount() == 0
+                && !m.isSynthetic()
+                && !Modifier.isStatic(m.getModifiers())
+                && m.getDeclaringClass() != Object.class;
```
qualidafial: May want to exclude bridge methods too

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
```diff
@@ -0,0 +1,266 @@
...
+        }
+
+        @Override
+        public Map<String, ? extends PojoProperty<T>> getProperties() {
+            return properties;
+        }
+
+        protected abstract ImmutablesPojoProperty<T> createProperty(String name, Method m);
```
qualidafial: Could be package private

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
```diff
@@ -0,0 +1,266 @@
...
+        protected final Class<?> defn;
+        protected final Supplier<?> builder;
+
+        BasePojoProperties(Type type, Class<?> defn, Supplier<B> builder) {
+            super(type);
+            this.defn = defn;
+            this.builder = builder;
+            for (Method m : defn.getMethods()) {
+                if (isProperty(m)) {
+                    final String name = m.getName();
```
qualidafial: Immutables supports both `getName()` and `name()` method naming patterns out of the box--we should probably run this value through a function that strips the `get`/`is` prefix if present.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
```diff
@@ -0,0 +1,266 @@
...
+            } catch (IllegalAccessException | NoSuchMethodException e) {
+                throw new IllegalArgumentException("Failed to inspect method " + m, e);
+            }
+        }
+
+        private MethodHandle findBuilderSetter(final Class<?> builderClass, String name, Type type)
+        throws IllegalAccessException, NoSuchMethodException {
+            final NoSuchMethodException failure;
+            try {
+                return MethodHandles.lookup().unreflect(builderClass.getMethod(name, GenericTypes.getErasedType(type)));
```
qualidafial: Similar to getters, we should fall back to the `setName()` method if `name()` setter not found.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
```diff
@@ -0,0 +1,266 @@
...
+            } catch (NoSuchMethodException e) {
+                failure = e;
+            }
+            for (Method m : builderClass.getMethods()) {
+                if (m.getName().equals(name) && m.getParameterCount() == 1) {
+                    return MethodHandles.lookup().unreflect(m);
```
qualidafial: So if we don't find the method with the exact parameter type, search for single-parameter method with the right name and hope the parameter type is right?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
```diff
@@ -0,0 +1,266 @@
...
+                        MethodHandles.lookup().unreflect(m),
+                        MethodHandles.lookup().findVirtual(impl, setterName(name), MethodType.methodType(impl, GenericTypes.getErasedType(propertyType))));
+            } catch (IllegalAccessException | NoSuchMethodException e) {
+                throw new IllegalArgumentException("Failed to inspect method " + m, e);
+            }
+        }
+
+        @SuppressWarnings("unchecked")
+        private Class<M> implClass() {
+            return (Class<M>) builder.get().getClass();
```
qualidafial: This will create a garbage builder instance on every invocation. But since it's only invoked once per method and cached, that's probably ok.

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
```diff
@@ -57,13 +64,21 @@ public static QualifiedType of(Type type, Annotation qualifier) {
 
     /**
      * Creates a QualifiedType for {@code type} with the given qualifiers.
+     *
+     * @param type the type
+     * @param qualifiers the qualifiers
+     * @return the QualifiedType
```
qualidafial: Probably a source of merge conflicts here. We changed to `of(type).with(qualifiers...)` style

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -183,6 +184,33 @@ public This bindBean(String prefix, Object bean) {
...
 
+    /**
+     * Binds named parameters from object properties on the argument.
+     *
+     * @param pojo source of named parameter values to use as arguments
+     *
+     * @return modified statement
+     */
+    @Beta
+    public This bindProperties(Object pojo) {
```
qualidafial: What do you think about `bindPojo()` as a more concise name?

core/src/test/java/org/jdbi/v3/core/mapper/ImmutablesTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        h.registerRowMapper(ImmutablesMapperFactory.mapImmutable(SubValue.class, ImmutableSubValue.class, ImmutableSubValue::builder));
+    }
+
+    // tag::example[]
+    @Value.Immutable
+    public interface Train {
+        String name();
+        int carriages();
+        boolean observationCar();
```
qualidafial: Can we add some tests for the `getX()` naming style too?

core/src/main/java/org/jdbi/v3/core/argument/internal/PojoPropertyArguments.java
```diff
@@ -0,0 +1,69 @@
...
+     * @param bean the bean to inspect and bind
+     * @param ctx the statement context
+     */
+    public PojoPropertyArguments(String prefix, Object bean, StatementContext ctx) {
+        super(prefix, bean);
+        this.ctx = ctx;
+        final RowMapper<? extends Object> mapper = ctx.findRowMapperFor(bean.getClass())
+                .orElseThrow(() -> new UnableToCreateStatementException("Couldn't find registered property mapper for " + bean.getClass()));
+        if (!(mapper instanceof BeanMapper<?>)) {
+            throw new UnableToCreateStatementException("Registered mapper for " + bean.getClass() + " is not a property based mapper");
```
qualidafial: What's the difference between PojoPropertyArguments and BeanPropertyArguments?

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
qualidafial: This needs to be documented that you can only use this method if you've registered a property-based mapper

core/src/test/java/org/jdbi/v3/core/mapper/ImmutablesTest.java
```diff
@@ -0,0 +1,131 @@
...
+                .mapTo(ModifiableFooBarBaz.class)
+                .findOnly())
+            .extracting("id", "foo", "bar", "baz")
+            .containsExactly(1, Optional.of("foo"), OptionalInt.of(42), OptionalDouble.of(1.0));
+
+        assertThat(h.createQuery("select * from fbb")
+                .mapTo(ImmutableFooBarBaz.class)
+                .findOnly())
+            .extracting("id", "foo", "bar", "baz")
+        // FIXME: https://github.com/joel-costigliola/assertj-core/pull/1360
```
qualidafial: This still an issue?

postgres/pom.xml
```diff
@@ -54,6 +54,12 @@
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.immutables</groupId>
+            <artifactId>value</artifactId>
+            <scope>provided</scope>
```
qualidafial: Should be test scope, as there is no compile dependency in src/main

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindProperties.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.jdbi.v3.sqlobject.customizer.internal.BindPropertiesFactory;
+
+/**
+ * Binds the properties of an object to a SQL statement.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.PARAMETER})
+@SqlStatementCustomizingAnnotation(BindPropertiesFactory.class)
+public @interface BindProperties {
```
qualidafial: If we decide to rename `bindProperties()` to `bindPojo()` then we should likewise rename this annotation.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindProperties.java
```diff
@@ -0,0 +1,71 @@
...
+    public H2DatabaseRule dbRule = new H2DatabaseRule().withPlugin(new SqlObjectPlugin());
+
+    private Handle h;
+
+    private Dao dao;
+
+    @Before
+    public void setUp() {
+        h = dbRule.getSharedHandle();
+        h.registerRowMapper(ImmutablesMapperFactory.mapImmutable(Train.class, ImmutableTrain.class, ImmutableTrain::builder));
```
qualidafial: Feels like we want annotations to register Immutables mappers, e.g. `@RegisterImmutablesMapper(value = Train.class)` annotation. The impl and builder method name could be default attributes that look for `Immutable*` and `Immutable*.builder()`, respectively.

qualidafial: While we're on the subject.. those automatic values feel like good parameter defaults on the immutable and modifiable factory methods themselves.

qualidafial: We could name the annotations `@RegisterRowMapper.Immutable` and `@RegisterRowMapper.Modifiable` just to be smart asses :trollface:

core/src/test/java/org/jdbi/v3/core/mapper/ImmutablesTest.java
stevenschlansker: It's merged but not released yet :/

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindProperties.java
qualidafial: Yeah it's a bit of a misnomer to mix Immutable and Modifiable together, but that's sort of what the Immurables library hands us with their `@Value.Immutable` and `@Value.Modifiable` annotations. I'm not sure how to advertise the "Modifiable" support without being confusing.

qualidafial:
TBH the Modifiable vs Immutable thing is making me think we might be better off moving the Immutable mappers to their own module so we can explicitly advertise that module as Immutables library integration. Otherwise users are likely to conflate the API names as generic terms, vs the library specific context their intended for.

stevenschlansker:
The reason I didn't move it to its own module is that it (1) doesn't require a new runtime library and (2) currently shares a fair amount of code with the BeanMapper and friends.  Moving it to a separate module will make much more of the currently hidden code public, and since we can't move BeanMapper, it'll become API.  At least if it's in `core` we can use package private and protected.

stevenschlansker:
I'll make sure to add javadoc to clarify what the publicly visible classes are.

core/src/test/java/org/jdbi/v3/core/mapper/ImmutablesTest.java
stevenschlansker: Pinged the maintainer asking for a release.

qualidafial:
If we move it to a separate module we can always share public classes in internal packages.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutablesMapperFactory.java
```diff
@@ -0,0 +1,59 @@
...
+        this.defn = defn;
+        this.impl = impl;
+        this.properties = properties;
+    }
+
+    public static <T, I extends T, B> RowMapperFactory mapImmutable(Class<T> defn, Class<I> immutable, Supplier<B> builder) {
+        return new ImmutablesMapperFactory<>(defn, immutable, ImmutablesPropertiesFactory.immutable(defn, builder));
+    }
+
+    public static <T, M extends T> RowMapperFactory mapModifiable(Class<T> defn, Class<M> modifiable, Supplier<M> constructor) {
```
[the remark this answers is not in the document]
stevenschlansker: It is a little weird, but I'm not sure what to do better here.  I've changed these to `forModifiable` and `forImmutable` which is maybe incrementally better, but fundamentally these are the literal names from the library, and as kris-nova points out we should try to match it.  Open to suggestions.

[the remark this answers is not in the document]
TheRealMarnes: I think the fact that the class is named for the library makes it clear enough that users should interpret its content in that lib's context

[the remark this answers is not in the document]
TheRealMarnes: Then again, "immutables" is pretty generic so one could easily not recognize it as a library integration...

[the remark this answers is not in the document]
stevenschlansker: I think I'll make it a multiple-instantiable `JdbiPlugin` instead.  Should be nicer and it gets rid of the `getBeanInfo` hack too...

stevenschlansker:
Right, but `BeanMapper` is unfortunately stuck as public api.  So any changes we make that are needed in other packages will leak into public API accidentally.  It probably should have itself been `internal` but far too late for that...

TheRealMarnes:
And you can't cut BeanMapper loose or make it only a front for the real impl, reasonably implement all the changes you need internally with as much as duplication or delegation as may be necessary? I didn't look up the exact code the discussion is about, so honest question. If doing a true parallel implementation would be arguably silly, then we'll just have to leave it in core and make it an [idiot in a hurry thing](https://en.wikipedia.org/wiki/A_moron_in_a_hurry).

Besides, let's not pretend users are supposed to blindly use something without glancing at its javadoc just because the name sounds about right ¯\\\_(ツ)\_/¯

core/pom.xml
stevenschlansker: Unfortunately moving it to `<scope>test</scope>` breaks Eclipse's annotation processor, as it doesn't expect to find test-only processors.  So I've made it `<scope>provided</scope><optional>true</optional>`, hopefully that works?

core/src/main/java/org/jdbi/v3/core/argument/internal/PojoPropertyArguments.java
stevenschlansker: `BeanPropertyArguments` is public API and in a different package, and as such it's really hard to extend it without accidentally introducing public API.  Even a new constructor shows up... will see if there's a better way to do this.

This probably also should have been `internal`.

stevenschlansker: Cool, I fixed it by making the new stuff `protected`, and this is now simply a stub subclass (`stubclass` :trollface: ).  I marked `BeanPropertyArguments` deprecated.  We can clean it all up eventually.

electrum:
Dependencies for annotation processors should now be added to the compiler plugin: https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#annotationProcessorPaths

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoMapper.java
```diff
@@ -11,17 +11,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jdbi.v3.core.internal;
+package org.jdbi.v3.core.mapper.reflect.internal;
```
stevenschlansker: Thanks Git...

stevenschlansker:
Who can even keep track of all this stuff... :P

stevenschlansker:
Ok.  I've cleaned up the public API -- now there's only the newly renamed `bindPojo`, `@BindPojo`, and an `ImmutablesPlugin` that does the previous hacky bit of registering a row mapper with properties handoff to the argument binder.  All the other new added code is in `internal`.

I moved the `immutables` dependency to the compiler annotation path, which solves all the scope and optional questions nicely.  I made it unconditionally active since it's simpler in the pom that way, and it only triggers when the compiler finds the annotation anyway.

stevenschlansker:
Still going to add `get` / `is` prefix support, but that'll have to come later -- gotta run.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindProperties.java
stevenschlansker: The new approach is to have an `ImmutablesPlugin` with `forImmutable` and `forModifiable` static factory methods.  how's that?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ImmutablesMapperFactory.java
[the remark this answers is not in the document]
stevenschlansker: This stuff is no longer public.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestBindProperties.java
stevenschlansker: Instead of making a new annotation, I made a plugin.  Hopefully that's a good solution.

core/src/test/java/org/jdbi/v3/core/mapper/ImmutablesTest.java
stevenschlansker: Sounds like it's not going to happen for a bit, so we shouldn't block on it, unfortunately. https://github.com/joel-costigliola/assertj-core/pull/1360#issuecomment-451627976

core/src/main/java/org/jdbi/v3/core/mapper/reflect/PropertiesMapper.java
```diff
@@ -0,0 +1,49 @@
...
+package org.jdbi.v3.core.mapper.reflect;
+
+import org.jdbi.v3.core.mapper.ColumnMapper;
+import org.jdbi.v3.core.mapper.NoSuchMapperException;
+import org.jdbi.v3.core.mapper.RowMapper;
+import org.jdbi.v3.core.mapper.RowMapperFactory;
+import org.jdbi.v3.core.mapper.reflect.internal.PojoProperties;
+import org.jdbi.v3.core.mapper.reflect.internal.PojoProperties.PojoProperty;
+
+class PropertiesMapper<T> extends BeanMapper<T> {
```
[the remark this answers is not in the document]
stevenschlansker: OK.  That leaves `BeanMapper` as a deprecated shell, which is just fine by me :)

stevenschlansker:
I added support for bean-style naming.  Now, in theory, this code could entirely replace the `Introspector` based approach... (but obviously this PR does not do it).  Based on @qualidafial 's suggestion, I moved the code into `internal` packages via `PojoPropertyArguments` and `PojoMapper` but left the original classes deprecated so as not to change public api.  I believe I've now addressed all feedback and I'm looking for one last review!

TheRealMarnes:
I've pushed a master-merge to the best of my knowledge for you

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
```diff
@@ -41,22 +41,16 @@
     private final Set<Annotation> qualifiers;
 
     /**
-     * Creates a QualifiedType for {@code type} with no qualifiers. In practice, the returned object is treated the same
-     * as using {@code type} raw.
-     *
-     * @param type the unqualified type
-     * @return the unqualified QualifiedType
+     * Creates a {@code QualifiedType<T>} for a {@code Class<T>} with no extra qualifiers.
+     * This alone isn't useful; you should call {@link #with} as well.
```
stevenschlansker: This is in fact useful (and in fact necessary), for when you have a `Type` and need an unqualified `QualifiedType`.  For example:

```java
    public Optional<Argument> findFor(Type type, Object value) {
        return findFor(QualifiedType.of(type), value);
    }
```

docs/src/adoc/index.adoc
```diff
@@ -3848,6 +3848,22 @@ class MyBean {
...
+[source,java]
+----
+QualifiedType<MyJson> qualifier = QualifiedType.of(MyJson.class).with(Json.class);
+
+h.createUpdate("insert into myjsons(json) values(:json)")
+    .bindByType("json", new MyJson(), qualifier)
+    .execute();
+
+MyJson result = h.createQuery("select json from myjsons")
+    .mapTo(MyJson.class, Json.class)
```
stevenschlansker: This example is inconsistent: how come we use `QualifiedType` on the `bindByType` call but break it apart for `mapTo`.  Shouldn't they be symmetric?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoMapper.java
```diff
@@ -100,8 +100,9 @@ public T map(ResultSet rs, StatementContext ctx) throws SQLException {
 
                 findColumnIndex(paramName, columnNames, columnNameMatchers, () -> debugName(property))
                     .ifPresent(index -> {
+                        @SuppressWarnings("unchecked")
                         ColumnMapper<?> mapper = ctx.findColumnMapperFor(property.getQualifiedType())
-                            .orElseGet(() -> defaultColumnMapper(property));
+                            .orElseGet(() -> (ColumnMapper) defaultColumnMapper(property));
```
stevenschlansker: Introduces a raw type warning.

TheRealMarnes:
The comments and replies got a little screwed up there somehow :/

Do you wanna join us on slack so we can coordinate more closely on some of these details? None of this is helpful to do in public and it feels like trying to talk through a wall. We'd get the fine lines sorted out much quicker talking more directly and freely.

If you don't like my attempt to handle the merge for you (I did that because it seemed to annoy you so), feel free to throw the commits away and handle the merge the way you want it done, or squash them, or whatever. You can do on your branch as you like, and discarding some work isn't a sin.

qualidafial:
@stevenschlansker One thing that Marnes pointed out is that ImmutablesPlugin is stateful, and so you have to install the plugin once per immutable / modifiable type. I don't care for this for a few reasons:

- The usage here is quite different from the how the rest of the library feels. Using plugins in a way that we usually use config classes.
- Since Immutables support is only provided through the plugin, you have to register on the `Jdbi` instance. This forecloses the possibility for a `@RegisterImmutableType(defn = Foo.class, impl = ImmutableFoo.class)` SQL Object annotation.

The fact that ImmutablesPlugin has any fields at all suggests to me that it really wants to be a config class instead.

I propose:

- Rename `PojoProperties` to `PojoType`, and move it and all the types it references into public packages. (still `@Beta` though)
- Add a `PojoTypes` registry config class as `@Beta` API, with a `register(type, PojoType)` method.
- Refactor `bindPojo()` to lookup the `PojoType` from the `PojoTypes` registry, instead of looking up a row mapper and checking `instanceof PojoMapper`, etc.
- Add `PojoMapperFactory` and register it by default. Have it automatically map any type for which a `PojoType` is registered.
- Add an `Immutables` config class to act as an Immutables-specific facade for the `PojoTypes` registry. When you call `Immutables.registerImmutable(Foo.class, ImmutableFoo.class)` or `.registerModifiable(Foo.class, ModifiableFoo.class)`, it constructs the appropriate `PojoType` implementation and registers it twice with `PojoTypes`--once for the definition class, once for the implementation class.
- Introduce SQL Object annotations for each supported POJO type, with attributes to fit the use case:
  - `@RegisterImmutableType(defn = Foo.class, impl = ImmutableFoo.class)`
  - `@RegisterBeanType(MyBean.class)`
- [MAYBE?] Alter the SQL Object default to check if there is a `PojoType` registered for the parameter type. If call `bindPojo()` instead of `bindByType()`?

stevenschlansker:
Hm, indeed, somehow I managed to comment on code that isn't in the final diff.  Not sure how that happened.

I'm actually already halfway to your proposed `PojoTypes` (it's called `PojoPropertiesFactories`), so that's a fine proposal.  That's how I got rid of the cast you mention (it's not in the latest revision already)

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
stevenschlansker: I've made it an `@see` instead

core/src/main/java/org/jdbi/v3/core/argument/internal/PojoPropertyArguments.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    protected Optional<TypedValue> getValue(String name, StatementContext ctx2) {
+        @SuppressWarnings("unchecked")
+        PojoProperty<Object> property = (PojoProperty<Object>) properties.getProperties().get(name);
+
+        if (property == null) {
+            return Optional.empty();
+        }
+
+        return Optional.of(new TypedValue(property.getQualifiedType(), property.get(obj)));
```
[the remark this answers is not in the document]
stevenschlansker: done

stevenschlansker:
Also, sorry, I can't really be on slack from court.  I can only get a half hour or so of work in during the intermittent breaks.

stevenschlansker:
OK I've made changes that are largely in line with your proposal above.  I'd actually already set up almost all of the hard work...  It turns out the reason I did the plugin approach before is that it's somewhat of a PITA to just do random configuration on tests, it works much better with plugins.  So I've added some helpers to the test rules to make it a bit easier to register immutable types via configuring.

stevenschlansker:
And to be clear, now that I realize what's happened -- it wasn't the merge itself that bugged me, it was that it brought in a Javadoc conflict that introduced warnings.  But I've fixed it all up, and hopefully the `@see` solves your concern (people should be aware of that) without calling an unqualified type useless.

core/src/main/java/org/jdbi/v3/core/argument/internal/PojoPropertyArguments.java
```diff
@@ -0,0 +1,64 @@
...
+ * When we can remove that class from public API, this class will easily replace it.
+ */
+public class PojoPropertyArguments extends MethodReturnValueNamedArgumentFinder {
+    private final PojoProperties<?> properties;
+    private final StatementContext ctx;
+
+    public PojoPropertyArguments(String prefix, Object obj, StatementContext ctx) {
+        this(prefix,
+                obj,
+                ctx.getConfig(PojoTypes.class).propertiesOf(obj.getClass())
```
qualidafial: Can we name this method `findFor()` to stay consistent with other registries?

core/src/main/java/org/jdbi/v3/core/mapper/immutables/JdbiImmutables.java
```diff
@@ -0,0 +1,135 @@
...
+import org.jdbi.v3.core.mapper.reflect.internal.ImmutablesPropertiesFactory;
+import org.jdbi.v3.core.mapper.reflect.internal.PojoProperties;
+import org.jdbi.v3.core.mapper.reflect.internal.PojoTypes;
+import org.jdbi.v3.meta.Beta;
+
+/**
+ * Configures support for an Immutables generated {@code Immutable} or {@code Modifiable} value type.
+ */
+@Beta
+public class JdbiImmutables implements JdbiConfig<JdbiImmutables> {
```
qualidafial: This is really nice

core/src/main/java/org/jdbi/v3/core/mapper/immutables/JdbiImmutables.java
```diff
@@ -0,0 +1,135 @@
...
+import org.jdbi.v3.core.config.JdbiConfig;
+import org.jdbi.v3.core.internal.JdbiOptionals;
+import org.jdbi.v3.core.internal.exceptions.Unchecked;
+import org.jdbi.v3.core.mapper.reflect.internal.ImmutablesPropertiesFactory;
+import org.jdbi.v3.core.mapper.reflect.internal.PojoProperties;
+import org.jdbi.v3.core.mapper.reflect.internal.PojoTypes;
+import org.jdbi.v3.meta.Beta;
+
+/**
+ * Configures support for an Immutables generated {@code Immutable} or {@code Modifiable} value type.
```
qualidafial: Should probably link to the immutables.github.io project to make it clear that we're talking about a specific project, and not the abstract concept of immutability.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java
```diff
@@ -51,31 +30,12 @@
  * properties.
  *
  * The mapped class must have a default constructor.
+ *
+ * @deprecated this class should not be public API, use {@link org.jdbi.v3.core.statement.SqlStatement#bindBean(Object)} instead.
```
qualidafial: Copypasta: `bindBean()` is not a substitute for `BeanMapper` :)

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/BeanPropertiesFactory.java
```diff
@@ -0,0 +1,192 @@
...
+            @Override
+            public QualifiedType<?> getQualifiedType() {
+                Parameter setterParam = Optional.ofNullable(descriptor.getWriteMethod())
+                    .map(m -> m.getParameterCount() > 0 ? m.getParameters()[0] : null)
+                    .orElse(null);
+
+                return QualifiedType.of(
+                    Optional.ofNullable(descriptor.getReadMethod())
+                        .map(Method::getGenericReturnType)
+                        .orElseGet(() -> descriptor.getWriteMethod().getGenericParameterTypes()[0]))
```
qualidafial: I'm not sure if we were doing this before, but should we resolve these types against the bean class? The properties might parameterized from a supertype.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
```diff
@@ -0,0 +1,286 @@
...
+        protected ImmutablesPojoProperty<T> createProperty(String name, Method m) {
+            final Type propertyType = GenericTypes.resolveType(m.getGenericReturnType(), getType());
+            try {
+                return new ImmutablesPojoProperty<T>(
+                        name,
+                        QualifiedType.of(propertyType).with(Qualifiers.getQualifiers(m)),
+                        m,
+                        isSetMethod(name),
+                        MethodHandles.lookup().unreflect(m),
+                        MethodHandles.lookup().findVirtual(impl, setterName(name), MethodType.methodType(impl, GenericTypes.getErasedType(propertyType))));
```
qualidafial: We should test this assumption that modifiables always generate `setX(x)` vs just `x(x)` if the original getter method is just `x()`. The docs only give `setX()` examples but that might just be an omission.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoMapper.java
```diff
@@ -0,0 +1,174 @@
...
+        });
+    }
+
+    @SuppressWarnings({ "rawtypes", "unchecked" }) // Type loses <T>
+    protected PojoMapper<?> createNestedMapper(StatementContext ctx, PojoProperty<T> property, String nestedPrefix) {
+        final Type propertyType = property.getQualifiedType().getType();
+        return new PojoMapper(
+                GenericTypes.getErasedType(propertyType),
+                ctx.getConfig(PojoTypes.class).propertiesOf(type)
+                    .orElseThrow(() -> new UnableToProduceResultException("Couldn't find properties for nested type " + propertyType, ctx)),
```
qualidafial: Whoa, does this mean we can mix and match pojo types for nested mappers?

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -179,10 +181,36 @@ public This bindBean(Object bean) {
...
         return bindNamedArgumentFinder(new BeanPropertyArguments(prefix, bean));
     }
 
+    /**
+     * Binds named parameters from object properties on the argument.
+     *
+     * @param pojo source of named parameter values to use as arguments
+     *
+     * @return modified statement
+     */
```
qualidafial: Should document that the pojo type must already be registered, and add `@see` to the possible registration methods. Ditto for the next method.

core/src/test/java/org/jdbi/v3/core/mapper/ImmutablesTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+    @Test
+    public void simpleTest() {
+        jdbi.getConfig(JdbiImmutables.class).registerImmutable(Train.class);
+        try (Handle handle = jdbi.open()) {
+            handle.execute("create table train (name varchar, carriages int, observation_car boolean)");
+
+            assertThat(
+                handle.createUpdate("insert into train(name, carriages, observation_car) values (:name, :carriages, :observationCar)")
+                    .bindPojo(ImmutableTrain.builder().name("Zephyr").carriages(8).observationCar(true).build())
```
qualidafial: choo choo

pom.xml
```diff
@@ -500,6 +507,13 @@
                     <configuration>
                         <source>${project.build.targetJdk}</source>
                         <target>${project.build.targetJdk}</target>
+                        <annotationProcessorPaths>
+                            <annotationProcessorPath>
+                                <groupId>org.immutables</groupId>
+                                <artifactId>value</artifactId>
+                                <version>${dep.immutables.version}</version>
+                            </annotationProcessorPath>
+                        </annotationProcessorPaths>
```
qualidafial: Is this annotation processor config still required?

qualidafial:
Whoops.. probably should have gathered those into a single review.

Some final thoughts: should we add new API for registering a class as a bean type, e.g. `Configurable.registerBeanClass()`? Then we can realistically deprecate `@BindBean` in favor of `@BindPojo`, and deprecate `BeanMapper`

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoMapper.java
TheRealMarnes: I'd expect no less from a recursive system tbh. Cool that it's now implemented. :+1:

core/src/main/java/org/jdbi/v3/core/argument/internal/PojoPropertyArguments.java
TheRealMarnes: I think that makes sense here, the method signature clarifies "find _what_ for this" enough IMO.

core/src/test/java/org/jdbi/v3/core/mapper/ImmutablesTest.java
stevenschlansker: 🚂

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoMapper.java
stevenschlansker: Yup!  Eventually this could be a more flexible replacement for the entire existing bean binding infrastructure.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java
```diff
@@ -120,158 +80,18 @@ public static RowMapperFactory factory(Class<?> type, String prefix) {
...
-        this.type = type;
-        this.prefix = prefix.toLowerCase();
-        try {
-            info = Introspector.getBeanInfo(type);
-        } catch (IntrospectionException e) {
-            throw new IllegalArgumentException(e);
-        }
+    private BeanMapper(Class<T> type, PojoProperties<T> properties, String prefix) {
+        super(type, properties, prefix);
+        strictColumnMapping = false;
```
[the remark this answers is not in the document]
stevenschlansker: No, this is orthogonal, but with a confusingly similar name.  This is declaring that we must find a column mapper for all property types we encounter.  In the original BeanMapper, if we can't figure out how to handle a type, we just assume it's a vendor type and call `rs.getObject`: https://github.com/jdbi/jdbi/blob/master/core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java#L192-L193

I think this behavior is a shitty v2 holdover, personally, so the new code opts to instead throw an exception rather than fallback to `getObject` and hope.

The boolean allows the new implementation to behave one way and the existing `BeanMapper` another.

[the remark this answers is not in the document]
stevenschlansker: I'll change it to `strictColumnTypeMapping` how's that?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/BeanPropertiesFactory.java
```diff
@@ -0,0 +1,192 @@
...
+        BeanPojoProperties(Type type) {
+            super(type);
+            try {
+                this.info = Introspector.getBeanInfo(GenericTypes.getErasedType(type));
+            } catch (IntrospectionException e) {
+                throw new IllegalArgumentException("Failed to inspect bean " + type, e);
+            }
+            final Map<String, BeanPojoProperty<T>> props = new LinkedHashMap<>();
+            for (PropertyDescriptor property : info.getPropertyDescriptors()) {
+                final BeanPojoProperty<T> bp = new BeanPojoProperty<>(property);
```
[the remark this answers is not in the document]
stevenschlansker: No need to, the `Introspector` apparently does that for us.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/BeanPropertiesFactory.java
stevenschlansker: We weren't doing that before.  I agree with you it would make sense to do so, but can we save that for another PR?  I'm trying to match existing bean infrastructure behavior as closely as possible to reduce risk of this PR.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
```diff
@@ -0,0 +1,286 @@
...
+            .builder()
+            .expiration(10, TimeUnit.MINUTES)
+            .expirationPolicy(ExpirationPolicy.ACCESSED)
+            .build();
+
+    private ImmutablesPropertiesFactory() {}
+
+    @SuppressWarnings("unchecked")
+    public static <T, B> Function<Type, PojoProperties<T>> immutable(Class<T> defn, Supplier<B> builder) {
+        return t -> (PojoProperties<T>) IMMUTABLE_PROPERTIES.computeIfAbsent(t, x -> new ImmutablePojoProperties<>(t, defn, builder));
```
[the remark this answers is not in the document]
stevenschlansker: All of the caches are keyed off a `Class<?>` which obviously loses generic parameter info.  A property type could be generic based on the type signature, and this gives us the opportunity to consider the full (possibly `Generic`)`Type`.  Right now it does not add much benefit but later this would allow us to support parameterized pojo types.

pom.xml
stevenschlansker: Yes, at least Eclipse does not seem to consider test-only dependencies for annotation processing, and this tells it to actually use it.  I think javac may not need it but it's a PITA to not have your IDE work.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
stevenschlansker: It's possible to customize it by changing the annotation `@Value.Style`, however if you do that then we have no clue what the prefix will be (as the annotation is not retained) so I'm not sure how I would actually handle the situation.  Currently it's simply documented to not support custom setter styles.  I could easily craft a builder that does not follow the `setX` pattern but I see it as out of scope to handle that in this first pass, if that's ok, as I simply don't see a clean way to do it without further cooperation from Immutables.

stevenschlansker: Specifically:
```java
@Value.Style:
    /**
     * Modifiable object "setter" method. Used for mutable implementations.
     * Do not confuse with {@link #init()}
     * @return naming template
     */
    String set() default "set*";
```

stevenschlansker:
I'm torn on your final suggestion.  On one hand, it removes the need to know the magic `.getConfig(JdbiImmutables.class).register...` to just `.register...`

On the other hand, `Configurable` is starting to become a dumping ground for lots of different configuration types.  It leads to e.g. `registerRowMapper` and this new `registerPojoType` being commingled in a place that otherwise neither would live in.  They're all just facades:

```java
    default This registerArrayType(SqlArrayType<?> arrayType) {
        return configure(SqlArrayTypes.class, c -> c.register(arrayType));
    }
```

How do we draw a line between features that are "important" and "core" enough to deserve this treatment, versus those that aren't?  It also inherently prefers implementing features in core, since otherwise you cannot add it to Configurable, which is probably not the incentive we want.

On the other hand, it's clearly shorter and nicer to not have to type `JdbiImmutables` :)

Thoughts?

TheRealMarnes:
My personal opinion is that redundant shorthands are to be avoided in any code not directly purposed for said feature. It just creates dependency spaghetti, expresses interconnectedness that goes against the actual design/blurs the borders between components, and ultimately only saves you having to type a few letters. I wouldn't add any for immutables, and if I could, I'd delete the existing ones that don't relate directly to the core mechanics (e.g. you can't get around having arguments and mappers, but most of the rest is non-critical luxury).

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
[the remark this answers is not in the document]
stevenschlansker: Actually, it works right now.  There's even a test! :) see `parameterizedTest`.

stevenschlansker:
In that case, how about let's not do it as part of this PR.  We can choose to add e.g. `@RegisterImmutable` at a later point.

I've found myself more and more registering types in `JdbiPlugin`s anyway, rather than using the SqlObject annotations like `@RegisterConstructorMapper`.  Generally, when I have a type, I always want it to be mapped the same way - a `PersonDto` is a `PersonDto` whether it's read from a `PersonDao` or an `ImposterDao`.  I think that's true of most uses.  Not to say that we should restrict the ability to do so - but I think the SqlObject annotations aren't as useful for the common use case, you're better off doing it on the whole `jdbi` anyway.

I'm also +1 to deprecating `@BindBean` and friends, but I think we should defer that a bit too.  Let's let this new API bake in beta and get some use and battle testing, and at that point, we can start more forcefully nudging people to the new API.

I believe I've addressed all feedback, or suggested deferring it.  :finnadie:

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/BeanPropertiesFactory.java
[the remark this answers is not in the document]
stevenschlansker: oh, I see, this actually ends up being merged in from my other PR.  so it will get sifted out via `shouldSeeProperty`.

qualidafial:
If we're going to let it bake, let's not deprecate BeanMapper yet. Registering a bean mapper at the Jdbi level is still valid.

Other than that, I'm :+1: to merge

stevenschlansker:
image

qualidafial:
@stevenschlansker let's un-deprecate BeanMapper until we have a replacement lined up for mapper registration

stevenschlansker:
https://github.com/jdbi/jdbi/commit/96d95bda61cf4c107e83816486a9364fcf96982a

PR #1284, opened 2018-10-21
Update to commons-text version 1.6

PascalSchumacher:


TheRealMarnes:
Looks like you'll need to keep the section about commons-lang3 and set it to 3.8.1

stevenschlansker:
Is there a particular reason for this update?  (i.e. security fix, new feature in use?)
Any users who use commons-text 1.4 or 1.5 or commons-lang3 3.7.0 will now have their build broken by this commit, so it's important to understand what the gain is.  Thanks!

qualidafial:
I'm glad this came up, and I'd like to discuss it further. Broadly we have two options for managing dependencies in Jdbi:

- Always upgrade to the latest version
- Update dependencies only when we need bug fixes or new APIs

Whichever way we go, it's gonna create problems for somebody.

- Always upgrading deps in Jdbi will expand the scope of work users have to do when they upgrade Jdbi itself, since they will have to explicitly manage those transitive dependencies.
- Only upgrading when we want a bugfix or new API _might_ keep the scope of work smaller, but depending on the user's plugins, it might create _more_ work for them if _they_ want a newer version of the other dependency.

I'm currently trying to upgrade dependencies in a microservice. I'm stalled because one dependency uses Guava 18 deprecated API, and another dependency requires Guava 26, which removes the old deprecated APIs.

Honestly, if I had to choose which throat to choke, I would pick the maintainer who let his dependencies fall behind.

TheRealMarnes:
I tend to choose breakage by progress over anything else. Growth pains come with growth. We shouldn't stop growing just because it hurts a little. Not growing is much worse in the long run. If something hurts too much too often, we'll find a way to deal with it.

qualidafial:
I think I can tolerate us breaking a user's build, if they have to upgrade their Jdbi dependency for the breakage to happen in the first place. That's sort of par for the course.

The bar for breaking runtime behavior is much higher.

stevenschlansker:
OK by me.  It's always a little frustrating when "one thing leads to another" and one library update cascades, but I'm all for being ahead of the curve.

PR #1285, opened 2018-10-21
Update mockito-core to version 2.23.0

PascalSchumacher:


stevenschlansker:
That's an interesting idea.  However, there are some caveats.  Since jdbi is a library, the dependency versions are a *baseline* and are expected to be managed by projects that incorporate jdbi into a final build.  An update to any dependent library from version X to Y raises the baseline - any downstream users who are on version < Y will now have failing builds.

This certainly isn't a reason to avoid updating, and if you'd like to set up the bot and give it a spin, I'm all for it!  But we have to critically consider dependency updates, they are not necessarily harmless in a library context.

Since mockito is a test-only dependency, it is not transitive, so this concern does not apply to this PR.

PR #1287, opened 2018-10-23
Fix ClassCastException on `Error` thrown by `LocalTransactionHandler`.

GrahamDennis:
LocalTransactionHandler would attempt to cast the caught Throwable to
Exception, which would fail for non-Exception Throwables like Error.

This bug was introduced in c87c667a6d8c64795a83845b6e79cb728aa40ef6

### DIFF
--- a/core/src/test/java/org/jdbi/v3/core/transaction/TestLocalTransactionHandler.java
+++ b/core/src/test/java/org/jdbi/v3/core/transaction/TestLocalTransactionHandler.java
@@ -23,6 +23,7 @@ import org.mockito.junit.MockitoJUnit;
 import org.mockito.junit.MockitoRule;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 public class TestLocalTransactionHandler {
     @Rule
@@ -61,12 +62,11 @@ public class TestLocalTransactionHandler {
         Mockito.when(c.getAutoCommit()).thenReturn(true);
         Mockito.when(h.getConnection()).thenReturn(c);
 
-        try {
+        // Verify that we don't get a ClassCastException when an Error is thrown within a transaction.
+        assertThatThrownBy(() ->
             new LocalTransactionHandler().inTransaction(h, x -> {
                 throw error;
-            });
-        } catch (Throwable throwable) {
-            assertThat(throwable).isSameAs(error);
-        }
+            }))
+            .isSameAs(error);
     }
 }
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/jdbi/v3/core/transaction/TestLocalTransactionHandler.java
```diff
@@ -53,4 +53,20 @@ public void testRollbackThrow() throws Exception {
...
+        Error error = new Error("Transaction throws!");
+
+        Mockito.when(c.getAutoCommit()).thenReturn(true);
+        Mockito.when(h.getConnection()).thenReturn(c);
+
+        try {
+            new LocalTransactionHandler().inTransaction(h, x -> {
+                throw error;
+            });
+        } catch (Throwable throwable) {
```
TheRealMarnes: could you do this with assertj's `assertThatThrownBy` instead of a regular catch, please?

GrahamDennis: done!

core/src/test/java/org/jdbi/v3/core/transaction/TestLocalTransactionHandler.java
```diff
@@ -53,4 +54,19 @@ public void testRollbackThrow() throws Exception {
...
     }
+
+    @Test
+    public void testThrowError() throws Exception {
+        Error error = new Error("Transaction throws!");
+
+        Mockito.when(c.getAutoCommit()).thenReturn(true);
+        Mockito.when(h.getConnection()).thenReturn(c);
+
+        // Verify that we don't get a ClassCastException when an Error is thrown within a transaction.
```
TheRealMarnes: Well if you're going to write a comment to explain the test case, best also put it in the appropriate place, a `describedAs` call. But I'm not sure a comment is the way to go. An assertion like `isNotInstanceOf(ClassCastException.class)` would achieve the same but in a tangible way.

TheRealMarnes: Well if you're going to write a comment to explain the test case, best also put it in the appropriate place, a `describedAs` call. But I'm not sure a comment is the way to go. An assertion like `isNotInstanceOf(ClassCastException.class)` would give the same info but in a tangible way.

GrahamDennis:
Also I was mistaken. This was bug introduced in 57800cdea62937b48e305cca74a6f90e6fe3f2e7 not c87c667a6d8c64795a83845b6e79cb728aa40ef6.

core/src/test/java/org/jdbi/v3/core/transaction/TestLocalTransactionHandler.java
GrahamDennis: I've removed the comment and not added `.isNotInstanceOf(ClassCastException.cass)` as that would be redundant with the `.isSameAs(error)` check.

TheRealMarnes:
Great, thanks!

qualidafial:
We should not be catching Throwable in the first place. Error are unrecoverable and should be left unroll the stack without being caught.

TheRealMarnes:
Good point. I remember we catch Throwable in some places by necessity of the compiler, but this seems to be possible to do with more specific types. I'll make a PR to clean up our catches.

PR #1292, opened 2018-10-27
Readme change regarding java 11

28Smiles:
You should consider updating the readme to include informations about using Jdbi with Java 11 since you now also use Java 11 for your runners

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Jdbi with it.
 
 Jdbi 3 requires Java 8.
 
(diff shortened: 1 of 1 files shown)


TheRealMarnes:
I don't know what happened regarding java 11. @qualidafial @stevenschlansker ?

stevenschlansker:
Thanks for the reminder.  I'm going to be a bit more confident in the language, so I'll make a different commit.

PR #1298, opened 2018-11-03
avoid NPEs in builtin argfactories

TheRealMarnes:
I decided to go for the strategy that calls `setNull` if value is null (instead of throwing npe). The binding behavior when not null is identical to before.


TheRealMarnes:
AbstractArgumentFactory:79
```
 return Optional.of(value == null
    ? new NullArgument(sqlType)
    : build((T) value, config));
```

Turns out our base argfac does a null check before calling implementations to build their Argument. That kinda defeats the purpose of making our argumentfactories null-savvy.

I'd still merge this for the sake of housekeeping, but I think we need to discuss how this is supposed to work, because we have redundant mechanisms and it isn't obvious how it all fits together IMO.

Issue #1301, opened 2018-11-05
What is the jdbi3 eqivivalent of jdbi2 BinderFactory?


qwelyt:
I'm trying to move from jdbi2 to jdbi3 but I'm stumped on how to migrate our `BinderFactory` mappings.

As per http://jdbi.org/jdbi2/sql_object_api_argument_binding/ we can specify our own `BinderFactory` when we have an object that we want to bind in some specific way. As far as I can tell, `BinderFactory` has been removed from jdbi3 (and jdbi3 sqlobjects).

Given this code as an example

```java
@BindingAnnotation(BindSomething.SomethingBinderFactory.class)
@Retention(RetentionPolicy.RUNTIME)
@Target({ElementType.PARAMETER})
public @interface BindSomething { 
  public static class SomethingBinderFactory implements BinderFactory {
    public Binder build(Annotation annotation) {
      return new Binder<BindSomething, Something>()  {
        public void bind(SQLStatement q, BindSomething bind, Something arg)  {
          q.bind("id", arg.getId());
          q.bind("name", arg.getName());
        }
      };
    }
  }
}


public static interface BindSomethingExample {
  @SqlUpdate("insert into something (id, name) values (:id, :name)")
  void insert(@BindSomething Something s);
}
```

How would I go about doing this in JDBI3?

If there is a tutorial somewhere, I would gladly follow it. But I don't seem to be able to find any. http://jdbi.org/#_binding_arguments seems to be the closes, but it only deals with the "fluent" style.  
The migration "guide" ( http://jdbi.org/#_upgrading_from_v2_to_v3 ) doesn't mention the removal of `BinderFactory` either.

TheRealMarnes:
If you look at e.g. BindBean or BindList, you can follow their trail into SqlStatementCustomizerFactory instances:
```
@Override
public SqlStatementParameterCustomizer createForParameter(Annotation annotation,
                                                          Class<?> sqlObjectType,
                                                          Method method,
                                                          Parameter param,
                                                          int index,
                                                          Type type) {
    BindBean bind = (BindBean) annotation;
    return (stmt, bean) -> {
        String prefix = bind.value();
        if (prefix.isEmpty()) {
            stmt.bindBean(bean);
        } else {
            stmt.bindBean(prefix, bean);
        }
    };
}
```

Are these the lines of code you are looking for? 👋

TheRealMarnes:
[...]

PR #1305, opened 2018-11-07
Add generics to SqlStatementParameterCustomizer

qwelyt:
Previously you would have to perform a cast in your
SqlStatementParameterCustomizer.apply method to get the object you knew
you were passing in. This change really only removes the need to do the
casting.

MapToFactory needs to be explicitly told to use Object. Leaving the
generics blank caused compiler errors. Seeing how it was using Object
before no change has actually been made. It's more explicit, but that's
about it.

I asked for this here: https://github.com/jdbi/jdbi/issues/1301#issuecomment-436562644
Thought the discussion, if there should be any, could be here instead of in a issue that has nothing to do with it really.

### DIFF
--- a/kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/KotlinSqlStatementCustomizerFactory.kt
+++ b/kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/KotlinSqlStatementCustomizerFactory.kt
@@ -27,7 +27,7 @@ class KotlinSqlStatementCustomizerFactory : ParameterCustomizerFactory {
                                     method: Method,
                                     parameter: Parameter,
                                     paramIdx: Int,
-                                    type: Type): SqlStatementParameterCustomizer {
+                                    type: Type): SqlStatementParameterCustomizer<Any> {
 
         val bindName = if (parameter.isNamePresent) {
             parameter.name
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/SqlStatementParameterCustomizer.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/SqlStatementParameterCustomizer.java
@@ -20,12 +20,12 @@ import org.jdbi.v3.core.statement.SqlStatement;
 /**
  * Customize a {@link SqlStatement} according to the value of an annotated parameter.
  */
-public interface SqlStatementParameterCustomizer {
+public interface SqlStatementParameterCustomizer<T> {
     /**
      * Applies the customization to the SQL statement using the argument passed to the method.
      * @param stmt the statement being customized
      * @param arg the argument passed to the method
      * @throws SQLException will abort statement creation
      */
-    void apply(SqlStatement<?> stmt, Object arg) throws SQLException;
+    void apply(SqlStatement<?> stmt, T arg) throws SQLException;
 }
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/MapToFactory.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/MapToFactory.java
@@ -25,7 +25,7 @@ import org.jdbi.v3.sqlobject.customizer.SqlStatementParameterCustomizer;
 
 public class MapToFactory implements SqlStatementCustomizerFactory {
     @Override
-    public SqlStatementParameterCustomizer createForParameter(Annotation annotation,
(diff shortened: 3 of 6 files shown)


qwelyt:
https://github.com/jdbi/jdbi/issues/1301#issuecomment-436587268
> Hmm, I don't think we should do that. Nothing can stop you from making your BindEvent's implementation generic for Event and then putting it on a param of any other type. Jdbi shouldn't pretend to be able to safely give you a T when that isn't what happens in reality. The Object param isn't nice to users, I agree, but it is technically the correct thing to do, the right representation of the situation. Generics should only be used where you can actually guarantee type correctness. As soon as a blind cast is involved in a chain of generics, the chain is compromised and should be removed to avoid creating a false sense of safety.
> 
> image

While I agree that you could be lulled into a false sense of security, you would still get a `ClassCastException` if you tried to get some object that you are not actually sending in.
If you want to highlight that this might not succeed, you could append `throws ClassCastException`.

For me and my usage cases you will always have to deal with the cast, no matter how it is implemented, and you would always risk casting to the wrong thing no matter what with the current implementation (ie, JDBI can't provide type safety).

TheRealMarnes:
Personally not happy with this change. This makes a binding annotation specific for a given parameter type, but there's nothing that prevents you from putting that annotation on a parameter of another type. If we can't guarantee that the types match, then jdbi shouldn't pretend that they do. Generics shouldn't be used as a rug to hide type safety issues under without solving them. The warning generated by this change attests to that:
image

If we can't compile-time ensure that the input object is of type T, then we shouldn't compile-time promise that the output object given to the SqlStatementParameterCustomizer will be of type T.

qwelyt:
> append it where? 

I was thinking the `void apply(...)` method.  
While you *can* specify RuntimeExceptions (we do sometimes, not 100% sure why, historical things) I understand why you wouldn't want to.

I'm not all that familiar with JDBI, but in v2 you specified the type of the object you expected to get in your `Binder` eg
```java
   public @interface BindSomething {
      class SomethingBinderFactory implements BinderFactory {
         public Binder build(final Annotation annotation) {
            return (Binder<BindSomething, Something>) (q, bind, something) -> {
....
```
Was this erroneous in the same way that adding the parameter now would be?

TheRealMarnes:
> I was thinking the void apply(...) method.

But that wouldn't help. Jdbi calls the method, so jdbi would have to catch the exception. It should be the user who gets forced to consider the possibility that the type doesn't match if they misplace the annotation somewhere.

Either that, or a runtime-valid way of checking argument types needs to be added, like an `accepts(Object)` or `getType(): Class<?>` method, which is all more "foggy" than just receiving an Object, which is the reality of the situation. Generics won't solve anything here because they can't be preserved across the annotation-on-a-parameter gap.

As for your Binder question, it suffers the same problem. The fact remains that the annotation can be put on the wrong object, and then the call will throw an exception _somewhere_. It's better to make this visible to the user so they can deal with it, rather than trying to hide it.

TheRealMarnes:
Having to do a cast is by itself not a big issue, but it can't be a blind cast. There would have to be some mechanism that puts the responsibility of casting correctly with the same code that demands the cast in the first place. That would be a breaking change, and in the end it would still saddle you up with the same work, just in a less atomic and less concise structure: the type check and the cast operation must happen, so just doing them together in one or two lines of code on an Object param is a lot simpler than having to implement an additional method and losing the ability to use lambdas.

You're very welcome. Sorry if I gave the impression of being annoyed or something, I wasn't/am not, it's just a side-effect of trying to be clear and right down to business. :)

It's a good thing that these discussions happen and that both sides challenge each other. Thanks for getting involved in jdbi.

I'll assume this can be closed now.

qwelyt:
Are you thinking that this could provide the type safety? Or just hide the cast?  
If it just hides the cast, I agree with @TheRealMarnes , error hidden within a lib is *not* fun, and it should probably not be implemented.

I'm a lazy developer. I want to write as few lines as needed.  
But I'm also a *lazy* developer. I don't want to go bug-hunting to understand why something breaks *inside* a lib. If the ClassCastException is just hidden by the lib, I have to work extra hard to find out why it breaks rather than just see that it was *my* cast that broke stuff.

Though if the type can be preserved, then Yes, I would very much like the feature.

TheRealMarnes:
The type can't be preserved across the annotation-parameter gap no matter what. We can sort of do it though, by either making the creator of the annotation responsible for the cast himself by demanding a Class instance from it, or by using reflection to get a ParameterizedType that at least allows jdbi to detect mismatches before they happen so it can throw a useful exception.

qualidafial:
I'm not convinced it's worth adding and supporting a class just to avoid a cast in client code.

PR #1306, opened 2018-11-09
Fixed transaction support for Spring

tsydd:
jdbi doesn't support `@Transactional` annotation out of the box. To enable  it, we need to delegate opening and releasing connections to Spring. I've introduced `ConnectionHandler` and it's implementation for Spring

PR #1308, opened 2018-11-10
ConnectionFactory: add a closeConnection with default

### LINKED ISSUE #1304: Provide an out-of-the-box way to use JDBI with an existing connection.
Originally discussed on the mailing list here: https://groups.google.com/forum/#!topic/jdbi/x5Y5gqfnChg

There's currently no clean way to use JDBI with plugins & existing JDBC connection / transaction. This is because to install plugins it is necessary to use Jdbi instance (no official way to install plugins directly into Handle) and Jdbi instance closes underlying connection on every method use/with method invocation.

This functionality may be needed when inserting JDBI into legacy code, for example.

One workaround is to create JDBC connection wrapper that simply ignores invocation of close() -- but it would be nice if something like this is provided out-of-the-box so that there's no need to roll your own workaround every time.

stevenschlansker:
Fixes #1304

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/ConnectionFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/ConnectionFactory.java
@@ -17,13 +17,24 @@ import java.sql.Connection;
 import java.sql.SQLException;
 
 /**
- * Supplies {@link Connection} instances to a created {@link Handle}.
+ * Supplies {@link Connection} instances to a created {@link Handle} and allows
+ * custom close handling.
  */
 @FunctionalInterface
 public interface ConnectionFactory {
     /**
+     * Open a connection.
      * @return a Connection
      * @throws SQLException if anything goes wrong
      */
     Connection openConnection() throws SQLException;
+
+    /**
+     * Close a connection.
+     * @param conn the connection to close
+     * @throws SQLException if anything goes wrong
+     */
+    default void closeConnection(Connection conn) throws SQLException {
+        conn.close();
+    }
 }
--- a/core/src/main/java/org/jdbi/v3/core/Jdbi.java
+++ b/core/src/main/java/org/jdbi/v3/core/Jdbi.java
@@ -290,7 +290,7 @@ public class Jdbi implements Configurable<Jdbi> {
             }
 
             StatementBuilder cache = statementBuilderFactory.get().createStatementBuilder(conn);
-            Handle h = new Handle(config.createCopy(), transactionhandler.get(), cache, conn);
+            Handle h = new Handle(config.createCopy(), connectionFactory::closeConnection, transactionhandler.get(), cache, conn);
             for (JdbiPlugin p : plugins) {
                 h = p.customizeHandle(h);
             }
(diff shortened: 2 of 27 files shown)


TheRealMarnes:
So this is to allow the user to install a NOP ConnectionCloser so that a pool or framework can handle the connection lifetime, or how should I see it?

core/src/main/java/org/jdbi/v3/core/ConnectionFactory.java
```diff
@@ -17,13 +17,24 @@
...
      * @throws SQLException if anything goes wrong
      */
     Connection openConnection() throws SQLException;
+
+    /**
+     * Close a connection.
+     * @param conn the connection to close
+     * @throws SQLException if anything goes wrong
+     */
+    default void closeConnection(Connection conn) throws SQLException {
```
TheRealMarnes: should be explicitly tested

TheRealMarnes:
Sorry, I'm still getting into the whole way reviews on github work :p

TheRealMarnes:
Isn't #1306 the exact same thing (but already with a test)?

TheRealMarnes:
https://github.com/jdbi/jdbi/issues/1304#issuecomment-437546222 might be worth considering

stevenschlansker:
The big difference is that I thought adding a new default method is a cleaner / smaller change than adding a new interface, which is the approach the other PR took.

stevenschlansker:
Unless there's a reason to add a new interface, I think I will just merge this PR and the other PR together into a franken-PR.

TheRealMarnes:
As in, take his changes but without adding a new interface?

stevenschlansker:
I merged @tsydd 's work into this branch, did a bit of cleanup, and made the `SingleConnectionHandler` non-public (you can just use `Jdbi.create(Connection)` instead).

Could I get a re-review, @TheRealMarnes, and turn that approval-frown upside-down?  Thanks!

TheRealMarnes:
This is now b-e-a-utiful

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -62,6 +62,15 @@ private Jdbi(ConnectionFactory connectionFactory) {
         this.connectionFactory = connectionFactory;
     }
 
+    /**
+     * @param connection db connection
+     *
+     * @return a Jdbi which works on single connection
+     */
+    public static Jdbi create(Connection connection) {
```
qualidafial: Let's `@Beta` this until we have some appropriate disclaimers in place. I'm thinking of people using this method naively assuming it works the same as a connection pool.

spring4/src/main/java/org/jdbi/v3/spring4/SpringConnectionFactory.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Override
+    public Connection openConnection() throws SQLException {
+        return DataSourceUtils.getConnection(dataSource);
+    }
+
+    @Override
+    public void closeConnection(Connection connection) throws SQLException {
+        DataSourceUtils.releaseConnection(connection, dataSource);
```
qualidafial: Oh hell yes

PR #1309, opened 2018-11-13
Show useful message when IllegalArgument is passed

prathik:
When Object value cannot be used by the setter
method a very generic IllegalArgumentException
message is present. This leads to a lot of time
debugging the error. This exception message
clearly calls out why the error happened.


stevenschlansker:
Thanks for improving the error reporting here.

PR #1313, opened 2018-11-18
getting started warning

TheRealMarnes:


qualidafial:
A little too forcefully worded. Something like:

> Classes in `internal` packages are not public API and may change without notice.

PR #1316, opened 2018-11-24
Support @JdbiConstructor on static factory methods

### LINKED ISSUE #1314: Support @JdbiConstructor on static factory methods
For classes with public constructor, one can `@RegisterConstructorMapper` or instantiate `ConstructorMapper.factory(SomeClass.class)` directly.

Please provide equivalent functionality for classes that do not have public constructor, but have public static factory method. 

Note: compare this to Jackson. I can put `@JsonCreator` on a constructor or on a static factory method.

### Existing alternatives

Currently, user has to implement `RowMapper` directly, which means manually mentioning all columns,  their types, finding corresponding column mappers.
Benefiting from `org.jdbi.v3.core.mapper.RowMapper#specialize` is even more elaborate.

### Proposed solution

Allow `@JdbiConstructor` to be placed on methods. Annotated method must be static. Then it's used just as the constructor would be used.

findepi:
Fixes #1314

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
@@ -16,7 +16,6 @@ package org.jdbi.v3.core.mapper.reflect;
 import java.beans.ConstructorProperties;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Parameter;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -35,7 +34,7 @@ import org.jdbi.v3.core.mapper.SingleColumnMapper;
 import org.jdbi.v3.core.qualifier.QualifiedType;
 import org.jdbi.v3.core.statement.StatementContext;
 
-import static org.jdbi.v3.core.mapper.reflect.JdbiConstructors.findConstructorFor;
+import static org.jdbi.v3.core.mapper.reflect.JdbiConstructors.findFactoryFor;
 import static org.jdbi.v3.core.mapper.reflect.ReflectionMapperUtil.anyColumnsStartWithPrefix;
 import static org.jdbi.v3.core.mapper.reflect.ReflectionMapperUtil.findColumnIndex;
 import static org.jdbi.v3.core.mapper.reflect.ReflectionMapperUtil.getColumnNames;
@@ -55,22 +54,22 @@ public class ConstructorMapper<T> implements RowMapper<T> {
     private static final String DEFAULT_PREFIX = "";
 
     private static final String UNMATCHED_CONSTRUCTOR_PARAMETERS =
-        "Constructor '%s' could not match any parameter to any columns in the result set. "
+        "Instance factory '%s' could not match any parameter to any columns in the result set. "
             + "Verify that the Java compiler is configured to emit parameter names, "
             + "that your result set has the columns expected, annotate the "
             + "parameter names explicitly with @ColumnName, or annotate nullable parameters as @Nullable";
 
     private static final String UNMATCHED_CONSTRUCTOR_PARAMETER =
-        "Constructor '%s' parameter '%s' has no matching columns in the result set. "
+        "Instance factory '%s' parameter '%s' has no matching columns in the result set. "
             + "Verify that the Java compiler is configured to emit parameter names, "
             + "that your result set has the columns expected, annotate the "
             + "parameter names explicitly with @ColumnName, or annotate nullable parameters as @Nullable";
 
     private static final String UNMATCHED_COLUMNS_STRICT =
-        "Mapping constructor-injected type %s could not match parameters for columns: %s";
+        "Mapping instance factory %s could not match parameters for columns: %s";
 
     private static final String MISSING_COLUMN_MAPPER =
-        "Could not find column mapper for type '%s' of parameter '%s' for constructor '%s'";
+        "Could not find column mapper for type '%s' of parameter '%s' for instance factory '%s'";
 
     /**
      * Use the only declared constructor to map a class.
@@ -122,7 +121,7 @@ public class ConstructorMapper<T> implements RowMapper<T> {
      * @return the mapper
      */
     public static <T> RowMapper<T> of(Class<T> type) {
-        return ConstructorMapper.of(findConstructorFor(type));
+        return ConstructorMapper.of(type, DEFAULT_PREFIX);
     }
 
     /**
@@ -134,7 +133,7 @@ public class ConstructorMapper<T> implements RowMapper<T> {
      * @return the mapper
      */
     public static <T> RowMapper<T> of(Class<T> type, String prefix) {
-        return ConstructorMapper.of(findConstructorFor(type), prefix);
+        return new ConstructorMapper<>(findFactoryFor(type), prefix);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


findepi:
@anusudarsan

core/src/main/java/org/jdbi/v3/core/mapper/reflect/JdbiConstructors.java
```diff
@@ -27,6 +27,29 @@ private JdbiConstructors() {
...
+            Stream.of(constructors)
+                .filter(constructor -> constructor.isAnnotationPresent(JdbiConstructor.class))
+                .map(ConstructorInstanceFactory::new),
+            Stream.of(type.getDeclaredMethods())
+                .filter(method -> method.isAnnotationPresent(JdbiConstructor.class))
+                .map(method -> new StaticMethodInstanceFactory<>(type, method)))
+            .collect(Collectors.toList());
+
+        if (annotatedFactories.size() > 1) {
+            throw new IllegalArgumentException(type + " must have at most one constructor or static factory method annotated @JdbiConstructor");
```
qualidafial: s/must/may/

core/src/main/java/org/jdbi/v3/core/mapper/reflect/JdbiConstructors.java
```diff
@@ -27,6 +27,29 @@ private JdbiConstructors() {
...
+        List<InstanceFactory<T>> annotatedFactories = Stream.concat(
+            Stream.of(constructors)
+                .filter(constructor -> constructor.isAnnotationPresent(JdbiConstructor.class))
+                .map(ConstructorInstanceFactory::new),
+            Stream.of(type.getDeclaredMethods())
+                .filter(method -> method.isAnnotationPresent(JdbiConstructor.class))
+                .map(method -> new StaticMethodInstanceFactory<>(type, method)))
+            .collect(Collectors.toList());
+
+        if (annotatedFactories.size() > 1) {
```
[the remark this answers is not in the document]
qualidafial: I disagree on the switch statement--IMO that makes the code less readable.

I'm more concerned about the readability of this method. I'd like to see it refactored to something like:

```java
static <T> InstanceFactory<T> findFactoryFor(Class<T> type) {
        () -> findOnlyConstructorAnnotated(type, JdbiConstructor.class),
        () -> findOnlyConstructorAnnotated(type, ConstructorProperties.class)
        () -> findOnlyStaticMethodAnnotated(type, JdbiConstructor.class))
        .orElseGet(() -> getOnlyConstructor(type));
}
```

core/src/test/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapperTest.java
```diff
@@ -298,4 +299,36 @@ public void nestedPrefixParametersStrict() {
...
+            assertThat(pass).isEqualTo(42); // Make sure this constructor is not used directly
+            this.s = s;
+            this.i = i;
+        }
+
+        @JdbiConstructor
+        static StaticFactoryMethodBean create(String s, int i) {
+            return new StaticFactoryMethodBean(s, i, 42);
+        }
+    }
```
qualidafial: Can we a test for multiple static methods with the `@JdbiConstructor` annotation?

findepi:
> I think github is showing me a messed up diff for ConstructorMapper with badly associated changes so I'm not entirely sure what happened there (did you change anything aside from the field rename?)

@TheRealMarnes in that file there are these changes:
- change field type & name -- i can revert field name change, then the diff will be more pleasant. Should I?
- change `.of(...)` methods implementations (different delegation chain now). I can extract this to a separate preparatory commit, but i don't think this would improve readability
- `InstantiationException` removed from `catch` clause at the end

core/src/main/java/org/jdbi/v3/core/mapper/reflect/JdbiConstructors.java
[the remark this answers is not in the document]
TheRealMarnes: I like it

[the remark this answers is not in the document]
findepi: @TheRealMarnes i added this (slightly modified version, without `Stream#map`)

findepi:
shall i squash the commits?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
```diff
@@ -185,13 +184,13 @@ public T map(ResultSet rs, StatementContext ctx) throws SQLException {
 
         RowMapper<T> mapper = specialize0(ctx, columnNames, columnNameMatchers, unmatchedColumns)
             .orElseThrow(() -> new IllegalArgumentException(String.format(
-                UNMATCHED_CONSTRUCTOR_PARAMETERS, constructor)));
+                UNMATCHED_CONSTRUCTOR_PARAMETERS, factory)));
```
qualidafial: These error message constants should be updated to say "constructor or static factory method" instead of just "constructor."

core/src/main/java/org/jdbi/v3/core/mapper/reflect/InstanceFactory.java
```diff
@@ -0,0 +1,51 @@
...
+    public Parameter[] getParameters() {
+        return executable.getParameters();
+    }
+
+    @Nullable
+    public <A extends Annotation> A getAnnotation(Class<A> annotationClass) {
+        return executable.getAnnotation(annotationClass);
+    }
+
+    public abstract T newInstance(Object... params);
```
qualidafial: Nit: since class is package private, all the above methods should be package private too

core/src/main/java/org/jdbi/v3/core/mapper/reflect/JdbiConstructors.java
qualidafial: Let's use `may` in each message except for line 96.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/JdbiConstructors.java
```diff
@@ -27,6 +28,31 @@ private JdbiConstructors() {
         throw new UnsupportedOperationException("utility class");
     }
 
+    static <T> InstanceFactory<T> findFactoryFor(Class<T> type) {
```
qualidafial: Let's make this method public, document it, and deprecate `findConstructorFor(Class)`

qualidafial:
On the question of traditional vs squash merges, I tend to prefer traditional merges, primarily because squash merges change the git history and make my git think that a local branch hasn't been merged, even though the same changes have been merged in a squash commit.

It's a soft preference though.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/InstanceFactory.java
findepi: as per https://github.com/jdbi/jdbi/pull/1316#discussion_r236820179, i made this class public

findepi:
AC

core/src/main/java/org/jdbi/v3/core/mapper/reflect/InstanceFactory.java
qualidafial: Ah, I didn't think this one through--didn't mean to expose `InstanceFactory` as API. In my head I was thinking it would return an `Executable` but that's probably pointless since it has to be cast to `Constructor` or `Method` to invoke it.

Let's make this class package private again, and hide that static method in `JdbiConstructors like it was before. Sorry for the churn.

findepi:
@qualidafial AC (https://github.com/jdbi/jdbi/pull/1316#discussion_r237178396)

qualidafial:
Thanks for contributing!

PR #1318, opened 2018-11-24
Support generic Map<String, V> mapping

TheRealMarnes:
After like a year in limbo, here it finally is.

I remember seeing an issue somewhere where at least 2 people were asking for this, but for the life of me I can't seem to find it anymore...


TheRealMarnes:
Oh btw, I did not register this factory by default so that it would not be a behavioral change. That's also why the class is public. In jdbi4 I think we should hide it and make it installed by default, as per the TODO.

core/src/main/java/org/jdbi/v3/core/result/ResultBearing.java
```diff
@@ -171,6 +171,18 @@ static ResultBearing of(Supplier<ResultSet> resultSetSupplier, StatementContext
...
 
+    /**
+     * Maps this result set to a {@link Map} of {@link String} and the given value class.
+     *
+     * @param valueType the class to map the result set columns to
+     * @return a {@link Map} of String and the given type.
+     * @see Configurable#registerColumnMapper(ColumnMapper)
+     */
+    @Beta
+    default <T> ResultIterable<Map<String, T>> mapToGenericMap(Class<T> valueType) {
```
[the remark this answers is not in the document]
TheRealMarnes: sounds good

core/src/main/java/org/jdbi/v3/core/mapper/GenericMapMapperFactory.java
```diff
@@ -0,0 +1,136 @@
...
+
+                    for (int i = 0; i < columnNames.size(); i++) {
+                        @SuppressWarnings("unchecked")
+                        X value = (X) columnMapper.map(r, i + 1, ctx);
+                        row.put(columnNames.get(i), value);
+                    }
+
+                    return row;
+                };
+            } else {
```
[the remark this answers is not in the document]
stevenschlansker: +1, using `isPresent` is a code smell.

core/src/main/java/org/jdbi/v3/core/mapper/GenericMapMapperFactory.java
```diff
@@ -0,0 +1,136 @@
...
+                String name = columnLabel == null ? columnName : columnLabel;
+                name = caseChange.apply(name);
+
+                boolean added = names.add(name);
+                if (!added) {
+                    throw new RuntimeException("column \"" + name + "\" appears twice in this resultset!");
+                }
+            }
+
+            return new ArrayList<>(names);
```
stevenschlansker: You could avoid a list copy here if you return it as just a `Collection` and iterate over it; you already built it as a linked list.

core/src/main/java/org/jdbi/v3/core/result/ResultBearing.java
```diff
@@ -27,17 +38,6 @@
 import org.jdbi.v3.core.statement.StatementContext;
 import org.jdbi.v3.meta.Beta;
 
-import java.lang.annotation.Annotation;
```
stevenschlansker: We really need to ensure everyone is using a common import order, but until we get there, let's try to avoid unnecessary shuffling.  It detracts from the PR and causes code churn.

core/src/main/java/org/jdbi/v3/core/result/ResultBearing.java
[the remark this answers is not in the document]
stevenschlansker: Should this also possibly accept a `GenericType<T>`?

sqlobject/src/test/java/org/jdbi/v3/sqlobject/GenericMapMapperFactoryTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+public class GenericMapMapperFactoryTest {
+    private static final String QUERY = "select 1.0 as one, 2.0 as two, 3.0 as three from (values(null))";
```
stevenschlansker: You can get rid of the `from ...` clause entirely.

core/src/main/java/org/jdbi/v3/core/mapper/GenericMapMapperFactory.java
```diff
@@ -0,0 +1,136 @@
...
+        Type valueType = maybeValueType.get();
+
+        // ... that is not Object...
+        if (valueType == Object.class) {
+            return Optional.empty();
+        }
+        Optional<ColumnMapper<?>> maybeColumnMapper = config.get(ColumnMappers.class).findFor(valueType);
+
+        // ... for which there is a ColumnMapper
+        return maybeColumnMapper.map(t -> forValueType(valueType));
```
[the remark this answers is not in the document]
TheRealMarnes: Can't believe I didn't think of it

core/src/main/java/org/jdbi/v3/core/mapper/GenericMapMapperFactory.java
TheRealMarnes: need to iterate and `get` by index though

core/src/main/java/org/jdbi/v3/core/result/ResultBearing.java
TheRealMarnes: I've been thinking recently I'd like to give an auto-formatter plugin a try.

qualidafial: Consider this my vote for Google java style guide

core/src/main/java/org/jdbi/v3/core/result/ResultBearing.java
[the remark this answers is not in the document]
TheRealMarnes: API-wise, should the generictype be `Map<String, V>` or `V`?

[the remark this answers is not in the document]
TheRealMarnes: I don't entirely follow with the 4 different ways an API user can specify a type to map to so I'll probably get it wrong somewhere...

TheRealMarnes:
Handled everything that was mentioned :)

I'm a little flabberghasted by how the generics ended up working out, but if it works, it works...

core/src/main/java/org/jdbi/v3/core/mapper/GenericMapMapperFactory.java
stevenschlansker: Yes, it is a little weird.  But so is an unnecessary copy :)

stevenschlansker:
:shipit:

core/src/main/java/org/jdbi/v3/core/mapper/GenericMapMapperFactory.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private static class GenericMapMapper<T> implements RowMapper<Map<String, T>> {
+        private final ColumnMapper<T> mapper;
+
+        private GenericMapMapper(ColumnMapper<T> mapper) {
+            this.mapper = mapper;
+        }
+
+        private static GenericMapMapper<?> wildcard(ColumnMapper<?> mapper) {
```
stevenschlansker: This is a bit of an oddball that seems to exist only to *throw away* generic information.  How come it's not `<T> GMM<T> of(CM<T> mapper)`?

TheRealMarnes: image

It was the cleanest solution possible. I spent a lot of time shuffling code and parameters around yesterday, this is the only form I was acceptably happy with (the fewest ugly hacks and fake generics).

TheRealMarnes: I don't find `.map((Function<ColumnMapper<?>, ? extends GenericMapMapper<?>>) GenericMapMapper::new)` any better tbh

TheRealMarnes: After adding that in, I inlined the intermediate factory method (turning it into 2 different `findFor` calls again), and with that I can all of a sudden use the constructor instead of the factory method again, and basically all casts and stuff are gone.

Much confusion. ¯\\\_(ツ)\_/¯

TheRealMarnes:
Thanks for the extra pushes that turned out quite right, guys. Seems I'm out of my groove a little...

core/src/main/java/org/jdbi/v3/core/mapper/GenericMapMapperFactory.java
TheRealMarnes: Well I mean, the final result now makes perfect sense, but I wonder why it gave me so many errors along the way, half of which didn't make much sense, that made me write so much roundabout code in the first place...

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
```diff
@@ -31,8 +31,9 @@
  * keys are column names, while map values are the values in those columns. Map keys are converted to lowercase by
  * default.
  *
- * @see GenericMapMapperFactory for generics support (i.e. to get {@code Map<String, X>})
+ * @deprecated see {@link GenericMapMapperFactory} for generics support (i.e. to get {@code Map<String, T>})
```
qualidafial: I'm not sure I agree with deprecating. `Map<String, T>` means that all columns are the same type, but `Map<String, Object>` means the columns could all be different types.

docs/src/adoc/index.adoc
```diff
@@ -3025,8 +3025,11 @@ is a single `Map.Entry`. However, what if the `Map` we want to return is
 actually a single row or even a single column?
 
 Jdbi's link:{jdbidocs}/core/mapper/MapMapper.html[MapMapper^] maps each row to a
-`Map<String, Object>`, where column names (converted to lowercase) are mapped to
-column values.
+`Map<String, Object>`, where column names are mapped to column values.
+
+[NOTE]
+Jdbi's default setting is to convert Map keys to lowercase. This behavior can be
```
qualidafial: Can we phrase this as "convert column names to lowercase for Map keys"

docs/src/adoc/index.adoc
```diff
@@ -3040,6 +3043,19 @@ type is treated as a single value instead of a collection:
...
 
+Starting with Jdbi 3.6.0, there's the
+link:{jdbidocs}/core/mapper/GenericMapMapper.html[GenericMapMapper^],
+which provides the same feature but allows value types other than `Object`
+as long as a suitable ColumnMapper is registered:
+
+[source,java]
+----
+@SqlQuery("select 1.0 as LOW, 2.0 as MEDIUM, 3.0 as HIGH")
+@RegisterRowMapper(GenericMapMapper.class)
```
qualidafial: Needs to register the factory, not the mapper

core/src/main/java/org/jdbi/v3/core/mapper/MapMapper.java
TheRealMarnes: Oh yes, that's true

TheRealMarnes:
Latest comments addressed, thanks ^^'

core/src/main/java/org/jdbi/v3/core/mapper/GenericMapMapperFactory.java
```diff
@@ -0,0 +1,120 @@
...
+            .filter(map -> String.class.equals(GenericTypes.findGenericParameter(map, Map.class, 0).orElse(null)))
+            .flatMap(map -> GenericTypes.findGenericParameter(map, Map.class, 1))
+            .filter(value -> !Object.class.equals(value))
+            .flatMap(config.get(ColumnMappers.class)::findFor)
+            .map(GenericMapMapper::new);
+    }
+
+    // invoked manually or by fluent api
+    @Beta
+    public static <T> GenericMapMapper<T> getMapperForValueType(Class<T> valueType, ConfigRegistry config) {
```
qualidafial: This should return `RowMapper<String, T>`

core/src/main/java/org/jdbi/v3/core/mapper/GenericMapMapperFactory.java
```diff
@@ -0,0 +1,120 @@
...
+        return config.get(ColumnMappers.class)
+            .findFor(valueType)
+            .map(GenericMapMapper::new)
+            .orElseThrow(() -> new RuntimeException("no column mapper found for type " + valueType));
+    }
+
+    // invoked manually or by fluent api
+    @Beta
+    public static <T> GenericMapMapper<T> getMapperForValueType(GenericType<T> valueType, ConfigRegistry config) {
```
qualidafial: Ditto

core/src/main/java/org/jdbi/v3/core/mapper/GenericMapMapperFactory.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.function.UnaryOperator;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericType;
+import org.jdbi.v3.core.generic.GenericTypes;
+import org.jdbi.v3.core.statement.StatementContext;
+import org.jdbi.v3.meta.Beta;
+
+@Beta
+// TODO jdbi4: integrate with MapMapper? check if registering by default is compatible with other Map-mapping features
+public class GenericMapMapperFactory implements RowMapperFactory {
```
qualidafial: Can we get javadoc for this class?

PR #1322, opened 2018-11-29
add Jool's Unchecked as lib, minimize irrelevant exception clutter

TheRealMarnes:
I recommend reviewing this commit by commit and skipping the first 3, which are just jool copypasta and cutting...


TheRealMarnes:
Well sure, but I have experience with neither of those, so it'll be an adventure.

TheRealMarnes:
This seems to be what we need: https://maven.apache.org/plugins/maven-shade-plugin/examples/class-relocation.html

Looks fairly simple actually.

stevenschlansker:
Per our conversation, let's keep our small existing class for now, and make readability / compatibility improvements as appropriate.  `jool` is a cool library but using a single utility class doesn't justify the added dependency.  Thanks!

PR #1323, opened 2018-11-30
Disable toolchain profile for ci to see if we can actually use our build matrix

stevenschlansker:
I'm sure there was some reason it's this way in the first place, but doesn't enabling the `toolchains` profile for ci ensure we run our Travis build repeatedly using the exact same JDK?

### DIFF
--- a/sqlobject/src/test/java/org/jdbi/v3/sqlobject/locator/TestSqlLocator.java
+++ b/sqlobject/src/test/java/org/jdbi/v3/sqlobject/locator/TestSqlLocator.java
@@ -32,7 +32,7 @@ public class TestSqlLocator {
     public PgDatabaseRule dbRule = new PgDatabaseRule().withPlugin(new SqlObjectPlugin());
 
     @Test
-    public void testLocateConfigDriven() {
+    public void testLocateConfigDriven() throws Exception {
         Jdbi jdbi = dbRule.getJdbi();
         jdbi.useHandle(h -> {
             h.execute("create table something (id int, name text)");
--- a/src/build/ci-install.sh
+++ b/src/build/ci-install.sh
@@ -2,6 +2,4 @@
 
 OPTS="-DskipTests=true -Dbasepom.check.skip-all=true -Dmaven.javadoc.skip=true -B"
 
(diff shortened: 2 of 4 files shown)


stevenschlansker:
@qualidafial do you have any idea why oracle would be giving us a 'not authorized'?  I can't see what credentials are used, which makes it hard to troubleshoot.

I can't believe this jar isn't available publicly.  Should we just drop Oracle support, or build it out of tree?  This breaks again and again and again.

stevenschlansker:
I'm moving it out of tree.  :middle_finger: oracle

stevenschlansker:
Ugh, that breaks building documentation.  So that's a little obnoxious.  I guess I'll give you a chance to see if you see something wrong with the configuration here, before I go and start breaking things?

stevenschlansker:
Great, now Travis agrees: master doesn't build on jdk11.

https://travis-ci.org/jdbi/jdbi/jobs/466297395
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project jdbi3-sqlobject: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/jdbi/jdbi/sqlobject/src/test/java/org/jdbi/v3/sqlobject/locator/TestSqlLocator.java:[48,38] unreported exception java.lang.Exception; must be caught or declared to be thrown
```

stevenschlansker:
Presto bingo, now it does.  What do you guys think?  I think there's little point of having the Travis build matrix build `openjdk11` unless we use the actual 11, not just toolchain it back to 9.  WDYT?

qualidafial:
Not having reviewed the code yet: I've always found the toolchain stuff a little confusing, so if you think we can get by without it I'm alright with that in principle.

qualidafial:
Do we also want to get rid of the `toolchains.xml` stuff?

qualidafial:
And the relevant plugins?

qualidafial:
Oops, I misread the context of this PR. Thought we were getting rid of toolchains altogether, but I see the goal here is just to get them out of the way for CI builds.

PR #1325, opened 2018-11-30
Remove OracleDB support with extreme prejudice.

### LINKED ISSUE #1324: Remove oracle module?
_rant incoming_

Setting up Maven to talk to the password-protected maven.oracle.com repository is a pain in the ass.

That's why everything related to Oracle database is swept behind a profile--so contributors don't have to go through a tedious setup procedure that is meaningless to them just to start contributing.

This frequently causes release errors because `mvn versions:set` will skip the Oracle module if the profile is not activated.

Not to mention, some asshole replaced the orai18n JAR on their repository with the corresponding POM, which forced us to comment out everything Oracle in our TravisCI build scripts.

I'm getting sick of dealing with this.

I propose permanently removing the `jdbi3-oracle12` artifact from the build.

If Oracle can't be arsed to put the JDBC drivers on Maven central, I don't see why we should bother supporting them.

qualidafial:
Fixes #1324


RELEASE_NOTES
```diff
@@ -1,14 +1,20 @@
...
       data types that map to the same Java type. e.g. VARCHAR, NVARCHAR, and Postgres MACADDR all
       map to String, but are bound and mapped with different JDBC APIs.
     - Support for NVARCHAR columns, using the @NVarchar qualifying annotation
     - Support for Postgres MACADDR columns, using the @MacAddr qualifying annotation.
     - Support for HSTORE columns, using the @HStore annotation
+  - Discontinued API
+    - Due to ongoing stability problems arising to Oracle's Maven servers, development on the
+      jdbi3-oracle12 artifact has been discontinued, sources removed from the
+      project, and the earth salted for good measure. For posterity, jdbi3-bom dependency
+      management pins jdbi3-oracle12 to version 3.5.2.
```
stevenschlansker: We can release it separately, so we don't necessarily have to call it "discontinued", we can just treat it as a totally separate artifact.

bom/pom.xml
```diff
@@ -91,7 +91,7 @@
             <dependency>
                 <groupId>org.jdbi</groupId>
                 <artifactId>jdbi3-oracle12</artifactId>
-                <version>${project.version}</version>
+                <version>3.5.2</version>
```
stevenschlansker: If we do decide to continue it as a separate project, should we just remove this management entirely rather than pinning it?

TheRealMarnes:
I agree with @stevenschlansker. It can be moved out to its own project with a dependency on jdbi, and occasionally fixed by good samaritans to keep working with jdbi updates.

Changes look great to me otherwise. Devguide seems to lose a lot of weight.

RELEASE_NOTES
qualidafial: Updated

bom/pom.xml
qualidafial: Removed

TheRealMarnes:
image

README.md
```diff
@@ -58,6 +52,7 @@ This project is licensed under the
 * **Henning Schmiedehausen (@hgschmie)**
 * **Matthew Hall (@qualidafial)**
 * **Artem Prigoda (@arteam)**
+* **Marnick L'Eau (@TheRealMarnes)**
```
TheRealMarnes: 😊

stevenschlansker:
https://www.youtube.com/watch?v=kPIdRJlzERo

PR #1326, opened 2018-12-03
Allow instantiation of ParsedSql and ParsedParameters

sarxos:
Hello @jdbi Team :)

This change changes `ParsedSql` constructor visibility from private to protected.

Why did I change this? Let me briefly describe.

We recently migrated from JDBI v2 to v3 and we are really astonished of the number of enhancements, but while migrating the code I encountered small blocker.

In our solution I use [`SqlParser`](http://jdbi.org/#_sqlparser) to process SQL queries from annotated [SQL objects](http://jdbi.org/#_sql_objects). This processing consist of some cheap operations (e.g. prepend DAO name at the query beginning) and some pretty expensive operations (i.e. parse whole SQL and rewrite schemas found in query). Due to expensive nature of SQL parsing resultant SQL should be cached.

What I would like to do is to cache intermediate SQL together with named parameters, e.g. have cache like this one:

```
intermediateSql = "select a, b, c from {SCHEMA}.test where a > :a", parameters = [ a ]
intermediateSql = "insert into {SCHEMA}.test (a) values (:a)", parameters = [ a ]
```

And have custom implementation of ParsedSql which can be created by:

```java
public class MyParser implements SqlParser {
  public ParsedSql parse(String sql, StatementContext ctx) {
    Item item = cache.computeIfAbsent(sql, s -> doExpensiveSqlParsing(s, ctx));
    String intermediateSql = item.intermediateSql;
    ParsedParameters parameters = item.parameters;
    String finalSql = replaceAll("{SCHEMA}", schema);
    return new ParsedSql(finalSql, parameters);
  }
```

Since constructor is private and because `ParsedSql` can be created only via `ParsedSql.Builder` which takes SQL in parts, the `ParsedSql` instance cannot be instantiated from fully rendered SQL and known parameters. I have to split it into parts and merge using parameters one-by-one. This is problematic because it makes `?` a magic string which cannot be used in queries (because it's used to split it).

With current solution I can only accept `?` being magic or cache `ParsedSql` objects per schema which cause the number of items in cache to raise because there is no limit for a number of applications (and thus the number of schemas located on DB server, due to multitenant environment where application databases are isolated from each other). This will cause cache to grow bigger together with a number of applications, e.g.:

```
select a, b, c from app0001.test where a > :a; names = [ a ]
select a, b, c from app0002.test where a > :a; names = [ a ]
select a, b, c from app0003.test where a > :a; names = [ a ]
select a, b, c from app0004.test where a > :a; names = [ a ]
insert into app0001.test (a, b, c) values (:a, :b, :c); names = [a, b, c]
insert into app0002.test (a, b, c) values (:a, :b, :c); names = [a, b, c]
insert into app0003.test (a, b, c) values (:a, :b, :c); names = [a, b, c]
insert into app0004.test (a, b, c) values (:a, :b, :c); names = [a, b, c]
etc...
```

Please let me know if this change (private to protected) is acceptable. The other option for me would be to have `Builder` more fluent so I can set both SQL and parameters at once, not in parts, e.g.:

```java
ParsedSql sql = ParsedSql.builder()
  .setSql(finalSql)
  .setParameters(parameterNames)
  .build();
```

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
@@ -13,11 +13,12 @@
  */
 package org.jdbi.v3.core.statement;
 
+import static java.util.Collections.unmodifiableList;
+
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
-
-import static java.util.Collections.unmodifiableList;
 
 /**
  * The parsed parameters from an SQL statement.
@@ -77,5 +78,37 @@ public class ParsedParameters {
             + "positional=" + positional
             + ", parameterNames=" + parameterNames
             + '}';
+    }
+
+    /**
+     * A static factory of named {@link ParsedParameters} instances.
+     * List given in argument must contain parameter names from the
+     * related statement. They must be bare names, free of SQL
+     * variable processing syntax such as a prefixed colon or other
+     * delimiters.
+     *
+     * @param names the parameter names from SQL statement
+     * @return New {@link ParsedParameters} instance
+     * @throws IllegalArgumentException if names list contains positional parameter
+     */
+    public static ParsedParameters named(List<String> names) {
+        if (names.contains(ParsedSql.POSITIONAL_PARAM)) {
+            throw new IllegalArgumentException("Named parameters "
+                + "list must not contain positional parameter "
+                + "\"" + ParsedSql.POSITIONAL_PARAM + "\"");
+        }
+        return new ParsedParameters(false, names);
+    }
+
+    /**
+     * A static factory of positional {@link ParsedParameters} instances.
+     * The count given in the argument must indicate how many positional
+     * parameters are available in the statement.
+     *
+     * @param count the number of positional parameters in statement
+     * @return New {@link ParsedParameters} instance
+     */
+    public static ParsedParameters positional(int count) {
+        return new ParsedParameters(true, Collections.nCopies(count, "?"));
     }
 }
--- a/core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
@@ -21,7 +21,7 @@ import java.util.Objects;
  * The SQL and parameters parsed from an SQL statement.
  */
 public class ParsedSql {
-    private static final String POSITIONAL_PARAM = "?";
+    static final String POSITIONAL_PARAM = "?";
 
     private final String sql;
     private final ParsedParameters parameters;
@@ -73,6 +73,26 @@ public class ParsedSql {
     }
 
     /**
+     * A static factory of {@link ParsedSql} instances. The statement
+     * may contain only positional parameters
+     * (the {@value #POSITIONAL_PARAM} character). If your SQL
+     * code contains named parameters (for example variables preceded
+     * by a colon) then you have to replace them with positional
+     * parameters and specify the mapping in the
+     * {@link ParsedParameters}. You cannot mix named and positional
+     * parameters in one SQL statement.
+     *
+     * @param sql the SQL code containing only positional parameters
+     * @param parameters the ordered list of named parameters, or positional parameters
+     * @return New {@link ParsedSql} instance
+     * @see ParsedParameters#positional(int)
+     * @see ParsedParameters#named(List)
+     */
+    public static ParsedSql of(String sql, ParsedParameters parameters) {
+        return new ParsedSql(sql, parameters);
+    }
+
+    /**
      * @return a new ParsedSql builder.
      */
     public static Builder builder() {
@@ -110,7 +130,7 @@ public class ParsedSql {
         public Builder appendPositionalParameter() {
             positional = true;
             parameterNames.add(POSITIONAL_PARAM);
-            return append("?");
+            return append(POSITIONAL_PARAM);
         }
 
         /**
@@ -123,7 +143,7 @@ public class ParsedSql {
         public Builder appendNamedParameter(String name) {
             named = true;
             parameterNames.add(name);
-            return append("?");
+            return append(POSITIONAL_PARAM);
         }
 
         /**
--- a/core/src/test/java/org/jdbi/v3/core/statement/TestParsedParameters.java
+++ b/core/src/test/java/org/jdbi/v3/core/statement/TestParsedParameters.java
@@ -1,0 +1,54 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.statement;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.util.Arrays;
+import java.util.List;
+import org.junit.Test;
+
+public class TestParsedParameters {
+
+    @Test
+    public void testFactoryNamedParameters() {
+
+        final List<String> names = Arrays.asList("a", "b", "c");
+        final ParsedParameters parameters = ParsedParameters.named(names);
+
+        assertThat(parameters).isNotNull();
+        assertThat(parameters.isPositional()).isFalse();
+        assertThat(parameters.getParameterCount()).isEqualTo(3);
+        assertThat(parameters.getParameterNames()).containsExactly("a", "b", "c");
+    }
+
+    @Test
+    public void testFactoryNamedAndPositionalParametersMix() {
+        assertThatThrownBy(() -> ParsedParameters.named(Arrays.asList("a", "b", "?")))
+            .isInstanceOf(IllegalArgumentException.class)
+            .hasMessage("Named parameters list must not contain positional parameter \"?\"");
+    }
+
+    @Test
+    public void testFactoryPositionalParameters() {
+
+        final ParsedParameters parameters = ParsedParameters.positional(3);
+
+        assertThat(parameters).isNotNull();
+        assertThat(parameters.isPositional()).isTrue();
+        assertThat(parameters.getParameterCount()).isEqualTo(3);
+        assertThat(parameters.getParameterNames()).containsOnly("?");
+    }
+}
--- a/core/src/test/java/org/jdbi/v3/core/statement/TestParsedSql.java
+++ b/core/src/test/java/org/jdbi/v3/core/statement/TestParsedSql.java
@@ -1,0 +1,91 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.statement;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Arrays;
+import java.util.List;
+import org.junit.Test;
+
+public class TestParsedSql {
+
+    @Test
+    public void testFactoryNamedParameters() {
+
+        final List<String> names = Arrays.asList("a", "b", "c");
+
+        final String sql = "insert into test (a, b, c) values (?, ?, ?)";
+        final ParsedParameters parameters = ParsedParameters.named(names);
+        final ParsedSql parsedSql = ParsedSql.of(sql, parameters);
+
+        assertThat(parsedSql).isNotNull();
+        assertThat(parsedSql.getSql()).isEqualTo(sql);
+        assertThat(parsedSql.getParameters().isPositional()).isFalse();
+        assertThat(parsedSql.getParameters().getParameterNames()).containsExactly("a", "b", "c");
+    }
+
+    @Test
+    public void testFactoryPositionalParameters() {
+
+        final String sql = "insert into test (a, b, c) values (?, ?, ?)";
+        final ParsedParameters parameters = ParsedParameters.positional(3);
+        final ParsedSql parsedSql = ParsedSql.of(sql, parameters);
+
+        assertThat(parsedSql).isNotNull();
+        assertThat(parsedSql.getSql()).isEqualTo(sql);
+        assertThat(parsedSql.getParameters().isPositional()).isTrue();
+        assertThat(parsedSql.getParameters().getParameterNames()).containsOnly("?");
+    }
+
+    @Test
+    public void testBuilderWithNamedParameters() {
+
+        final List<String> names = Arrays.asList("a", "b", "c");
+
+        final ParsedSql parsedSql = ParsedSql.builder()
+            .append("insert into test (a, b, c) values (")
+            .appendNamedParameter(names.get(0))
+            .append(", ")
+            .appendNamedParameter(names.get(1))
+            .append(", ")
+            .appendNamedParameter(names.get(2))
+            .append(")")
+            .build();
+
+        assertThat(parsedSql).isNotNull();
+        assertThat(parsedSql.getSql()).isEqualTo("insert into test (a, b, c) values (?, ?, ?)");
+        assertThat(parsedSql.getParameters().isPositional()).isFalse();
+        assertThat(parsedSql.getParameters().getParameterNames()).containsExactly("a", "b", "c");
+    }
+
+    @Test
+    public void testBuilderWithPositionalParameters() {
+
+        final ParsedSql parsedSql = ParsedSql.builder()
+            .append("insert into test (a, b, c) values (")
+            .appendPositionalParameter()
+            .append(", ")
+            .appendPositionalParameter()
+            .append(", ")
+            .appendPositionalParameter()
+            .append(")")
+            .build();
+
+        assertThat(parsedSql).isNotNull();
+        assertThat(parsedSql.getSql()).isEqualTo("insert into test (a, b, c) values (?, ?, ?)");
+        assertThat(parsedSql.getParameters().isPositional()).isTrue();
+        assertThat(parsedSql.getParameters().getParameterNames()).containsOnly("?");
+    }
+}


TheRealMarnes:
Hmm, I recall having had a very similar situation once. I gave your situation some thought from an API point of view, and then checked the solution I found once, and saw what I expected: I wrote a custom SqlParser to do it (based on existing jdbi sources)... but I can well understand that would be overkill development work if you just wanna be able to copy ParsedSqls.

I'm not entirely happy to see the constructor changed from private. It makes sense to be able to instantiate them more directly since all they are, technically at this time as well as conceptually, is a list of params and a string to bind them to. Modifying the builder doesn't feel like a clean fix to me because the builder serves to build it piece-wise, and setting the whole thing at once would be dissonant with that idea.

What I would be happier with is perhaps a second, different builder, which would actually be very naturally refactored to a simple factory method.

sarxos:
Hi @TheRealMarnes 

Thank you for your thoughts. A builder with factory method is indeed a cleaner option. Will upload new patchset tomorrow.

qualidafial:
Hang on a second--your problem is that you want your SQL parser to substitute in e.g. a schema or table name? If so, then you should be using `TemplateEngine` to do that substitution, and then you can use the default `SqlParser`

TheRealMarnes:
Oh yeah, that too. I was kinda thinking ahead to the more general case of "I have some advanced processing I need to do that doesn't fit in TemplateEngine or SqlParser, and I wanna cache the work", because things usually go beyond people's original examples anyway...

qualidafial:
I would try the below code, and profile it to confirm whether the rendering / parsing pipeline is really your bottleneck:

```java
SqlQuery("select a, b, c from <schema>.test where a > :a")
List<Test> getTestsAfter(@Bind("a") DateTime a, @Define("schema") String schema);
```

qualidafial:
Assuming none of the above solve your problem, I would be okay exposing static factory methods, e.g. `ParsedSql.of(String sql, ParsedParameters params)`, `ParsedParameters.of(String... parameterNames)`?

sarxos:
Hi @qualidafial,

Thank you for your suggestion. Unfortunately, using template engine was one of the options we wanted to start with from the beginning, but we dropped it because it requires template variable (in this case a schema) to be `@Define`d in annotated method of SQL object or passed down to `SqlStatement` via `.define(variable, value)` call. This is not always possible because applications running in our clusters usually don't care or don't know what is the schema they operate on or how it's constructed. Their SQL objects are implemented with assumption that they do not have to know it.

With template engine in use, a method like this:

```java
@SqlUpdate("insert into test (a) values (:a)")
void insert(@Bind("a") int a);
```

Would have to be modified to:

```java
@SqlUpdate("insert into <schema>.test (a) values (:a)")
void insert(@Define("schema") String schema, @Bind("a") int a);
```

And the number of such constructions in our code can be count in hundreds :( The change required to update all of them together with a code where these methods are invoked would be massive. That's why we dropped this solution.

I'm afraid it's a kind of architecture legacy we inherited over the years. In the past we had `Handle`s linked directly to the schema via `DBI` object which knew the schema, but with growing number of applications sharing the same DB server this will be a problem because at the lowest level the connections pool is obviously limited.

We above in mind, we came into conclusion that `SqlParser` approach will be the least API-impacting solution since framework can supply schemas into queries. By doing this neither SQL objects nor implementations where these objects are used, does not have to be changed at all.

sarxos:
@qualidafial,

Before introducing more changes I went to your [contributing](https://github.com/jdbi/jdbi/blob/master/CONTRIBUTING.md) doc and would like to clarify your proposals in regards to the statement from this document:

> We favor constructors — especially the default one — over factory methods where possible. Adding factory methods is not discouraged, but restricting the visibility of useful constructors without technical reason is.

And:

> Jdbi places serious emphasis on not breaking compatibility. Remember these simple rules and think twice before making any classes or class members public!

With above in mind I'm a little bit concerned about introducing public factory methods because I'm fully aware that both `ParsedParameters` and `ParsedSql` are core classes and you limited their constructors visibility on purpose.

My question in regards to above one is - shall these new factory methods be protected or are you ok with public ones? For example:

```java
public class ParsedSql {
  protected static ParsedSql of(String sql, ParsedParameters params) {
    // ...
  }
}
```

And:

```java
public class ParsedParameters {
  protected static ParsedParameters of(boolean positional, List<String> names) {
    // ...
  }
}
```

TheRealMarnes:
A protected factory method isn't much use if the constructor is private: you wouldn't be able to subclass it and you're not supposed to put your own classes in our packages, so you wouldn't be able to call it.

sarxos:
@TheRealMarnes, yes, you are not supposed to put own code in jdbi packages, but this is still doable. It will most likely discourage people from using these protected factories, but will not close the doors to those who are looking for any viable solution. I think it's still better than reflecting constructors and making them accessible, but in general I completely agree with your statement on such solution uselessness, however I had to ask this question, at least in regards to what is stated in contributing guide. I would also like to have them public and will start with that.

TheRealMarnes:
I think we'd rather open doors wider than we'd like to, than open them only a bit but so little that someone will force them open :)

sarxos:
@TheRealMarnes, indeed, I had to force them open too many times to not understand your point :) Please take a look at the code. I added some tests to cover new and existing code.

sarxos:
I would also like to replace "?" in

sarxos:
I would also like to replace "?" in 

https://github.com/jdbi/jdbi/blob/757df210ce8546c86204db9e7fc5af05fd8194a4/core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java#L132

And 

https://github.com/jdbi/jdbi/blob/757df210ce8546c86204db9e7fc5af05fd8194a4/core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java#L145

With `POSITIONAL_PARAM`, i.e.:

```java
return append(POSITIONAL_PARAM);
```

I guess with existing implementation Sonar or PMD detects these as "magic string".

TheRealMarnes:
I'm not sure that's functionally correct... The net result is the same, but in terms of expressing intent, I'm not sure that's what it should do. @qualidafial ?

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
```diff
@@ -78,4 +78,22 @@ public String toString() {
...
+     * value). The second argument is a list which contains parameters
+     * available in SQL. In case when positional parameters are used the
+     * names list should contain only positional parameters, or more
+     * strictly, a "?" characters.
+     *
+     * @param positional true if statement uses positional parameters, false otherwise
+     * @param names the parameter names from SQL statement
+     * @return New {@link ParsedParameters} object
+     */
+    public static ParsedParameters of(boolean positional, List<String> names) {
```
TheRealMarnes: I think API-wise (without considering functionality yet) this factory could be better: either the params are positional and you need a List (Set? LinkedSet?) of names, or they are positional and you need to know how many (as an int) there are. I'd split this into two, more expressive methods.

TheRealMarnes: I think API-wise (without considering functionality yet) this factory could be better: either the params are positional and you need a List (maybe a List/Set hybrid for uniqueness?) of names, or they are positional and you need to know how many (as an int) there are. I'd split this into two, more expressive methods.

core/src/test/java/org/jdbi/v3/core/statement/TestParsedParameters.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.Test;
+
+public class TestParsedParameters {
+
+    @Test
+    public void testFactoryNamedParameters() {
+
+        final List<String> names = new ArrayList<>();
```
TheRealMarnes: please use more concise methods like `Arrays.asList`

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
```diff
@@ -72,6 +72,25 @@ public String toString() {
             + '}';
     }
 
+    /**
+     * A static factory for {@link ParsedSql} object. It takes two
+     * arguments. The first one is SQL which must contain only
+     * positional parameters (i.e. the {@value #POSITIONAL_PARAM}
+     * characters). If your code contains named parameters (i.e.
```
TheRealMarnes: typo: character

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
```diff
@@ -72,6 +72,25 @@ public String toString() {
...
+     * the ones followed by a colon) then you have to replace them
```
TheRealMarnes: aren't they _preceded_ by a colon? genuinely asking

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
```diff
@@ -78,4 +78,22 @@ public String toString() {
...
+
+    /**
+     * A static factory for {@link ParsedParameters} objects. It takes two
+     * arguments. The first argument is a boolean flag which indicates if
+     * statement, to which this object refers to, is using positional
+     * (<code>true</code> value) or named parameters (a <code>false</code>
+     * value). The second argument is a list which contains parameters
+     * available in SQL. In case when positional parameters are used the
+     * names list should contain only positional parameters, or more
+     * strictly, a "?" characters.
```
TheRealMarnes: typo: character (but I suspect this javadoc will be changed anyway)

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
```diff
@@ -72,6 +72,25 @@ public String toString() {
             + '}';
     }
 
+    /**
+     * A static factory for {@link ParsedSql} object. It takes two
```
TheRealMarnes: typo: objects
"It takes two arguments" is kinda redundant to say

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
```diff
@@ -72,6 +72,25 @@ public String toString() {
...
+     * arguments. The first one is SQL which must contain only
+     * positional parameters (i.e. the {@value #POSITIONAL_PARAM}
+     * characters). If your code contains named parameters (i.e.
+     * the ones followed by a colon) then you have to replace them
+     * by positional parameters and store mapping as
```
TheRealMarnes: I'd reword this to something like
"and specify the mapping in the {@code parameters} (see _links to factory methods here_). You cannot mix named and positional parameters in one SQL statement."

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
```diff
@@ -72,6 +72,25 @@ public String toString() {
...
+     * {@link ParsedParameters} which is taken as the second
+     * argument. Please also note that you cannot mix named and
+     * positional parameters in SQL statement.
+     *
+     * @param sql the sql code containing only positional parameters
+     * @param parameters the list of named parameters (ordered)
```
TheRealMarnes: ", or an appropriate positional ParsedParameters object"

core/src/test/java/org/jdbi/v3/core/statement/TestParsedSql.java
```diff
@@ -0,0 +1,102 @@
...
+            .appendPositionalParameter()
+            .append(", ")
+            .appendPositionalParameter()
+            .append(")")
+            .build();
+
+        assertThat(parsedSql).isNotNull();
+        assertThat(parsedSql.getSql()).isEqualTo("insert into test (a, b, c) values (?, ?, ?)");
+        assertThat(parsedSql.getParameters().isPositional()).isTrue();
+        assertThat(parsedSql.getParameters().getParameterNames()).contains("?", "?", "?");
```
TheRealMarnes: `.containsOnly("?")`

core/src/test/java/org/jdbi/v3/core/statement/TestParsedSql.java
```diff
@@ -0,0 +1,102 @@
...
+        names.add("?");
+
+        final String sql = "insert into test (a, b, c) values (?, ?, ?)";
+        final ParsedParameters parameters = ParsedParameters.of(true, names);
+        final ParsedSql parsedSql = ParsedSql.of(sql, parameters);
+
+        assertThat(parsedSql).isNotNull();
+        assertThat(parsedSql.getSql()).isEqualTo(sql);
+        assertThat(parsedSql.getParameters().isPositional()).isTrue();
+        assertThat(parsedSql.getParameters().getParameterNames()).containsExactly("?", "?", "?");
```
TheRealMarnes: `.containsOnly("?")`

core/src/test/java/org/jdbi/v3/core/statement/TestParsedSql.java
```diff
@@ -0,0 +1,102 @@
...
+            .appendNamedParameter(names.get(1))
+            .append(", ")
+            .appendNamedParameter(names.get(2))
+            .append(")")
+            .build();
+
+        assertThat(parsedSql).isNotNull();
+        assertThat(parsedSql.getSql()).isEqualTo("insert into test (a, b, c) values (?, ?, ?)");
+        assertThat(parsedSql.getParameters().isPositional()).isFalse();
+        assertThat(parsedSql.getParameters().getParameterNames()).contains("a", "b", "c");
```
TheRealMarnes: `containsExactly` is preferred

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
sarxos: Having `Set` instead of a list is a no-go IMO because it can't be accessed by index. In regards to your other comments I may suggest the following:

```java
 public static ParsedParameters positional(int count)
 public static ParsedParameters named(List<String> names)
```

Where `positional(int count)` will create list with a given size and fill it with "?" characters. Please let me know if this is ok.

core/src/test/java/org/jdbi/v3/core/statement/TestParsedParameters.java
sarxos: Will do.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Will do.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Yes, you are correct. My bad. Will fix this.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
sarxos: Will do.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Will do.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
TheRealMarnes: Yeah those are the factory methods I meant.

Set wouldn't work, but e.g. NavigableSet or LinkedHashSet would. I have a feeling the other guys will say to just use a List instead of those exotic types though, so it's fine.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Will do.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Will change to "the ordered list of named or positional parameters". Is this ok?

core/src/test/java/org/jdbi/v3/core/statement/TestParsedSql.java
sarxos: Will do.

core/src/test/java/org/jdbi/v3/core/statement/TestParsedSql.java
sarxos: Will do.

core/src/test/java/org/jdbi/v3/core/statement/TestParsedSql.java
sarxos: I'm surprised I didn't use it :)

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
TheRealMarnes: Come to think of it, uniqueness probably isn't even wanted there because the same name should be able to appear there multiple times, shouldn't it?

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
TheRealMarnes: "the ordered list of named parameters, or positional parameters" then, to avoid dubious grammar

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
sarxos: Yes. Same name can appear multiple times on a list and parameter index is mapped to its position in SQL. That's why `Set` cannot be used in the first place.

sarxos: Done.

core/src/test/java/org/jdbi/v3/core/statement/TestParsedParameters.java
sarxos: Done.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Done.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Done.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Done.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Done.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Done.

core/src/test/java/org/jdbi/v3/core/statement/TestParsedSql.java
sarxos: Done.

core/src/test/java/org/jdbi/v3/core/statement/TestParsedSql.java
sarxos: Done.

core/src/test/java/org/jdbi/v3/core/statement/TestParsedSql.java
sarxos: Done.

sarxos:
@TheRealMarnes Updated code is ready for review.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
```diff
@@ -78,4 +79,28 @@ public String toString() {
             + ", parameterNames=" + parameterNames
             + '}';
     }
+
+    /**
+     * A static factory of named {@link ParsedParameters} instances.
+     * List given in argument must contain parameter names from SQL
+     * (i.e. variables preceded by a colon).
```
TheRealMarnes: I think this could be dubiously interpreted as having to include the colon in the names contained in the list

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
```diff
@@ -78,4 +79,28 @@ public String toString() {
...
+     * @return New {@link ParsedParameters} instance
+     */
+    public static ParsedParameters named(List<String> names) {
+        return new ParsedParameters(false, names);
+    }
+
+    /**
+     * A static factory of positional {@link ParsedParameters} instances.
+     * The count given in the argument must indicate how many positional
+     * parameters is available in SQL.
```
TheRealMarnes: *are available in the statement

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
```diff
@@ -72,6 +72,26 @@ public String toString() {
             + '}';
     }
 
+    /**
+     * A static factory of {@link ParsedSql} instances. The first
+     * argument, i.e. SQL, must contain only positional parameters
+     * (i.e. the {@value #POSITIONAL_PARAM} character). If your SQL
+     * code contains named parameters (i.e. variables preceded by a
+     * colon) then you have to replace them with a positional
```
TheRealMarnes: *with positional

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
sarxos: I suggest:

> A static factory of named {link ParsedParameters} instances. List given in argument must contain parameter names from SQL. Names from a list must be stripped of a colon which stands before variable in unprocessed SQL.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
sarxos: Will change.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Will change.

sarxos:
@TheRealMarnes I addressed your last comments. Please let me know if latest version is ok. In case of any additional issues I can fix them tomorrow.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
TheRealMarnes: If I'm not mistaken, there's only talk of a colon at all because that's just the syntax jdbi uses by default with the ColonPrefixSqlParser, not because of a jdbc standard or anything. So the text should be agnostic about that.

> List must contain parameter names from the related statement. They must be "bare" names, independent of SQL variable processing syntax such as a prefixed colon or other delimiters.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
```diff
@@ -72,6 +72,26 @@ public String toString() {
             + '}';
     }
 
+    /**
+     * A static factory of {@link ParsedSql} instances. The first
+     * argument, i.e. SQL, must contain only positional parameters
+     * (i.e. the {@value #POSITIONAL_PARAM} character). If your SQL
+     * code contains named parameters (i.e. variables preceded by a
```
TheRealMarnes: this should actually be `e.g.`, not `i.e.`, because the colon is merely a jdbi default, not a hard fact in this context

qualidafial:
Wow, I missed quite a lot of discussion since yesterday. Still trying to parse all the discussion, but I do want to respond to this:

> Thank you for your suggestion. Unfortunately, using template engine was one of the options we wanted to start with from the beginning, but we dropped it because it requires template variable (in this case a schema) to be @Defined in annotated method of SQL object or passed down to SqlStatement via .define(variable, value) call. This is not always possible because applications running in our clusters usually don't care or don't know what is the schema they operate on or how it's constructed.

You can call `define()` on any implementation of `Configurable`, including `Jdbi` or `Handle`.

Also, if you are defining your own template engine, you are not restricted just to defined attributes, although that is the primary use case.

The template engine is passed the statement context, which means you could create your own JdbiConfig implementation and access it via `ctx.getConfig(MyConfigType.class)`.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
qualidafial: e.g. stands for "exempli gratia," meaning "for example"

i.e. stands for "id est," meaning "in other words"

:)

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
```diff
@@ -78,4 +79,29 @@ public String toString() {
...
+    /**
+     * A static factory of positional {@link ParsedParameters} instances.
+     * The count given in the argument must indicate how many positional
+     * parameters are available in SQL.
+     *
+     * @param count the number of positional parameters in SQL statement
+     * @return New {@link ParsedParameters} instance
+     */
+    public static ParsedParameters positional(int count) {
+        return new ParsedParameters(true, Collections.nCopies(count, "?"));
```
qualidafial: Neat, I didn't know about the `nCopies` method

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
```diff
@@ -72,6 +72,26 @@ public String toString() {
             + '}';
     }
 
+    /**
+     * A static factory of {@link ParsedSql} instances. The first
+     * argument, i.e. SQL, must contain only positional parameters
```
qualidafial: More to the point, `?` is a magic token in JDBC.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
TheRealMarnes: is that to correct me or informational for sarxos? I think it should be e.g. because being preceded by a colon is just one possible syntax, no?

Maybe the first i.e. in the paragraph can be replaced by "the statement" too, to go easier on the mind...

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
```diff
@@ -78,4 +79,29 @@ public String toString() {
...
+     * A static factory of named {@link ParsedParameters} instances.
+     * List given in argument must contain parameter names from SQL.
+     * Names from a list must be stripped of a colon which stands
+     * before variable in unprocessed SQL.
+     *
+     * @param names the parameter names from SQL statement
+     * @return New {@link ParsedParameters} instance
+     */
+    public static ParsedParameters named(List<String> names) {
+        return new ParsedParameters(false, names);
```
qualidafial: Since this method routes around the the usual builder interface, we should validate that `names` does not contain `POSITIONAL_PARAM`

sarxos: Done.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: I understand this is just a note. Not something I should describe in javadoc?

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
sarxos: Indeed. It's really nice because it does not create collection filled with N elements, but a singleton list where every index from 0..N-1 points the same element and effectively reduce memory consumption.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: These "i.e." and "e.g." was removed from javadoc.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
sarxos: Added `if` statement which throws `IllegalArgumentException` if `names.contains("?")`

sarxos:
New patchset is ready for review.

sarxos:
Hi @qualidafial,

In regards to your comment on custom `TemplateEngine`.

> Also, if you are defining your own template engine, you are not restricted just to defined attributes, although that is the primary use case.

I briefly tested this approach and it initially looked better, but it failed on corner cases (e.g. string template placeholders in SQL objects/methods) because our parser used to substitute schemas in tables is only able to process valid SQL :( To w/a this I would have to remove SQL parser from equation and rework all our SQLs instead, to put `<schema>` before each table, but I do not want to because of the two reasons:

1. Our SQL parser does some other things as well.
2. With custom `TemplateEngine` this is much less work when compared to `@Define("schema")` solution, but still a lot of.

For now I think we will stick to the custom `SqlParser` because it addresses all sunny and rainy day scenarios together with some corner cases I was able to identify. With custom `TemplateEngine` only a subset of these is resolved.

sarxos:
@TheRealMarnes I addressed your last comments.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
```diff
@@ -78,4 +79,35 @@ public String toString() {
...
+     * Names from a list must be stripped of a colon which stands
+     * before variable in unprocessed SQL.
+     *
+     * @param names the parameter names from SQL statement
+     * @return New {@link ParsedParameters} instance
+     * @throws IllegalArgumentException if names list contains positional parameter
+     */
+    public static ParsedParameters named(List<String> names) {
+        if (names.contains(ParsedSql.POSITIONAL_PARAM)) {
+            throw new IllegalArgumentException("Named parameters "
```
TheRealMarnes: I think `UnableToCreateStatementException` would be better here

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
```diff
@@ -78,4 +79,35 @@ public String toString() {
...
+                + "list must not contain positional parameter "
+                + ParsedSql.POSITIONAL_PARAM);
```
TheRealMarnes: enclose the character in quotes, please

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
```diff
@@ -21,7 +21,7 @@
  * The SQL and parameters parsed from an SQL statement.
  */
 public class ParsedSql {
-    private static final String POSITIONAL_PARAM = "?";
+    protected static final String POSITIONAL_PARAM = "?";
```
TheRealMarnes: This can be default/package access instead of protected

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
```diff
@@ -78,4 +79,35 @@ public String toString() {
...
+        }
+        return new ParsedParameters(false, names);
+    }
+
+    /**
+     * A static factory of positional {@link ParsedParameters} instances.
+     * The count given in the argument must indicate how many positional
+     * parameters are available in SQL.
```
TheRealMarnes: *available in the statement

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
```diff
@@ -78,4 +79,35 @@ public String toString() {
             + ", parameterNames=" + parameterNames
             + '}';
     }
+
+    /**
+     * A static factory of named {@link ParsedParameters} instances.
+     * List given in argument must contain parameter names from SQL.
```
TheRealMarnes: *List must contain parameter names from the related statement. They must be bare names, free of SQL variable processing syntax such as a prefixed colon or other delimiters.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
```diff
@@ -72,6 +72,26 @@ public String toString() {
             + '}';
     }
 
+    /**
+     * A static factory of {@link ParsedSql} instances. The first
```
TheRealMarnes: *The statement may contain only

sarxos: Done.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
sarxos: Done.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
sarxos: Done.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
sarxos: Yes, it can be, but does not have to, because this variable is static. There is no value added in having this default-accessible.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
sarxos: Why is that? It's not statement being created here.

TheRealMarnes: Well sort of. The exceptions indicate in roughly which part of the overall jdbi usage process something has gone wrong. If you disagree, I'm fine with passing this up, I think we're not really icky about changing exceptions at a later time.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
TheRealMarnes: The added value is that package-protected is less visible than protected (which is accessible to package + subclasses).

sarxos: But there are no subclasses of `ParsedSql` due to a private nature of its only constructor. This class cannot be derived.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
sarxos: What you stated is valid, however I do not think that changing exception class in this factory is a solution. From what I see in the code the process is started in `SqlStatement.internalExecute()` and involves the following steps: rendering, parsing, statement creation and customization, but what is interesting is the fact that only statement creation and customization is guarded by `try/catch` block which catch `SQLException` and wraps it with `UnableToCreateStatementException` (because it had to be caught).

If we consider the fact that custom implementations of `TemplateEngine`, `SqlParser`, `StatementCustomizer`, etc, are not enforced to throw `UnableToCreateStatementException`, then, with above in mind, and in regards to your comment, we can conclude that any custom implementation which throws non-statement creation exception, breaks indication you described. With existing impl this indication is only assumed - it works only when default components are used. Using custom components does not guarantee this indication at all.

I believe this could be fixed by doing general exception handling in `SqlStatement`.`internalExecute()` and wrapping exception thrown by custom implementations the same way `SQLException` is done. This would give us real, customization-resistant indication of where and why process failed.

I would like to leave this `IllegalArgumentException` as it is, but I'm completely ok if JDBI team disagree and decide change this later. I'm not really attached to my code :)

sarxos:
New patchset uploaded with fix for failing unit test.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
TheRealMarnes: Then why should be `protected` at all? :)

sarxos: For better readability.

core/src/main/java/org/jdbi/v3/core/statement/ParsedParameters.java
TheRealMarnes: Well yeah, but just because we can't control every external factor doesn't mean we should neglect to control our internal factors. But let's just leave it as it is now, it's not a big deal and I don't want to keep nitpicking on you over things that are of little import.

sarxos: Ok, marking this one as resolved.

sarxos:
Yet another patchset. This one is to make linter happy.

core/src/main/java/org/jdbi/v3/core/statement/ParsedSql.java
TheRealMarnes: I don't agree... Please change it. Then we'll merge.

sarxos: Done. I'm making this one as resolved.

sarxos:
Next patchset with protected static variable access modified to default one as per @TheRealMarnes request.

TheRealMarnes:
Excellent, thanks very much :)

🚢

TheRealMarnes:
Oh yeah, as usual, we all forgot about appending to RELEASE_NOTES.

You can submit a PR for that if you feel this change could be interesting to people. I won't be able to nitpick as much :)

sarxos:
@TheRealMarnes I will submit, but please don't be as much demanding as with this one. This is extremely demotivating.

TheRealMarnes:
I know, sorry, but the javadoc of our public API needs to be written in the most concise, semantically correct, and unambiguous english possible, and we try to strive for the team's and our own combined subjective idea of excellence in the finest details in code too.

Check my #1318 for example, my own code gets tons of remarks too :) They're not to bully or annoy you, we just want to avoid anything we may regret later, however unlikely it may be.

sarxos:
@TheRealMarnes I'm far from assuming it was to annoy me. You are just a very detailed reviewer, and this is nothing bad, it's good actually. Thank you for merging this!

qualidafial:
@sarxos Thank you for sticking with us and responding to all our concerns.

I think it's important to understand that whatever qualities of Jdbi 3 you enjoy are largely the result of the kind of process you've seen unfold in this PR.

I absolutely understand your frustration in having us keep coming back with more comments, and I appreciate your willingness to making adjustments as requested.

In the future, if you so choose, it might make the process a little faster if we had write access to your PR branch--there's a checkbox in the right-hand column when you open the PR. Then instead of sending feedback, waiting for the submitter to respond, reviewing again to ensure the changes are correct, etc, we could just tweak the PR, push, and merge.

PR #1331, opened 2018-12-09
boot out jexl test dep

TheRealMarnes:
jexl was added as a test dep to jdbi just to write a test that appears to verify it's possible to implement a statementcustomizer...

This change ditches the jexl dep (:tada:) and brings what I believe is a more informative test.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -244,18 +244,6 @@
             </dependency>
 
             <dependency>
-                <groupId>org.apache.commons</groupId>
-                <artifactId>commons-jexl</artifactId>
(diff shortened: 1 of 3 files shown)


TheRealMarnes:
@stevenschlansker ? You seem involved in this being added so I'd like your opinion...

stevenschlansker:
Didn't get much time to put an opinion on :P but this thing is so old, I have no idea why it's in there -- if I'm on the blame, it's more likely a rewrite / jdbi3 merge rather than me introducing it in the first place.

PR #1332, opened 2018-12-10
Spike: shade jdbi3-core dependencies

qualidafial:
Shaded dependencies:

* antlr-runtime
* guava
* expiringmap


vavr/src/main/java/org/jdbi/v3/vavr/VavrGenericMapUtil.java
```diff
@@ -64,13 +64,39 @@ private static Type resolveMaplikeEntryType(Type maplikeType, Type keyParam, Typ
     }
 
     private static Type resolveMapEntryType(Type keyType, Type valueType) {
-        return resolveMapEntryType(TypeToken.of(keyType), TypeToken.of(valueType));
-    }
+        return new ParameterizedType() {
```
qualidafial: I feel like this use case is asking for a `parameterizedType(rawType, typeParamers...)` method in `GenericTypes`

PR #1334, opened 2018-12-11
jackson2 module

stevenschlansker:
adds `@Json` type qualifier

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -18,6 +18,7 @@
     - Support for NVARCHAR columns, using the @NVarchar qualifying annotation
     - Support for Postgres MACADDR columns, using the @MacAddr qualifying annotation
     - Support for HSTORE columns, using the @HStore annotation
+    - @Json type qualifier with Jackson 2.0 bindings
   - Oracle DB support changes
     - Due to ongoing stability problems with Oracle's Maven servers, we have split the
       jdbi3-oracle12 artifact out of the main project, to a new home at
--- a/core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
@@ -1,0 +1,29 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.internal;
+
+public class JdbiClassUtils {
+    private JdbiClassUtils() {
+        throw new UtilityClassException();
+    }
+
+    public static boolean isPresent(String klass) {
+        try {
+            Class.forName(klass);
+            return true;
+        } catch (ReflectiveOperationException ignored) {
+            return false;
+        }
+    }
+}
--- a/jackson2/pom.xml
+++ b/jackson2/pom.xml
@@ -14,25 +14,16 @@
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-
     <parent>
         <artifactId>jdbi3-parent</artifactId>
         <groupId>org.jdbi</groupId>
         <version>3.6.0-SNAPSHOT</version>
     </parent>
 
-    <artifactId>jdbi3-postgres</artifactId>
-    <name>jdbi3 Postgres</name>
-    <description>jdbi PostgreSQL specific features</description>
+    <artifactId>jdbi3-jackson2</artifactId>
+    <name>jdbi3 jackson2</name>
+    <description>Jdbi Jackson 2.0 integration</description>
     <url>http://jdbi.org/</url>
-
-    <licenses>
-        <license>
-            <name>Apache License 2.0</name>
-            <url>http://www.apache.org/licenses/LICENSE-2.0.html</url>
-            <distribution>repo</distribution>
-        </license>
-    </licenses>
 
     <dependencies>
         <dependency>
@@ -42,24 +33,21 @@
         <dependency>
             <groupId>org.jdbi</groupId>
             <artifactId>jdbi3-json</artifactId>
-            <optional>true</optional>
         </dependency>
 
         <dependency>
-            <groupId>com.google.code.findbugs</groupId>
-            <artifactId>jsr305</artifactId>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
         </dependency>
 
-        <dependency>
-            <groupId>org.postgresql</groupId>
-            <artifactId>postgresql</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.jdbi</groupId>
-            <artifactId>jdbi3-core</artifactId>
-            <classifier>tests</classifier>
-            <scope>test</scope>
-        </dependency>
         <dependency>
             <groupId>org.jdbi</groupId>
             <artifactId>jdbi3-testing</artifactId>
@@ -67,34 +55,24 @@
         </dependency>
         <dependency>
             <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-postgres</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-postgres</artifactId>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
             <artifactId>jdbi3-sqlobject</artifactId>
             <scope>test</scope>
         </dependency>
-
         <dependency>
             <groupId>com.opentable.components</groupId>
             <artifactId>otj-pg-embedded</artifactId>
             <scope>test</scope>
         </dependency>
-
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
-
-    <build>
-        <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
-                <configuration>
-                    <compilerArgs>
-                        <arg>-parameters</arg>
-                    </compilerArgs>
-                </configuration>
-            </plugin>
-        </plugins>
-    </build>
 </project>
--- a/jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Config.java
+++ b/jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Config.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.jackson2;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import org.jdbi.v3.core.config.JdbiConfig;
+import org.jdbi.v3.meta.Beta;
+
+/**
+ * Configuration class for Jackson 2 integration.
+ */
+@Beta
+public class Jackson2Config implements JdbiConfig<Jackson2Config> {
+    private ObjectMapper mapper;
+
+    public Jackson2Config() {
+        this.mapper = new ObjectMapper();
+    }
+
+    private Jackson2Config(Jackson2Config other) {
+        this.mapper = other.mapper;
+    }
+
+    /**
+     * Set the {@link ObjectMapper} to use for json conversion.
+     * @param mapper the mapper to use
+     * @return this
+     */
+    public Jackson2Config setMapper(ObjectMapper mapper) {
+        this.mapper = mapper;
+        return this;
+    }
+
+    /**
+     * @return the object mapper to use for json conversion
+     */
+    public ObjectMapper getMapper() {
+        return mapper;
+    }
+
+    @Override
+    public Jackson2Config createCopy() {
+        return new Jackson2Config(this);
+    }
+}
--- a/jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Plugin.java
+++ b/jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Plugin.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.jackson2;
+
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.json.Json;
+import org.jdbi.v3.json.JsonConfig;
+import org.jdbi.v3.json.internal.AbstractJsonPlugin;
+import org.jdbi.v3.meta.Beta;
+
+/**
+ * Jackson 2 integration plugin.
+ *
+ * Adds support for {@code @Json} qualifying annotation.
+ *
+ * @see Json
+ */
+@Beta
+public class Jackson2Plugin extends AbstractJsonPlugin {
+    @Override
+    public void customizeJdbi(Jdbi jdbi) {
+        jdbi.getConfig(JsonConfig.class).setImpl(new JacksonJsonImpl());
+        super.customizeJdbi(jdbi);
+    }
+}
--- a/jackson2/src/main/java/org/jdbi/v3/jackson2/JacksonJsonImpl.java
+++ b/jackson2/src/main/java/org/jdbi/v3/jackson2/JacksonJsonImpl.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.jackson2;
+
+import java.io.IOException;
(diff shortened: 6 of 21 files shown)


jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Config.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.jdbi.v3.core.config.JdbiConfig;
+import org.jdbi.v3.meta.Beta;
+
+/**
+ * Configuration class for Jackson 2 integration.
+ */
+@Beta
+public class Jackson2Config implements JdbiConfig<Jackson2Config> {
+    private ObjectMapper mapper;
```
qualidafial: Do we want to default this to a `new ObjectMapper()` for basic out of the box functionality?

jackson2/src/test/java/org/jdbi/v3/jackson2/TestJackson2Plugin.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    interface JsonDao {
+        @SqlUpdate("insert into whozits (whozit) values(?)")
+        int insert(@Json Whozit value);
+
+        @SqlQuery("select whozit from whozits")
+        @Json
+        List<Whozit> select();
+    }
```
qualidafial: I'd also like to see some tests with a `@Json` property of e.g. a bean class.

jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Config.java
stevenschlansker: I waffled a bit on that one.  Personally, I always configure my Mappers, so this prevents me from accidentally forgetting to configure it.  It also means whatever mapper caching Jackson does is shared with the rest of the application.  I'm happy to change this if anyone feels strongly, though.

qualidafial: I usually configure the module 2, but for those who don't need to I'd rather have it "just work" out fo the box.

An alternative to having a default is to no-arg constructor to the plugin that uses a `new ObjectMapper()`

jackson2/src/main/java/org/jdbi/v3/jackson2/JsonArgumentFactory.java
```diff
@@ -0,0 +1,39 @@
...
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.statement.UnableToCreateStatementException;
+
+@Json
+class JsonArgumentFactory implements ArgumentFactory {
+    @Override
+    public Optional<Argument> build(Type type, Object value, ConfigRegistry config) {
+        return Optional.of((p, s, c) -> {
+            try {
+                s.setObject(p, config.get(Jackson2Config.class).getMapper().writeValueAsString(value), Types.OTHER);
```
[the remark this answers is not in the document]
stevenschlansker: Agreed.  I only use Postgres and I didn't want to make the initial pass overly complicated.

jackson2/src/test/java/org/jdbi/v3/jackson2/TestJackson2Plugin.java
```diff
@@ -0,0 +1,88 @@
...
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+import org.jdbi.v3.testing.JdbiRule;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TestJackson2Plugin {
+// tag::jackson2example[]
```
qualidafial: Dropping the whole unit test into the docs doesn't exactly convey how the plugin should be used.

Not a blocker for merge, but I think users would find more value in a conversational "first you add the dep: <example>, then you install the plugin: <example>, then if you annotate your bound properties or SQL objects like this: <example> then code like this Just Works :tm: <example>

stevenschlansker:
I addressed all the review comments.

jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Plugin.java
```diff
@@ -0,0 +1,54 @@
...
+    public Jackson2Plugin(ObjectMapper mapper) {
+        if (mapper == null) {
+            throw new IllegalArgumentException("Jackson 2 integration requires an ObjectMapper");
+        }
+        this.mapper = mapper;
+    }
+
+    @Override
+    public void customizeJdbi(Jdbi jdbi) {
+        if (mapper != null) {
```
qualidafial: We're already checking for non-null in the constructor

jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Config.java
TheRealMarnes: I think it feels more in line with how the rest of jdbi works if we default this to a `new`

jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Config.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * @return the object mapper to use for json conversion
+     */
+    public ObjectMapper getMapper() {
+        return mapper;
+    }
+
+    @Override
+    public Jackson2Config createCopy() {
+        return new Jackson2Config().setMapper(mapper);
```
TheRealMarnes: perhaps a copy constructor to be consistent with all other Config classes? Would be a shame to divert from the pattern and have to refactor it later anyway.

jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Plugin.java
```diff
@@ -0,0 +1,54 @@
...
+@Beta
+public class Jackson2Plugin implements JdbiPlugin {
+
+    private final ObjectMapper mapper;
+
+    public Jackson2Plugin() {
+        this(new ObjectMapper());
+    }
+
+    public Jackson2Plugin(ObjectMapper mapper) {
```
TheRealMarnes: kinda feels like this is different from how the other plugins work, delegating all their configuration to Config classes alone instead of having special constructors that will eventually become a burden when more configuration is added

qualidafial: I'm ambivalent on this one.

I do agree that it's a departure from how other plugins work, so maybe it's worth making it consistent for consistency's sake.

qualidafial: @TheRealMarnes by "ugly mess" do you mean overloading constructors for more config parameters? I don't see that happening. The `ObjectMapper` is the focal point of this module so it's kind of special in that sense.

TheRealMarnes: That's what they always say at first :)

But yeah, the sum of _maybe_ wanting to add more in the future because of expansion and the deviation from an established pattern, makes me say we shouldn't do this.

RELEASE_NOTES
```diff
@@ -11,6 +11,7 @@
     - ParsedSql can be created with ParsedSql.of(String, ParsedParameters) factory
     - ParsedParameters can be created with ParsedSql.positional(int) and
       ParsedSql.named(List<String>) factories.
+    - @Json type qualifier with Jackson 2.0 bindings
```
qualidafial: Should move this to the beta API section

jackson2/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>${jackson.version}</version>
```
qualidafial: Should we manage this in the parent POM instead?

jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Plugin.java
```diff
@@ -0,0 +1,52 @@
...
+ * Adds support for {@code @Json} qualifying annotation.
+ *
+ * @see Json
+ */
+@Beta
+public class Jackson2Plugin implements JdbiPlugin {
+
+    private final ObjectMapper mapper;
+
+    public Jackson2Plugin() {
```
qualidafial: Should we produce a `META-INF/services` SPI file for this plugin?

jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Plugin.java
qualidafial: On reflection I think I agree with Marnes on this.

We can always change our mind in the future and put this constructor back, but we can't take it back once it's published API.

stevenschlansker: I never intended to continue adding new constructors if there's any more configuration to do.  I only had this constructor because it really is an integral part of the integration, and currently there's no need for further configuration - so you don't need to know about the config class at all.  I guess consistency is an argument, but I do think ObjectMapper is integral to this module in a way that e.g. changing null binding or configuring case mapping isn't.

TheRealMarnes:
I'll just TLDR here: why don't you make the Json mapping itself an impl-agnostic core feature with an ObjectMapper-ish interface, and let Jackson only be a separate module that provides a Plugin that registers the mapper instance? Then users could just as well use Gson or whatever other json mapper impl, anything that can (de)serialize a String and Type.

stevenschlansker:
I considered doing that, as well as factoring out the strategy we use to bind Postgres json.  But I was really just trying to get the feature out the door and in use before making it overly general.  Do we think anyone will actually contribute Gson support?  I'm not opposed or disagreeing, it's just that you're asking for the scope of this to increase a lot...

TheRealMarnes:
Well if you made the json mapping a core feature with an SPI, adding any backing impl like jackson or gson would be 5 lines of code max... I'm not intent on adding gson so much as I am worried about unnecessarily tying a generic json feature to a single specific library that can't be substituted by another when there's really no need for such a hard dependency.

TheRealMarnes:
The change could actually be as simple as taking the core code out of this module into a jdbi3-json, and moving the jackson implementation code to a jdbi3-jackson module with a dependency on jdbi3-json. Then users would add jdbi3-jackson or another impl, or jdbi3-json and roll their own.

stevenschlansker:
So if I added a new module `jdbi-annotations`, moved the `@Json` type qualifier there, but left the only implementation as jackson2 -- is that good enough?

stevenschlansker:
Ah, I guess it could be a `json` module as well.  We're getting so many modules...

TheRealMarnes:
Making jdbi3-json as a separate module is entirely optional, but I think we should start scaling horizontally instead of tacking more and more into core. I don't really see added modules as a burden, there are only so many integrations we can provide.

TheRealMarnes:
But yeah, I would make the entire "this data should be handled and mapped as json" part of this PR an agnostic feature on itself (either in core or a new module), and provide the "... by jackson" part of the sentence as an SPI implementation to delegate the String/Type to only.

jackson2/pom.xml
stevenschlansker: I'd put it here since no other modules should ever depend on jackson, so there is no reason for them to inherit its management.  Do you think we should put it in the parent anyway?

TheRealMarnes: I think it's good practice to always put dependency versions in the parent regardless of usage location, just to avoid ever getting different versions of the same thing or spreading version management out over multiple places.

jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Plugin.java
TheRealMarnes: I really don't see this constructor as added value over the config class, especially because we can provide a default anyway...

If we go ahead with splitting jackson out of the feaure itself as discussed further below though, then this constructor would become a necessity: you could have a `new JacksonPlugin()` or `new GsonPlugin()`, but would require a `new JsonPlugin(mapperImpl)`.

jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Config.java
stevenschlansker: It's all internal and can be changed on a whim with no problem.  Why is the copy constructor important?  I'm experimenting a bit and don't you agree that having named chained setters is usually preferable to long constructors?  (You were advocating *against* constructors with long parameter lists in this very review!)

jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Plugin.java
stevenschlansker: I did this.

stevenschlansker:
I split the implementation from the interface, but the changes are significant enough that the code needs re-review.

jackson2/pom.xml
stevenschlansker: Right, and that's exactly what I'd imagined here -- someday there will be a Jackson 3 and that will be a new `jackson3` module, and you don't want the `jackson2` dependency management creeping in.  Or someone (foolishly) decides to contribute a `jackson1` module.  So would you consider this such an exception?

TheRealMarnes: Well you could leave it in the parent because there's no (expectation of a) competing version module _yet_, or you could move it inside the module preemptively... The difference though is that with the former you'll notice the necessary change at dev time, while with the latter another jackson version might be added to another module for whatever reason and then there'll be conflicts _at runtime_ for users. It feels like opening the door to an (unlikely) unwelcome surprise. I would prefer it in the parent until there's a competing module, but it's not like I'm gonna make a big deal out of it if you feel strongly and make the decision to not do that ¯\\\_(ツ)\_/¯

core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
```diff
@@ -0,0 +1,29 @@
...
+public class JdbiClassUtils {
+    private JdbiClassUtils() {
+        throw new UtilityClassException();
+    }
+
+    public static boolean isPresent(String klass) {
+        try {
+            Class.forName(klass);
+            return true;
+        } catch (ReflectiveOperationException e) {
```
TheRealMarnes: should be named `ignored` to satisfy the "log or throw but not both nor neither" rule

TheRealMarnes:
I'm happy with the code, but I'm wondering if there's a language barrier here. Do we have the same idea of what a copy constructor is? To me it is a constructor like `private X(X other) { this.foo = other.foo; etc}`, is it that to you too? Did you begrudgingly put regular all-arg constructors on the configs because you thought I was asking for that, or did you make the decision not to do copy constructors as I see them and jsut go with a regular constructor?

stevenschlansker:
There must be some misunderstanding.  I thought I was doing exactly as you outlined:
https://github.com/jdbi/jdbi/pull/1334/commits/83790281d3b234f42b5df5aace13dcb9930d4638
```diff
+    private JsonConfig(JsonMapperImpl impl) {
+        this.impl = impl;
+    }
```

TheRealMarnes:
That explains the friction I felt :)

No actually I'm in favor of copy constructors as I outlined them just now, i.e. not how they are now implemented nor with the setters from before. Working on the other JdbiConfigs showed those are the best solution, and I suspect that you might agree. :)

stevenschlansker:
I don't know for sure that I agree, but since I don't feel strongly, I'm willing to take your assurance that it's better :)

stevenschlansker:
I'm still somewhat sad about the loss of `Jackson2Module(ObjectMapper)` but it's only a small wart to have to immediately configure it.  I still can't shake the feeling that it's 'essential' configuration in the way most of the other configurables aren't.  But I'm inclined to merge this as is (without it) and maybe add it later if need be.

jackson2/src/main/java/org/jdbi/v3/jackson2/Jackson2Config.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+@Beta
+public class Jackson2Config implements JdbiConfig<Jackson2Config> {
+    private ObjectMapper mapper;
+
+    public Jackson2Config() {
+        this(new ObjectMapper());
+    }
+
+    private Jackson2Config(ObjectMapper mapper) {
```
TheRealMarnes: so yeah let's change this real quick

json/src/main/java/org/jdbi/v3/json/JsonConfig.java
```diff
@@ -0,0 +1,46 @@
...
+import org.jdbi.v3.json.internal.JsonMapperImpl;
+import org.jdbi.v3.meta.Beta;
+
+@Beta
+public class JsonConfig implements JdbiConfig<JsonConfig> {
+    private JsonMapperImpl impl;
+
+    public JsonConfig() {}
+
+    private JsonConfig(JsonMapperImpl impl) {
```
TheRealMarnes: and this one too

stevenschlansker:
I think there's a strong argument that exposing `Jackson2Config` is a heavier weight API than another constructor.  But I'm more interested in focusing on shipping this PR right now than worrying about this issue.

TheRealMarnes:
I'm curious, what do you see now that wasn't there before? Something about the language barrier, or did eclipse give you the same additional support as intellij when changing the constructors to the new kind?

stevenschlansker:
I just totally missed your point.

TheRealMarnes:
Hah, ok. Could I have done something better?

TheRealMarnes:
And yeah, true, the additional exposed API is much heavier than the constructor, but none of that is redundant. Redundant API exposure feels 1000x heavier to me than new API exposure. But whatever, add the constructor in the future if people ask or something :)

stevenschlansker:
I was just frustrated with how many times around this PR is taking and misunderstood.  If you give me a green checkmark all will be forgiven ;)

stevenschlansker:
I squished it.

TheRealMarnes:
I got the same remark a few days ago from sarxos, but I'm very decidedly with both feet in the "rather safe than sorry" camp. I'll gladly concede something I request in a review if given a firm no, but as long as something is up for debate, I'll keep fighting until I get a hard counterargument. I already feel like crap for having added the MessageFormatTemplateEngine without knowing until now about the integer format issue. In fact I feel stupid not having picked up over all these years that String.format's `%s` is exactly the string formatting tool I've been looking for all this time after all, always having used MessageFormat instead to get around an issue that was purely a misunderstanding.

stevenschlansker:
That's the right attitude to have.

TheRealMarnes:
Could have just asked me to do that with the button, we wouldn't have to wait for another build 😛

stevenschlansker:
Oops, my finger slipped ;)

jackson2/src/main/java/org/jdbi/v3/jackson2/JacksonJsonImpl.java
```diff
@@ -0,0 +1,55 @@
...
+    public String toJson(Type type, Object value, StatementContext ctx) {
+        try {
+            return getMapper(ctx).writeValueAsString(value);
+        } catch (JsonProcessingException e) {
+            throw new UnableToCreateStatementException(e, ctx);
+        }
+    }
+
+    @Override
+    public <T> T fromJson(Type type, String json, StatementContext ctx) {
```
qualidafial: With the `Type` parameter, the best this method can promise to return is `Object`

stevenschlansker: You're right, I'll fix this.

stevenschlansker: https://github.com/jdbi/jdbi/pull/1344

json/src/main/java/org/jdbi/v3/json/JsonConfig.java
```diff
@@ -0,0 +1,46 @@
...
+public class JsonConfig implements JdbiConfig<JsonConfig> {
+    private JsonMapperImpl impl;
+
+    public JsonConfig() {}
+
+    private JsonConfig(JsonConfig other) {
+        this.impl = other.impl;
+    }
+
+    public JsonConfig setImpl(JsonMapperImpl impl) {
```
qualidafial: What do you guys think of renaming the interface to `JsonMapper` and naming this config property `jsonMapper`?

TheRealMarnes: Sounds good to me

postgres/src/main/java/org/jdbi/v3/postgres/JsonArgumentFactory.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.argument.ArgumentFactory;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.json.Json;
+
+class JsonArgumentFactory implements ArgumentFactory {
+    @Override
+    public Optional<Argument> build(Type type, Object value, ConfigRegistry config) {
+        if (type != Json.class) {
```
qualidafial: What's happening here? `Json.class` isn't the type, is the qualifier. Shouldn't this be an annotation on the factory class declaration?

stevenschlansker: This is the "poor man's SPI" to decouple Postgres support from Json support.
The Json binder matches parameterized types e.g. `@Json Foo`, makes them a Json string, then re-binds them as the type `Json.class` (bit of a hack, but it works!)

It expects the database engine, in this case Postgres, to provide the specific binding strategy.

I guess this could be an ArgumentFactory setting on the config instead if that's clearer.  But we already have an argument factory registry...

qualidafial: It sounds like `QualifiedArgumentFactory` wants to be API.

PR #1335, opened 2018-12-11
Remove Guava reflection code, use GeAnTyRef instead!

stevenschlansker:
https://github.com/leangen/geantyref


sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
```diff
@@ -57,7 +60,7 @@ static ResultReturner forOptionalReturn(Class<?> extensionType, Method method) {
      * @return an instance that takes a ResultIterable and constructs the return value
      */
     static ResultReturner forMethod(Class<?> extensionType, Method method) {
-        Type returnType = GenericTypes.resolveType(method.getGenericReturnType(), extensionType);
+        Type returnType = GenericTypeReflector.getReturnType(method, extensionType);
```
qualidafial: If this can't keep using `GenericTypes.resolveType()` then this is a breaking change.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlCallHandler.java
```diff
@@ -21,13 +21,15 @@
...
+import io.leangen.geantyref.GenericTypeReflector;
+
 public class SqlCallHandler extends CustomizingStatementHandler<Call> {
     private final boolean returnOutParams;
 
     public SqlCallHandler(Class<?> sqlObjectType, Method method) {
         super(sqlObjectType, method);
 
-        Type returnType = GenericTypes.resolveType(method.getGenericReturnType(), sqlObjectType);
+        Type returnType = GenericTypeReflector.getReturnType(method, sqlObjectType);
```
qualidafial: ditto

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlUpdateHandler.java
```diff
@@ -42,7 +43,7 @@ public SqlUpdateHandler(Class<?> sqlObjectType, Method method) {
         boolean isGetGeneratedKeys = method.isAnnotationPresent(GetGeneratedKeys.class);
 
         QualifiedType returnType = QualifiedType.of(
-            GenericTypes.resolveType(method.getGenericReturnType(), sqlObjectType),
+            GenericTypeReflector.getReturnType(method, sqlObjectType),
```
qualidafial: Ditto

vavr/src/main/java/org/jdbi/v3/vavr/VavrGenericMapUtil.java
```diff
@@ -57,20 +56,13 @@ static Type resolveMultimapEntryType(Type mapType) {
...
     private static Type resolveMapEntryType(Type keyType, Type valueType) {
-        return resolveMapEntryType(TypeToken.of(keyType), TypeToken.of(valueType));
-    }
-
-    private static <K, V> Type resolveMapEntryType(TypeToken<K> keyType, TypeToken<V> valueType) {
-        return new TypeToken<Tuple2<K, V>>() {}
-                .where(new TypeParameter<K>() {}, keyType)
-                .where(new TypeParameter<V>() {}, valueType)
-                .getType();
+        return TypeFactory.parameterizedClass(Tuple2.class, keyType, valueType);
```
qualidafial: This is another argument in favor of re-exposing this method through e.g. `GenericType.parameterized(Class, Type...)`

qualidafial: In fact this whole class could go away.

core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
```diff
@@ -108,11 +104,25 @@ private GenericTypes() {
...
      * @return the resolved type.
+     * @deprecated use {@link #resolveType(TypeVariable, Type)} instead
      */
+    @SuppressWarnings("unchecked")
+    @Deprecated
     public static Type resolveType(Type type, Type contextType) {
-        return TypeToken.of(contextType)
-                .resolveType(type)
-                .getType();
+        return resolveType((TypeVariable<? extends Class<?>>) type, contextType);
```
qualidafial: We need to have a backup plan here if `type` isn't a `TypeVariable`.

stevenschlansker: 100% of the uses so far are of `TypeVariable`s, there are no uses of this method.  I only left it so as not to break the public API.  I think it might have been a mistake to have it take `Type` in the first place not `TypeVariable`.

stevenschlansker: Oh, or maybe this is my place to have it emulate the behavior of Guava.

qualidafial: > 100% of the uses so far are of `TypeVariable`s

I assumed that the only reason for the changes in calculating return types in SQL Object was because this stopped working under GeAnTyRef (god that's awful to type out)

qualidafial:
I disagree. The whole point was to give users some common primitive reflection operations so they could reason about types in factories.

vavr/src/main/java/org/jdbi/v3/vavr/VavrGenericMapUtil.java
stevenschlansker: Nice

stevenschlansker:
I think I figured out how to make it compatible!

stevenschlansker:
Ok, I don't have a fix yet, but I at least understand where the difference lies.
When you ask the GenericTypeReflector to map type parameters, it doesn't look at the supertype's type declarations.  `class Dao extends BaseDao<String>` loses the mapping from T -> String.  I wonder if this is a bug, maybe I'll file an issue...

core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
```diff
@@ -108,11 +105,37 @@ private GenericTypes() {
...
-                .resolveType(type)
-                .getType();
+        if (type instanceof TypeVariable) {
+            return resolveType((TypeVariable<? extends Class<?>>) type, contextType);
+        }
+        do {
+            try {
+                return GenericTypeReflector.mapTypeParameters(GenericTypeReflector.annotate(type), GenericTypeReflector.annotate(contextType)).getType();
+            } catch (RuntimeException e) {
+                if (contextType instanceof Class<?>) { // XXX HACK
```
stevenschlansker: This bit is obviously not right

core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
stevenschlansker: Yeah, you're right, I am wrong.  I made that change because the name is nicer but didn't realize it was hiding this behavior change.

stevenschlansker:
OK @qualidafial sorry for the confusion here and thanks for reviewing all my stuff, I think I've figured out what's going on, and it's due to a subtle incorrectness in the original version.  The invocation `GenericTypes.resolveType(method.getGenericReturnType(), MethodOwner.class)` does not provide sufficient information to `GenericTypes` to correctly resolve all possible generic return types.

Consider
```java
class BaseMethody<T> {
    <S extends T> S get();
}

class Methody extends BaseMethody<String> {
    
}
```

If you `resolveType(Methody.class.getMethod("get").getGenericReturnType(), Methody.class);`, the generic return type is `S`.  But when you attempt to resolve that in the context of `Methody.class`, there is no declared type variable `S` -- it's on the *method*. And we've lost the method reference (since we only pass in the return type).  So the combination of method generic return type and declaring class is not sufficient to resolve potential return types, and I believe that's where this API mismatch is coming from.

I suspect we never hit this case since the `<S extends T>` return type is nonsense to jdbi in most cases, but I'm fairly sure this is an indication that our current `GenericTypes.resolveType` is not quite right.  Does this make sense to you, or am I spouting more nonsense?

stevenschlansker:
If nothing else, maybe this is justification that deprecating the existing method and replacing it with the `TypeVariable` version is actually a reasonable change, since the original method signature allows a class of unanswerable invocations.

stevenschlansker:
The other difference arises because Guava's type introspector code considers supertypes when it looks for the type mapping, whereas geantyref does not.  I'll open an issue and see what the maintainer thinks.

qualidafial:
In your example, I would expect to get back a type variable `S extends String`.

But I think you've missed my point. SQL Object is already using the existing `resolveType(Type type, Type contextType)` to resolve (possibly parameterized) generic return types to resolve the specific return type.

I've seen this used to write a generic base DAO:

```java
@UseClasspathSqlLocator // subclass provides all .sql files
interface BaseDao<T, ID> {
  @SqlUpdate
  @GetGeneratedKeys
  T insert(@BindBean T entity);

  @SqlQuery
  T getById(ID id);

  @SqlUpdate
  @GetGeneratedKeys
  T update(@BindBean T entity);

  @SqlUpdate
  void deleteById(ID id);
}
```

Which can then be subclassed for each entity class:

```java
@RegisterBeanMapper(User.class)
interface UserDao extends BaseDao<User, UUID> {
  // additional methods if desired
}
```

If we can't modify the underlying implementation `GenericTypes.resolveType()` without leaving existing usage intact, then we've broken it--for us, and for outside users of that API.

stevenschlansker:
I agree, I want to handle that case.  There's a candidate fix here: https://github.com/leangen/geantyref/pull/5

stevenschlansker:
> In your example, I would expect to get back a type variable S extends String.

Unfortunately, you don't.  You get back a `TypeVariable[S]`, the bound `S extends String` is only implied by `Methody` and it's not correct to return `S extends String` from a Method of the base class, so the core reflection doesn't find it.  And the link that would let us figure out the puzzle, `<S extends T>`, is only on the method declaration (neither the generic return type nor the implementing class, which are the two parameters we are given).

qualidafial:
Now I'm curious how Guava is able to untangle it.

stevenschlansker:
I am too.  I'm investigating that presently.

kaqqao:
Hello everyone :)

While working on figuring out what can be done to support your use case in geantyref, I also read through this issue to better understand it.

I still have to understand all the implications (especially around [captures](https://github.com/leangen/geantyref/wiki/CaptureType)), but here are some of my thoughts typed out, in case you find them useful:

For the given example:

```java
class BaseMethody<T> {
    <S extends T> S get();
}

class Methody extends BaseMethody<String> {
    
}

Method method = BaseMethody.class.getDeclaredMethod("get");
Type type = GenericTypeReflector.getReturnType(m, Methody.class);
```
Geantyref will give you the expected `S extends String`.

`TypeVaribale` maintains the reference to the element that declared it. So in this case
```java
((TypeVariableImpl) type).getGenericDeclaration()
```
will give you the reference to the `BaseMethody#get`.

kaqqao:
I added a feature that should satisfy your use-case in the current master branch of Geantyref.

The relevant methods are `GenericTypeReflector.resolveExactType` (which throws an exception if an unresolvable variable is encountered) and `GenericTypeReflector.resolveType` (which allows partial resolution).

If you get a chance, please give it a spin to verify it works as expected.
I'll release the next version of Geantyref with this feature in a couple of days, or as soon as you give me the thumbsup.

Cheers :)

kaqqao:
Geantyref version 1.3.7 released, I hope it will work for you. If there's anything, feel free to contact me :)

TheRealMarnes:
Thanks! Just rebased and updated, it seems you removed some API and made it public. I'm not into the details so I'll let steven figure out the necessary changes.

stevenschlansker:
@kaqqao I have integrated the new method with this PR.  It looks like it's good to go now!  Thank you very much :smile_cat:

stevenschlansker:
Thanks @qualidafial and @TheRealMarnes for your patience here, I think this PR is good to review now.  Good-bye vendored code!

TheRealMarnes:
@qualidafial review and merge, please

core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
```diff
@@ -116,9 +113,20 @@ private GenericTypes() {
      * @return the resolved type.
      */
     public static Type resolveType(Type type, Type contextType) {
-        return TypeToken.of(contextType)
-                .resolveType(type)
-                .getType();
+        return GenericTypeReflector.resolveType(GenericTypeReflector.annotate(type), GenericTypeReflector.annotate(contextType)).getType();
```
kaqqao: Please use `GenericTypeReflector.resolveType(Type, Type)` instead of wrapping first using `annotate`. The end result is the same, but it avoids one extra traversal through the type structure, so it should be somewhat faster.

Details, in case you care:
When calling `GenericTypeReflector.resolveType(AnnotatedType, AnnotatedType)`, the `contextType` first gets its generics expanded (so that each `Class` with type parameters gets wrapped into a `ParameterizedType`), and when calling `GenericTypeReflector.resolveType(Type, Type)` this is done at the same time the `Type` is wrapped into an `AnnotatedType`, so one traversal is skipped.

stevenschlansker: Thanks!

qualidafial:
Reviewing

core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
```diff
@@ -116,9 +113,20 @@ private GenericTypes() {
...
+
+    /**
+     * Resolves the <code>type</code> parameter in the context of <code>contextType</code>. For example, if
+     * <code>type</code> is <code>List.class.getMethod("get", int.class).getGenericReturnType()</code>, and
+     * <code>contextType</code> is <code>List&lt;String&gt;</code>, this method returns <code>String.class</code>
+     * @param type the type to be resolved in the scope of <code>contextType</code>
+     * @param contextType the context type in which <code>type</code> is interpreted to resolve the type.
+     * @return the resolved type.
+     */
+    public static Type resolveType(TypeVariable<? extends Class<?>> type, Type contextType) {
```
qualidafial: I'm unclear on what this new method is doing. How do the semantics of this method vary from the existing one?

I'm also a little concerned that this new method overload takes over 7 of the 15 existing references to the existing `resolveType` method. Is this intentional?

Would you be open to renaming this method, e.g. to `resolveTypeVariable()` to prevent ambiguity?

core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
```diff
@@ -147,13 +155,22 @@ public static Type resolveMapEntryType(Type mapType) {
...
+     * For example: {@code parameterizedClass(Map.class, Integer.class, String.class)}
+     * returns the type {@code Map<Integer, String>}.
+     *
+     * @param clazz     Type class of the type to create
+     * @param arguments Type arguments for the variables of {@code clazz}, or null if these are not
+     *                  known.
+     * @return A {@link ParameterizedType}, or simply {@code clazz} if {@code arguments} is
+     * {@code null} or empty.
+     */
+    public static Type parameterizedClass(Class<?> clazz, Type... arguments) {
```
qualidafial: Nice!

Could we rename this to `parameterizeClass` (without the `d`) to make the method name contain a verb?

core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
kaqqao: For reference, the `resolveType(TypeVariable, Type)` overload shouldn't be needed. The usual `resolveType(Type, Type)` covers it. There is probably a small difference in performance, but shouldn't be anything relevant.

stevenschlansker: It came in one of the intermediate steps, before the new geantyref methods existed.  I'll remove it.

stevenschlansker:
@qualidafial thanks for the review, took care of your comments, rebased + squished on master.  Anything else?

core/src/main/java/org/jdbi/v3/core/generic/GenericTypes.java
```diff
@@ -147,13 +142,22 @@ public static Type resolveMapEntryType(Type mapType) {
...
+     * returns the type {@code Map<Integer, String>}.
+     *
+     * @param clazz     Type class of the type to create
+     * @param arguments Type arguments for the variables of {@code clazz}, or null if these are not
+     *                  known.
+     * @return A {@link ParameterizedType}, or simply {@code clazz} if {@code arguments} is
+     * {@code null} or empty.
+     */
+    public static Type parameterizeClass(Class<?> clazz, Type... arguments) {
+        return TypeFactory.parameterizedClass(clazz, arguments);
```
qualidafial: I :heart: this new method so much

stevenschlansker:
Hip Hip Hooray!

PR #1336, opened 2018-12-11
Pet peeve: UtilityClassException

TheRealMarnes:
Reusable semantic exceptions > generic exceptions with a repetitive message

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/internal/UtilityClassException.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/UtilityClassException.java
@@ -13,15 +13,8 @@
  */
 package org.jdbi.v3.core.internal;
 
-import java.util.Optional;
-import java.util.stream.Stream;
-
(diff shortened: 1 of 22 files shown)


core/src/main/java/org/jdbi/v3/core/internal/UtilityClassException.java
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.internal;
+
+public class UtilityClassException extends UnsupportedOperationException {
+    public UtilityClassException() {
+        super("utility class");
```
qualidafial: We should ```throw new UtilityClassException()``` here instead :trollface:

PR #1337, opened 2018-12-11
Throwables -> Unchecked

TheRealMarnes:
Mirroring the jool api, only the parts we use (and of course everyone is welcome to add more as needed), with small changes to the way sneaky throwing mechanism to make it suit our standards more (returning a dummy exception and forcing the fake return value to be thrown).

Only thing we need to discuss is the TODO in Sneaky: https://github.com/jOOQ/jOOL/issues/230

On one hand I don't think jdbi should be deciding to interfere in such things in non-reversible ways, and the InterruptedException isn't _wrapped_ so it doesn't really count as a catch-point where standards presumably dictate setting the `interrupted` flag. On the other, it seems to be standard/expected/helpful to do so anyway. ¯\\\_(ツ)\_/¯


core/src/main/java/org/jdbi/v3/core/internal/exceptions/Sneaky.java
```diff
@@ -0,0 +1,63 @@
...
+
+        if (t instanceof IOException) {
+            throw new UncheckedIOException((IOException) t);
+        }
+
+        /* TODO original jool code, wat do? https://github.com/jOOQ/jOOL/issues/230
+        if (t instanceof InterruptedException) {
+            Thread.currentThread().interrupt();
+        }
+        */
```
qualidafial: If you catch an `InterruptException` that you cannot process, you _must_ call `Thread.currentThread().interrupt()` to re-interrupt the thread.

qualidafial: Good summary on https://praveer09.github.io/technology/2015/12/06/understanding-thread-interruption-in-java/

TheRealMarnes:
> who's gonna get the magic issue #1337
— @qualidafial , 2018

TheRealMarnes:
@qualidafial I uncommented the interruption part, but I also just added an idea I got at work, to unwrap InvocationTargetException. Yay/nay?

core/src/main/java/org/jdbi/v3/core/internal/exceptions/CheckedBiFunction.java
```diff
@@ -0,0 +1,19 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
qualidafial: We need to preserve the copyright statements from the original project.

qualidafial:
I like the unwrap idea.

PR #1341, opened 2018-12-14
add fair warning to MessageFormatTemplateEngine

TheRealMarnes:


stevenschlansker:
I never understood why it was there in the first place, really, so I'm fine with that.

TheRealMarnes:
Can this be merged with the note as-is then? I moved it back to the bottom as we usually do with `@deprecated`

Issue #1340, opened 2018-12-12
BOM is incompatible with Spring Boot 2.1


ivandeskov:
Hi guys,

I've just started exploring jdbi and it's a great library, thanks very much for your efforts!

I noticed a problem when using the bom with Spring Boot 2.1. Boot is bringing Spring 5.1.2 in and jdbi3-parent has declared Spring v4.2.4 in its dependency management section, so I ended up with spring-core and spring-context 4.2.4 on my classpath

The issue is that spring-context 4.2.4 is incompatible with boot 2.1 and this causes the following error at startup:

> Exception in thread "main" java.lang.AbstractMethodError: org.springframework.boot.context.config.ConfigFileApplicationListener.supportsSourceType(Ljava/lang/Class;)Z
	at org.springframework.context.event.GenericApplicationListenerAdapter.supportsSourceType(GenericApplicationListenerAdapter.java:79)
	at org.springframework.context.event.AbstractApplicationEventMulticaster.supportsEvent(AbstractApplicationEventMulticaster.java:282)
	at org.springframework.context.event.AbstractApplicationEventMulticaster.retrieveApplicationListeners(AbstractApplicationEventMulticaster.java:214)
	at org.springframework.context.event.AbstractApplicationEventMulticaster.getApplicationListeners(AbstractApplicationEventMulticaster.java:185)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:125)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:119)
	at org.springframework.boot.context.event.EventPublishingRunListener.environmentPrepared(EventPublishingRunListener.java:75)
	at org.springframework.boot.SpringApplicationRunListeners.environmentPrepared(SpringApplicationRunListeners.java:54)
	at org.springframework.boot.SpringApplication.prepareEnvironment(SpringApplication.java:347)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:306)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1260)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1248)
	at com.aexp.rcomply.rcomplyapi.CaiapiApplication.main(CaiapiApplication.java:11)

Please let us know if I can provide any further information about this.

Thanks!

TheRealMarnes:
[...]

PR #1342, opened 2018-12-14
compatibility note

TheRealMarnes:
Because of #1340

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -4084,6 +4084,10 @@ Then configure the Jdbi factory bean in your Spring container, e.g.:
     `@Inject` annotations on the target class instead of configuring it in
     your `beans.xml`.
 
+[NOTE]
+This module has been tested to be compatible
+with Spring 5.1.3 (used by Spring Boot 2.1.1) as well.
+
 ==== Installing plugins
 
 Plugins may be automatically installed by scanning the classpath for
--- a/spring5/pom.xml
+++ b/spring5/pom.xml
@@ -14,53 +14,59 @@
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-
     <parent>
         <artifactId>jdbi3-parent</artifactId>
         <groupId>org.jdbi</groupId>
         <version>3.6.0-SNAPSHOT</version>
     </parent>
 
-    <artifactId>jdbi3-spring4</artifactId>
-    <name>jdbi3 Spring 4</name>
-    <description>jdbi Spring 4 integration</description>
+    <artifactId>jdbi3-spring5</artifactId>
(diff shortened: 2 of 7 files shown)


docs/src/adoc/index.adoc
```diff
@@ -4046,6 +4046,10 @@ Then configure the Jdbi factory bean in your Spring container, e.g.:
     `@Inject` annotations on the target class instead of configuring it in
     your `beans.xml`.
 
+[NOTE]
+This module has been *experimentally* shown to be compatible
```
stevenschlansker: What additional validation would we need to do to drop the '*experimentally*' here?  Spring 5 is the GA version, if we can claim to support it, we should.

TheRealMarnes: don't know. It seems to me that if its basic usage works and everything compiles, it probably just works, but I'm not the expert on it so I'm not gonna state it as fact without a thorough test. I guess I would set up a jdbi-spring5 that _only_ contains tests and doublecheck tx handling and stuff there?

TheRealMarnes: On that note btw, what's up with #1252?

stevenschlansker: I'm not sure exactly.  Nobody in the core project understands Spring transaction management (or wants to use it) and it's a lot of code to merge that we don't really understand.  And it's not currently quite ready to ship anyway.  So it's stuck in limbo...

TheRealMarnes: Hmm, yeah, I see what you mean. The first half isn't so bad, but then you get to the real spring tx integration part...

TheRealMarnes:
Just pushed a stub of an idea I had, that currently doesn't work yet. Any idea if it's possible to just take the tests of another module and run them again in the current module as if they were a dependency? I thought the `<classifier>tests</classifier>` stuff would do just that...

spring5/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+        <version>3.6.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jdbi3-spring5</artifactId>
+    <name>jdbi3 Spring 5</name>
+    <description>jdbi Spring 4/5 forward compatibiility testing</description>
+    <url>http://jdbi.org/</url>
+
+    <properties>
+        <dep.spring.version>5.1.3.RELEASE</dep.spring.version>
```
qualidafial: Does this actually succeed in changing the dependency version?

TheRealMarnes: according to intellij > maven > show effective pom, it does

spring5/src/test/java/org/jdbi/v3/spring5/TestVersion.java
```diff
@@ -0,0 +1,14 @@
...
+import org.junit.Test;
+import org.springframework.core.SpringVersion;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TestVersion {
+    @Test
+    public void thisIsSpring5() {
+        assertThat(SpringVersion.getVersion())
+            .isEqualTo("5.1.3.RELEASE");
```
TheRealMarnes: @qualidafial

TheRealMarnes:
Ok, done. Turned out to be surprisingly easy. Tests extended so they're re-run, and overrode the spring version with a test to make sure.

ez pz

PR #1343, opened 2018-12-14
jdbi3-gson ;)

TheRealMarnes:
Because everyone needs a ~friend~ competitor.

I think jackson vs gson is a popular enough point of debate and matter of personal choice for people that it's worth supporting both, for people who prefer one over the other. Feel free to freeze or reject, it's a copypasteplementation, so not a lot of work.

### DIFF
--- a/gson2/pom.xml
+++ b/gson2/pom.xml
@@ -1,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+~   Licensed under the Apache License, Version 2.0 (the "License");
+~   you may not use this file except in compliance with the License.
+~   You may obtain a copy of the License at
+~
+~   http://www.apache.org/licenses/LICENSE-2.0
+~
+~   Unless required by applicable law or agreed to in writing, software
+~   distributed under the License is distributed on an "AS IS" BASIS,
+~   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+~   See the License for the specific language governing permissions and
+~   limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>jdbi3-parent</artifactId>
+        <groupId>org.jdbi</groupId>
+        <version>3.6.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>jdbi3-gson2</artifactId>
+    <name>jdbi3 gson2</name>
+    <description>Jdbi Gson 2.8 integration</description>
+    <url>http://jdbi.org/</url>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-json</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-core</artifactId>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-sqlobject</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-testing</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-json</artifactId>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-postgres</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
+            <artifactId>jdbi3-postgres</artifactId>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.opentable.components</groupId>
+            <artifactId>otj-pg-embedded</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/gson2/pom.xml
+++ b/gson2/pom.xml
@@ -12,15 +12,13 @@
 ~   See the License for the specific language governing permissions and
 ~   limitations under the License.
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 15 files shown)


gson2/src/test/java/org/jdbi/v3/gson2/TestGson2Plugin.java
```diff
@@ -0,0 +1,160 @@
...
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+import org.jdbi.v3.testing.JdbiRule;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TestGson2Plugin {
```
stevenschlansker: Should this test now be an abstract test superclass in `jdbi3-json` and have two implementing tests, one in `jackson` and one in `gson`?

TheRealMarnes: Hmm, I guess it should?

stevenschlansker: I think that's preferable to copy+pasting the code...

gson2/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+    <parent>
+        <artifactId>jdbi3-parent</artifactId>
+        <groupId>org.jdbi</groupId>
+        <version>3.6.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>jdbi3-gson2</artifactId>
+    <name>jdbi3 gson2</name>
+    <description>Jdbi Gson 2.8 integration</description>
```
stevenschlansker: Is this `2.0` or `2.x` integration, or is there something special about `2.8`?

stevenschlansker:
I think you're being a little unfair to Jackson here -- just because it behaves differently than Gson doesn't mean it's wrong, and there is absolutely no reason we should need to have polymorphic deserialization going on.

Take a look at https://github.com/jdbi/jdbi/pull/1343/commits/333a27ac246704585f39875713bd8bd5264a8ecc where I cleaned it up a bit and see if that's better.

The things to note are, first Jackson was not finding the parameter names because (1) it wasn't enabled in the `json` module compiler arguments and (2) you need the `jackson-module-parameter-names` addon (as Jackson 2.x predates parameter names by quite a while, it wasn't a built-in feature to start with).

Additionally, when you remove the `@JsonCreator` annotation, Jackson is forced to guess more about your class layout.  With a constructor e.g. `Nested1(int a)`, Gson assumes it's a single-property object, and Jackson assumes that it's a wrapper value for a primitive int.  I think that ambiguity is understandable, even if it's not what we want here, and I've fixed it by making both nested objects have two properties so it cannot be mistaken for a value type.

TheRealMarnes:
Shows how familiar I am with jackson, I guess :)

jackson2/src/test/java/org/jdbi/v3/jackson2/TestJackson2Plugin.java
```diff
@@ -13,167 +13,22 @@
...
-        @JsonCreator
-        public N2(@JsonProperty("b") String b) {
-            this.b = b;
-        }
-
-        public String getB() {
-            return b;
-        }
+    public void before() {
+        jdbi = db.getJdbi().installPlugin(new Jackson2Plugin());
```
stevenschlansker: This is where I wish I could do `jdbi = db.getJdbi().installPlugin(new Jackson2Plugin(new ObjectMapper().registerModule(new ParameterNamesModule())))` but I guess splitting it across two lines isn't so bad.

pom.xml
```diff
@@ -83,7 +84,7 @@
         <dep.antlr.version>3.4</dep.antlr.version>
         <dep.checkstyle.version>8.10</dep.checkstyle.version>
         <dep.dokka.version>0.9.17</dep.dokka.version>
-        <dep.jackson2.version>2.6.0</dep.jackson2.version>
+        <dep.jackson2.version>2.8.5</dep.jackson2.version>
```
stevenschlansker: didn't do parameter names before 2.8.5

TheRealMarnes:
I'm not really sure about the version numbers and what to do with them. I just took the latest gson number and that was that. What do you think is the best thing to do?

jackson2/src/test/java/org/jdbi/v3/jackson2/TestJackson2Plugin.java
TheRealMarnes: I won't stop you if you want to add it then

stevenschlansker:
I think it's fine, it was more of a question.  Because right now if someone currently uses gson 2.7 and brings in jdbi3-gson2 it will break until you upgrade to 2.8.  Probably no big deal, but worth considering.  Also pushed a fix for the build, hopefully it goes green now.

stevenschlansker:
also note the bonus cleanup https://github.com/jdbi/jdbi/pull/1343/commits/946699740c8ba094e49c444663314d5f2e2879d5

TheRealMarnes:
Yiss, I saw, nice.

As for version mismatches, I'm still convinced that with everything else being equal to us, we should always go for the latest.

I've been up for 28 hours so I can't really think much anymore... I think the best thing to do with those version number references is to just have them equal to the dependency version. `2.x` doesn't tell me much when I wanna know the version it's programmed to.

TheRealMarnes:
Just pushed a simple change... The modules are named after the major version of each, so I figure the version number callouts should do the same. Agreeable?

json/src/main/java/org/jdbi/v3/json/internal/JsonColumnMapperFactory.java
```diff
@@ -33,7 +33,8 @@
                 () -> c.findColumnMapperFor(QualifiedType.of(Json.class)).map(JsonColumnMapperFactory::bludgeon),
                 () -> c.findColumnMapperFor(String.class))
                     .orElseThrow(() -> new IllegalStateException("No column mapper found for '@Json String' or 'String', this really shouldn't happen..."));
-            return c.getConfig(JsonConfig.class).getJsonMapper().fromJson(type, jsonStringMapper.map(r, i, c), c);
+            final String jsonValue = jsonStringMapper.map(r, i, c);
+            return jsonValue == null ? null : c.getConfig(JsonConfig.class).getJsonMapper().fromJson(type, jsonValue, c);
```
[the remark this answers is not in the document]
TheRealMarnes: Will do tonight

TheRealMarnes:
Well reading the code, it seems there are two independent features but the implementing code seems to conflate them.

One is object/json mapping. This can be done independently of the db with String handling: we can set it up in the `json` module with a public `JsonPlugin` that registers an argument/mapper that look for the JsonMapper instance (which would default to an exception stub) and otherwise follows the normal String/T flow.

The second is db-specific support for json columns like in postgres. This adds an argument/mapper specifically for the `Json` type and only specifies the new db binding method, leaving the json mapping to the `Json` module. The jackson and gson modules merely provide a non-stub impl of JsonMapper as they do now.

Aside from that, I think JsonMapper should be moved out of `internal` because 1) people can implement it themselves if they want, and 2) if someone writes a custom argument/mapper for Json handling, then they'll need to invoke this interface. 

This is all triggered by something I saw and find fishy. JsonColumnMapperFactory is set up to look for dedicated `Json` handlers first, then for String handlers. If none are found, throw exception because String handlers should really be present anyway, and implying that Json handlers are optional. This sounds all right to me.

But JsonArgumentFactory specifically looks only for dedicated Json handlers and throws an exception instructing to install PostgresPlugin. This should revert to plain String handlers too. You don't need to be using postgres to want to map objects to/from json in your db, but the feature overall feels to me like it's currently geared for that assumption.

I think I should just make the changes and PR them, this is a lot of text for what's really a small change.

qualidafial:
Hang on, isn't the `@Json` column mapper factory going to hijack any `@Json String` column mapper lookups?

TheRealMarnes:
The factories provided by the Json plugin will only convert values to/from json text. Their first step will be to deny any requests that are already for (`@Json`) Strings, otherwise they'd just make an infinite loop.

PR #1348, opened 2018-12-16
Let user configure a Clock

### LINKED ISSUE #1279: Let user set a Clock in jdbi for time(zone) data
From #1268

TheRealMarnes:
fixes #1279 

See tests for why this can be useful.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlLoggerUtil.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlLoggerUtil.java
@@ -14,7 +14,9 @@
 package org.jdbi.v3.core.statement;
 
 import java.sql.SQLException;
+import java.time.Clock;
 import java.time.Instant;
+import org.jdbi.v3.core.Time;
 import org.jdbi.v3.core.internal.UtilityClassException;
 
 class SqlLoggerUtil {
@@ -23,18 +25,20 @@ class SqlLoggerUtil {
     }
 
     static <T> T wrap(SqlLoggable<T> r, StatementContext ctx, SqlLogger logger) throws SQLException {
+        Clock clock = ctx.getConfig(Time.class).getClock();
+
         try {
-            ctx.setExecutionMoment(Instant.now());
+            ctx.setExecutionMoment(Instant.now(clock));
             logger.logBeforeExecution(ctx);
 
             T result = r.invoke();
 
-            ctx.setCompletionMoment(Instant.now());
+            ctx.setCompletionMoment(Instant.now(clock));
             logger.logAfterExecution(ctx);
 
             return result;
         } catch (SQLException e) {
-            ctx.setExceptionMoment(Instant.now());
+            ctx.setExceptionMoment(Instant.now(clock));
             logger.logException(ctx, e);
             throw e;
         }
--- a/core/src/main/java/org/jdbi/v3/core/statement/StatementContext.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/StatementContext.java
@@ -491,7 +491,7 @@ public class StatementContext implements Closeable {
      */
     @Nullable
     public ZonedDateTime getExecutionTimestamp() {
-        return executionMoment.atZone(getConfig(Time.class).getClock().getZone());
+        return executionMoment == null ? null : executionMoment.atZone(getConfig(Time.class).getClock().getZone());
     }
 
     /**
@@ -514,7 +514,7 @@ public class StatementContext implements Closeable {
      */
     @Nullable
     public ZonedDateTime getCompletionTimestamp() {
-        return completionMoment.atZone(getConfig(Time.class).getClock().getZone());
+        return completionMoment == null ? null : completionMoment.atZone(getConfig(Time.class).getClock().getZone());
     }
 
     /**
@@ -537,7 +537,7 @@ public class StatementContext implements Closeable {
      */
     @Nullable
     public ZonedDateTime getExceptionTimestamp() {
-        return exceptionMoment.atZone(getConfig(Time.class).getClock().getZone());
+        return exceptionMoment == null ? null : exceptionMoment.atZone(getConfig(Time.class).getClock().getZone());
     }
 
     /**
--- a/core/src/main/java/org/jdbi/v3/core/statement/StatementContext.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/StatementContext.java
@@ -21,6 +21,7 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.time.Instant;
+import java.time.ZonedDateTime;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -31,8 +32,10 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collector;
+import javax.annotation.Nullable;
 import org.jdbi.v3.core.CloseException;
 import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.Time;
 import org.jdbi.v3.core.argument.Argument;
 import org.jdbi.v3.core.argument.Arguments;
 import org.jdbi.v3.core.array.SqlArrayArgumentStrategy;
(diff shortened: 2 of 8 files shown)


core/src/main/java/org/jdbi/v3/core/statement/SqlLoggerUtil.java
```diff
@@ -23,18 +25,20 @@ private SqlLoggerUtil() {
     }
 
     static <T> T wrap(SqlLoggable<T> r, StatementContext ctx, SqlLogger logger) throws SQLException {
+        Clock clock = ctx.getConfig(Time.class).getClock();
+
         try {
-            ctx.setExecutionMoment(Instant.now());
+            ctx.setExecutionMoment(Instant.now(clock));
```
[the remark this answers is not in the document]
TheRealMarnes: well this is the preferred way to do it, because Clock only offers a method to get an Instant, but the other 9 Temporal types all have a `now` method with an overload that takes a Clock, so it's more consistent that way.

TheRealMarnes:
Well we do generate timestamps in `@Timestamped` :) And being able to set the reference clock can be useful for SqlLogger, Timestamped, and others when you don't want the system default clock to be used. I might want to run my application in my user timezone but all my database interactions in UTC, for example.

It's a pet peeve of mine to never use the no-arg `now()` methods. I just consider them something of a shoehorned lazy solution in any project, since they depend on a shared systemwide setting instead of a workable dependency abstraction system. I also think it's bad form for a library to just make the presumption that I want it to use a system setting without even having a choice in the matter.

TheRealMarnes:
It could also help in binding and mapping when timezone data needs to be filled in (possibly upcoming work). Instead of just blindly taking the system zone, this gives users some control over that too.

core/src/main/java/org/jdbi/v3/core/statement/SqlLoggerUtil.java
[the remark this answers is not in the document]
stevenschlansker: ok

stevenschlansker:
At the risk of getting off into the weeds, why would we need to fill in timezone data?  Currently the paradigm is simple - all data that goes to the database is passed in to jdbi, possibly transformed, and then shipped off into the database.  Producing our own data (whether timestamps or anything more complicated) is adding in a new thing jdbi does that it historically didn't do.

And I'm already a little bit uncomfortable about e.g. `@Timestamped` because for example it calls itself `now` but *behaves differently than SQL now*.  SQL now is the transaction start time.  And it's not clear to me how having

```java
@Timestamped
@SqlUpdate("insert into ... values(:now)")
void insert()
```

```java
@SqlUpdate("insert into ... values(now())")
void insert()
```

It's wordier, and behaves in a surprisingly subtly different way.

TheRealMarnes:
I don't think we should, but I could swear jdbi actually did somewhere. I know there was a big deal at some point about timezones getting messed up by jdbi mappings. I'm all for avoiding that altogether, of course.

As for `:now` vs `now()`, I suspect it's a matter of database language support, sticking to java types, and I've got a branch ready with a feature I'd like to suggest where Timetamped allows the user to the Temporal type to use (1 of 9), so they can have any of 9 differently typed database fields to put their timestamps in (e.g. an Instant, ZonedDateTime, just a Date, etc).

stevenschlansker:
But at least in Postgres, `now()` covers all those cases, or at least I think it does!  Why is it valuable to have this all in Java code?  To justify its inclusion, it should:

* do something new, e.g. what does having an implicitly-bound `ZonedDateTime.now(clock)` allow that database `now()` with an appropriate timezone can't?
* be convincingly superior to binding an explicit argument just like any other

TheRealMarnes:
I wasn't there when `@Timestamped` was added so honestly I don't know the full story. It just seems like a feature that's there to stay to me, and I saw some weakness in it so I thought I'd hop in and try to improve it.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestTimestamped.java
```diff
@@ -31,94 +38,122 @@
...
 
-        dbRule.getJdbi().setTimingCollector((elapsed, ctx) ->
-                assertThat(ctx.getBinding().findForName("now", ctx)).isPresent());
-
+        logNextTimestamp("now");
         personDAO.insert(p);
+        verifyLastTimestamp(T0);
 
-        dbRule.getJdbi().setTimingCollector(TimingCollector.NOP_TIMING_COLLECTOR);
+        when(clock.instant()).thenReturn(T0.plusSeconds(10).toInstant());
```
qualidafial: In previous projects I've seen folks implement a `FakeClock` class, with methods like `advance(Duration)` to simplify simulating passage of time.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestTimestamped.java
```diff
@@ -223,25 +258,15 @@ public boolean equals(Object o) {
...
-            Person person = (Person) o;
-
-            if (id != person.id) {
-                return false;
-            }
-            if (!firstName.equals(person.firstName)) {
-                return false;
-            }
-            return lastName != null ? lastName.equals(person.lastName) : person.lastName == null;
+            Person other = (Person) o;
```
qualidafial: Let's start using Lombok `@Data` classes in tests, instead of writing all this boilerplate by hand.

TheRealMarnes: Working on this in a new branch/pr

TheRealMarnes:
I added in a little commit to add zoned alternatives to the timing getters on StatementContext, and to add javadoc to all of them

core/src/main/java/org/jdbi/v3/core/statement/StatementContext.java
```diff
@@ -475,26 +478,71 @@ public void setConcurrentUpdatable(final boolean concurrentUpdatable) {
...
     public Instant getExceptionMoment() {
         return exceptionMoment;
     }
 
+    /**
+     * @return the {@link ZonedDateTime} at which query execution ended, if it did so with an exception
+     */
+    @Nullable
+    public ZonedDateTime getExceptionTimestamp() {
+        return exceptionMoment.atZone(getConfig(Time.class).getClock().getZone());
```
stevenschlansker: This sure looks like it's going to cause a NPE if `exceptionMoment` isn't set yet.

core/src/main/java/org/jdbi/v3/core/statement/StatementContext.java
```diff
@@ -475,26 +478,71 @@ public void setConcurrentUpdatable(final boolean concurrentUpdatable) {
...
+    @Nullable
     public Instant getExceptionMoment() {
         return exceptionMoment;
     }
 
+    /**
+     * @return the {@link ZonedDateTime} at which query execution ended, if it did so with an exception
+     */
+    @Nullable
+    public ZonedDateTime getExceptionTimestamp() {
```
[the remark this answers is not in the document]
stevenschlansker: I disagree - I always log everything as UTC.  Having timezones in logs only adds to the confusion.

core/src/main/java/org/jdbi/v3/core/statement/StatementContext.java
TheRealMarnes: Good catch, sorry!

stevenschlansker:
I'm getting increasingly concerned about the growth of time-related functionality in jdbi.  I think that 'formatting timestamps to make them nice for logging' is decidedly outside of the project's charter.

TheRealMarnes:
But none of this is about formatting. It's about passing more meaningful information to the user and avoiding the hardcoded use of system settings.

This started with some features being added that call `.now()`, which relies on the system clock. It's bad practice, doubly so for a library, to not use a Clock for this because you might not want to or be able to set the Clock you want as the system clock. That's all I changed really.

These new methods are about returning the now available, more informative ZonedDateTimes insteaad of Instants. An Instant is a number of millis since epoch, a ZonedDateTime is much more usable and debuggable info that can be converted to an Instant but not back. If we are able to construct a zdt (and we are now), we should offer that over an Instant.

core/src/main/java/org/jdbi/v3/core/statement/StatementContext.java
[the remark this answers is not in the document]
TheRealMarnes: Then you don't get the zdt, you get the already existing Instants.

stevenschlansker:
Instant conveys exactly the same information, just without a timezone.  A consumer of the Instant can always call `atZone` on it to place it in a timezone of their choice.  In fact, you'll notice that the `Clock` interface itself only provides `clock.instant()` and not `clock.zonedDateTime()`

stevenschlansker:
@qualidafial do you have any thoughts here?

TheRealMarnes:
I think I'm starting to see your point. The timezone handling in SqlLogger can be handled by a Clock instance managed in the logger itself instead of centralizing it in jdbi (or users can add their own Time config if it really starts to annoy them to have to repeat the Clock injection in multiple places). Since the timings on the context are Instants, there's no assumptions being made by jdbi, so that's fine.

Timestamped still annoys me though. It returns a ZDT but it's set to whatever the system timezone is. This should either use a configurable Clock, or set an Instant without making any timezone assumptions. Right now to do this users would have to re-implement all of Timestamped.

I'd be ok with ditching this PR if we ensure in the future we keep avoiding the use of system settings without settable alternatives (basically meaning never using ZDT, always Instant) and we change the Timestamped impl in the next major.

Sorry if I've been a little anal over this, I've just had so much frustration in the past over libraries being hardcoded to just read whatever system setting and go with that. Making a settable Clock was the most obvious solution, but you have a point that there aren't many time features yet, the logging one is already properly built with neutral Instants, and only Timestamped (a feature you think is odd in the first place) is kind of doing what I believe is wrong.

stevenschlansker:
For the record I'm not opposed to the 'mocking out the clock' portion of this PR, so if you want to get that part shipped, feel free to reopen just that piece.

stevenschlansker:
Does anyone actually use `@Timestamped`?  Would deprecating it in favor of SQL `now()` make sense?

TheRealMarnes:
I guess in hindsight I'd be happy with something as simple as adding TimestampedConfig that holds a timezone to be used by Timestamped instead of the system zone. Then the feature becomes conceptually clean to me. Whether we'll keep it long-term or not is a separate topic.

It could be made more useful/worth keeping if I also implement my simple change that lets users configure the Temporal type they want Timestamped to inject, so that it can be made to inject the current Instant, ZDT, LocalDate, LocalTime, YearMonth, whatever. Then at least it's more-than-equivalent to SQL `now()`.

core/src/main/java/org/jdbi/v3/core/statement/StatementContext.java
```diff
@@ -475,26 +478,71 @@ public void setConcurrentUpdatable(final boolean concurrentUpdatable) {
...
         return executionMoment;
     }
 
+    /**
+     * @return the {@link ZonedDateTime} at which query execution began
+     */
+    @Nullable
+    public ZonedDateTime getExecutionTimestamp() {
+        return executionMoment == null ? null : executionMoment.atZone(getConfig(Time.class).getClock().getZone());
+    }
```
qualidafial: I think we could ditch these overloads. If the user wants them rendered converted to a zone they can do that themselves.

TheRealMarnes: Well that was kind of the entire point of adding the Clock inside jdbi, that the whole process would be more convenient.

But I pretty much agree with steven by now, this PR is dead (long live this PR)

stevenschlansker:
Adding a `TimestampedConfig` is a reasonable compromise.

stevenschlansker:
I guess `now()` in particular may be a postgres-ism but the SQL standard `CURRENT_DATE`, `CURRENT_DATETIME`, etc should enjoy wide support and reasonable behavior.

PR #1349, opened 2018-12-17
setting for enums by name/ordinal

### LINKED ISSUE #1197: ArgumentFactory for enums by ordinal
**Scenario**

**Feature proposal**

### LINKED ISSUE #1273: Let user configure if enums should be mapped by name or ordinal
From #1268 

### LINKED ISSUE #1354: @ByName and @ByOrdinal for enum values

TheRealMarnes:
fixes #1273
fixes #1197 as far as I'm concerned (I don't think the factories are needed as API)
fixes #1354

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/EnumConfig.java
+++ b/core/src/main/java/org/jdbi/v3/core/EnumConfig.java
@@ -1,0 +1,73 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core;
+
+import java.lang.annotation.Annotation;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import org.jdbi.v3.core.config.JdbiConfig;
+
+/**
+ * Configuration for behavior related to {@link Enum}s.
+ */
+public class EnumConfig implements JdbiConfig<EnumConfig> {
+    private static final Set<Class<? extends Annotation>> ALLOWED_HANDLINGS = new HashSet<>(Arrays.asList(EnumByName.class, EnumByOrdinal.class));
+    private Class<? extends Annotation> handling;
+
+    public EnumConfig() {
+        handling = EnumByName.class;
+    }
+
+    private EnumConfig(EnumConfig other) {
+        handling = other.handling;
+    }
+
+    /**
+     * Applies to both binding and mapping.
+     *
+     * @return true if enums are handled by name, false if enums are handled by ordinal
+     */
+    @Deprecated
+    public boolean isEnumHandledByName() {
+        return handling == EnumByName.class;
+    }
+
+    /**
+     * Applies to both binding and mapping.
+     *
+     * @param byName true if enums should be handled by name, false if enums should be handled by ordinal
+     */
+    @Deprecated
+    public void setEnumHandledByName(boolean byName) {
+        handling = byName ? EnumByName.class : EnumByOrdinal.class;
+    }
+
+    public Class<? extends Annotation> getEnumHandling() {
+        return handling;
+    }
+
+    public EnumConfig setEnumHandling(Class<? extends Annotation> handling) {
+        if (!ALLOWED_HANDLINGS.contains(handling)) {
+            throw new IllegalArgumentException(handling + " is not an accepted enum handling annotation");
+        }
+        this.handling = handling;
+        return this;
+    }
+
+    @Override
+    public EnumConfig createCopy() {
+        return new EnumConfig(this);
+    }
+}
--- a/core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
@@ -54,7 +54,9 @@ public class Arguments implements JdbiConfig<Arguments> {
         register(new EssentialsArgumentFactory());
         register(new JavaTimeZoneIdArgumentFactory());
         register(new NVarcharArgumentFactory());
-        register(new EnumArgumentFactory());
+        register(new EnumUnqualifiedArgumentFactory());
+        register(new EnumByOrdinalArgumentFactory());
+        register(new EnumByNameArgumentFactory());
         register(new OptionalArgumentFactory());
     }
 
--- a/core/src/main/java/org/jdbi/v3/core/argument/EnumByNameArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/EnumByNameArgumentFactory.java
@@ -11,29 +11,28 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jdbi.v3.core.mapper;
+package org.jdbi.v3.core.argument;
 
 import java.lang.reflect.Type;
+import java.sql.Types;
 import java.util.Optional;
-
+import org.jdbi.v3.core.EnumByName;
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.internal.Enums;
 
-import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
-
-/**
- * Produces enum column mappers, which map enums from varchar columns using {@link Enum#valueOf(Class, String)}.
- *
- * @deprecated there is no reason for this to be API
- */
-@Deprecated
-// TODO jdbi4: delete
-public class EnumByNameMapperFactory implements ColumnMapperFactory {
+@EnumByName
+class EnumByNameArgumentFactory implements ArgumentFactory {
     @Override
-    public Optional<ColumnMapper<?>> build(Type type, ConfigRegistry config) {
-        Class<?> clazz = getErasedType(type);
+    public Optional<Argument> build(Type type, Object rawValue, ConfigRegistry config) {
+        if (!Enums.isEnum(type)) {
+            return Optional.empty();
+        }
 
-        return clazz.isEnum()
-                ? Optional.of(EnumMapper.byName(clazz.asSubclass(Enum.class)))
-                : Optional.empty();
+        if (rawValue == null) {
+            return Optional.of(new NullArgument(Types.VARCHAR));
+        }
+
+        Enum<?> enumValue = (Enum<?>) rawValue;
+        return config.get(Arguments.class).findFor(String.class, enumValue.name());
     }
 }
--- a/core/src/main/java/org/jdbi/v3/core/argument/EnumByOrdinalArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/EnumByOrdinalArgumentFactory.java
@@ -11,29 +11,29 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jdbi.v3.core.mapper;
+package org.jdbi.v3.core.argument;
 
 import java.lang.reflect.Type;
+import java.sql.Types;
 import java.util.Optional;
-
+import org.jdbi.v3.core.EnumByOrdinal;
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.internal.Enums;
 
-import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
-
-/**
- * Produces enum column mappers, which map enums from varchar columns using {@link Enum#valueOf(Class, String)}.
- *
- * @deprecated there is no reason for this to be API
- */
-@Deprecated
-// TODO jdbi4: delete
-public class EnumByNameMapperFactory implements ColumnMapperFactory {
+@EnumByOrdinal
+class EnumByOrdinalArgumentFactory implements ArgumentFactory {
     @Override
-    public Optional<ColumnMapper<?>> build(Type type, ConfigRegistry config) {
-        Class<?> clazz = getErasedType(type);
+    public Optional<Argument> build(Type type, Object rawValue, ConfigRegistry config) {
+        if (!Enums.isEnum(type)) {
+            return Optional.empty();
+        }
 
-        return clazz.isEnum()
-                ? Optional.of(EnumMapper.byName(clazz.asSubclass(Enum.class)))
-                : Optional.empty();
+        if (rawValue == null) {
+            return Optional.of(new NullArgument(Types.INTEGER));
+        }
+
+        Enum<?> enumValue = (Enum<?>) rawValue;
+
+        return config.get(Arguments.class).findFor(Integer.class, enumValue.ordinal());
     }
 }
--- a/core/src/main/java/org/jdbi/v3/core/argument/EnumUnqualifiedArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/EnumUnqualifiedArgumentFactory.java
@@ -15,20 +15,17 @@ package org.jdbi.v3.core.argument;
 
 import java.lang.reflect.Type;
 import java.util.Optional;
-
-import org.jdbi.v3.core.argument.internal.strategies.LoggableBinderArgument;
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.internal.Enums;
 
-class EnumArgumentFactory implements ArgumentFactory {
+/**
+ * catches any non-qualified enums, qualifies them, and passes them back to the chain
+ */
+class EnumUnqualifiedArgumentFactory implements ArgumentFactory {
     @Override
-    public Optional<Argument> build(Type expectedType, Object rawValue, ConfigRegistry config) {
-        // Enums must be bound as VARCHAR
-        // TODO use the same configuration as EnumMapperFactory for consistency
-        if (rawValue instanceof Enum) {
-            Enum<?> enumValue = (Enum<?>) rawValue;
-            return Optional.of(LoggableBinderArgument.bindAsString(enumValue.name()));
-        } else {
-            return Optional.empty();
-        }
+    public Optional<Argument> build(Type type, Object rawValue, ConfigRegistry config) {
+        return Enums.isEnum(type)
+            ? config.get(Arguments.class).findFor(Enums.qualifyByNameOrOrdinal(type, config), rawValue)
+            : Optional.empty();
     }
 }
--- a/core/src/main/java/org/jdbi/v3/core/internal/Enums.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/Enums.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.internal;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import java.util.Set;
+import org.jdbi.v3.core.EnumByName;
+import org.jdbi.v3.core.EnumByOrdinal;
+import org.jdbi.v3.core.EnumConfig;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericTypes;
+import org.jdbi.v3.core.qualifier.QualifiedType;
+import org.jdbi.v3.core.qualifier.Qualifiers;
+
+public class Enums {
+    private Enums() {
+        throw new UtilityClassException();
+    }
+
+    // all enums are T in Enum<T>
+    public static boolean isEnum(Type type) {
+        return GenericTypes.getErasedType(type).isEnum();
+    }
+
+    public static QualifiedType qualifyByNameOrOrdinal(Type type, ConfigRegistry config) {
+        if (!isEnum(type)) {
+            throw new IllegalArgumentException("type " + type.getTypeName() + " is not an Enum");
+        }
+
+        // Ts of Enum<T> can only be classes
+        @SuppressWarnings("unchecked")
+        Class<? extends Enum<?>> enumClass = (Class) type;
+        Annotation qualifier = getAppropriateQualifier(enumClass, config);
+
+        return QualifiedType.of(enumClass).with(qualifier);
+    }
+
+    private static Annotation getAppropriateQualifier(Class<? extends Enum> enumClass, ConfigRegistry config) {
+        Class<? extends Annotation> qualifier = classByNameOrOrdinal(enumClass);
+
+        if (qualifier == null) {
+            qualifier = config.get(EnumConfig.class).getEnumHandling();
+        }
+
+        return AnnotationFactory.create(qualifier);
+    }
+
+    private static Class<? extends Annotation> classByNameOrOrdinal(Class<? extends Enum> enumClass) {
+        Set<Annotation> onClass = Qualifiers.getQualifiers(enumClass);
+
+        boolean byNameOnClass = onClass.stream().anyMatch(EnumByName.class::isInstance);
+        boolean byOrdinalOnClass = onClass.stream().anyMatch(EnumByOrdinal.class::isInstance);
+
+        if (byNameOnClass && byOrdinalOnClass) {
+            throw new IllegalArgumentException(String.format(
+                "enum %s is annotated with both @%s and @%s",
+                enumClass.getName(),
+                EnumByName.class.getSimpleName(),
+                EnumByOrdinal.class.getSimpleName()
+            ));
+        } else if (byNameOnClass) {
+            return EnumByName.class;
+        } else if (byOrdinalOnClass) {
+            return EnumByOrdinal.class;
+        } else {
+            return null;
+        }
+    }
+}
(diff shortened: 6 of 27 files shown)


core/src/main/java/org/jdbi/v3/core/Handles.java
```diff
@@ -50,8 +50,9 @@ public boolean isForceEndTransactions() {
      * @param forceEndTransactions whether to enforce transaction termination
      *                             discipline.
      */
-    public void setForceEndTransactions(boolean forceEndTransactions) {
+    public Handles setForceEndTransactions(boolean forceEndTransactions) {
```
stevenschlansker: This is a backwards incompatible change and cannot be merged.

core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
```diff
@@ -114,11 +117,12 @@ private Arguments register(QualifiedArgumentFactory factory) {
      * we don't have a type for.
      * @param untypedNullArgument the argument to bind
      */
-    public void setUntypedNullArgument(Argument untypedNullArgument) {
+    public Arguments setUntypedNullArgument(Argument untypedNullArgument) {
```
stevenschlansker: Another non-permissible change

core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
```diff
@@ -15,17 +15,21 @@
...
 class EnumArgumentFactory implements ArgumentFactory {
     @Override
     public Optional<Argument> build(Type expectedType, Object rawValue, ConfigRegistry config) {
-        // Enums must be bound as VARCHAR
-        // TODO use the same configuration as EnumMapperFactory for consistency
         if (rawValue instanceof Enum) {
             Enum<?> enumValue = (Enum<?>) rawValue;
-            return Optional.of(new LoggableToStringOrNPEArgument<>(enumValue, Enum::name));
+
+            StatementBinder<Enum<?>> binder = config.get(Arguments.class).getBindEnumsByName()
```
stevenschlansker: 👍

core/src/main/java/org/jdbi/v3/core/Handles.java
stevenschlansker: https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.15

TheRealMarnes: I had such a feeling...

TheRealMarnes:
Changes removed

core/src/main/java/org/jdbi/v3/core/config/Configurable.java
```diff
@@ -100,6 +100,30 @@ default This setSqlParser(SqlParser parser) {
...
 
+    /**
+     * convenience method to set the binding and mapping policy for enums to use ordinal values.
+     *
+     * @return this
+     */
+    @SuppressWarnings("unchecked")
+    default This handleEnumsByOrdinal() {
+        getConfig(Arguments.class).setBindEnumsByName(false);
+        getConfig(ColumnMappers.class).setMapEnumsByName(false);
```
qualidafial: Feels like we want an `Enums` config class with a single property for this.

core/src/main/java/org/jdbi/v3/core/mapper/EnumByOrdinalMapperFactory.java
```diff
@@ -13,21 +13,21 @@
...
 import java.lang.reflect.Type;
 import java.util.Optional;
-
 import org.jdbi.v3.core.config.ConfigRegistry;
 
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
 /**
- * Produces enum column mappers, which map enums from numeric columns according to ordinal value.
+ * Produces enum column mappers, which map enums from numeric columns according to {@link Enum#ordinal()}.
```
qualidafial: Would be nice to add a `{@link}` to the new config property

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapperFactory.java
```diff
@@ -22,9 +22,11 @@
  */
 class EnumMapperFactory implements ColumnMapperFactory {
     private static final EnumByNameMapperFactory BY_NAME = new EnumByNameMapperFactory();
+    private static final EnumByOrdinalMapperFactory BY_ORDINAL = new EnumByOrdinalMapperFactory();
```
qualidafial: Not a blocker for this PR, but what do you think of having `@ByName` and `@ByOrdinal` qualifying annotations which can override the preferred strategy on a case-by-case basis?

core/src/main/java/org/jdbi/v3/core/config/Configurable.java
TheRealMarnes: Well I thought people might want to set them up to behave differently by default in read and write, like if they mainly read from one set of tables with a convention for one way, and mostly write to other tables with the other convention. That and I think `Enums` is even more of a dead-end config class than Time...

I'm 50/50 for both, doesn't really matter to me. Shall I change it?

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapperFactory.java
TheRealMarnes: Yeah, that was brought up before, I'm in favor of it. #1354

core/src/main/java/org/jdbi/v3/core/mapper/EnumByOrdinalMapperFactory.java
TheRealMarnes: After we decide on the location of said new preference

core/src/main/java/org/jdbi/v3/core/config/Configurable.java
stevenschlansker: I also think it should be a simple property.  I don't think having read and write behavior differ makes all that much sense.

TheRealMarnes:
Remarks handled, looks good to go to me

TheRealMarnes:
Added a callout to Enums in the devguide.

TheRealMarnes:
Also fixes #1197 as far as I'm concerned

TheRealMarnes:
Working on a potential related improvement for these but I need a little time to figure it out, will remove the label when done/discarded

TheRealMarnes:
To help with reviewing this:
- both binding and mapping are handled by 3 factories: an unqualified master that checks the `Enums` setting and composes a qualified type that it passed back to the chain, to 2 specialized factories that pass the enum in a converted form (name/ordinal) to the chain again for the actual binding
- the argument part was never exposed but the mapping part was made "API", so I just left it all there to be removed in the future as it's redundant and not needed for the user
- we kinda need #1363 to make this feature nicer to use in the fluent api

The diff is kinda big but I think it's easy to breeze through content-wise. I can break it up into more atomic commits if you want.

core/src/main/java/org/jdbi/v3/core/argument/EnumByNameArgumentFactory.java
```diff
@@ -14,20 +14,26 @@
...
+import java.sql.Types;
 import java.util.Optional;
-import org.jdbi.v3.core.argument.internal.strategies.LoggableToStringOrNPEArgument;
+import org.jdbi.v3.core.EnumByName;
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericTypes;
 
-class EnumArgumentFactory implements ArgumentFactory {
+@EnumByName
+class EnumByNameArgumentFactory implements ArgumentFactory {
```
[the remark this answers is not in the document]
qualidafial: In fact you could probably do without EnumUnqualifiedArgumentFactory class acting as a switchboard as follows:

```java
// after the enum type check
Set<Annotation> enumByName = singleton(EnumQualifiers.enumByName());
if (enumByName.equals(qualifiedType.getQualifiers())
    || enumByName.equals(Qualifiers.getQualifiers(enumClass))
    || config.get(Enums.class).enumsHandledByName()) {
  // this enum is bound by name
}
```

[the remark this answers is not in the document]
TheRealMarnes: So you something like this?
image

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
```diff
@@ -42,50 +39,7 @@
      * @return an enum mapper that matches on {@link Enum#ordinal()}
      */
     public static <E extends Enum<E>> ColumnMapper<E> byOrdinal(Class<E> type) {
-        return new ByOrdinal<>(type);
+        return new EnumByOrdinalMapper<>(type);
     }
 
-    private static class ByName<E extends Enum<E>> extends EnumMapper<E> {
```
stevenschlansker: This was all nicely localized in one easily read file.  Now it's in five :(

core/src/main/java/org/jdbi/v3/core/mapper/EnumByOrdinalMapperFactory2.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Optional;
+import org.jdbi.v3.core.EnumByOrdinal;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericTypes;
+
+/**
+ * Produces enum column mappers, which map enums from numeric columns according to their {@link Enum#ordinal()}.
+ */
+@EnumByOrdinal
+class EnumByOrdinalMapperFactory2 implements ColumnMapperFactory {
```
stevenschlansker: How come we duplicate all these mapper factories and deprecate the old ones for removal?  Can we just deprecate them "for hiding" and continue to use retrofitted existing implementations?  This is a very high LOC add for a small feature.

core/src/main/java/org/jdbi/v3/core/EnumConfig.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package org.jdbi.v3.core;
+
+import org.jdbi.v3.core.config.JdbiConfig;
+
+/**
+ * Configuration for behavior related to {@link Enum}s.
+ */
+public class EnumConfig implements JdbiConfig<EnumConfig> {
+    private boolean handleEnumsByName = true;
```
stevenschlansker: Would it be useful to have this be a `private Annotation defaultEnumQualifier`, a more general mechanism that doesn't explicitly encode a choice between the two built in strategies?

core/src/main/java/org/jdbi/v3/core/mapper/EnumByOrdinalMapperFactory2.java
TheRealMarnes: I duplicated them to avoid tangling/spaghetti. I deprecated the old ones for hiding, and when the time comes you can just hit `Delete` on the old ones and all will be right. I like to sever connections between new and old and leave as little shared space as possible, or at least extract the common parts to an independent space where it can be used by old and new without them touching each other. This way you can just forget about deprecated things and delete them at a glance, they take up physical space but have no bearing on things.

I've never found LOC to be a very meaningful measurement compared to things like coupling, scope size/code piling, or excessive abstractions.

I think we should skip this pr until later if there's this much food for arguments in it.

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
TheRealMarnes: I'm not a fan of lobbing things together for non-technical reasons? Single responsibility, minimalism, yadda yadda?

core/src/main/java/org/jdbi/v3/core/EnumConfig.java
TheRealMarnes: Are you talking about the API or internal workings? We can change the boolean internals to something tertiary when the need arises. The binary nature of the API did bother me too though. I was thinking an enum would be a less surprising thing for users to have to use than `setEnumHandling(EnumByName.class)`

TheRealMarnes: Actually, I take that back. `setEnumHandling(EnumByName.class)` is probably pretty neat, less redundant stuff for users to have to link together since they'll just be using the annotations as points of contact for everything to do with the feature.

core/src/main/java/org/jdbi/v3/core/mapper/EnumByOrdinalMapperFactory2.java
TheRealMarnes: Would extracting ByName and ByOrdinal to independent package-private classes so they don't need to be duplicated satisfy you more?

core/src/main/java/org/jdbi/v3/core/EnumConfig.java
TheRealMarnes: I think I would like a marker interface on those annotations then though, so you can't pass just any `Annotation` to the methods. That'd be like going back to java 5.

TheRealMarnes: We can't make that method value-safe though, users would be able to pass in just any annotation class. I'd go with an enum that collects the different annotations as values, only to be used for safety on the setter.

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
stevenschlansker: I don't see this approach as particularly minimal, unfortunately.  I'm taking a pass at reducing some of the copy+paste.

core/src/main/java/org/jdbi/v3/core/EnumConfig.java
```diff
@@ -0,0 +1,73 @@
...
+        handling = other.handling;
+    }
+
+    /**
+     * Applies to both binding and mapping.
+     *
+     * @return true if enums are handled by name, false if enums are handled by ordinal
+     */
+    @Deprecated
+    public boolean isEnumHandledByName() {
```
stevenschlansker: This is adding already deprecated code?

core/src/main/java/org/jdbi/v3/core/EnumConfig.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import org.jdbi.v3.core.config.JdbiConfig;
+
+/**
+ * Configuration for behavior related to {@link Enum}s.
+ */
+public class EnumConfig implements JdbiConfig<EnumConfig> {
+    private static final Set<Class<? extends Annotation>> ALLOWED_HANDLINGS = new HashSet<>(Arrays.asList(EnumByName.class, EnumByOrdinal.class));
```
stevenschlansker: This restricts the added qualifying annotations, which removes the ability to bind your own handling.  Is there a specific reason to forbid custom handling?  The purpose of allowing the user to set their own choice was to allow e.g. vendor extensions if they need to bind it as a special type of object, and this forbids that..

stevenschlansker: Maybe instead we should just encourage them to register their own handler as a higher priority instead? 🤷‍♀️

sqlobject/src/test/java/org/jdbi/v3/sqlobject/AdvancableClock.java
```diff
@@ -16,27 +16,32 @@
 import java.time.Clock;
 import java.time.Instant;
 import java.time.ZoneId;
+import java.time.ZonedDateTime;
 import java.time.temporal.TemporalUnit;
 
-class MockClock extends Clock {
-    private Instant now = Instant.now();
+public class AdvancableClock extends Clock {
```
stevenschlansker: How come this class name changed?  This seems unrelated to the meat of this PR.  It seems unrelated to time handling?

TheRealMarnes: Just a tiny collateral improvement, came across it somehow and thought it needed a spit shine. Instead of throwing exceptions and assuming system timezone data, it should just take up the timezone as part of its function. Turns it into a functional component instead of just a hack to get around something. I can take the change out if it bothers you but it's so small it may as well just be merged along ¯\\\_(ツ)_\/¯

core/src/main/java/org/jdbi/v3/core/argument/EnumByNameArgumentFactory.java
```diff
@@ -14,21 +14,25 @@
...
 import java.util.Optional;
-
-import org.jdbi.v3.core.argument.internal.strategies.LoggableBinderArgument;
+import org.jdbi.v3.core.EnumByName;
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.internal.Enums;
 
-class EnumArgumentFactory implements ArgumentFactory {
+@EnumByName
+class EnumByNameArgumentFactory implements ArgumentFactory {
```
qualidafial: I think it would be much simpler to change this class (which is package-private, so this would not break compatibility) to implement `QualifiedArgumentFactory` instead of `ArgumentFactory`.

The precedence rules could live in e.g. `EnumConfig.resolveEnumStrategy(QualifiedType)`, which would return the correct `EnumStrategy` enum value based on the precedence rules:

```java
public EnumStrategy resolveEnumStrategy(QualifiedType type) {
  EnumStrategy strategy = resolveEnumStrategy(type.getQualifiers());

  if (strategy == null) {
    Class<?> enumType = getErasedType(type.getType());
    strategy = resolveEnumStrategy(getQualifiers(enumType));
  }

  if (strategy == null) {
    strategy = defaultStrategy;
  }
  
  return strategy;
}

private EnumStrategy resolveEnumStrategy(Collection<Annotation> qualifiers) {
  if (qualifiers.contains(EnumByName.class)) {
    return EnumStrategy.BY_NAME;
  }

  if (qualifiers.contains(EnumByOrdinal.class)) {
    return EnumStrategy.BY_ORDINAL;
  }

  return null;
}
```

Then the argument factory and mapper factory can delegate to that config class to resolve the enum strategy, and just pick the existing argument / mapper implementation according to the resolved strategy.

core/src/main/java/org/jdbi/v3/core/argument/EnumByOrdinalArgumentFactory.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.lang.reflect.Type;
+import java.sql.Types;
+import java.util.Optional;
+import org.jdbi.v3.core.EnumByOrdinal;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.internal.Enums;
+
+@EnumByOrdinal
+class EnumByOrdinalArgumentFactory implements ArgumentFactory {
```
qualidafial: Note that this factory becomes redundant if my suggestion to implement `QualifiedArgumentFactory` is used.

core/src/main/java/org/jdbi/v3/core/argument/EnumUnqualifiedArgumentFactory.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.lang.reflect.Type;
+import java.util.Optional;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.internal.Enums;
+
+/**
+ * catches any non-qualified enums, qualifies them, and passes them back to the chain
+ */
+class EnumUnqualifiedArgumentFactory implements ArgumentFactory {
```
qualidafial: And this class would also be redundant.

core/src/main/java/org/jdbi/v3/core/internal/Enums.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Set;
+import org.jdbi.v3.core.EnumByName;
+import org.jdbi.v3.core.EnumByOrdinal;
+import org.jdbi.v3.core.EnumConfig;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericTypes;
+import org.jdbi.v3.core.qualifier.QualifiedType;
+import org.jdbi.v3.core.qualifier.Qualifiers;
+
+public class Enums {
```
qualidafial: @TheRealMarnes Just a note on design philosophy of config classes that I'm not sure we've communicated to you:

We've been naming most config classes with pluralized names so that inline usage with the `Configurable.configure()` method reads as "configure row mappers", "configure arguments", etc. So I think I'd like to rename `EnumConfig` to `Enums`. Thus:

```
handle.createUpdate(sql)
    .configure(Enums.class, config -> config.setDefaultEnumStrategy(EnumStrategy.BY_ORDINAL))
    .execute();
```

core/src/main/java/org/jdbi/v3/core/mapper/EnumByNameMapperFactory2.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Optional;
+import org.jdbi.v3.core.EnumByName;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericTypes;
+
+/**
+ * Produces enum column mappers, which case-insensitively map enums from textual columns according to their {@link Enum#name()}.
+ */
+@EnumByName
+class EnumByNameMapperFactory2 implements ColumnMapperFactory {
```
qualidafial: I think we'd be better served by refactoring `EnumMapperFactory` (which is package private) to implement `QualifiedColumnMapperFactory`, and follow the same logic I suggested for `EnumArgumentFactory`. Then these new classes are redundant.

core/src/main/java/org/jdbi/v3/core/mapper/EnumByOrdinalMapperFactory2.java
qualidafial: This class likewise becomes redundant if `EnumMapperFactory` implements `QualifiedColumnMapperFactory`

core/src/main/java/org/jdbi/v3/core/mapper/EnumByNameMapperFactory.java
```diff
@@ -22,7 +22,11 @@
 
 /**
  * Produces enum column mappers, which map enums from varchar columns using {@link Enum#valueOf(Class, String)}.
+ *
+ * @deprecated there is no reason for this to be API
  */
+@Deprecated
+// TODO jdbi4: delete
```
qualidafial: The deprecation message is inaccurate. This and `EnumByOrdinalMapperFactory` were made public API to enable changing the default enum strategy via a registration:

Core:

```
jdbi.registerColumnMapper(EnumByOrdinalMapperFactory.class);
```

SQL Object:

```
@RegisterColumnMapperFactory(EnumByOrdinalMapperFactory.class)
```

With that said, we never added an equivalent for `@RegisterArgumentFactory`, so it wasn't a full solution.

Let's change the deprecation message to redirect users to the new API options that replace this class: the config class `default*` property, or the annotations.

core/src/main/java/org/jdbi/v3/core/mapper/EnumByOrdinalMapper.java
```diff
@@ -0,0 +1,44 @@
...
+    EnumByOrdinalMapper(Class<E> type) {
+        this.constants = type.getEnumConstants();
+        this.type = type;
+    }
+
+    @Override
+    public E map(ResultSet rs, int columnNumber, StatementContext ctx) throws SQLException {
+        Integer ordinal = ctx.findColumnMapperFor(Integer.class)
+            .orElseThrow(() -> new UnableToProduceResultException("an Integer column mapper is required to map Enums from ordinals", ctx))
+            .map(rs, columnNumber, ctx);
```
qualidafial: Feels like overkill looking up the mapper from the registry when we can just call `getInt()`. The mappers for core java types are very unlikely to be overridden.

core/src/main/java/org/jdbi/v3/core/mapper/EnumByNameMapper.java
```diff
@@ -0,0 +1,53 @@
...
+    private final ConcurrentMap<String, E> lookups = new ConcurrentHashMap<>();
+
+    EnumByNameMapper(Class<E> type) {
+        this.type = type;
+    }
+
+    @Override
+    public E map(ResultSet rs, int columnNumber, StatementContext ctx) throws SQLException {
+        String name = ctx.findColumnMapperFor(String.class)
+            .orElseThrow(() -> new UnableToProduceResultException("a String column mapper is required to map Enums from names", ctx))
```
qualidafial: Ditto my comment in `EnumByOrdinalMapper.map()`

core/src/main/java/org/jdbi/v3/core/mapper/EnumByOrdinalMapper.java
qualidafial: If you feel strongly about keeping this, I'd vote move this lookup to `specialize()` so it's not repeated on every row.

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
```diff
@@ -13,17 +13,14 @@
...
-
 /**
  * Column mapper for Java {@code enum} types.
  * @param <E> the enum type mapped
+ *
+ * @deprecated there is no reason for this to be API
  */
+@Deprecated
+// TODO jdbi4: delete
 public abstract class EnumMapper<E extends Enum<E>> implements ColumnMapper<E> {
```
qualidafial: This class is still the only access point for `EnumMapper.byName()` and `EnumMapper.byOrdinal()`, e.g. for use with `ResultBearing.map(ColumnMapper<T>): ResultIterable<T>`

core/src/test/java/org/jdbi/v3/core/mapper/TestEnums.java
```diff
@@ -83,7 +84,7 @@ public void testMapInvalidEnumValue() {
 
         assertThatThrownBy(() -> h.createQuery("select * from something order by id")
             .mapToBean(SomethingElse.class)
-            .findFirst()).isInstanceOf(IllegalArgumentException.class);
+            .findFirst()).isInstanceOf(UnableToProduceResultException.class);
```
qualidafial: Why did this change?

sqlobject/src/test/java/org/jdbi/v3/sqlobject/EnumByXTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+            FooDao dao = h.attach(FooDao.class);
+
+            dao.insert(Foo.BAR);
+            assertThat(h.createQuery("select ordinal from enums").mapTo(Integer.class).findOnly()).isEqualTo(0);
+
+            Foo value = dao.select();
+            assertThat(value).isEqualTo(Foo.BAR);
+        });
+    }
```
qualidafial: Probably also want to check `@EnumByName` when the default is set to by-ordinal.

qualidafial:
Those are all my comments for now.

TheRealMarnes:
closing to merge locally and polish up for a fresh pr with a cleaner slate

core/src/test/java/org/jdbi/v3/core/mapper/TestEnums.java
TheRealMarnes: Because the new implementations throw this exception instead of a general IAE

PR #1352, opened 2018-12-17
spring5 compatibility test module

TheRealMarnes:
Just needs a fix for that weird build error where travis gets hung up on maven downloading xmlunit or spring-something in the mavencentral run

### DIFF
--- a/spring5/pom.xml
+++ b/spring5/pom.xml
@@ -14,53 +14,59 @@
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
-
     <parent>
         <artifactId>jdbi3-parent</artifactId>
         <groupId>org.jdbi</groupId>
         <version>3.6.0-SNAPSHOT</version>
     </parent>
 
-    <artifactId>jdbi3-spring4</artifactId>
-    <name>jdbi3 Spring 4</name>
-    <description>jdbi Spring 4 integration</description>
+    <artifactId>jdbi3-spring5</artifactId>
+    <name>jdbi3 Spring 5</name>
+    <description>jdbi Spring 4/5 forward compatibiility testing</description>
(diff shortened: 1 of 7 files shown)


spring5/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>jdbi3-parent</artifactId>
+        <groupId>org.jdbi</groupId>
+        <version>3.6.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>jdbi3-spring5</artifactId>
```
stevenschlansker: we probably should not publish this artifact, lest people search on `search.maven.org` and see, "oh, of course I should use `jdbi3-spring5`!"

spring5/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+    <description>jdbi Spring 4/5 forward compatibiility testing</description>
+    <url>http://jdbi.org/</url>
+
+    <properties>
+        <dep.spring.version>5.1.3.RELEASE</dep.spring.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework</groupId>
```
stevenschlansker: compile scope dependency on an artifact with no compiled sources

spring5/pom.xml
TheRealMarnes: I thought it wouldn't generate an artifact without any sources to compile into it... :/

TheRealMarnes:
Gave your feedback a try, dunno if it'll make the cut

spring5/pom.xml
TheRealMarnes: I suppressed the install and deploy plugins further below, does that resolve this or is there something ekse?

spring5/pom.xml
TheRealMarnes: I set it to test scope

TheRealMarnes:
Still seems to be getting that weird build timeout

TheRealMarnes:
Uh, I hope the build works...

PR #1360, opened 2018-12-20
TimestampedConfig to customize timezone

TheRealMarnes:
As discussed, only a minor change.

Instead of _only_ assuming and writing the system timzone to the db in the generated offsetdatetime, the user now can now also configure which timezone they want it to be


TheRealMarnes:
Also includes overdue devdoc and removal of deprecated code from the test

PR #1361, opened 2018-12-20
Support defining `null` values on templates.

stevenschlansker:
This is legitimate and useful, for example ST will emit
`context [anonymous] 1:119 attribute id isn't defined`
which you can avoid by defining a null value.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
@@ -17,11 +17,13 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.time.temporal.ChronoUnit;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 
 import javax.annotation.Nullable;
+
 import org.jdbi.v3.core.config.JdbiConfig;
 import org.jdbi.v3.meta.Beta;
 
@@ -39,7 +41,7 @@ public final class SqlStatements implements JdbiConfig<SqlStatements> {
     private final Collection<StatementCustomizer> customizers = new CopyOnWriteArrayList<>();
 
     public SqlStatements() {
-        attributes = new ConcurrentHashMap<>();
+        attributes = Collections.synchronizedMap(new HashMap<>());
         templateEngine = new DefinedAttributeTemplateEngine();
         sqlParser = new ColonPrefixSqlParser();
         sqlLogger = SqlLogger.NOP_SQL_LOGGER;
@@ -47,7 +49,7 @@ public final class SqlStatements implements JdbiConfig<SqlStatements> {
     }
 
     private SqlStatements(SqlStatements that) {
-        this.attributes = new ConcurrentHashMap<>(that.attributes);
+        this.attributes = Collections.synchronizedMap(that.getAttributes()); // already copied
         this.templateEngine = that.templateEngine;
         this.sqlParser = that.sqlParser;
         this.sqlLogger = that.sqlLogger;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -101,7 +104,9 @@ public Object getAttribute(String key) {
      * @return the defined attributes.
      */
     public Map<String, Object> getAttributes() {
-        return attributes;
+        synchronized (this) {
+            return new HashMap<>(attributes);
```
qualidafial: This will break anybody who calls e.g. `getAttributes().put(key, value)`

qualidafial:
Could probably replace all the `synchronized` blocks by just wrapping the map in `Collections.synchronizedMap()`

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
stevenschlansker: Yeah, it's true, although that's a super suspect use.  Returning a mutable map essentially requires us to continue using `ConcurrentHashMap` (synchronized map isn't good enough -- you need to know that e.g. you have to synchronize iteration) which forbids null values.

We could fix this with a null guard value.  That likely increases the complexity a fair bit.  Do you think it's worth it?  We never promised anything about the mutability or updateability of the returned collection, and as far as I know only JAXB is crazy enough to assume it.

stevenschlansker:
It doesn't accept null values.  I've switched to using `synchronizedMap` although it's essentially equivalent to the code as it stood before.  What do you not like about this change?  Currently, we are finding that the dominating factor in our test cases' performance is blocked on unbuffered `stderr` io printing out thousands and thousands of these errors, so it's not really a hypothetical concern to me.  If you can think of a better way to fix it I'm all ears.

TheRealMarnes:
The change I don't like is how you turn a threadsafe map into a non-safe one and then add sync locks to make it safe again. That's very odd to see.

I'm sorry, I'm missing something in your explanation, I don't see what the issue is :/
image

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
TheRealMarnes: tbh I'm with steven on that. It's best to not assume you can just `get()` a collection and then `add()` stuff to it with any useful result.

TheRealMarnes:
If you _must_ do the plain-map-with-a-sync approach, I'd prefer to see a meaningfully named ReentrantLock instead of `synchronized` to make it clear why we're doing such a strange thing. As for synchronizing on iterations, you're never supposed to be able to safely old-style iterate a threadsafe collection, there's `forEach` to do that.

stevenschlansker:
If there is a concurrent implementation that holds null values I'm all ears.  I didn't see any in the standard java library.

Most of our configuration classes are already (intentionally!) not thread-safe:
>  Generally, configuration should be set on a context before that context is used, and not changed later. Some configuration classes may be thread safe but most are not.

> As for synchronizing on iterations, you're never supposed to be able to safely old-style iterate a threadsafe collection, there's forEach to do that.

Right, but my point is that if someone wrote, old-style:

```java
for (Entry<String, Object> attr : ctx.getAttributes().entrySet()) {
    System.out.println(attr);
}
```

I didn't want making this change to expose the attributes map suddenly to inconsistent synchronization.  Copying it is safer.

> I'm sorry, I'm missing something in your explanation, I don't see what the issue is :/

You're missing half the story there.  By switching to a synchronized HashMap, we can store null values.  The error was never coming from the Map itself (the previous code just called `remove`) the problem is coming from StringTemplate.  While rendering, it is considered an error to lookup a value that is not defined, but it is *not* an error to lookup a value that is defined but null.  And that error is hardcoded to print to `stderr` right now, which is annoying.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
qualidafial: Ok. I would suggest returning `Collections.unmodifiableMap()` in that case, to cement that assumption.

stevenschlansker:
I think you must have missed my updated changeset -- it uses `synchronizedMap`.

PR #1362, opened 2018-12-20
SqlArrayType: introduce a factory, which simplifies using it nicely

stevenschlansker:
I was trying to register my own array type, and it's kinda a pain right now, as the test shows...

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -11,6 +11,9 @@
     - ParsedSql can be created with ParsedSql.of(String, ParsedParameters) factory
     - ParsedParameters can be created with ParsedSql.positional(int) and
       ParsedSql.named(List<String>) factories.
+    - SQL array registration improvements:
+      - SqlArrayType.of(String, Function) / SqlArrayTypeFactory.of(Class, String, Function) factory methods
+      - Configurable.registerArrayType(Class, String, Function) convenience method
   - New beta API
     - Type qualifiers for binding and mapping. Use annotations to distinguish between different SQL
       data types that map to the same Java type. e.g. VARCHAR, NVARCHAR, and Postgres MACADDR all
--- a/core/src/main/java/org/jdbi/v3/core/array/SqlArrayType.java
+++ b/core/src/main/java/org/jdbi/v3/core/array/SqlArrayType.java
@@ -13,6 +13,8 @@
  */
 package org.jdbi.v3.core.array;
 
+import java.util.function.Function;
+
 /**
  * Strategy for converting elements of array-like arguments into SQL array elements.
  *
@@ -35,4 +37,15 @@ public interface SqlArrayType<T> {
      * @return the converted element
      */
     Object convertArrayElement(T element);
+
+    /**
+     * Create a SqlArrayType from the given type and convert function.
+     *
+     * @param typeName the vendor sql type to use
+     * @param conversion convert the array element to the jdbc representation
+     * @return the created array type
+     */
+    static <T> SqlArrayType<T> of(String typeName, Function<T, ?> conversion) {
+        return new SqlArrayTypeImpl<T>(typeName, conversion);
+    }
 }
--- a/core/src/main/java/org/jdbi/v3/core/array/SqlArrayTypeFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/array/SqlArrayTypeFactory.java
@@ -15,8 +15,10 @@ package org.jdbi.v3.core.array;
 
 import java.lang.reflect.Type;
 import java.util.Optional;
+import java.util.function.Function;
 
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericTypes;
 
 /**
  * Factory interface to produce {@link SqlArrayType} instances.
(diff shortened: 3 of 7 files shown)


stevenschlansker:
Oops, this loses the generic type information.  Hm...

RELEASE_NOTES
```diff
@@ -11,6 +11,7 @@
     - ParsedSql can be created with ParsedSql.of(String, ParsedParameters) factory
     - ParsedParameters can be created with ParsedSql.positional(int) and
       ParsedSql.named(List<String>) factories.
+    - SqlArrayType.of(String, Function) / SqlArrayTypeFactory.of(Class, String, Function) factory methods
```
TheRealMarnes: is this enough information? Seems like a note only frequent users of SqlArrayType will find meaning in

core/src/main/java/org/jdbi/v3/core/array/SqlArrayType.java
```diff
@@ -35,4 +37,11 @@
      * @return the converted element
      */
     Object convertArrayElement(T element);
+
+    /**
+     * Create a SqlArrayType from the given type and convert function.
+     */
+    static <T> SqlArrayType<T> of(String typeName, Function<T, Object> convert) {
```
TheRealMarnes: `conversion` works better semantically IMO (in all the other places too)

stevenschlansker:
Arrays are great for e.g. "multi-get"

```java
@SqlQuery("select * from table where id=any(:ids)")
List<Row> multiGet(Collection<UUID> ids);
```

TheRealMarnes:
So in that case you would register a factory for UUID.class with UUID;;toString as the converter and "uuid" as the typename? And then in the fluent api you can specify you want to bind an array?

stevenschlansker:
Actually in the case of postgres, the JDBC driver understands the java `UUID` type, so you can just use `Function.identity()`.  Also it works with `Collection` types in addition to arrays.  But yep that's otherwise right.

qualidafial:
This probably merits a `Configurable.registerArrayType` override.

stevenschlansker:
added

RELEASE_NOTES
qualidafial: would add line for new `Configurable.registerArrayType` method too

core/src/main/java/org/jdbi/v3/core/array/SqlArrayType.java
```diff
@@ -35,4 +37,14 @@
...
      */
     Object convertArrayElement(T element);
+
+    /**
+     * Create a SqlArrayType from the given type and convert function.
+     *
+     * @param typeName the vendor sql type to use
+     * @conversion convert the array element to the jdbc representation
+     */
+    static <T> SqlArrayType<T> of(String typeName, Function<? super T, ?> conversion) {
```
qualidafial: Not sure about the `? super T` wildcard. e.g. if you call with a lambda:

```java
SqlArrayType.<MyType>of(
    "mytype",
    value -> /* can't call MyType methods because the lower bound is Object */)
```

core/src/main/java/org/jdbi/v3/core/array/SqlArrayTypeFactory.java
```diff
@@ -32,4 +34,18 @@
...
+    /**
+     * Create a SqlArrayTypeFactory for the given {@code elementType} that binds using a
+     * type name and convert function.
+     * @param elementType the array element type
+     * @param typeName the database type name to bind as
+     * @param conversion the function to convert to jdbc representation
+     * @return the built factory
+     * @see SqlArrayType#of(String, Function)
+     */
+    static <T> SqlArrayTypeFactory of(Class<T> elementType, String typeName, Function<? super T, ?> conversion) {
```
qualidafial: ditto for this conversion function

core/src/main/java/org/jdbi/v3/core/array/SqlArrayTypes.java
```diff
@@ -72,7 +73,8 @@ public SqlArrayTypes setArgumentStrategy(SqlArrayArgumentStrategy argumentStrate
      * @return this
      */
     public SqlArrayTypes register(Class<?> elementType, String sqlTypeName) {
-        return register(VendorSupportedArrayType.factory(elementType, sqlTypeName));
+        final Optional<SqlArrayType<?>> arrayType = Optional.of(SqlArrayType.of(sqlTypeName, Function.identity()));
```
qualidafial: Shouldn't we use `SqlArrayTypeFactory.of()` here instead of duplicating the type check?

core/src/test/java/org/jdbi/v3/core/array/TestCustomArrayType.java
```diff
@@ -30,7 +30,7 @@
 public class TestCustomArrayType {
 
     private void init(Jdbi db) {
-        db.registerArrayType(new UserIdArrayType());
+        db.registerArrayType(SqlArrayTypeFactory.of(UserId.class, "int", UserId::getId));
```
qualidafial: Can we unroll this to use the new `Configurable` method instead?

core/src/main/java/org/jdbi/v3/core/array/SqlArrayType.java
stevenschlansker: Ah, bummer.  I was thinking that accepting any supertype is fine, for example a `Function<Object, Object>` is just fine for binding a `String`.  But if it breaks lambdas, that's not ok...

PR #1365, opened 2018-12-21
Allow AnnotationType.class in place of @AnnotationType for zero-attribute annotation types

### LINKED ISSUE #1363: Enable qualifiers to be used through the fluent API
Atm qualifiers can only be used in SqlObject. In the fluent API it requires the use of an `internal` API.

Is it possible to wrap each given value/type in a QualifiedType that inspects the target class for qualifiers? Or a convenient way to enable qualifiers in the fluent api per invocation of e.g. `bind`?

qualidafial:
Fixes #1363 

* Add `QualifiedType.of(type, Class<? extends Annotation>...)` overloads
* Add  `ResultBearing.of(type, Class<? extends Annotation>...)` overloads
* Remove `Qualifiers.nVarchar()`. Just use `QualifiedType.of(String.class, NVarchar.class)` now.
* Remove `PostgresQualifiers.hStore()` and `.macAddr()`. Use `HStore.class`, and `MacAddr.class` now, respectively.


TheRealMarnes:
@stevenschlansker see, the spring5 thing is still making the build time out :/

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
```diff
@@ -55,13 +56,24 @@ public static QualifiedType of(Type type, Annotation qualifier) {
...
+    }
+
     /**
      * Creates a QualifiedType for {@code type} with the given qualifiers.
      */
     public static QualifiedType of(Type type, Annotation... qualifiers) {
         return of(type, Arrays.asList(qualifiers));
     }
 
+    public static QualifiedType of(Type type, Class<? extends Annotation>... qualifiers) {
```
TheRealMarnes: missing javadoc?

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
```diff
@@ -93,6 +105,13 @@ public static QualifiedType of(GenericType<?> type, Annotation... qualifiers) {
         return QualifiedType.of(type.getType(), qualifiers);
     }
 
+    /**
+     * Creates a QualifiedType for {@code type} with the given qualifiers.
+     */
+    public static QualifiedType of(GenericType<?> type, Class<? extends Annotation>... qualifiers) {
```
TheRealMarnes: having so many overloads to squarely combine Type and GenericType with the various forms of Annotation feels pretty bloated. Wouldn't a style like `QualifiedType.of(type).with(annotations)` with a copy-setter be nicer?

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
qualidafial: whoops

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
qualidafial: :thinking:

qualidafial:
I can remove the overloads in `QualifiedType` without much difficulty, but I don't see an easy way to do that in `ResultBearing.mapTo()` overloads. Maybe we just support the `Class<? extends Annotation>...` variants and assume that the others will just pass in a `QualifiedType`?

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
qualidafial: Done

PR #1366, opened 2018-12-21
Revert "Merge pull request #1352 from jdbi/spring5"

TheRealMarnes:
build fix for mystical timeout during mvn build


TheRealMarnes:
@qualidafial @stevenschlansker I'm just gonna merge it as soon as the build finishes so master gets green again, and then we should look into why this change 90% of the time makes the build time out

stevenschlansker:
It was green when I merged it :(

screen shot 2018-12-21 at 10 55 26 am

TheRealMarnes:
I know :/ It fails and succeeds randomly

PR #1368, opened 2018-12-21
Introduce defineBinds, which uses the bindings to create definitions for templates

stevenschlansker:
This is useful when e.g. using `bindBean` and templating to generate optional updates like

`<if(name)>name = :name<endif>`

where you want to use the same properties both as definitions and bindings


core/src/test/java/org/jdbi/v3/core/statement/TestDefineBinds.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.Test;
+
+public class TestDefineBinds {
+    @Rule
+    public H2DatabaseRule db = new H2DatabaseRule();
+
+    @Test
+    public void testDefineBinds() {
+        assertThat(
+            db.getSharedHandle().createQuery("<a><b>lect tru<b> from values(:a, :b)")
```
TheRealMarnes: I don't see the importance of the `values` part?

stevenschlansker:
It's not as useful with a single argument, but imagine you use either `bindBean` or you have a list of ten parameters.

`@Bind @Define` gets a bit tiring, and `@BindBean` would then need parallel `@DefineBean` etc.  And with Immutables it then extends to `@BindProperties` also needs `@DefineProperties`... this gives us a way to fix it once for all those cases.

core/src/test/java/org/jdbi/v3/core/statement/TestDefineBinds.java
stevenschlansker: It's an error to bind unused parameters, so this ensures jdbi considers them as used (otherwise we fail the test case)

stevenschlansker:
Is that a sufficient justification or would you like to explore alternative implementation ideas further before merging?

TheRealMarnes:
I don't think there are alternatives: either you repeat the calls/annotations or you use a both-in-one. The names are a bit silly and just as wordy as repetition to me, but using any other word would introduce extra jargon, so nah. I'm fine with this.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -1471,6 +1472,16 @@ public final This defineList(String key, List<?> values) {
         return define(key, value);
     }
 
+    /**
+     * Define all bound arguments that don't already have a definition.
+     * @return this
+     */
+    @Beta
+    public This defineBinds() {
```
qualidafial: This name is not intuitive to me. Maybe `defineNamedBindings()`?

core/src/main/java/org/jdbi/v3/core/statement/internal/DefineBindsTemplateEngine.java
```diff
@@ -0,0 +1,41 @@
...
+        this.delegate = delegate;
+    }
+
+    @Override
+    public String render(String template, StatementContext ctx) {
+        final Set<String> binds = new HashSet<>(ctx.getBinding().getNames());
+        binds.removeAll(ctx.getAttributes().keySet());
+        for (String boundButNotDefined : binds) {
+            ctx.getBinding().findForName(boundButNotDefined, ctx)
+                .map(a -> a instanceof NullArgument ? null : a)
```
qualidafial: Meh: `.filter(a -> !(a instanceof NullArgument))` would be equivalent.

core/src/main/java/org/jdbi/v3/core/statement/Binding.java
```diff
@@ -77,10 +84,16 @@ public void addNamedArgumentFinder(NamedArgumentFinder args) {
     }
 
     /**
-     * Look up an argument by position
+     * @return the set of known binding names
+     */
+    public Collection<String> getNames() {
+        return Collections.unmodifiableSet(names);
```
qualidafial: This could be calculated on invocation instead of unconditionally allocating a HashSet for it for every statement.

core/src/main/java/org/jdbi/v3/core/statement/internal/DefineBindsTemplateEngine.java
```diff
@@ -0,0 +1,41 @@
...
+                .ifPresent(a -> ctx.define(boundButNotDefined, a));
```
qualidafial:
```java
String name = "Robert'); drop table students;--";
handle.createUpdate(sql)
    .bind("name", name)
    .execute();
```

Making an unsanitized argument value available to the template engine is a major SQL injection attack vector.

What do you think about defining attribute value instead as a boolean `false` if the argument is null, `true` if non-null:

```java
ctx.getBinding().findForName(boundButNotDefined, ctx)
   .ifPresent(a -> ctx.define(boundButNotDefined, !(a instanceof NullArgument)));
```

This way a full-fledged template engine has a Boolean property to key off of, but we're not opening up our users to get pwned by a maliciously crafted parameter.

stevenschlansker: It's true, all I really need is the presence or absence of the value.  I could make it an option, with the default to just bind a boolean.  Or I could just hardcode it to always map to boolean, which is less useful generally, but still solves my particular use case.

core/src/main/java/org/jdbi/v3/core/statement/internal/DefineBindsTemplateEngine.java
stevenschlansker: It is subtly different, in that if there is an existing binding under the name, the code as-is will clear it.  Changing to `filter` means an existing definition will survive.  But I was not relying on this behavior, and your suggestion reads better, so I agree that's a positive change.

stevenschlansker: Nevermind, I'm wrong.  `Optional.map` turns `null` into empty.  I was thinking of `Stream.map`.

core/src/main/java/org/jdbi/v3/core/statement/internal/DefineBindsTemplateEngine.java
stevenschlansker: I've made it configurable, and the default is to only bind it as a boolean.  Hopefully that's in line with your suggestion.

core/src/main/java/org/jdbi/v3/core/statement/internal/DefineBindsTemplateEngine.java
TheRealMarnes: Doesn't this presume user-made factories will re-use our `NullArgument`?

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingMode.java
```diff
@@ -0,0 +1,51 @@
...
+@Beta
+public enum DefineNamedBindingMode {
+    /**
+     * Bind a boolean {@true} for present values, {@false} for null values.
+     * Default behavior.
+     */
+    BOOLEAN {
+        @Override
+        Optional<Object> apply(Argument arg) {
+            return Optional.of(arg instanceof NullArgument ? false : true);
```
TheRealMarnes: doesn't this presume user-made factories will re-use our `NullArgument`?

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsTemplateEngine.java
```diff
@@ -0,0 +1,39 @@
...
+        this.delegate = delegate;
+    }
+
+    @Override
+    public String render(String template, StatementContext ctx) {
+        final Set<String> binds = new HashSet<>(ctx.getBinding().getNames());
+        binds.removeAll(ctx.getAttributes().keySet());
+        for (String boundButNotDefined : binds) {
+            ctx.getBinding().findForName(boundButNotDefined, ctx)
+                .flatMap(mode::apply)
```
TheRealMarnes: I'm not seeing the point of the Optional here. We call `ctx.getBinding().getNames()` to get the name, so `ctx.getBinding().findForName` will never be able to return `empty()`, will it? I think it should be `orElseThrow` because we expect a value always.

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingMode.java
stevenschlansker: They should try to reuse `NullArgument` if possible.  You're right that this is a limitation, but I don't really see any other way to accomplish what I want here.

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsTemplateEngine.java
stevenschlansker: It's legitimate to implement a `NamedArgumentFinder` that has a name but is unable to find a value.  I don't see why you would do that, but if you do, this method should ignore it rather than exploding I think.

stevenschlansker:
@qualidafial I believe I've handled all your comments, any further thoughts?  Especially note that I added a new `StatementCustomizer` extension point which I think is a more flexible and reusable implementation than the original wrapping `TemplateEngine` approach.

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingMode.java
stevenschlansker: Added a note

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsTemplateEngine.java
TheRealMarnes: If you're certain an empty return can be legit, then ok. My concern was the functional expression of allowing empty values to go unnoticed, not so much whether they should never be empty or not.

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingMode.java
```diff
@@ -28,6 +28,9 @@
     /**
      * Bind a boolean {@true} for present values, {@false} for null values.
      * Default behavior.
+     *
+     * Note that due to implementation limitations, null arguments must be instances of
+     * {@link org.jdbi.v3.core.argument.NullArgument} to be recognized.
```
TheRealMarnes: People only implementing ArgumentFactories aren't likely to see this though. I think the note should go on the argfac interface and in the devdoc too.

stevenschlansker: But the limitation is specifically related to this strategy only.  It doesn't affect other users.  So shouldn't the note be local to the feature it affects?  The ArgumentFactory interface can't possibly be expected to document its interaction with all other features, otherwise it will be a dumping ground for everything.

TheRealMarnes: At least a `caution` in the devdoc (devguide) regarding argumentfactories then. Not just to say "use NullArgument to make feature x and y work", but generally "you really should use our NullArgument to ensure some features keep working".

TheRealMarnes: I don't think it's really unfair to have to javadoc it there. Offical features relying on it means it is part of the contract any ArgumentFactory must comply with, so an implementor must know.

stevenschlansker: I added a note to ArgumentFactory.

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsTemplateEngine.java
stevenschlansker: I'm not certain it's legit.  But it doesn't feel like this is the right place to start enforcing a new policy like that, and not mixed into an otherwise unrelated PR.

stevenschlansker: If the binding is truly accidentally forgotten, it should blow up later during binding.

core/src/main/java/org/jdbi/v3/core/statement/internal/DefineBindsTemplateEngine.java
qualidafial: I'm ok with that

qualidafial: Please add a warning about SQL injection to the interface and the config parameter setter so implementers are given every opportunity not to shoot themselves in the foot.

stevenschlansker:
@qualidafial any last thoughts before merge?

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingMode.java
```diff
@@ -0,0 +1,53 @@
...
+        @Override
+        Optional<Object> apply(Argument arg) {
+            return Optional.of(arg instanceof NullArgument ? false : true);
+        }
+    },
+    /**
+     * Bind the Argument's {@code toString} value as the template parameter.
+     * Please be cautious not to allow SQL injection!
+     */
+    TO_STRING {
```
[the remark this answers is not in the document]
TheRealMarnes: The eternal plague of existing code influencing the evolution of new code, instead of making new things from scratch without bias from outside...

core/src/main/java/org/jdbi/v3/core/argument/DelegatingArgumentFactory.java
```diff
@@ -43,6 +44,6 @@
     }
 
     <T> void register(Class<T> klass, int sqlType, StatementBinder<T> binder) {
-        builders.put(klass, (T value) -> new LoggableSetNullOrBinderArgument<>(value, sqlType, binder));
+        builders.put(klass, (T value) -> value == null ? new NullArgument(sqlType) : new LoggableBinderArgument<>(value, binder));
```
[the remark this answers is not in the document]
stevenschlansker: I don't love it.  But it works, and we can always fix it later if we come up with a better approach.

core/src/main/java/org/jdbi/v3/core/statement/Binding.java
```diff
@@ -77,10 +94,21 @@ public void addNamedArgumentFinder(NamedArgumentFinder args) {
     }
 
     /**
-     * Look up an argument by position
+     * @return the set of known binding names
+     */
+    @SuppressFBWarnings("NP_NULL_PARAM_DEREF") // false positive
+    public Collection<String> getNames() {
+        if (names == null) {
+            names = new HashSet<>(named.keySet());
```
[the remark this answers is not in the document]
stevenschlansker: Heh, I misunderstood you.  I did think it was a bit overly complicated for how it's actually getting used...

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingMode.java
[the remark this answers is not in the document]
stevenschlansker: Hm, I guess that works.  I don't disagree with your reasoning however there is only so far we can go to protect users from themselves.  `define` in general is suspect to this sort of issue and this isn't something new, you can get yourself in trouble just the same with defining any `String` values at all.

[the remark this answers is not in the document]
stevenschlansker: And we (almost) have to make them `enum`s -- otherwise you can't attach them to the annotation.  Unfortunately it's not possible to send e.g. an arbitrary Function in.  We'd have to take a `Class` and specify it has a nullary constructor, etc, much like `JdbiConfig`, which is also a big bummer.  Ends up being a fair amount of boilerplace to support something that almost nobody will extend.  I'm not necessarily against changing it, just pointing out that there is a nontrivial reason we went with the `enum` type, and if we are going to object to the only other implementation maybe I should just make it hard-coded for now until anyone actually asks for it...

core/src/main/java/org/jdbi/v3/core/argument/BeanPropertyArguments.java
```diff
@@ -59,6 +61,12 @@
         })
         .build();
 
+    private static boolean shouldSeeProperty(PropertyDescriptor pd) {
+        // 'class' isn't really a property
+        final Method read = pd.getReadMethod();
+        return read == null || read.getDeclaringClass() != Object.class;
```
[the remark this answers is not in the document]
stevenschlansker: I agree we might want to change this, but this mimics the existing behavior best.  For example if you only provide a setter, the BeanIntrospector will find a property with a null getter.  I experimented by changing this as you suggest and it breaks some existing tests, notably:

```java
    @Test
    public void testBindNonPublicGetter() {
        assertThatThrownBy(() -> new BeanPropertyArguments("foo", new NonPublicGetter()).find("foo.bar", ctx))
            .isInstanceOf(UnableToCreateStatementException.class);
    }
```
we test that having a setter without a getter causes an explosion.  Ignoring non-gettable properties changes this behavior.  If we do accept it as a property here, we later blow up since we can't actually read it, preserving existing behavior.

core/src/main/java/org/jdbi/v3/core/argument/DelegatingArgumentFactory.java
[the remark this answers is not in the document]
stevenschlansker: It also highlighted that the `LoggableSetNullOrBinderArgument` class was really identical to the code we already have, if you pull out the `if` statement.  The dangers of overly aggressive cleanup!

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingMode.java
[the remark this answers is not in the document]
stevenschlansker: I've dropped the customizable implementation, now it only defines booleans for now.  We can always add it back later when we understand its uses better.  #YAGNI

stevenschlansker:
@qualidafial , I've fixed up the `getNames` code and dropped the enum entirely.  Now it only binds booleans if the parameter is present, since that was my original use.  I'll leave it as a later extension to add customizable strategies if we find it gets used.  Look ok now?

docs/src/adoc/index.adoc
```diff
@@ -684,6 +684,12 @@ handle.createQuery("select * from emotion where emoticon = <sticker>;")
     .list();
 ----
 
+Bindings and definitions are usually separate.  You can define all bindings that don't already
+have template definitions using the `stmt.defineNamedBindings()` or `@DefineNamedBindings` customizers.
+
+By default, to prevent SQL injection, this will bind a boolean indicating only whether the name was bound or not.
```
qualidafial: the boolean indicates whether the bound parameter is non-null, not whether it was bound

docs/src/adoc/index.adoc
```diff
@@ -684,6 +684,12 @@ handle.createQuery("select * from emotion where emoticon = <sticker>;")
...
+If you need the full string value, see the `DefineNamedBindingsMode` enum.
```
qualidafial: This is obsolete

stringtemplate4/src/test/java/org/jdbi/v3/core/statement/TestDefineNamedBindingsST.java
```diff
@@ -0,0 +1,48 @@
...
+
+public class TestDefineNamedBindingsST {
+    @Rule
+    public H2DatabaseRule db = new H2DatabaseRule();
+
+    @Test
+    public void testDefineBoolean() {
+        db.getSharedHandle().setTemplateEngine(new StringTemplateEngine());
+        assertThat(
+            db.getSharedHandle().createQuery("select <a> from values(:a<if(b)>, :b shouldn't be defined<endif>)")
```
qualidafial: What about a test where the conditional is in the WHERE clause? e.g.

```
<if(b)>WHERE b = :b<endif>
```

qualidafial:
Just a few more comments, otherwise I'm happy with this.

docs/src/adoc/index.adoc
TheRealMarnes: > by default

implies there's a choice in the matter, but — I may be overlooking it — I don't see an enum or implementable interface to have other behavior?

docs/src/adoc/index.adoc
```diff
@@ -684,6 +684,10 @@ handle.createQuery("select * from emotion where emoticon = <sticker>;")
     .list();
 ----
 
+Bindings and definitions are usually separate.  You can link them in a limited manner
+using the `stmt.defineNamedBindings()` or `@DefineNamedBindings` customizers, which will
+define a boolean indicating whether the value is present and not a `NullArgument` (bind of a `null`).
```
TheRealMarnes: I think users who know nothing about this feature won't be informed clearly enough by this paragraph. `define a boolean` without a `for each` of some sorts implies a single boolean, and `the` in `the value` refers to something unspecified.

Maybe something like this?

>Bindings and definitions are usually independent. You can use the `stmt.defineNamedBindings()` or `@DefineNamedBindings` customizers to make e.g. conditionals easier for `TemplateEngines` to process: a definition/attribute will be added for each binding, with the same name (unless something is already defined with said name), and with the value being a boolean that indicates if the bound object is a `NullArgument` (`false`) or not (`true`).

stevenschlansker: I used some of your wording and included a concrete example

stevenschlansker:
:fireworks:

PR #1369, opened 2018-12-22
Qualifiers benchmarking

stevenschlansker:
second pass at results,

```
# JMH version: 1.21
# VM version: JDK 11.0.1, OpenJDK 64-Bit Server VM, 11.0.1+13
# VM invoker: /usr/lib/jvm/java-11-openjdk-11.0.1.13-4.fc29.x86_64/bin/java
# VM options: <none>
# Warmup: 5 iterations, 2 s each
# Measurement: 5 iterations, 5 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
```

Base qualifier benchmark, no cache:
```
Benchmark                                     Mode  Cnt      Score     Error  Units
QualifiersBenchmark.getQualifiersAnnotated    avgt    5    218.109 ±   7.360  ns/op
QualifiersBenchmark.getQualifiersUnannotated  avgt    5    141.886 ±   9.804  ns/op
QualifiersBenchmark.mapQualifiedBean          avgt    5  22283.394 ± 601.425  ns/op
QualifiersBenchmark.mapUnqualifiedBean        avgt    5  22025.249 ± 659.755  ns/op
```

Benchmark again, with added qualifier cache:
```
Benchmark                                     Mode  Cnt      Score      Error  Units
QualifiersBenchmark.getQualifiersAnnotated    avgt    5     29.904 ±    0.754  ns/op
QualifiersBenchmark.getQualifiersUnannotated  avgt    5     29.448 ±    0.610  ns/op
QualifiersBenchmark.mapQualifiedBean          avgt    5  19843.370 ± 1592.108  ns/op
QualifiersBenchmark.mapUnqualifiedBean        avgt    5  19739.789 ±  204.387  ns/op
```

Looks like there's a clear win on the micro scale.


TheRealMarnes:
```
Benchmark                                     Mode  Cnt      Score      Error  Units
QualifiersBenchmark.getQualifiersAnnotated    avgt    5    267.734 ±   56.382  ns/op
QualifiersBenchmark.getQualifiersUnannotated  avgt    5    147.198 ±    4.463  ns/op
QualifiersBenchmark.mapQualifiedBean          avgt    5  30892.585 ± 2680.355  ns/op
QualifiersBenchmark.mapUnqualifiedBean        avgt    5  30566.726 ± 4941.357  ns/op
```
```
Benchmark                                     Mode  Cnt      Score     Error  Units
QualifiersBenchmark.getQualifiersAnnotated    avgt    5     47.285 ±   1.672  ns/op
QualifiersBenchmark.getQualifiersUnannotated  avgt    5     46.055 ±   0.550  ns/op
QualifiersBenchmark.mapQualifiedBean          avgt    5  27049.600 ± 683.405  ns/op
QualifiersBenchmark.mapUnqualifiedBean        avgt    5  27115.527 ± 873.620  ns/op
```

stevenschlansker:
Caching Strategy 2:

```
Benchmark                                     Mode  Cnt      Score     Error  Units
QualifiersBenchmark.getQualifiersAnnotated    avgt    5     24.996 ±   1.731  ns/op
QualifiersBenchmark.getQualifiersUnannotated  avgt    5     25.038 ±   0.768  ns/op
QualifiersBenchmark.mapQualifiedBean          avgt    5  17713.706 ± 392.089  ns/op
QualifiersBenchmark.mapUnqualifiedBean        avgt    5  17656.570 ± 910.178  ns/op
```

qualidafial:
This is nifty. If we get the build working (and ensure the benchmark module isn't deployed) I think this will make a great addition.

TheRealMarnes:
We'd need to add A vs B implementations of everything we want to test to the module, right? Or is there another idea?

qualidafial:
Used for A/B comparison, but can also be used to detect performance changes from release to release.

As we put together a meaningful collection of benchmarks, we can start tracking performance changes over time.

stevenschlansker:
Depends on https://github.com/jdbi/jdbi/pull/1399 now

benchmark/src/main/java/org/jdbi/v3/core/qualifier/QualifiersBenchmark.java
```diff
@@ -0,0 +1,130 @@
+/*
+ * Copyright (c) 2014, Oracle America, Inc.
```
TheRealMarnes: what's with this license header? Why is it for oracle?

benchmark/src/main/java/org/jdbi/v3/core/qualifier/QualifiersBenchmark.java
```diff
@@ -0,0 +1,130 @@
...
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.AverageTime)
+@Measurement(time=5)
+@Warmup(time=2)
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@Fork(1)
+public class QualifiersBenchmark {
+
+    JdbiRule db;
+    Jdbi jdbi;
```
TheRealMarnes: I assume these can't be private because of the test framework?

benchmark/src/main/java/org/jdbi/v3/core/qualifier/BogusReference.java
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.qualifier;
+
+import org.junit.rules.ExternalResource;
+
+// We end up needing junit in compile scope since we borrow JdbiRule, so fool the dependency analyzer.
+public class BogusReference {
```
TheRealMarnes: Even shorter and smartassly semantic :trollface: to just do:
```
@Ignore
public class BogusReference {}
```

pom.xml
```diff
@@ -92,6 +93,7 @@
         <dep.plugin.checkstyle.version>3.0.0</dep.plugin.checkstyle.version>
         <dep.plugin.pmd.version>3.9.0</dep.plugin.pmd.version>
         <dep.plugin.sortpom.version>2.8.0</dep.plugin.sortpom.version>
+        <dep.plugin.spotbugs.version>3.1.10</dep.plugin.spotbugs.version>
```
TheRealMarnes: just a collateral improvement?

TheRealMarnes:
So what's the idea for this module? Do we exclude it from running during normal builds to not bog them down, and pile up A/B reference implementations for testing and documentation purposes in it?

pom.xml
stevenschlansker: Yeah, I tripped over some spotbugs issue and a version upgrade fixed it.  We were on a pretty old version.  I could make a separate PR if you like but it's literally this one line change.

benchmark/src/main/java/org/jdbi/v3/core/qualifier/QualifiersBenchmark.java
stevenschlansker: the `jmh` project generator left it, whoops.

stevenschlansker:
I don't think its necessary to suppress the build, unless you think there's a particular reason to.  Wouldn't that just mean that we're likely to break it all the time?  It's important to have CI test that the code compiles.

TheRealMarnes:
I'll reword my concern: we're not running every benchmark we ever set up during every build, are we?

stevenschlansker:
Benchmarks never run as part of the build.  It only produces an executable that may run the benchmarks, but you have to invoke it yourself.

stevenschlansker:
It's not used in any way other than ad-hoc right now.  The intent is we either start trending (by collecting results for every release on the same machine) or doing A/B tests before merging purported speed improvements.  Either way it shows as a demo of how to do it.

@qualidafial any concerns?  otherwise I'll merge it just to have it in-tree

benchmark/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+                                    <!--
+                                        Shading signed JARs will fail without this.
+                                        http://stackoverflow.com/questions/999489/invalid-signature-file-when-attempting-to-run-a-jar
+                                    -->
+                                    <artifact>*:*</artifact>
+                                    <excludes>
+                                        <exclude>META-INF/*.SF</exclude>
+                                        <exclude>META-INF/*.DSA</exclude>
+                                        <exclude>META-INF/*.RSA</exclude>
+                                    </excludes>
```
qualidafial: I swear there's a profile in basepom we could be using that does all this. I want to say it's activated from the existence of a file named `.build-executable`

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
```diff
@@ -28,6 +31,10 @@
  */
 @Beta
 public class Qualifiers {
+    private static final ExpiringMap<Object, Set<Annotation>> ANNOS = ExpiringMap.builder()
```
qualidafial: Nit: shouldn't the key parameter be `AnnotatedElement` instead of `Object`?

testing/src/main/java/org/jdbi/v3/testing/JdbiRule.java
```diff
@@ -134,7 +134,7 @@ public JdbiRule withPlugin(final JdbiPlugin plugin) {
     }
 
     @Override
-    protected void before() throws Throwable {
+    public void before() throws Throwable {
```
qualidafial: Strictly speaking, changing this from `protected` to `public` is a breaking change, since third-party implementors (if any) could already override it with `protected` access.

I think the fallout is minimal, but it's probably worth a mention in the release notes.

Issue #1374, opened 2018-12-25
Code error in the documentation


Ahmed-Itani:
Hi,

I have found a code error in the site documentation under section "Binding Arguments" for the below code. The link http://jdbi.org/#_binding_arguments_2 

**Documentation code**
```
@SqlQuery("select name from users where id in (<userIds>)")
List<String> getFromIds(@BindList("userIds") userIds)
```

**Working Code**
```
@SqlQuery("select name from users where id in (<userIds>)")
List<String> getFromIds(@BindList("userIds") List<Long> userIds)
```

TheRealMarnes:
We're a bit overdue for a release and this typo has been fixed in source 2 months ago, apparently :) Thanks for reporting!

PR #1375, opened 2018-12-25
 #1374 explain BindList a bit more

TheRealMarnes:
I thought this section should explain _why_ users need to do something as odd as using attribute syntax to perform what the API calls a binding. Came about because  #1374 made me look at this part of the manual. Recently voiced concern about the security of using attributes comes to mind too.

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -366,7 +366,7 @@ handle.createUpdate("insert into contacts (id, name) values (:id, :name)")
       .execute();
 ----
 
-You can bind multiple values at once, from either a `List<T>` or a list of arguments. Note that this requires your binding to be written using the `<boundValue>` form (as opposed to `:boundValue` for most of the other bindings.)
+You can bind multiple values at once, from either a `List<T>` or a vararg:
 
 [source,java]
 ----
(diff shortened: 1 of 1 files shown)


docs/src/adoc/index.adoc
```diff
@@ -360,7 +360,15 @@ handle.createUpdate("insert into contacts (id, name) values (:id, :name)")
...
 
-You can bind multiple values at once, from either a `List<T>` or a list of arguments. Note that this requires your binding to be written using the `<boundValue>` form (as opposed to `:boundValue` for most of the other bindings.)
+You can bind multiple values at once, from either a `List<T>` or a vararg.
+
+Note that this requires your binding to be written like an attribute (e.g. `<boundValue>`)
+as opposed to a binding (`:boundValue`). Jdbi will _bind_ each value in the passed collection under a
+generated key name, and _render_ the placeholder in the query to a comma-separated list of those keys
+using the configured `TemplateEngine`. At a later stage in query execution, the `SqlParser` will bind
+the rendered comma-separated list of keys to its values. This means you should *not* embed any SQL syntax
+like string delimiters in your values, and this feature carries no risks like SQL injection;
```
stevenschlansker: These two statements seem contradictory.  If "this feature carries no risks like SQL injection" then why does it "mean you should *not* embed any SQL syntax".  If there's no risk of injection, then it should be *safe* to have sql syntax embedded in your parameters.

TheRealMarnes: you don't need to include sql syntax like string delimiters, like you would need to in other cases of `define`ing a variable, because the actual value is _bound_ despite being `define`d. Yeah it's a tricky thing to explain.

TheRealMarnes: It will be safe, but your string will end up being `"'foo'"` instead of `"foo"` is the point

TheRealMarnes: The value is _bound_ in the usual safe and syntax-free way despite it being `define`d, which usually implies you have to provide delimiters, escapes, etc of your own

PR #1378, opened 2018-12-27
fix #1377: BindList conditionals with null and StringTemplate

### LINKED ISSUE #1377: Add another option to process null bind list
Current for this sql use BindList and StringTemplate4

```sql
@SqlQuery("select * from resource where app_id = :appId <if(resourceIds)> and resource_id in 
                     +"(<resourceIds>) <endif> "
@UseStringTemplateEngine 
List<Object> query(@Bind("appId") String appId,@Define("resourceIds") @BindList(value = "resourceIds", onEmpty = BindList.EmptyHandling.NULL) List<String> resourceIds)
```

when resourceIds input is null,the sql will render to
```sql
select * from resource where app_id=xxx and resource_id in (null)
```
but i want it render  to
```sql
select * from resource where app_id=xxx
```
currently a workaround way is convert input null to a empty list，but not everyone invoke this method know this trick

can BindList.EmptyHandling add another option？we can process null list like this: 
 ```sql  
stmt.define(name, null); 
```

### PR DESCRIPTION
**Pull request by TheRealMarnes**
fixes #1377 

I haven't really considered this change from a conservative or nitpicky point of view yet, but all this really is, is making a constant for `if (list == null) list = emptyList()`, so it can't really be a problem? The only "risk" to this is that you leave the text representation of an empty list in the query up to the TemplateEngine, but that's what allows it to treat it as a falsy conditional in the first place.

@shelocks

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
@@ -57,24 +57,24 @@ public @interface BindList {
         /**
          * <p>Output "" (without quotes, i.e. nothing).</p>
          *
-         * <code>select * from things where x in ()</code>
+         * {@code select * from things where x in ()}
          */
         VOID,
         /**
          * <p>Output "null" (without quotes, as keyword), useful e.g. in postgresql where "in ()" is invalid syntax.</p>
          *
-         * <code>select * from things where x in (null)</code>
+         * {@code select * from things where x in (null)}
          */
         NULL,
-        /**
-         * Throw IllegalArgumentException.
-         */
-        THROW,
         /**
          * <p>Define an empty list, leaving the resulting query text up to the {@link org.jdbi.v3.core.statement.TemplateEngine} to decide.</p>
          *
          * This value was specifically added to <a href="https://github.com/jdbi/jdbi/issues/1377">make conditionals work better with <code>StringTemplate</code></a>.
          */
-        EMPTY_LIST
+        EMPTY_LIST,
+        /**
+         * Throw IllegalArgumentException.
+         */
+        THROW
     }
 }
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
@@ -67,11 +67,11 @@ public @interface BindList {
          */
         NULL,
         /**
-         * <p>Define an empty list, leaving the resulting query text up to the {@link org.jdbi.v3.core.statement.TemplateEngine} to decide.</p>
+         * <p>Define a {@code null} value, leaving the resulting query text up to the {@link org.jdbi.v3.core.statement.TemplateEngine} to decide.</p>
          *
          * This value was specifically added to <a href="https://github.com/jdbi/jdbi/issues/1377">make conditionals work better with <code>StringTemplate</code></a>.
          */
-        EMPTY_LIST,
+        JAVA_NULL,
         /**
          * Throw IllegalArgumentException.
          */
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
@@ -71,7 +71,7 @@ public @interface BindList {
          *
          * This value was specifically added to <a href="https://github.com/jdbi/jdbi/issues/1377">make conditionals work better with <code>StringTemplate</code></a>.
          */
-        JAVA_NULL,
+        NULL_VALUE,
         /**
          * Throw IllegalArgumentException.
          */
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
@@ -55,25 +55,25 @@ public @interface BindList {
      */
     enum EmptyHandling {
         /**
-         * output "" (without quotes, i.e. nothing)
-         * <p>
-         * select * from things where x in ()
+         * <p>Output "" (without quotes, i.e. nothing).</p>
+         *
+         * <code>select * from things where x in ()</code>
          */
         VOID,
         /**
-         * output "null" (without quotes, as keyword), useful e.g. in postgresql where "in ()" is invalid syntax
-         * <p>
-         * select * from things where x in (null)
+         * <p>Output "null" (without quotes, as keyword), useful e.g. in postgresql where "in ()" is invalid syntax.</p>
+         *
+         * <code>select * from things where x in (null)</code>
          */
         NULL,
         /**
-         * throw IllegalArgumentException
+         * Throw IllegalArgumentException.
          */
         THROW,
         /**
-         * <p>define an empty list, leaving the resulting query text up to the {@link org.jdbi.v3.core.statement.TemplateEngine} to decide.</p>
+         * <p>Define an empty list, leaving the resulting query text up to the {@link org.jdbi.v3.core.statement.TemplateEngine} to decide.</p>
          *
-         * @apiNote this value was specifically added to <a href="https://github.com/jdbi/jdbi/issues/1377">make conditionals work better with <code>StringTemplate</code></a>
+         * This value was specifically added to <a href="https://github.com/jdbi/jdbi/issues/1377">make conditionals work better with <code>StringTemplate</code></a>.
          */
         EMPTY_LIST
     }
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
@@ -69,6 +69,12 @@ public @interface BindList {
         /**
          * throw IllegalArgumentException
          */
-        THROW;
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java [hunk 1]
```diff
@@ -55,20 +55,26 @@
...
+         *
+         * <code>select * from things where x in (null)</code>
          */
         NULL,
         /**
-         * throw IllegalArgumentException
+         * Throw IllegalArgumentException.
          */
-        THROW;
+        THROW,
```
stevenschlansker: I think non-exceptional behavior should go above `THROW` in the list.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
```diff
@@ -55,20 +55,26 @@
...
+        /**
+         * <p>Define an empty list, leaving the resulting query text up to the {@link org.jdbi.v3.core.statement.TemplateEngine} to decide.</p>
+         *
+         * This value was specifically added to <a href="https://github.com/jdbi/jdbi/issues/1377">make conditionals work better with <code>StringTemplate</code></a>.
+         */
+        EMPTY_LIST
```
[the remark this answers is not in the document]
TheRealMarnes: Out of the 3 options, it seems to be the clearest and most semantic: this is about BindList so we define a List value, skipping the definition is functionally misleading because we did high-level bind/define something so the Map should show that, and defining `null`... I dunno, it may work, I'll test it, but I don't like the way a null value entry corresponds in usage to the key not being present at all when a value check is done by the readers of the Map. It's misleading, you know? Is it null because it's not defined or because a null was put into it, you can't tell.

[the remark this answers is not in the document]
TheRealMarnes: Also, I had no idea what to name an option that defines `null` when we already called the `"null"` one `NULL` 🤷‍♂️ So since binding an empty list works just as well, and avoids the ever-dreaded `null`, I went with that...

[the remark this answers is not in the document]
TheRealMarnes: Just pushed a commit with this change. It works at least. I would prefer the empty list variant, but it's a soft preference. If you have a hard preference for the other one, I'll roll with it.

stringtemplate4/src/test/java/org/jdbi/v3/sqlobject/BindListTest.java
```diff
@@ -215,4 +208,35 @@ public void testSomethingByIteratorHandleDefault() {
...
+        List<String> names = dao.get(null);
+
+        assertThat(names)
+            .describedAs("ST did not evaluate null as truthy, query did not select by `name is null`")
+            .hasSize(4);
+    }
+
+    private interface ConditionalDao {
+        // `in (null)` doesn't work on h2
+        @SqlQuery("select name from something <if(name)> where name is <name> <endif>")
```
qualidafial: Should this be `name in (<name>)`?

qualidafial: Also, can we get a positive test case for this `<if()`> statement?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
```diff
@@ -55,20 +55,26 @@
...
          */
         NULL,
         /**
-         * throw IllegalArgumentException
+         * <p>Define a {@code null} value, leaving the resulting query text up to the {@link org.jdbi.v3.core.statement.TemplateEngine} to decide.</p>
+         *
+         * This value was specifically added to <a href="https://github.com/jdbi/jdbi/issues/1377">make conditionals work better with <code>StringTemplate</code></a>.
          */
-        THROW;
+        JAVA_NULL,
```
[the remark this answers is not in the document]
TheRealMarnes: `NULL_STRING` vs `NULL_VALUE` maybe? The java perspective on this null matters most, because of the templateengine using it.

stringtemplate4/src/test/java/org/jdbi/v3/sqlobject/BindListTest.java
TheRealMarnes: The comment on it explains why the query isn't `in ()`. It doesn't really matter, it's only about what BindList with StringTemplate renders to, not the query parsing. That's also why I have 2 control queries in the test, to demonstrate the essence of it.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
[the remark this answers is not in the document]
stevenschlansker: I just didn't see any evidence that the empty list was actually ever used as anything except a "falsy" value.  If our goal is to insert a falsy value, I expect `null` is more widely accepted as falsy, `"[]" == null` could easily be a StringTemplate-ism.

Also, I agree that inserting a literal null is icky, we can simply omit defining it in that case.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/internal/BindListFactory.java
```diff
@@ -49,10 +48,14 @@ public SqlStatementParameterCustomizer createForParameter(Annotation annotation,
                 case NULL:
                     stmt.define(name, "null");
                     return;
+                case JAVA_NULL:
+                    stmt.define(name, null);
```
stevenschlansker: We could consider simply no-op'ing right here.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/internal/BindListFactory.java
```diff
@@ -49,10 +48,14 @@ public SqlStatementParameterCustomizer createForParameter(Annotation annotation,
...
+                    return;
                 case THROW:
```
stevenschlansker: Should these behaviors be a dynamic method on the Enum, rather than a switch statement?  It eliminates the `default` case.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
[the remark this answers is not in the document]
TheRealMarnes: Yeah, it was intended as a StringTemplate-ism, hence the javadoc. We explicitly support StringTemplate, so I figured it ok.

I kinda don't wanna just omit the definition because it's still a null but now also only as a side-effect instead of explicit behavior.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/internal/BindListFactory.java
stevenschlansker: (And then the naming question is simpler, we could simply call it `NONE` or `NOOP` or something like that)

TheRealMarnes: I'd rather not because that makes it rely on circumstance instead of explicit behavior

TheRealMarnes:
remarks handled

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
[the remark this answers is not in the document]
stevenschlansker: Well, `map.put(key, null)` is not generally safe.  Depending on the map implementation it will throw.  If you really want to do something we should probably `map.remove(key)` instead, as that should always have the desired effect.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
```diff
@@ -55,20 +57,47 @@
      */
     enum EmptyHandling {
         /**
-         * output "" (without quotes, i.e. nothing)
-         * <p>
-         * select * from things where x in ()
+         * <p>Output "" (without quotes, i.e. nothing).</p>
```
stevenschlansker: Why all these `<p>` tags?  They really ugly up the javadoc.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/internal/BindListFactory.java
```diff
@@ -42,23 +39,10 @@ public SqlStatementParameterCustomizer createForParameter(Annotation annotation,
...
-                case THROW:
-                    throw new IllegalArgumentException(arg == null
-                    ? "argument is null; null was explicitly forbidden on this instance of BindList"
-                            : "argument is empty; emptiness was explicitly forbidden on this instance of BindList");
-                default:
-                    throw new IllegalStateException(VALUE_NOT_HANDLED_MESSAGE);
-                }
+                bindList.onEmpty().define(stmt, name);
+            } else {
+                stmt.bindList(name, IterableLike.toList(arg));
```
stevenschlansker: 👍 nice

sqlobject/src/main/java/org/jdbi/v3/sqlobject/customizer/BindList.java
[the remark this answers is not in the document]
TheRealMarnes: We'll change that when it happens then

PR #1384, opened 2018-12-31
Document LIKE clauses

### LINKED ISSUE #969: Developer guide: provide binding example for LIKE clauses with bound parameters.
Discussed on [mailing list](https://groups.google.com/forum/#!topic/jdbi/izn9__2a27w):

Jdbi ignores parameter tokens inside varchar literals, to match JDBC. This leaves some ambiguity about what to do with `LIKE` clauses using a bound parameter.

Add examples to the developer guide demonstrating varchar concatenation as the recommended approach:

* Substring search:

```sql
SELECT * FROM contacts
WHERE name LIKE '%' || name || '%'
```

* Prefix search:

```sql
SELECT * FROM contacts
WHERE name LIKE name || '%'
```

* Suffix search:

```sql
SELECT * FROM contacts
WHERE name LIKE '%' || name
```

TheRealMarnes:
fixes #969

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -597,6 +597,44 @@ LocalDate releaseDate = handle.createQuery(
     .findOnly();
 ----
 
+==== LIKE clauses
+
+Due to a built-in safety mechanism, `LIKE` clauses and argument binding
+don't work together intuitively and must be written a particular way. The following
+is what users would intuitively try to do:
+
+[source,java]
+----
+handle.createQuery("select name from things
+where name like '%:search%'")
+    .bind("search", "foo")
+    .mapTo(String.class)
+    .list()
+----
+
+Firstly, this query would try to select `where name like '%:search%'`
+_literally_, without binding any arguments. This is because proper `SqlParser`
+implementations do not bind arguments _inside string literals_.
+
+Secondly, Jdbi has your back. This query will throw an exception,
+because we don't allow unused bindings by default. Of course, if you use
+the parameter somewhere else in the query and the `SqlParser` binds it,
+then this check won't save you.
+
+The solution is to use SQL string concatenation:
+
+[source,java]
+----
+handle.createQuery("select name from things
+where name like '%' || :search || '%'")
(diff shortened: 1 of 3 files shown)


docs/src/adoc/index.adoc
```diff
@@ -597,6 +597,44 @@ LocalDate releaseDate = handle.createQuery(
     .findOnly();
 ----
 
+==== LIKE clauses
+
+Due to a built-in safety mechanism, `LIKE` clauses and argument binding
+don't work together intuitively and must be written a particular way. The following
```
stevenschlansker: I somewhat disagree with the wording here.  This is very intuitive behavior if you understand how the driver works -- bound parameters must be proper first class parameters to the query, not some templated string.  It'd be extremely surprising for it to template *into* string literals.

I'm going to do a followup here to reword this since I didn't get a chance to review this in the first place.

TheRealMarnes: > if you understand how the driver works

:)

But fair enough, manuals are there to be amended

docs/src/adoc/index.adoc
```diff
@@ -597,6 +597,44 @@ LocalDate releaseDate = handle.createQuery(
...
+handle.createQuery("select name from things
+where name like '%:search%'")
+    .bind("search", "foo")
+    .mapTo(String.class)
+    .list()
+----
+
+Firstly, this query would try to select `where name like '%:search%'`
+_literally_, without binding any arguments. This is because proper `SqlParser`
+implementations do not bind arguments _inside string literals_.
```
stevenschlansker: This isn't a `SqlParser` limitation, this is a consequence of how jdbc works.  We could change the parser any which way we want but the driver will not support doing this.

TheRealMarnes: We do actually specifically write our parsers to not even attempt it regardless of db capability. The text doesn't assign blame, it just says it will fail in sqlparsing already (to fail sooner than later) by design because jdbi is explicitly meant to do so, even if your maybe custom-made or patched db would support it.

PR #1385, opened 2018-12-31
Security Update for Jackson

brianm:
There is a security issue with jackson prior to 2.8.11.1


brianm:
Ugh, sadly it is not, apparently not all artifacts were pushed for this version of jackson

TheRealMarnes:
Looks like central stops at 2.8.11. Guess we'll have to do a minor version bump to 2.9, and we might as well go to the latest 2.9.8 then.

PR #1386, opened 2018-12-31
fix null byte array typing

### LINKED ISSUE #1038: Null byte[] arguments become ARRAY type
I just ran into an issue while trying to use nullable binary types, with byte[] fields on a bound bean object. Derby was spewing errors about unsupported ARRAY types.

I did a little digging and it looks like the problem occurs in SqlArrayArgumentFactory, where the null byte[] is assumed to be an ARRAY type:
https://github.com/jdbi/jdbi/blob/90af5ec3ac74e05919d7e90eb87ce2af8e7ba9ed/core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgumentFactory.java#L50-L52

However, if the value is not null, it will move on to the BuiltInArgumentFactory and use the correct type.

TheRealMarnes:
fixes #1038

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgument.java
+++ b/core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgument.java
@@ -13,37 +13,24 @@
  */
 package org.jdbi.v3.core.array;
 
-import java.lang.reflect.Array;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-
-import org.jdbi.v3.core.statement.StatementContext;
+import java.util.stream.Stream;
 import org.jdbi.v3.core.argument.Argument;
 import org.jdbi.v3.core.internal.IterableLike;
+import org.jdbi.v3.core.statement.StatementContext;
 
 class SqlArrayArgument<T> implements Argument {
-
     private final String typeName;
     private final Object[] array;
 
-    @SuppressWarnings("unchecked")
     SqlArrayArgument(SqlArrayType<T> arrayType, Object newArray) {
         this.typeName = arrayType.getTypeName();
 
-        List<Object> elements = new ArrayList<>(
-                newArray.getClass().isArray() ? Array.getLength(newArray) : 10);
-        IterableLike.of(newArray).forEachRemaining(
-                e -> elements.add(arrayType.convertArrayElement((T) e)));
-        array = elements.toArray();
-    }
-
-    SqlArrayArgument(SqlArrayType<T> arrayType, Collection<T> list) {
-        this.typeName = arrayType.getTypeName();
-        this.array = list.stream().map(arrayType::convertArrayElement).toArray();
+        @SuppressWarnings("unchecked")
+        Stream<T> stream = (Stream<T>) IterableLike.stream(newArray);
+        array = stream.map(arrayType::convertArrayElement).toArray();
     }
 
     @Override
--- a/core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgumentFactory.java
@@ -21,7 +21,7 @@ import org.jdbi.v3.core.argument.Argument;
 import org.jdbi.v3.core.argument.ArgumentFactory;
 import org.jdbi.v3.core.argument.NullArgument;
 import org.jdbi.v3.core.config.ConfigRegistry;
-import org.jdbi.v3.core.generic.GenericTypes;
+import org.jdbi.v3.core.internal.IterableLike;
 
 /**
  * Bind a Java array or {@link Collection} to a SQL array using the
@@ -39,24 +39,16 @@ import org.jdbi.v3.core.generic.GenericTypes;
 public class SqlArrayArgumentFactory implements ArgumentFactory {
     @Override
     public Optional<Argument> build(Type type, Object value, ConfigRegistry config) {
-        Class<?> rawClass = GenericTypes.getErasedType(type);
+        return IterableLike.elementTypeOf(type)
+            .flatMap(config.get(SqlArrayTypes.class)::findFor)
+            .map(arrayType -> arrayArgument(value, arrayType));
+    }
 
-        if (!rawClass.isArray() && !Collection.class.isAssignableFrom(rawClass)) {
-            return Optional.empty();
-        }
-
+    private Argument arrayArgument(Object value, SqlArrayType<?> arrayType) {
         if (value == null) {
-            return Optional.of(new NullArgument(Types.ARRAY));
-        }
-
-        if (rawClass.isArray()) {
-            return config.get(SqlArrayTypes.class)
-                .findFor(rawClass.getComponentType())
-                .map(arrayType -> new SqlArrayArgument<>(arrayType, value));
+            return new NullArgument(Types.ARRAY);
         } else {
-            return GenericTypes.findGenericParameter(type, Collection.class)
-                .flatMap(config.get(SqlArrayTypes.class)::findFor)
-                .map(arrayType -> new SqlArrayArgument<>(arrayType, value));
+            return new SqlArrayArgument<>(arrayType, value);
         }
     }
 }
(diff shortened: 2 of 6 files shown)


TheRealMarnes:
Included is a bit of general cleanup in the involved classes, and a refactor of the factory itself to split the main method into its 2 parts that were lobbed together, for clarity.

core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgument.java
```diff
@@ -18,34 +18,27 @@
...
-    @SuppressWarnings("unchecked")
     SqlArrayArgument(SqlArrayType<T> arrayType, Object newArray) {
         this.typeName = arrayType.getTypeName();
 
-        List<Object> elements = new ArrayList<>(
-                newArray.getClass().isArray() ? Array.getLength(newArray) : 10);
-        IterableLike.of(newArray).forEachRemaining(
-                e -> elements.add(arrayType.convertArrayElement((T) e)));
+        List<Object> elements = new ArrayList<>(newArray.getClass().isArray() ? Array.getLength(newArray) : 10);
+        IterableLike.of(newArray).forEachRemaining(e -> {
```
stevenschlansker: What's going on in this refactoring?  It looks the same but it's way wordier now.  This doesn't seem like an improvement to me unless I'm missing something.

core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgumentFactory.java
```diff
@@ -41,27 +40,41 @@
...
-            return Optional.of(new NullArgument(Types.ARRAY));
+        if (rawType.isArray()) {
+            if (value == null) {
+                return Optional.of(new NullArgument(Types.ARRAY));
+            }
+
+            return config.get(SqlArrayTypes.class)
+                .findFor(rawType.getComponentType())
+                .map(arrayType -> new SqlArrayArgument<>(arrayType, value));
         }
```
stevenschlansker: I'm not sure I see the improvement in this refactoring either.  It also ends up being significantly longer.  What was wrong with the original code?

core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgument.java
TheRealMarnes: If you open the old code, not being familiar with it, it takes some serious reading to figure out. The lookup variable just obscured the process. The whole method top to bottom just goes "this is either an array or a list, we're gonna cast it to whichever it is, and call 2 different flows to finish up depending on which it is". For all intents and purposes the method was just 2 functionally distinct methods clobbered together to save on lines of text, and not really structured to express its binary nature either.

I just factored the 2 methods apart and sort of normalized them. The fact that they look very different now is a clear sign they shouldn't have been writting as one whole. They happen to share a few variables and the null part, but that's all they share.

It's all about clarifying intent, making expression match function.

TheRealMarnes: oh, and IDE support becomes pretty crap too when a variable is inferred to be "either this or that, who knows" :p

TheRealMarnes: Oh, this part. Reducing the scope of the SuppressWarnings

core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgumentFactory.java
TheRealMarnes: First 2 replies on the other remark were meant for this one

TheRealMarnes: If you open the old code, not being familiar with it, it takes some serious reading to figure out. The lookup variable just obscured the process. The whole method top to bottom just goes "this is either an array or a list, we're gonna cast it to whichever it is, and call 2 different flows to finish up depending on which it is". For all intents and purposes the method was just 2 functionally distinct methods clobbered together to save on lines of text, and not really structured to express its binary nature either.

I just factored the 2 methods apart and sort of normalized them. The fact that they look very different now compared to each other is a clear sign they shouldn't have been written as one whole. They happen to share a few variables and the null part, but that's all they share.

It's all about clarifying intent, making expression match function.

TheRealMarnes: oh, and IDE support becomes pretty crap too when a variable is inferred to be "either this or that, who knows" :p

core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgument.java
stevenschlansker: It turned two lines into five lines, that sure seems wordier to me.  You even extracted out a new variable `cast` that is totally useless.  This PR would have been a slam dunk if it just had the relevant fix at hand...

core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgumentFactory.java
stevenschlansker: I think the intent is a lot less clear now.  Before, it read nicely, "If the type doesn't match, exit.  If it's null, bind null.  Otherwise look up the element binder.  Then, depending on whether it's an array or collection, find the element type and bind it.  One less controversial improvement might be to move the "how do I find the element type" bit to `IterableLike` which would remove the final branch from the code.

core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgument.java
TheRealMarnes: ?? I only _added_ 1 line of minimal code, by extracting the cast to keep the scope of the SuppressWarnings to a minimum. The only other addition is a pair of brackets because multiline lambda. I really don't see how it's wordier. It's not more code. I literally just moved the suppresswarnings and extracted a variable for good practice.

core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgumentFactory.java
TheRealMarnes: I'll see if I can rewrite it to more of a golden middle way. I agree it is _larger_, but I generally find surface area much less of a mental burden/complexity than number of operations per method, ambiguity or sub-methods (the schroedinger variable being "one of two" types at the same time that require different handling every step of the way), and back-and-forth branching between 2 different processes that are conjoined at the hip for sharing a few lines of elementary code.

But my vote for the refactor bit (separate commit) is a soft vote, if your objection tends toward hard I can just stop caring about it, rewind the commit, and we'll move on.

core/src/main/java/org/jdbi/v3/core/internal/IterableLike.java
```diff
@@ -68,6 +72,16 @@ public static boolean isIterable(Object maybeIterable) {
         return Arrays.asList((Object[]) iterable).iterator();
     }
 
+    /**
+     * Given an iterable object (which may be a iterator, iterable, primitive
+     * or reference array), return a {@link Stream} over its (possibly boxed) elements.
+     *
+     * @return an iterator of the given array's elements
```
qualidafial: s/iterator/stream/

core/src/main/java/org/jdbi/v3/core/array/SqlArrayArgumentFactory.java
```diff
@@ -41,27 +39,24 @@
...
-            Class<?> elementType = erasedType.getComponentType();
-            return lookup.apply(elementType)
-                    .map(arrayType -> new SqlArrayArgument<>(arrayType, value));
-        }
-
-        return GenericTypes.findGenericParameter(type, Collection.class)
-                .flatMap(lookup)
+        if (rawClass.isArray()) {
+            return config.get(SqlArrayTypes.class)
+                .findFor(rawClass.getComponentType())
```
stevenschlansker: It still irks me slightly that the `findFor` bit is duplicated here.  I'll make one quick fixup then it's good to go.

stevenschlansker:
I pulled out the 'find element type' logic into `IterableLike`, since it already has related functionality.  I think the result is a lot clearer, and as a bonus, we now recognize `Iterator<T> `in addition to `Iterable<T>`.  Hopefully you like the change: https://github.com/jdbi/jdbi/pull/1386/commits/ecfe3c86ce3b3ed3d6da54abd1bf0550a15e0785

TheRealMarnes:
Wholeheartedly agreed, that's a neat change :+1:

stevenschlansker:
Oops, I broke it.  One more try...

PR #1390, opened 2019-01-01
interpret parentheses around parameters to support trailing query text

### LINKED ISSUE #1389: Can't use postgres casting operator with hash prefix parser
A query with something like `#recordId::UUID` throws an error because it tries to convert the whole thing to a named parameter. I know it can be done with the cast function, but it would be nice to be able to use the operator.

TheRealMarnes:
fixes #1389 IMO. @toadzky got any input on this?

Will need some more work to make it backward compatible. I'm thinking of a boolean flag `parseParentheses` (default false) that can be passed in by constructor and setter.

This change makes parameters more usable by allowing you to embed them inside an expression that would otherwise parse wrongly, like `#[var]::UUID`, where the parameter would be interpreted as being named `var::UUID` instead of just `var`.

I think it would be neat to add `$` as a supported marker character to have a nice set of 3, and give each marker its own different set of optional parentheses: `:()`, `#[]`, and `${}`, to suit different styles and have totally different characters to support as much unknown syntax as possible.

On-hold because it needs discussion.


core/src/main/antlr3/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g
```diff
@@ -11,6 +11,8 @@ lexer grammar ColonStatementLexer;
   }
 }
 
+fragment LEFT_PAREN: '(';
+fragment RIGHT_PAREN: ')';
```
qualidafial: Why parens here and square brackets in the hash lexer?

TheRealMarnes: It's more useful to have different bracket types going with different marker characters. If `:` is a problem in a query, you can switch to `#` (and maybe in the future we'll add a third like `$` with `{}`).

If you want to use bracketing and it turns out the parentheses are an issue, then solving that should be just as simple, by switching to one of the other parsers with a different bracket type. If they all use the same characters for brackets, then bracketing is an issue in a given query no matter the parser used, which isn't very helpful.

TLDR: having 2 or 3 syntaxes with _all different_ characters is going to enable users to get around more collisions than if they're partly different and partly the same.

qualidafial: I'm not sure I buy that reasoning. Any decision we make here needs to be grounded in vendor-specific SQL syntax.

For my part, I'm pretty hesitant to use `()` as braces since they're part of standard SQL syntax

TheRealMarnes: But the sqlparsers aren't vendor specific, they're general-purpose and all in `core`. I don't think it's ok to reason about vendor-specific issues that will ultimately never be completely unitable with each other (i.e. there will always be a conflict somewhere, and the need to do escaping and whatnot) in `core`.

TheRealMarnes: When someone uses the ColonPrefixSqlParser, the implication is "colon prefix", not "workarounds for postgres syntax that may interfere with stuff that isn't postgres"

qualidafial: The SQL parsers don't have to be vendor specific, but they should be agnostic enough to be usable with any vendor.

So if we're going to support an alternative parameter syntax, it should be based on a specific need--which is to say, a vendor syntax which has been found to be incompatible with the default `:name` syntax.

TheRealMarnes: But that's asking for conflicts. You can't build support for everything into 1 thing without ending up with complicated code and imposing restrictions or forcing workarounds/extra considerations on all for the sake of one. Isn't that the entire reason we have vendor modules in the first place? Why split everything up to make it as simple as possible in `core` with all dedicated support isolated in dedicated modules, but exempt sqlparsers from that?

toadzky: i'm pretty sure all the bracket/brace options are already part of sql. `[]` is arrays, `{}` is used for arrays and json, and `()` is scoping.

toadzky:
looks good to me, as long as everyone is on board with the different syntax stuff per parser stuff.

core/src/main/antlr3/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g
TheRealMarnes: > For my part, I'm pretty hesitant to use () as braces since they're part of standard SQL syntax

This isn't actually an issue though btw, they would only be interpreted as parameter brackets when actually surrounding what we already recognize as a parameter placeholder. When used anywhere else they just keep their sql meaning. They go _inside/on the edge of_ the variable name, so they only conflict with the query as much as the variable name itself does.

TheRealMarnes: `where name = :name;` is clear to everyone, I don't see why `where name = :(name);` would be an issue

stevenschlansker: Introducing paren-based quoting will conflict with otherwise valid SQL syntax.  For example, `[:(3)]` is a Postgres array slice, but would be ambiguous with the proposed syntax.

I agree that braces `{}` are common in arrays and json, but aren't those always quoted in SQL?  At least in Postgres the valid forms are `ARRAY[1, 2]` or `'{1, 2}'::int[]`, neither of which has bare `{}`, and same story for json.

Could we change the hash parser to not treat `::` as part of a valid name?  It's not a general fix, which is a bit of a bummer, but it's the only concrete problem we've heard of yet.  And it seems extremely unlikely to be part of a valid name.  (but it is technically a small breaking change)

We should try to avoid adding any need for additional escaping or quoting if we can.

TheRealMarnes: > If we must, I think {} braces would be the most broadly compatible.

If that's your bottom line (as discussed yesterday), then let's forget about this pr. I won't mind.

For me the priority is creating a minimal mechanism (magic character, escape char, maybe delimiters) and applying it predictably, repetitive as it may be, and avoiding having to make the mechanism more complex. Having such a mechanism but then also tacking unique exceptions on top of it that a user needs to go learn about and be able to spot when they interfere with something unintendedly, just to avoid having to actually _use_ the former mechanism, makes me wonder why it's there at all then. It's shunning your own tools and adding learning curve for very minor convenience, while creating room for new inconveniences.

Jdbi having its own special understanding of how to recognize/ignore parameters inside a query that goes beyond simple pattern substitution is a step toward creating a new language with its own grammar instead of integrating existing things everyone's already familiar with.

TLDR I prefer to _know_ a tool is dumb and will not try be smart, than to have to wonder if a tool will or will not correctly interpret a particular case of me trying to speak to it in its own language.

I think I'm gonna call it the French Grammar trope. :)

PR #1391, opened 2019-01-02
ColonStatementLexer: ignore postgres array slice e.g. `[3:5]`

### LINKED ISSUE #1387: Postgres Array Slice operator is treated as a named parameter with no name
Postgres array slices (found here: https://www.postgresql.org/docs/10/arrays.html) are 2 numbers separated a `:` surrounded by optional whitespace. From the docs:
> An array slice is denoted by writing **lower-bound:upper-bound** 
This causes the lexer to throw an ANTLR exception:

```
org.antlr.runtime.NoViableAltException: null
  at org.jdbi.v3.core.internal.lexer.ColonStatementLexer.mTokens(ColonStatementLexer.java:1074)
  at org.antlr.runtime.Lexer.nextToken(Lexer.java:89)
  ... 96 common frames omitted
  Causing: java.lang.IllegalArgumentException: NoViableAltException(':'@[])
  at org.jdbi.v3.core.internal.lexer.ColonStatementLexer.reportError(ColonStatementLexer.java:33)
  at org.antlr.runtime.Lexer.nextToken(Lexer.java:107)
  at org.jdbi.v3.core.statement.ColonPrefixSqlParser.internalParse(ColonPrefixSqlParser.java:80)
```

The parser should allow for standalone colons or provide a method to escape them.

stevenschlansker:
Fixes #1387 

Note that this change attempts to be conservative -- we only change lexing rules that happen *after* we already see a leading `[`, and simply ignore anything that looks like a slice.  I've verified that it is not valid syntax (parse error) to try to bind a parameter into an array slice, so there isn't any ambiguity as to whether `[:2]` is a open ended slice or a named parameter slice, as the latter is a Postgres syntax error.

### DIFF
--- a/RELEASE_NOTES
+++ b/RELEASE_NOTES
@@ -22,6 +22,8 @@
     - Support for Postgres MACADDR columns, using the @MacAddr qualifying annotation
     - Support for HSTORE columns, using the @HStore annotation
     - @Json type qualifier with Jackson 2 and Gson 2 bindings
+  - Improvements
+    - ColonStatementLexer: don't trip up over postgres array slice `array[3:4]` syntax
   - Oracle DB support changes
     - Due to ongoing stability problems with Oracle's Maven servers, we have split the
       jdbi3-oracle12 artifact out of the main project, to a new home at
--- a/core/src/main/antlr3/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g
+++ b/core/src/main/antlr3/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g
@@ -19,7 +19,9 @@ fragment COLON: {input.LA(2) != ':'}?=> ':';
 fragment DOUBLE_COLON: {input.LA(2) == ':'}?=> '::';
 fragment QUESTION: {input.LA(2) != '?'}?=> '?';
 fragment DOUBLE_QUESTION: {input.LA(2) == '?'}?=> '??';
-fragment NAME: 'a'..'z' | 'A'..'Z' | '0'..'9' | '_' | '.' | '?.' | '#';
+fragment DIGIT: '0'..'9';
+fragment NAME: 'a'..'z' | 'A'..'Z' | DIGIT | '_' | '.' | '?.' | '#';
+fragment BRACKET_OR_SLICE: '[' (' '* DIGIT* ' '* ':' ' '* DIGIT* ' '* ']')?;
 
 COMMENT: '/*' .* '*/';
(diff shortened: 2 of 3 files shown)


TheRealMarnes:
I think a better solution would be to say that "2" isn't a valid parameter name, and not recognize `:2` as a binding for that reason. This solves the issue with postgres slices and doesn't fill `core` with special postgresism considerations. It keeps the sqlparser syntax simple and obvious without setting a trend of special grammar rules.

stevenschlansker:
Precisely.

toadzky:
the exact slice syntax i'm using is a bit different because we are calculating the indexes using postgres array functions. it looks like this `actions [1:array_position(actions, planId) - 1]`. if i'm reading the antlr right, this would solve the simple case with literals but not the dynamic case with functions or (heaven forbid) some kind of sub-select.

stevenschlansker:
Darn, that increases the complexity a fair bit.  I'll play with it a bit more to see if I can hit that use case as well.  We may have to compromise on only supporting literals or nothing, though, depending on the resulting complexity.

toadzky:
the actual scenario is a recursive union tree walk. there is an array column that contains a list of ids, which may point to another row and we need to append those actions recursively to the result. so we grab a slice up to the id that points to another row.

TheRealMarnes:
We're defining `core` behavior here so I'm concerned about non-postgres usage. What this changes is that a named parameter, of which the name is only a number, cannot be used if it's surrounded by `[]`. So if e.g. a database allows some sort of js-like array expressions with parameters (`"[?]", param1: 2`), or anything else that fits the pretty simple format, and the param is named as a number, jdbi will forbid it.

On one hand, I find naming a parameter as only a number fishy to begin with, so I don't see this being a real problem in practice. On the other hand, I don't like the principle that postgresisms are creating surprising behavior in `core` and getting in the way of people who may well have nothing to do with postgres, when we have a `postgres` module to deal with postgresisms.

qualidafial:
Can I scope creep this PR to request that we treat `:` followed by a non-name character as a literal? Because right now it throws a `NoViableAltException`

stevenschlansker:
Does #1488 fix this issue as well?  (Or does it make it worse?)

qualidafial:
Worth noting that the `:` can be escaped as `\\:` in a Java String, or as `\:` in a text file.

stevenschlansker:
Sadly, this fix isn't going to work for the more advanced use case you cite.  So I'm going to close it for now.  We can always pick this work back up later.  For now, you'll have to escape array slices in Jdbi `\\:`.

PR #1393, opened 2019-01-04
Cleanup #1384

stevenschlansker:
Followup from #1384 since I never got a chance to review it.

Reduce a bit of the editorialization, make a new `Cookbook` section to start stashing "recipes"
Also make it clear that this is a *jdbc* limitation, not a SqlParser limitation.

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -290,6 +290,11 @@ Jdbi's configuration system enables you to easily set up different Arguments and
 [NOTE]
 _Binding_ is a term specifically related to _Arguments_ only. _Attributes_ are involved in _rendering_ a query _template_ using a _TemplateEngine_ to perform String-wise manipulations on a query before it is interpreted as SQL.
 
+[NOTE]
+The use of bound parameters is limited by JDBC support.  Notably, parameters usually cannot be used to change the structure of a query
+(for example the table or column name, `SELECT` or `INSERT`, etc.)
+nor may they be interpolated into string literals.  See <<TemplateEngine>> for more information.
+
 ==== Positional Arguments
 
 When a SQL statement uses `?` tokens, Jdbi can bind a values to parameters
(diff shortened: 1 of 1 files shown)


docs/src/adoc/index.adoc
```diff
@@ -597,44 +602,6 @@ LocalDate releaseDate = handle.createQuery(
     .findOnly();
 ----
 
-==== LIKE clauses
-
-Due to a built-in safety mechanism, `LIKE` clauses and argument binding
```
stevenschlansker: This isn't really a safety mechanism, it's just how jdbc works.  Parameters may not be bound into the middle of string literals (even if we decided to support the syntax, the driver won't do it.)

stevenschlansker:
To better support my assertion that this is expected and desirable behavior, and not surprising, refer to `ISO-9075-2:2016` (which is unfortunately paywalled, but maybe you can find an excerpt, which defines that `SELECT '?'` for example is a literal `?` not a bound parameter.

So jdbi seeing `SELECT ':foo'` as a literal `:foo` is exactly what JDBC does, and is required by ISO / IEC specification.  We're just copying that behavior.  Behaving the same way as the driver does should be the least astonishing approach generally.

stevenschlansker:
They also leave relevant guidance for implementors, relevant to our ongoing discussions about vendor-specific syntax like array slicing:
> Also interesting to note that the JDBC spec section 6.2 (Guidelines and Requirements) says:
>    Drivers should provide access to every feature implemented by the underlying data source, including features that extend the JDBC API. The intent is for applications using the JDBC API to have access to the same feature set as native applications.

> So based on that you should (not must) support the ?-operator, you just need to find a practical way to do it.

TheRealMarnes:
I agree on the inside-string-literals part, no discussion from me there. Jdbc sets it as a rule and so it shall be. This whole PR has my full support. 👍 

--

I feel we've already said everything there is to say, and I like the last point we came to: we'll go ahead with it for now under the assumption that the issue I fear, while possible, is not going to appear soon, so until the day that it does we can go ahead with the theoretically imperfect current idea. So as far as I'm concerned, the discussion is over. 😃

So read on only if you, like me, enjoy discussions for their own sake.

--

I would like to point out though that the text you posted says exactly what I have been saying all the time:

> The intent is for applications using the JDBC API to have access to the same feature set as native applications.

People can very simply tell jdbi to ignore anything it excessively interprets. They cannot tell jdbi to un-ignore something it decides to ignore for whatever reason. One group's tiny comfort of not having to escape their non-standard syntax to make it compatible with universal parsers should not cause everyone else to be saddled up with no longer being able to use their own non-standard syntax _at all_.

Core should allow maximum freedom and be unbiased. If there's an issue for a specific db, then the dedicated module for that db should take care of it with an override that isn't forced onto everyone else as an inconvenience. In the meantime, everybody can equally use the simple, universal solution of escaping the incompatible syntax and nobody needs to feel more disadvantaged than others.

The only consideration I can agree with is that postgres enjoys a special member privilege of being directly supported, while the thing it bans is for now only "theoretical" in a certain sense.

--

I'm a pessimist and perfectionist. I'm rather safe than sorry, and theory and principles trump practical considerations. Maximum convenience for all, but _never_ at the cost of pushing someone else down.

TheRealMarnes:
2 approvals, let's go

PR #1396, opened 2019-01-06
Polish QualifiedType and related code

### LINKED ISSUE #1395: Polish some API before release
There are 2 things I'd like to improve before we release some of our new API:

image

- the varargs warning on `with(...)`
- the fact that QualifiedType doesn't have a `T` so that a `@Json String` isn't recognized as a `String` and we need to do avoidable casts

TheRealMarnes:
fixes #1395

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
@@ -162,7 +162,7 @@ public class ColumnMappers implements JdbiConfig<ColumnMappers> {
      * @return a ColumnMapper for the given type, or empty if no column mapper is registered for the given type.
      */
     public Optional<ColumnMapper<?>> findFor(Type type) {
-        return findFor(QualifiedType.of(type));
+        return findFor(QualifiedType.of(type)).map(m -> (ColumnMapper<?>) m);
     }
 
     /**
@@ -176,13 +176,14 @@ public class ColumnMappers implements JdbiConfig<ColumnMappers> {
         // ConcurrentHashMap can enter an infinite loop on nested computeIfAbsent calls.
         // Since column mappers can decorate other column mappers, we have to populate the cache the old fashioned way.
         // See https://bugs.openjdk.java.net/browse/JDK-8062841, https://bugs.openjdk.java.net/browse/JDK-8142175
-        ColumnMapper<?> cached = cache.get(type);
+        @SuppressWarnings("unchecked")
+        ColumnMapper<T> cached = (ColumnMapper<T>) cache.get(type);
 
         if (cached != null) {
             return Optional.of(cached);
         }
 
-        Optional<ColumnMapper<?>> mapper = factories.stream()
+        Optional<ColumnMapper<T>> mapper = factories.stream()
                 .flatMap(factory -> toStream(factory.build(type, registry)))
                 .findFirst();
 
--- a/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
@@ -107,7 +107,7 @@ public class ColumnMappers implements JdbiConfig<ColumnMappers> {
      * @return this
      */
     @Beta
-    public ColumnMappers register(QualifiedType<?> type, ColumnMapper<?> mapper) {
+    public <T> ColumnMappers register(QualifiedType<T> type, ColumnMapper<T> mapper) {
         return this.register(QualifiedColumnMapperFactory.of(type, mapper));
     }
 
@@ -172,7 +172,7 @@ public class ColumnMappers implements JdbiConfig<ColumnMappers> {
      * @return a ColumnMapper for the given type, or empty if no column mapper is registered for the given type.
      */
     @Beta
-    public Optional<ColumnMapper<?>> findFor(QualifiedType<?> type) {
+    public <T> Optional<ColumnMapper<T>> findFor(QualifiedType<T> type) {
         // ConcurrentHashMap can enter an infinite loop on nested computeIfAbsent calls.
         // Since column mappers can decorate other column mappers, we have to populate the cache the old fashioned way.
         // See https://bugs.openjdk.java.net/browse/JDK-8062841, https://bugs.openjdk.java.net/browse/JDK-8142175
--- a/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
@@ -32,7 +32,7 @@ import static org.jdbi.v3.core.internal.JdbiStreams.toStream;
  */
 public class ColumnMappers implements JdbiConfig<ColumnMappers> {
     private final List<QualifiedColumnMapperFactory> factories = new CopyOnWriteArrayList<>();
-    private final ConcurrentHashMap<QualifiedType, ColumnMapper<?>> cache = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<QualifiedType<?>, ColumnMapper<?>> cache = new ConcurrentHashMap<>();
     private ConfigRegistry registry;
 
     public ColumnMappers() {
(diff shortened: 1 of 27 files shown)


TheRealMarnes:
I think we should leave this convenience method out:

image

It's a shorthand for a QualifiedType, but while the varargs warning can be fixed on qualtyp, it can't be fixed here because ResultBearing is an interface and cannot promise safe varargs handling. I don't think this is unreasonable instead:

image

@qualidafial @stevenschlansker ?

TheRealMarnes:
All changes pushed in sequential/semantic commits to clearly justify each change and make it easy to follow. I suggest checking #e28afef on its own, and just checking the global diff for the rest. All the changes relate to QualifiedType so backward compatibility is moot. I've avoided changes to imports too ^^ I think this should be a piece of cake to greenlight.

core/src/main/java/org/jdbi/v3/core/mapper/QualifiedColumnMapperFactory.java
```diff
@@ -13,27 +13,37 @@
...
-    Optional<ColumnMapper<?>> build(QualifiedType type, ConfigRegistry config);
+    <T> Optional<ColumnMapper<T>> build(QualifiedType<T> type, ConfigRegistry config);
 
     static QualifiedColumnMapperFactory adapt(ColumnMapperFactory factory) {
         Set<Annotation> qualifiers = getQualifiers(factory.getClass());
-        return (type, config) -> type.getQualifiers().equals(qualifiers)
-            ? factory.build(type.getType(), config)
-            : Optional.empty();
+
+        return new QualifiedColumnMapperFactory() {
```
TheRealMarnes: this and the other inline class are necessary to get a T reference to work with

core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java
```diff
@@ -186,10 +186,10 @@ public T map(ResultSet rs, StatementContext ctx) throws SQLException {
                             .map(m -> m.getParameterCount() > 0 ? m.getParameters()[0] : null)
                             .orElse(null);
 
-                        QualifiedType type = QualifiedType.of(propertyType(descriptor))
+                        QualifiedType<?> type = QualifiedType.of(propertyType(descriptor))
                             .with(getQualifiers(getter, setter, setterParam));
                         ColumnMapper<?> mapper = ctx.findColumnMapperFor(type)
-                            .orElse((r, n, c) -> r.getObject(n));
+                            .orElse((ColumnMapper) (r, n, c) -> r.getObject(n));
```
qualidafial: This probably wants a `@SuppressWarnings("raw")`

core/src/main/java/org/jdbi/v3/core/mapper/reflect/FieldMapper.java
```diff
@@ -146,10 +146,10 @@ public T map(ResultSet rs, StatementContext ctx) throws SQLException {
 
                     findColumnIndex(paramName, columnNames, columnNameMatchers, () -> debugName(field))
                         .ifPresent(index -> {
-                            QualifiedType type = QualifiedType.of(field.getGenericType())
+                            QualifiedType<?> type = QualifiedType.of(field.getGenericType())
                                 .with(getQualifiers(field));
                             ColumnMapper<?> mapper = ctx.findColumnMapperFor(type)
-                                .orElse((r, n, c) -> r.getObject(n));
+                                .orElse((ColumnMapper) (r, n, c) -> r.getObject(n));
```
qualidafial: Probably wants `@SuppressWarnings("raw")` here too

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
```diff
@@ -223,15 +223,15 @@ protected QualifiedType elementType(StatementContext ctx) {
...
         @Override
-        protected QualifiedType elementType(StatementContext ctx) {
+        protected QualifiedType<?> elementType(StatementContext ctx) {
             return elementType;
         }
     }
 
-    static class SingleValueReturner extends ResultReturner {
-        private final QualifiedType returnType;
+    static class SingleValueReturner<T> extends ResultReturner {
```
qualidafial: Any reason for introducing the `<T>` parameter on this and the next returner, but not the others?

TheRealMarnes: the others do `mapType` so become wildcards. Unless I'm mistaken

TheRealMarnes: `ConsumerResultReturner` doesn't get its type from a param, no room for a T
`VoidReturner` idem
`IteratorReturner` does `mapType`
`ResultIteratorReturner` idem
`StreamReturner` idem
`ResultIterableReturner` idem

qualidafial:
@stevenschlansker We're merging to free up other PRs before the release. Let us know if you have any comments and we'll address them separately.

core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
```diff
@@ -172,19 +172,21 @@ private ColumnMappers register(QualifiedColumnMapperFactory factory) {
      * @return a ColumnMapper for the given type, or empty if no column mapper is registered for the given type.
      */
     @Beta
-    public Optional<ColumnMapper<?>> findFor(QualifiedType type) {
+    public <T> Optional<ColumnMapper<T>> findFor(QualifiedType<T> type) {
         // ConcurrentHashMap can enter an infinite loop on nested computeIfAbsent calls.
         // Since column mappers can decorate other column mappers, we have to populate the cache the old fashioned way.
         // See https://bugs.openjdk.java.net/browse/JDK-8062841, https://bugs.openjdk.java.net/browse/JDK-8142175
-        ColumnMapper<?> cached = cache.get(type);
+        @SuppressWarnings("unchecked")
```
stevenschlansker: Should this `@SuppressWarnings` be on the whole method?  There's a second cast slightly further down.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java
stevenschlansker: I don't see this as actually resolved.  There's an `unchecked` suppression but not one for `rawtypes`.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/FieldMapper.java
stevenschlansker: Ditto, not seeing this as resolved.

core/src/test/java/org/jdbi/v3/core/qualifier/TestNVarchar.java
```diff
@@ -76,18 +76,18 @@ public void sqlStatementBindNVarchar() {
...
             assertThat(
                 handle.select("SELECT name FROM nvarchars ORDER BY id")
-                    .mapTo(new GenericType<String>() {}, NVarchar.class)
+                    .mapTo(QualifiedType.of(new GenericType<String>() {}).with(NVarchar.class))
                     .list())
                 .containsExactly("foo", "bar", "baz", "qux");
 
             List rawList = handle.select("SELECT name FROM nvarchars ORDER BY id")
-                .mapTo(NVARCHAR_STRING)
+                .mapTo((QualifiedType<?>) NVARCHAR_STRING)
```
stevenschlansker: Why is this cast necessary?  That seems like a real wart that you need to cast it to remove the generic type?

core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
TheRealMarnes: I'm very much against giving suppressions a wide scope, they may end up hiding actual problems or things you _could_ improve if only you noticed them.

core/src/test/java/org/jdbi/v3/core/qualifier/TestNVarchar.java
TheRealMarnes: It's not necessary at all, it's purely to help demonstrate the point of that test, to neglect the generic type.

TheRealMarnes:
Sorry about the merge conflicts. I am trying to wrestle through immutables, I wanna see it merged too. This one can't have caused much though, maybe 2 lines, and I don't think not doing doable work because something else is still open at the same time is a great way to use a ticket-oriented system. If you're blaming me for the quick merge, I didn't decide that or ask for it. And like you said, the intent of the "freeze" wasn't to utterly forbid all other work, especially not work very directly relevant to the release.

PR #1398, opened 2019-01-10
jdbi.org favicon

TheRealMarnes:
Everybody deserves to look nice :)

image

asciidoctor needs to see a doctor:

image


docs/src/adoc/index.adoc
```diff
@@ -10,6 +10,7 @@
 :source-highlighter: coderay
 :source-language: asciidoc
 :imagesdir: images
+:docinfo: private
```
stevenschlansker: I'm not sure I understand this change.  Is this not the default? 

https://github.com/asciidoctor/asciidoctor.org/blob/master/docs/_includes/docinfo.adoc

> Setting docinfo with no value is equivalent to setting the value to private.

TheRealMarnes: done

stevenschlansker: Can we not omit the key entirely?  (That was the change I expected to see)  Or is its presence important?

TheRealMarnes: I tried it, doesn't work. The doc implies as much.

TheRealMarnes:
I'll just hotfix that then, preferred it too

PR #1399, opened 2019-01-11
build: cleanup orphaned files, move config into `policy` module

stevenschlansker:
This simplifies contributor setup instructions by removing the need to set `maven.multiModuleProjectDirectory` and instead using maven dependency references to an artifact we inject into plugin dependencies.


policy/pom.xml
```diff
@@ -0,0 +1,27 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>basepom-oss</artifactId>
+        <groupId>org.basepom</groupId>
+        <version>27</version>
+    </parent>
+
+    <artifactId>jdbi3-policy</artifactId>
+    <name>jdbi3 Policy</name>
```
qualidafial: Without a `<version>` tag this will default to `27` from basepom

stevenschlansker: and `groupId` too, woops

qualidafial: Should this have used jdbi3-parent as the parent pom instead?

stevenschlansker:
Not sure why Travis didn't update the check, but it seems to be all green.

PR #1401, opened 2019-01-11
#1357 reduce test boilerplate with lombok

TheRealMarnes:


core/src/test/java/org/jdbi/v3/core/ValueType.java
```diff
@@ -14,42 +14,13 @@
 
 package org.jdbi.v3.core;
 
-public class ValueType {
-    public static ValueType valueOf(String value) {
-        return value == null ? null : new ValueType(value);
-    }
+import lombok.Data;
 
+@Data
```
qualidafial: Should be `@Value` not `@Data`

core/src/test/java/org/jdbi/v3/core/argument/TestBeanArguments.java
```diff
@@ -58,16 +61,10 @@ public void testBindNull() throws Exception {
         verify(stmt).setNull(3, Types.NUMERIC);
     }
 
+    @RequiredArgsConstructor
+    @Getter
```
qualidafial: `@Value` is equivalent

core/src/test/java/org/jdbi/v3/core/argument/TestBeanArguments.java
```diff
@@ -115,14 +112,10 @@ public void testBindNonPublicGetter() {
             .isInstanceOf(UnableToCreateStatementException.class);
     }
 
+    @Setter
+    @Getter(AccessLevel.PROTECTED)
     public static class NonPublicGetter {
```
qualidafial: I would probably leave this one how it was

core/src/test/java/org/jdbi/v3/core/argument/TestBeanArguments.java
TheRealMarnes: equals/hashcode is technically a behavioral change but I guess there are extremely few tests where it will matter

TheRealMarnes:
Closing because we'll be doing this in smaller steps

PR #1402, opened 2019-01-12
Latest PMD, rules fixups.

stevenschlansker:
A number of bugs have been fixed, and a couple of new rules to exclude, and failures to fix...

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -13,9 +13,14 @@
  */
 package org.jdbi.v3.core;
 
+import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
 import java.io.Closeable;
 import java.sql.Connection;
 import java.sql.SQLException;
+
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.config.Configurable;
 import org.jdbi.v3.core.extension.ExtensionMethod;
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -322,10 +325,14 @@ public Handle rollback() {
     public Handle rollbackToSavepoint(String savepointName) {
         final long start = System.nanoTime();
         transactions.rollbackToSavepoint(this, savepointName);
-        LOG.trace("Handle [{}] rollback to savepoint \"{}\" in {}ms", this, savepointName, (System.nanoTime() - start) / 1000000L);
+        LOG.trace("Handle [{}] rollback to savepoint \"{}\" in {}ms", this, savepointName, msSince(start));
         return this;
     }
 
+    private static long msSince(final long start) {
```
TheRealMarnes: I feel we're better off making a `Stopwtch` class if timing comes up in multiple places

stevenschlansker: I think it's clearer to have a simple one line function in the local file, rather than to add a whole new file.
I'm not opposed to a stopwatch class but I'd recommend we do that as a separate PR that considers the consequences properly rather than just on a whim in a cleanup.

TheRealMarnes: Could be as simple as this IMO:
```
public class Timer {
    private Instant start;
    
    public Timer start() {
        start = Instant.now();
        return this;
    }
    
    public long measure(ChronoUnit unit) {
        return unit.between(start, Instant.now());
    }
}
```

stevenschlansker: My point is that it adds a non-local reference for the reader when we can sensibly keep it local for now.  This PR isn't intended to add a generalizable Stopwatch.

PR #1403, opened 2019-01-12
checkstyle: Import order

stevenschlansker:
This is intended to be a reasonable compromise between Eclipse and IntelliJ standards.  Enforcing import order adds some pain but reduces diff churn due to fighting IDEs and IMO the benefits outweigh the unfortunate cost.


TheRealMarnes:
@qualidafial anything to add?

PR #1408, opened 2019-01-14
Cleanup enum PR

stevenschlansker:
Fold functionality back into fewer files, factor out common logic


core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.jdbi.v3.core.argument;
+
+import java.lang.reflect.Type;
+import java.util.Optional;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.internal.EnumStrategy;
+import org.jdbi.v3.core.qualifier.QualifiedType;
+
+class EnumArgumentFactory implements QualifiedArgumentFactory {
```
stevenschlansker: Note how this QualifiedArgumentFactory replaces all three implementations below, and delegates only the bits that actually vary to the `EnumStrategy`.

stevenschlansker:
There were also a number of tests with exactly the same filename, so I fixed that as well.

core/src/main/java/org/jdbi/v3/core/EnumConfig.java
```diff
@@ -14,55 +14,36 @@
...
 
-    /**
-     * Applies to both binding and mapping.
-     *
-     * @return true if enums are handled by name, false if enums are handled by ordinal
-     */
-    @Deprecated
-    public boolean isEnumHandledByName() {
-        return handling == EnumByName.class;
+    public Class<? extends Annotation> getDefaultQualifier() {
```
qualidafial: Using an annotation type feels odd to me. I get the sense @stevenschlansker that you're trying to design for extension here. However the set of options that plain JDK enums is strictly closed: name and ordinal.

If somebody wants to map/bind their enums in another way, we already have a registration procedure for that.

IMO this config option should just be a simple enum property, e.g. `EnumStrategy.BY_NAME` or `EnumStrategy.BY_VALUE`.

TheRealMarnes:
closing to merge locally and polish up for a fresh pr with a cleaner slate

PR #1409, opened 2019-01-14
Remove LoggableToStringOrNPEArgument

stevenschlansker:
This class probably should never have existed


core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
```diff
@@ -26,7 +28,7 @@
         // TODO use the same configuration as EnumMapperFactory for consistency
         if (rawValue instanceof Enum) {
             Enum<?> enumValue = (Enum<?>) rawValue;
-            return Optional.of(new LoggableToStringOrNPEArgument<>(enumValue, Enum::name));
+            return Optional.of(new LoggableBinderArgument<>(Objects.toString(enumValue, null), PreparedStatement::setString));
```
stevenschlansker: In this case, it could never be null -- `null instanceof ?` is always false

core/src/main/java/org/jdbi/v3/core/argument/JavaTimeZoneIdArgumentFactory.java
```diff
@@ -26,6 +28,6 @@ public JavaTimeZoneIdArgumentFactory() {
 
     @Override
     protected Argument build(ZoneId value, ConfigRegistry config) {
-        return new LoggableToStringOrNPEArgument<>(value);
+        return new LoggableBinderArgument<>(Objects.toString(value, null), PreparedStatement::setString);
```
stevenschlansker: in this case, it's almost certainly a bug -- binding a `null` `ZoneId` should be valid and simply bind null.

core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
stevenschlansker: it's going away with the enum cleanup anyway

TheRealMarnes:
> This class probably should never have existed

I hope you're not thinking or implying I wrote this class from inspiration and added it to the code for some reason? The original code that was deduplicated into this class to highlight the existing problem is what shouldn't have existed.

stevenschlansker:
> I hope you're not thinking or implying I wrote this class from inspiration and added it to the code for some reason?

No, I more meant that I wish we'd spent a bit of time thinking about the better long term solution, rather than refactoring to an intermediate state that needed further refactoring.

I'll make `ObjectArgument` delegate null handling to `NullArgument`.  You're right that's a weakness of the current approach but nobody's come up with a good solution yet...

TheRealMarnes:
Well an obvious solution is adding `default boolean isNull() { return false: }` to `Argument` and using that on null-aware arguments... Want me to offer that up as a POC to discuss? We could consider it a 3.6.0 release blocker to ensure `defineBinds` is more finished instead of needing post-release work.

stevenschlansker:
Let's not rush further changes here.  I think this is an edge case we shouldn't focus too hard on right now.

TheRealMarnes:
Not in this PR no, but separately

postgres/src/main/java/org/jdbi/v3/postgres/JavaTimeArgumentFactory.java
```diff
@@ -31,22 +31,20 @@
...
+        } else if (type == OffsetDateTime.class) {
+            sqlType = Types.TIMESTAMP_WITH_TIMEZONE;
+        } else {
+            return Optional.empty();
         }
-        if (type == OffsetDateTime.class) {
-            return Optional.of(new ObjectArgument(value, Types.TIMESTAMP_WITH_TIMEZONE));
-        }
-        return Optional.empty();
+        return Optional.of(ObjectArgument.of(value, sqlType));
```
TheRealMarnes: This structure smells like I'd delegate it to a `Map<Class, Integer>` instead of a code tree if you're gonna refactor it anyway

core/src/main/java/org/jdbi/v3/core/argument/JavaTimeZoneIdArgumentFactory.java
qualidafial: v4 Argument should probably add a `getValue()` method or something, so we can stop doing these hacks

qualidafial: Or possibly, we could refactor bindings so that the value and the Argument are stored in some other type

PR #1411, opened 2019-01-15
Fix handle close + transaction handling errors

### LINKED ISSUE #1405: Built-in transaction handler causes ClassCastException when an Error is thrown
The transaction handler causes a `ClassCastException` if an `Error` is thrown while executing the transaction.

If , for example, the template engine throws an error like so:

```java
jdbi.setTemplateEngine((sql, ctx) -> {
  throw new Error();
});
```

the error isn't propagated but causes this exception instead:

```
java.lang.ClassCastException: java.lang.Error cannot be cast to java.lang.Exception
	at org.jdbi.v3.core.transaction.LocalTransactionHandler.inTransaction(LocalTransactionHandler.java:145)
	at org.jdbi.v3.core.transaction.SerializableTransactionRunner.inTransaction(SerializableTransactionRunner.java:54)
	at org.jdbi.v3.core.Handle.inTransaction(Handle.java:393)
	at org.jdbi.v3.core.Jdbi.lambda$inTransaction$5(Jdbi.java:353)
	at org.jdbi.v3.core.Jdbi.withHandle(Jdbi.java:318)
	at org.jdbi.v3.core.Jdbi.inTransaction(Jdbi.java:353)
```

### LINKED ISSUE #1406: Handle.close() might cause Connection leaks if the autocommit state of the underlying connection cannot be determined
We're using a statement timeout in our JDBC connection String. On long running statements, this causes the JDBC driver or the database to close the connection which is used by a JDBI Handle.  When JDBI tries to close the Handle, an Exception is thrown in the call to isInTransaction(). The control flow of Handle.close() is interrupted by the Exception thrown in isInTransaction(), which means the close() Method of the Connection held by the Handle is never called, causing a Connection leak that is observable with JProfiler. 

The Situation as a [Stacktrace](https://github.com/jdbi/jdbi/files/2755028/Stacktrace.txt)

Since this occurs in a Payara-Micro Appserver I'm also attaching the relevant parts of the 
[glassfish-resources.xml](https://github.com/jdbi/jdbi/files/2755051/glassfish-resources.txt)

stevenschlansker:
Fixes #1405 
Fixes #1406

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -16,6 +16,8 @@ package org.jdbi.v3.core;
 import java.io.Closeable;
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
 
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.config.Configurable;
@@ -122,47 +124,55 @@ public class Handle implements Closeable, Configurable<Handle> {
      */
     @Override
     public void close() {
+        final List<Throwable> suppressed = new ArrayList<>();
         if (closed) {
             return;
         }
 
-        boolean wasInTransaction = isInTransaction()
-            && forceEndTransactions
-            && config.get().get(Handles.class).isForceEndTransactions();
+        boolean wasInTransaction = false;
+        if (forceEndTransactions && config.get().get(Handles.class).isForceEndTransactions()) {
+            try {
+                wasInTransaction = isInTransaction();
+            } catch (Exception e) {
+                suppressed.add(e);
+            }
+        }
 
         extensionMethod.remove();
         config.remove();
 
         if (wasInTransaction) {
-            rollback();
+            try {
+                rollback();
+            } catch (Exception e) {
+                suppressed.add(e);
+            }
         }
 
-        Exception suppressed = null;
         try {
             statementBuilder.close(getConnection());
         } catch (Exception e) {
-            suppressed = e;
+            suppressed.add(e);
         }
 
         try {
             closer.close(connection);
 
+            if (!suppressed.isEmpty()) {
+                final Throwable original = suppressed.remove(0);
+                suppressed.forEach(original::addSuppressed);
+                throw new CloseException("Failed to clear transaction status on close", original);
+            }
             if (wasInTransaction) {
-                TransactionException txe = new TransactionException("Improper transaction handling detected: A Handle with an open "
+                throw new TransactionException("Improper transaction handling detected: A Handle with an open "
                     + "transaction was closed. Transactions must be explicitly committed or rolled back "
                     + "before closing the Handle. "
                     + "Jdbi has rolled back this transaction automatically. "
                     + "This check may be disabled by calling getConfig(Handles.class).setForceEndTransactions(false).");
-                if (suppressed != null) {
-                    txe.addSuppressed(suppressed);
-                }
-                throw txe;
             }
         } catch (SQLException e) {
             CloseException ce = new CloseException("Unable to close Connection", e);
-            if (suppressed != null) {
-                ce.addSuppressed(suppressed);
-            }
+            suppressed.forEach(ce::addSuppressed);
             throw ce;
         } finally {
             LOG.trace("Handle [{}] released", this);
--- a/core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
+++ b/core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
@@ -135,7 +135,7 @@ public class LocalTransactionHandler implements TransactionHandler {
             if (!didTxnRollback.get()) {
                 handle.commit();
             }
-        } catch (Exception e) {
+        } catch (Throwable e) {
             try {
                 handle.rollback();
             } catch (Exception rollback) {
--- a/core/src/test/java/org/jdbi/v3/core/TestHandle.java
+++ b/core/src/test/java/org/jdbi/v3/core/TestHandle.java
@@ -14,6 +14,9 @@
 package org.jdbi.v3.core;
 
 import org.jdbi.v3.core.rule.H2DatabaseRule;
+import org.jdbi.v3.core.statement.UnableToCreateStatementException;
+import org.jdbi.v3.core.transaction.LocalTransactionHandler;
+import org.jdbi.v3.core.transaction.TransactionException;
 import org.jdbi.v3.core.transaction.TransactionIsolationLevel;
 import org.jdbi.v3.core.transaction.UnableToManipulateTransactionIsolationLevelException;
 import org.junit.Rule;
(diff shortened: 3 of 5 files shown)


core/src/test/java/org/jdbi/v3/core/TestHandle.java
```diff
@@ -77,4 +80,61 @@ public void unknownTransactionLevelIsOk() {
...
+        final BoomHandler handler = new BoomHandler();
+        dbRule.getJdbi().setTransactionHandler(handler);
+        final Handle h = dbRule.openHandle();
+
+        assertThat(h.isClosed()).isFalse();
+
+        handler.failRollback = true;
+        assertThatThrownBy(() -> h.useTransaction(h2 -> h2.execute("insert into true")))
+            .isInstanceOf(UnableToCreateStatementException.class);
+        assertThat(h.isInTransaction()).isTrue(); // rollback failed!
```
TheRealMarnes: Comments clarifying tests should be added with `.describedAs(String)` instead of java source comments

stevenschlansker:
@qualidafial if you have time to review this as well, I'd appreciate it, given how sensitive the code it touches is...

core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
```diff
@@ -135,7 +135,7 @@ public boolean isInTransaction(Handle handle) {
             if (!didTxnRollback.get()) {
                 handle.commit();
             }
-        } catch (Exception e) {
+        } catch (Exception | Error e) {
```
sarxos: Hi @stevenschlansker,

I think this should be `Throwable`. The `Error` case is very specific and I assume it was only provided in #1405 as an example to demonstrate the problem, but in a wild there is a bunch of `Throwable` monsters (especially in Scala) which are neither `Exception` nor `Error` but which can easily flip stack trace upside down.

Therefore, if there are no other reasons to limit a range of throwables being caught here, I think this should be changed to:

```java
} catch (Throwable t) {
```

Both `Error` and `Exception` are instances of `Throwable`, but the `Error` is **not** instance of `X extends Exception`. I guess this works only because IDE or compiler did not pick it up since generics are erased in compile time and `X extends Exception` disappears.

RELEASE_NOTES
```diff
@@ -35,6 +35,8 @@
       from the BOM.
   - API changes
     - SQLitePlugin now has the ServiceLoader manifest it deserves for automagical installation.
+  - Improvements
+    - local transaction handler: handle Errors and ensure we rollback
```
qualidafial: Needs to move to 3.7.0 section, in new RELEASE_NOTES.md file

core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -122,47 +124,59 @@ public Handle setStatementBuilder(StatementBuilder builder) {
...
         if (closed) {
             return;
         }
 
-        boolean wasInTransaction = isInTransaction()
-            && forceEndTransactions
-            && config.get().get(Handles.class).isForceEndTransactions();
+        boolean wasInTransaction = forceEndTransactions
+                && config.get().get(Handles.class).isForceEndTransactions();
+
```
qualidafial: Why did the `isInTransaction()` part get removed?

core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
qualidafial: The `X extends Exception` approach is how we get exception transparency in Jdbi calls. That is, `inTransaction()` can only throw whatever checked exceptions are thrown by your lambda. That's how we got away from wrapping every exception in e.g. `DBIException` like we did in v2.

core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -122,47 +124,59 @@ public Handle setStatementBuilder(StatementBuilder builder) {
...
+        if (wasInTransaction) {
+            try {
+                wasInTransaction &= isInTransaction();
```
qualidafial: Can we move this stuff into `isInTransaction()` instead? We could probably leave this method as is if `isInTransaction()` checked whether the connection was already closed--unless I'm missing something.

core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
qualidafial: From Error javadoc:

> An Error is a subclass of Throwable that indicates serious problems that a reasonable application should not try to catch.

Since we're no longer catching `Throwable` and casting to `X`, do we actually need this change?

stevenschlansker: > The X extends Exception approach is how we get exception transparency in Jdbi calls.
(although it really ought to have been `X extends Throwable`, but we didn't figure that out until too late to change it for v3)

TheRealMarnes: Why not catch throwable instead of only the subtypes error and exception?

core/src/main/java/org/jdbi/v3/core/Handle.java
stevenschlansker: It didn't; it's simply moved down below.  I reorganized the `A() && B && C` clause to `if (B && C) { A() }` since two of them are local booleans and `isInTransaction` potentially calls out to the remote database server or throws.

core/src/main/java/org/jdbi/v3/core/Handle.java
stevenschlansker: I don't think so.  If we move the exception catching into `isInTransaction` we are then forced to handle the exception then and there and invent a fallback `boolean` value -- do we pretend to be in a transaction?  or not?  neither is right all the time.  so I don't think the contract of `isInTransaction` should change -- if it can't determine it, it should throw.

stevenschlansker: This guards against arbitrary problems -- imagine an `EOFException` while trying to query the server's transaction status -- not just concurrent closes.

core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
stevenschlansker: I'm surprised `catch (Throwable e)` compiles.  The method `throws X (extends Exception)` and there are many throwables that don't match that signature... you must be right about funky erasure semantics here.

stevenschlansker:
I rebased this PR against master and went through all the comments.  One last review?

core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -122,47 +124,57 @@ public Handle setStatementBuilder(StatementBuilder builder) {
...
             if (wasInTransaction) {
                 TransactionException txe = new TransactionException("Improper transaction handling detected: A Handle with an open "
                     + "transaction was closed. Transactions must be explicitly committed or rolled back "
                     + "before closing the Handle. "
                     + "Jdbi has rolled back this transaction automatically. "
                     + "This check may be disabled by calling getConfig(Handles.class).setForceEndTransactions(false).");
-                if (suppressed != null) {
-                    txe.addSuppressed(suppressed);
-                }
+                suppressed.forEach(txe::addSuppressed);
```
sarxos: This lambda code is dead because `suppressed` list is **always** empty here. This is because for non-empty `supressed` list the `if` block in 161 is executed which throws `CloseException` in 164 making this one unreachable. Therefore, the only way for this line to run is to have `suppressed` list empty, but, because it's empty the lambda within `forEach` will never run.

I suggest to unify suppression propagation in a `catch` blocks to make it simpler:

```java
try {
  closer.close()
  if (wasInTransaction && suppressed.isEmpty()) {
    throw new TransactionException("Improper ...");
  }
} catch (Throwable t) {
  CloseException ce = new CloseException("Unable to close Connection", t);
  suppressed.forEach(ce::addSuppressed);
  throw ce;
} finally {
  closed = true;
}

if (!suppressed.isEmpty()) {
  CloseException ce = new CloseException("Failed to clear transaction status on close", suppressed.remove(0));
  suppressed.forEach(ce::addSuppressed);
  throw ce;  
}
```

The above code is just to present an idea. I was writing it without IDE so it may may not compile or may be buggy or violate some dependencies or requirements in call stack which I'm not aware of.

qualidafial: I think @sarxos has it right. This line of code is unreachable if `suppressed` is non-empty. So this line is superfluous.

core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -122,47 +124,57 @@ public Handle setStatementBuilder(StatementBuilder builder) {
...
                 throw txe;
             }
         } catch (SQLException e) {
             CloseException ce = new CloseException("Unable to close Connection", e);
-            if (suppressed != null) {
-                ce.addSuppressed(suppressed);
-            }
+            suppressed.forEach(ce::addSuppressed);
```
qualidafial: I believe this line is similarly unnecessary.

core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
```diff
@@ -135,7 +135,7 @@ public boolean isInTransaction(Handle handle) {
             if (!didTxnRollback.get()) {
                 handle.commit();
             }
-        } catch (Exception e) {
+        } catch (Throwable e) {
```
qualidafial: From the Throwable javadoc:

```
For the purposes of compile-time checking of exceptions, {@code
Throwable} and any subclass of {@code Throwable} that is not also a
subclass of either {@link RuntimeException} or {@link Error} are
regarded as checked exceptions.
```

So I don't understand how the compiler allows `throw e` in this context.

TheRealMarnes: My money is on generics erasure shenanigans confusing the compiler. Report it as a bug and win 5 million bucks for finding a flaw in java?

TheRealMarnes: Explicitly writing ` throw new Throwable();` anywhere in the method still gives a compile error...

TheRealMarnes: Exhibit A ^^

image

TheRealMarnes: Yeah, looks like the compiler can narrow the caught exception type on its own regardless of the declaration:
image

So it probably omits the "can't throw a Throwable here" because it knows you'll never catch a `Throwable`.

qualidafial: Huh. Well that's nifty.

TheRealMarnes: For completeness:
image

core/src/main/java/org/jdbi/v3/core/Handle.java
stevenschlansker: I think this one's actually needed, in case e.g. the `statementBuilder` throws and then so does `closer.close`.

PR #1415, opened 2019-01-15
build: dokka skips non-doc Travis same as javadoc

### LINKED ISSUE #1413: build: dokka shouldn't run in ci tests
it's slow and we skip building regular docs already

stevenschlansker:
Fixes #1413

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -532,6 +532,7 @@
                     <version>${dep.dokka.version}</version>
                     <configuration>
                         <jdkVersion>8</jdkVersion>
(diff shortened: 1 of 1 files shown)


stevenschlansker:
Confirmed in the build output:
```
[INFO] --- dokka-maven-plugin:0.9.17:dokka (default) @ jdbi3-kotlin ---
[INFO] Dokka skip parameter is true so no dokka output will be produced
```

TheRealMarnes:
🚢

PR #1423, opened 2019-01-19
fix #1422

### LINKED ISSUE #1422: test: Nondeterminstic `TestTimestampted` test failures
Observed in Travis build https://travis-ci.org/jdbi/jdbi/jobs/481605774

```
[ERROR] Failures: 
[ERROR]   TestTimestamped.shouldAllowCustomTimestampParameter:88 
Expecting:
 <2019-01-19T01:49:50.767+02:00>
to be between:
 [2019-01-18T23:49:50.766Z, 2019-01-18T23:49:50.767Z]
```

I'm not sure exactly what's going on with this failure.

TheRealMarnes:
fixes #1422

Add 1ms of margin to the assertions in case travis is having a fast day, because the subject offsetdatetime at offset +2 gets sorted after the test-end offsetdatetime in zulu if it has the same time component.

### DIFF
--- a/sqlobject/src/test/java/org/jdbi/v3/sqlobject/MockClock.java
+++ b/sqlobject/src/test/java/org/jdbi/v3/sqlobject/MockClock.java
@@ -41,7 +41,8 @@ public class MockClock extends Clock {
         return now.toInstant();
     }
 
-    public void advance(long amountToAdd, TemporalUnit unit) {
+    public Instant advance(long amountToAdd, TemporalUnit unit) {
         now = now.plus(amountToAdd, unit);
+        return instant();
     }
 }
--- a/sqlobject/src/test/java/org/jdbi/v3/sqlobject/MockClock.java
+++ b/sqlobject/src/test/java/org/jdbi/v3/sqlobject/MockClock.java
@@ -22,7 +22,7 @@ import java.time.temporal.TemporalUnit;
 public class MockClock extends Clock {
     private ZonedDateTime now;
 
-    public MockClock(ZonedDateTime now) {
+    private MockClock(ZonedDateTime now) {
         this.now = now;
     }
 
@@ -44,5 +44,13 @@ public class MockClock extends Clock {
     public Instant advance(long amountToAdd, TemporalUnit unit) {
         now = now.plus(amountToAdd, unit);
         return instant();
+    }
+
+    public static MockClock now() {
+        return at(ZonedDateTime.now());
+    }
+
+    public static MockClock at(ZonedDateTime now) {
+        return new MockClock(now);
     }
 }
--- a/sqlobject/src/test/java/org/jdbi/v3/sqlobject/MockClock.java
+++ b/sqlobject/src/test/java/org/jdbi/v3/sqlobject/MockClock.java
@@ -16,27 +16,32 @@ package org.jdbi.v3.sqlobject;
 import java.time.Clock;
 import java.time.Instant;
 import java.time.ZoneId;
+import java.time.ZonedDateTime;
 import java.time.temporal.TemporalUnit;
 
-class MockClock extends Clock {
-    private Instant now = Instant.now();
+public class MockClock extends Clock {
+    private ZonedDateTime now;
+
+    public MockClock(ZonedDateTime now) {
+        this.now = now;
+    }
 
     @Override
     public ZoneId getZone() {
-        return ZoneId.systemDefault();
+        return now.getZone();
     }
 
     @Override
     public Clock withZone(ZoneId zone) {
-        throw new UnsupportedOperationException();
+        return new MockClock(now.withZoneSameInstant(zone));
     }
 
     @Override
     public Instant instant() {
-        return now;
+        return now.toInstant();
     }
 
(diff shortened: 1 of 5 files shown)


stevenschlansker:
Ah, indeed.  The `OffsetDateTime` implementation falls back to comparing the *local* date time if they otherwise compare equal, which makes the `isBefore` predicate have surprising behavior.

Yuck.

The fix isn't beautiful but I think it's likely to work.  Thanks for tracking this down.

TheRealMarnes:
I don't really see the surprise? They compare by local datetime if their timezone is the same, and by Instant if not. So really what you get is ordering by absolute time, with ties resolved by timezone. Sounds pretty beautiful to me.

stevenschlansker:
For sorting, it's beautiful indeed.
For containment, it's less beautiful -- whether a `OffsetDateTime` range contains an instant at the edge depends on the timezone being before or after the reference, which isn't immediately obvious.  As is evidenced by the test being written wrong :)

TheRealMarnes:
too true

TheRealMarnes:
Urgh, here we go. My style settings from one project keep getting synced over the other.

TheRealMarnes:
@stevenschlansker I've pushed some additional changes to improve `MockClock` overall (implement full API, cleaner usage that doesn't leak implementation) and then make use of it in this test to remove the timing uncertainty by using it as the time source in the relevant factory.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/MockClock.java
```diff
@@ -16,27 +16,32 @@
...
 
     @Override
     public Instant instant() {
-        return now;
+        return now.toInstant();
     }
 
-    public Instant advance(long amountToAdd, TemporalUnit unit) {
-        return now = now.plus(amountToAdd, unit);
+    public void advance(long amountToAdd, TemporalUnit unit) {
```
stevenschlansker: Why not have this still return `Instant`?

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestInheritedAnnotations.java
```diff
@@ -57,7 +58,8 @@ public void testCrud() {
 
         assertThat(dao.findById(1)).contains(new Character(1, "Moiraine Sedai", inserted, inserted));
 
-        Instant modified = mockClock.advance(10, SECONDS);
```
stevenschlansker: This was just fine as it was

sqlobject/src/test/java/org/jdbi/v3/sqlobject/MockClock.java
TheRealMarnes: It returning an Instant was a nice coincidence with its implementation rather than the logical outcome of telling a clock to advance. It's easy to put back but I felt it oppurtunistic instead of deliberate.

TheRealMarnes: With its current impl it would return a ZonedDateTime instead, or need to convert it to an Instant to stay the same. That being weird code to have in that location tells me the premise itself is weird. Clocks already have proper API to return the time, and that's what should be used instead of conveniently returning it here because we can.

Does that convince you or do I put it back?

stevenschlansker: How is the premise weird here?  This is a simple accumulate operation like `+=`, except that we operate on a time rather than an int.  Or comparable to `AtomicLong.incrementAndGet`.  This isn't a coincidence, it was intentional, and it made using the class nicer.

I'm somewhat against returning a `ZonedDateTime` here.  I think that makes the API less tailored to our use case -- `jdbi` code should be avoiding timezone handling as much as possible, and we don't want to encourage its spread.  Additionally, as you point out, the `Clock` interface returns `Instant`s.  So it's weird for a `MockClock` to then return a `ZonedDateTime` instead when a normal Clock only does Instant.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestInheritedAnnotations.java
```diff
@@ -37,7 +38,7 @@
     @Rule
     public H2DatabaseRule dbRule = new H2DatabaseRule().withPlugin(new SqlObjectPlugin());
 
-    private MockClock mockClock = new MockClock();
+    private MockClock mockClock = new MockClock(ZonedDateTime.now());
```
stevenschlansker: How come we removed the default constructor?  Now all use sites have to do this repeatedly.

TheRealMarnes: To make it explicit in the tests themselves that you're relying on whatever the current time is, which is an uncertainty, rather than it being hidden as an implementation detail you could overlook when reading a test.

TheRealMarnes: Would factory methods `now()` and `at(...)` suit you?

PR #1424, opened 2019-01-20
Fix typo in Javadoc

dhardtke:
Found a small typo 🥇

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
@@ -22,7 +22,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 
-import edu.umd.cs.findbugs.annotations.SuppressWarnings;
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.result.ResultBearing;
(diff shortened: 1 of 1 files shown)


dhardtke:
It should of course say Javadoc instead of documentation - it's late, I should probably go to bed

core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
```diff
@@ -22,6 +22,7 @@
 import java.util.Map;
 import java.util.NoSuchElementException;
 
+import edu.umd.cs.findbugs.annotations.SuppressWarnings;
```
TheRealMarnes: This import needs to be removed, it changes the SuppressWarnings down below from the java lang one to the findbugs one.

TheRealMarnes:
Remove the stowaway import and we'll 🚢

Thanks!

core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
dhardtke: Got it. Weird, dunno how it got there in the first place.

TheRealMarnes: probably IDE auto-import

Issue #602, opened 2016-12-06
Support Postgres custom struct types


stevenschlansker:
We should at least document #589 if not provide a cool way to bind / map custom types with e.g. `ConstructorMapper`

spasa:
Hi, guys.

I would like to give it a try to implement this. Could I get some pointers?

Should, perhaps, implement it as a plugin?

stevenschlansker:
hi @spasa we'd love to see a pass at this!  I recommend first building it as a standalone project, to show it working without having to make modifications to jdbi itself.  Once you have a good idea of how to accomplish your goal, it could make sense to include it inside of `jdbi3-postgres` or make another module if it is large enough to be standalone.

spasa:
Hi @stevenschlansker, you can take a look at repo [postgres-types](https://github.com/spasa/postgres-types). There is a very little code, so I think it could go with `jdbi3-postgres` module. I still need to write some tests(for now just test it manually), put comments, doc... But, basically, the code in repo is my solution.

First, I did some research, and thought that I could use [SQLData](https://docs.oracle.com/javase/tutorial/jdbc/basics/sqlcustommapping.html), and implement custom types not just for postgres, but other databases too. Unfortunatelly, postgres driver doesn't support(implement) this, so we needed to go with PGobject.

This will be used by registering a type by calling next method `PostgresTypes.registerCustomType(String typeName, Class clazz)`. 

Now, just a few thing to mention. If we have, for example, stored procedure named `get_all_users()` which returns setof `user_type` custom type (for eg it has next fields `id, first_name, last_name`, so something like this `(1, 'John', 'Doe')` ), the result will vary depending on how we write a query:

- `SELECT get_all_users()` will return rows with `user_type` data type as a column, `(1, 'John', 'Doe')`. I wrote an column mapper which can be used in this case. 
`List<PGobject> result = handle.createQuery("SELECT get_all_users()")
                   .map(new PGObjectMapper())
                   .list();`
and after that, we can type cast it to any class that we used(need):
`List<User> users = (List<User>)(List<?>) result;`

- `SELECT * FROM get_all_users()` will return rows, but with all field from `user_type` as columns (we will have 3 columns(id, first_name and last_name)). In this case, any RowMapper or Constructor mapper could be used(just as with any other query).

In most of the cases, custom types are used as return type from stored procedures.

stevenschlansker:
[...]

PR #1425, opened 2019-01-20
add support for postgres custom types

spasa:
Hi guys,
this is my take on issue https://github.com/jdbi/jdbi/issues/602


postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArgumentFactory.java
```diff
@@ -0,0 +1,49 @@
...
+import org.jdbi.v3.core.argument.AbstractArgumentFactory;
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.statement.StatementContext;
+import org.postgresql.util.PGobject;
+
+/**
+ * Argument factory for {@link PGobject}.
+ */
+public class PGObjectArgumentFactory extends AbstractArgumentFactory<PGobject> {
```
TheRealMarnes: I'm a little worried about how everything so far in the postgres module is `public` and non-`internal`. Is it supposed to be that way for some reason or should we avoid doing it on any new additions? @stevenschlansker @qualidafial

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArgumentFactory.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public PGObjectArgumentFactory() {
+        super(Types.OTHER);
+    }
+
+    @Override
+    protected Argument build(PGobject value, ConfigRegistry config) {
+        return new Argument() {
```
TheRealMarnes: You need to use `LoggableBinderArgument` to ensure the object can be properly logged by `SqlLogger` (lambda `Argument`s don't implement `toString`).

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArgumentFactory.java
```diff
@@ -0,0 +1,49 @@
...
+
+            @Override
+            public void apply(int position, PreparedStatement statement, StatementContext ctx) throws SQLException {
+                ((org.postgresql.PGConnection) statement.getConnection()).addDataType(PostgresTypes.getTypeName(value.getClass()), value.getClass());
```
TheRealMarnes: Personal nitpick: I'd like to see the cast object extracted to a variable (`PGConnection conn = ... `), because this line is pretty long and hard to read, and I think there needs to be a `SuppressWarnings` on the cast, so extracting it to a variable also neatly limits the scope of the suppression.

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArrayArgumentFactory.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+public class PGObjectArrayArgumentFactory extends AbstractArgumentFactory<PGobject[]> {
+
+    public PGObjectArrayArgumentFactory() {
+        super(Types.ARRAY);
+    }
+
+    @Override
+    protected Argument build(PGobject[] value, ConfigRegistry config) {
+        return new Argument() {
```
TheRealMarnes: idem `LoggableBinderArgument`

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArrayArgumentFactory.java
```diff
@@ -0,0 +1,50 @@
...
+
+            @Override
+            public void apply(int position, PreparedStatement statement, StatementContext ctx) throws SQLException {
+                ((org.postgresql.PGConnection) statement.getConnection()).addDataType(PostgresTypes.getTypeName(value.getClass().getComponentType()), (Class<? extends PGobject>) value.getClass().getComponentType());
```
TheRealMarnes: idem casts

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectColumnMapper.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Column mapper for {@link PGobject}.
+ */
+public class PGObjectColumnMapper implements ColumnMapper<PGobject> {
+
+    @Override
+    public PGobject map(ResultSet rs, int columnNumber, StatementContext ctx) throws SQLException {
+        PostgresTypes.addTypesToConnection((PGConnection) ctx.getConnection());
+
+        return (PGobject) rs.getObject(columnNumber);
```
TheRealMarnes: I think `SuppressWarnings` is in order around here too

postgres/src/main/java/org/jdbi/v3/postgres/PostgresTypes.java
```diff
@@ -0,0 +1,59 @@
...
+        return TYPES.get(clazz);
+    }
+
+    /**
+     * Add handler for each registered PostgreSQL custom type
+     *
+     * @param connection connection on which to add all registered PostgreSQL custom types
+     */
+    public static void addTypesToConnection(PGConnection connection) throws SQLException {
+        Iterator it = TYPES.entrySet().iterator();
```
TheRealMarnes: I'd like to see modern iteration used here:
```
TYPES.forEach((clazz, name) -> connection.addDataType(name, clazz));
```

postgres/src/test/java/org/jdbi/v3/postgres/FooBarPGType.java
```diff
@@ -0,0 +1,108 @@
...
+        PGtokenizer t = new PGtokenizer(PGtokenizer.removePara(value), ',');
+
+        id = Integer.valueOf(t.getToken(0));
+        foo = t.getToken(1);
+        bar = t.getToken(2);
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = 7;
```
TheRealMarnes: please change this to simply `return Objects.hash(...)` to keep the boilerplate/LOC down

postgres/src/test/java/org/jdbi/v3/postgres/FooBarPGType.java
```diff
@@ -0,0 +1,108 @@
...
+            return true;
+        }
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final FooBarPGType other = (FooBarPGType) obj;
+        if (!Objects.equals(this.foo, other.foo)) {
```
TheRealMarnes: please change this to a chain of `Objects.equals(a, b) && Objects.equals(c, d) && ...` to keep the boilerplate/LOC down

postgres/src/test/java/org/jdbi/v3/postgres/TestPostgresTypes.java
```diff
@@ -0,0 +1,215 @@
...
+        assertThat(result).isEqualTo(new FooBarPGType(1, "foo1", "bar1"));
+    }
+
+    @Test
+    public void testReadListViaFluentAPI() {
+        List<PGobject> result = handle.createQuery("SELECT get_foo_bars()")
+                .map(new PGObjectColumnMapper())
+                .list();
+
+        List<FooBarPGType> fooBars = (List<FooBarPGType>) (List<?>) result;
```
TheRealMarnes: what's up with this triple cast?

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArrayArgumentFactory.java
spasa: @TheRealMarnes Would you care to elaborate a little bit this?
I do not know how `LoggableBinderArgument` works, but it seems to me that this just bind any argument as a String.

I couldn't find documentation about this. I found some usage in code (ie. `JavaTimeZoneIdArgumentFactory` ).

Keep in mind that I want to bind an Array here.

spasa: @TheRealMarnes Using LoggableBinderArgument now. I couldn't find any doc on this, so a double check on this would be greatly appreciated. Thanks

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArgumentFactory.java
spasa: @TheRealMarnes Using LoggableBinderArgument now. I couldn't find any doc on this, so a double check on this would be greatly appreciated. Thanks

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArgumentFactory.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class PGObjectArgumentFactory extends AbstractArgumentFactory<PGobject> {
+
+    public PGObjectArgumentFactory() {
+        super(Types.OTHER);
+    }
+
+    @Override
+    protected Argument build(PGobject value, ConfigRegistry config) {
+        return new LoggableBinderArgument<>(value, new StatementBinder<PGobject>() {
```
TheRealMarnes: Please use lambdas rather than full anonynmous classes to keep the code concise. The mantra for binders is `(p, i, v) -> {}`.

Good work so far otherwise though :)

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArrayArgumentFactory.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+public class PGObjectArrayArgumentFactory extends AbstractArgumentFactory<PGobject[]> {
+
+    public PGObjectArrayArgumentFactory() {
+        super(Types.ARRAY);
+    }
+
+    @Override
+    protected Argument build(PGobject[] value, ConfigRegistry config) {
+        return new LoggableBinderArgument<>(value, new StatementBinder<PGobject[]>() {
```
TheRealMarnes: idem lambdas

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArgumentFactory.java
spasa: @TheRealMarnes Sorry, missed it. I will fix them.
I can also give you write access, but I will try not to make miss those " finnicky details"

postgres/src/main/java/org/jdbi/v3/postgres/PostgresTypes.java
```diff
@@ -48,11 +50,21 @@ public static String getTypeName(Class clazz) {
...
+     * Add handler for PostgreSQL custom type
+     *
+     * @param connection connection on which to add PostgreSQL custom types
+     * @throws java.sql.SQLException
+     */
+    private static void addTypeToConnection(PGConnection connection, String type, Class<? extends PGobject> clazz) {
+        try {
+            connection.addDataType(type, clazz);
+        } catch (SQLException ex) {
+            LOG.error("Error while adding handler for PostgreSQL's custom type", ex);
```
TheRealMarnes: Did you write this method and add this logger just to swallow the exception so you could use the modern iteration? I'd like to point out two general bad practices in this idea. First, swallowing exceptions by only logging them when those exceptions aren't supposed to occur is sweeping things under the rug. Programs exploding when something goes wrong is a very, very good thing. Secondly, if you have a case like this where a checked exception gets in your way because it can't be thrown, the usual solution is to wrap it in a RuntimeException. In Jdbi we have a tool that mirrors `jool`, the `Unchecked` class, which will let you hide the exception without sweeping it under the rug or adding misinformative code:

`TYPES.forEach((clazz, type) -> Unchecked.biConsumer(connection::addDataType).accept(type, clazz));`

TheRealMarnes:
Alright, I've pushed some commits to handle all my own remarks for you :) Now we just need input from the postgres wizards

spasa:
@TheRealMarnes Thanks a lot for your help.

Should I be concerned that checks are not passing anymore?I trying to find the error, but it seems like the errors are not form postgres module

spasa:
@TheRealMarnes Nevermind, the errors are now in postgres module (test) and I will try to fix them

TheRealMarnes:
Well the current failures are all
```
[ERROR]   TestPostgresTypes.testReadListViaFluentAPI:107 
Expecting:
  <[(1,foo1,bar1), (2,foo2,bar2), (5,foo5,bar5), (6,foo6,bar6)]>
to contain exactly in any order:
  <[(1,foo1,bar1), (2,foo2,bar2)]>
but the following elements were unexpected:
  <[(5,foo5,bar5), (6,foo6,bar6)]>
```

So obviously the `containsExactlyInAnyOrder` on line 107 is the cause, but I think the result data/test setup is wrong and the test is failing as it should. Do you expect that list to contain the 5 and 6 results? It doesn't seem like it should to me but I may be wrong. The fact that you used class setup instead of test setup for a reason I don't see probably causes each test to pollute the db so that you now have superfluous return values.

TheRealMarnes:
It seems that was indeed it, the setup should have been per-test instead of just-once to keep the data clean. :)

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArrayArgumentFactory.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Override
+    protected Argument build(PGobject[] value, ConfigRegistry config) {
+        return new LoggableBinderArgument<>(value, (p, i, v) -> {
+            PGConnection pgConnection = (PGConnection) p.getConnection();
+            @SuppressWarnings("unchecked")
+            Class<? extends PGobject> componentType = (Class<? extends PGobject>) v.getClass().getComponentType();
+            String type = PostgresTypes.getTypeName(componentType);
+
+            pgConnection.addDataType(type, componentType);
```
stevenschlansker: similarly here, let's try to configure all data types on the connection once, perhaps in `PostgresPlugin.customizeConnection`.

postgres/src/main/java/org/jdbi/v3/postgres/PostgresTypes.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * @param typeName the PostgreSQL type to register
+     * @param clazz the class implementing the Java representation of the type;
+     * this class must implement {@link PGobject}.
+     */
+    public static void registerCustomType(String typeName, Class<? extends PGobject> clazz) {
+        TYPES.put(clazz, typeName);
+    }
+
+    public static String getTypeName(Class clazz) {
```
stevenschlansker: there's a couple of raw `Class` that should probably be `Class<?>` here

postgres/src/test/java/org/jdbi/v3/postgres/TestPostgresTypes.java
```diff
@@ -0,0 +1,213 @@
...
+    @Before
+    public void before() {
+        handle = postgresDbRule.getHandle();
+        handle.useTransaction(h -> {
+            h.execute("drop table if exists postgres_custom_types");
+            h.execute("create table postgres_custom_types(id integer not null, foo text, bar text, created_on timestamp)");
+
+            // create custom type
+            h.execute("drop function if exists get_foo_bars()");
+            h.execute("drop function if exists get_foo_bar(integer)");
```
stevenschlansker: I think each test gets a separate database - these `drop` statements should be unnecessary

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArrayArgumentFactory.java
spasa: @stevenschlansker I am not sure that I get it how would I implement this. Could you give me some more instructions on this?

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArgumentFactory.java
```diff
@@ -0,0 +1,45 @@
...
+public class PGObjectArgumentFactory extends AbstractArgumentFactory<PGobject> {
+
+    public PGObjectArgumentFactory() {
+        super(Types.OTHER);
+    }
+
+    @Override
+    protected Argument build(PGobject value, ConfigRegistry config) {
+        return new LoggableBinderArgument<>(value, (statement, index, val) -> {
+            PGConnection pgConnection = (PGConnection) statement.getConnection();
```
qualidafial: Connections can be wrapped, e.g. in a thread pool. Should use `statement.getConnection().unwrap(PGConnection.class)` instead of casting.

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArgumentFactory.java
```diff
@@ -0,0 +1,45 @@
...
+            String type = PostgresTypes.getTypeName(val.getClass());
+
+            pgConnection.addDataType(type, val.getClass());
```
qualidafial: Isn't `val.getClass()` always `PGobject.class`?

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArrayArgumentFactory.java
```diff
@@ -0,0 +1,49 @@
...
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.argument.internal.strategies.LoggableBinderArgument;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.postgresql.PGConnection;
+import org.postgresql.util.PGobject;
+
+/**
+ * Argument factory for array of {@link PGobject}.
+ */
+public class PGObjectArrayArgumentFactory extends AbstractArgumentFactory<PGobject[]> {
```
qualidafial: Jdbi already provides array support--you just have to register an array type and tell Jdbi the SQL type name that goes with the Java class:

```java
jdbi.registerArrayType("point", PGpoint.class);
```

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArrayArgumentFactory.java
qualidafial: Creating a Jdbi plugin is as simple as implementing the `JdbiPlugin` interface and passing it to `jdbi.installPlugin(plugin)`.

The plugin interface has three methods: `customizeJdbi()`, `customizeConnection()`, and `customizeHandle()`.

Since you're calling `PGConnection.addDataType` you would want to override `customizeConnection()` and do that work in the plugin, rather than on the argument factory.

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArrayArgumentFactory.java
```diff
@@ -0,0 +1,49 @@
...
+    protected Argument build(PGobject[] value, ConfigRegistry config) {
+        return new LoggableBinderArgument<>(value, (statement, index, val) -> {
+            PGConnection pgConnection = (PGConnection) statement.getConnection();
+            @SuppressWarnings("unchecked")
+            Class<? extends PGobject> componentType = (Class<? extends PGobject>) val.getClass().getComponentType();
+            String type = PostgresTypes.getTypeName(componentType);
+
+            pgConnection.addDataType(type, componentType);
+            Array array = statement.getConnection().createArrayOf(type, val);
+            statement.setArray(index, array);
```
qualidafial: Array is a database resource that must be freed after the statement has executed, or else the resource will be leaked.

```java
statement.getContext().addCleanable(array::free);
```

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArrayArgumentFactory.java
spasa: @qualidafial thanks for the help, I will try it.

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectColumnMapper.java
qualidafial: Could use `rs.getObject(columnNumber, PGobject.class)` instead

postgres/src/main/java/org/jdbi/v3/postgres/PostgresTypes.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.jdbi.v3.core.internal.exceptions.Unchecked;
+import org.postgresql.PGConnection;
+import org.postgresql.util.PGobject;
+
+/**
+ * Handler for PostgreSQL custom types.
+ */
+public class PostgresTypes {
```
qualidafial: Should this be a `JdbiConfig` class instead of static singleton?

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArgumentFactory.java
spasa: No, it is not. In case of our test, and `org.jdbi.v3.postgres.FooBarPGType` which extends `PGobject` class, `val.getClass()` will return `org.jdbi.v3.postgres.FooBarPGType.class`

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArgumentFactory.java
```diff
@@ -0,0 +1,45 @@
...
+        super(Types.OTHER);
+    }
+
+    @Override
+    protected Argument build(PGobject value, ConfigRegistry config) {
+        return new LoggableBinderArgument<>(value, (p, i, v) -> {
+            PGConnection pgConnection = (PGConnection) p.getConnection();
+            String type = PostgresTypes.getTypeName(v.getClass());
+
+            pgConnection.addDataType(type, v.getClass());
```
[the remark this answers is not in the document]
spasa: Yes, this is already gone, will push code in few minutes

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArrayArgumentFactory.java
spasa: @qualidafial at which point should we do this(`jdbi.registerArrayType("point", PGpoint.class)`)?

qualidafial:
@spasa Would you mind granting push access to your PR branch? https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

I'm happy to explain all the changes to you, but have limited time today and would like to get this one merged soon. Other than some API design concerns it's a slam dunk, functionality wise.

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectArgumentFactory.java
qualidafial: +1 to make package private

postgres/src/main/java/org/jdbi/v3/postgres/PGObjectColumnMapper.java
```diff
@@ -0,0 +1,33 @@
...
+import java.sql.SQLException;
+
+import org.jdbi.v3.core.mapper.ColumnMapper;
+import org.jdbi.v3.core.statement.StatementContext;
+import org.postgresql.util.PGobject;
+
+/**
+ * Column mapper for {@link PGobject}.
+ */
+public class PGObjectColumnMapper implements ColumnMapper<PGobject> {
```
qualidafial: This should probably be a ColumnMapperFactory so that we can explicitly column mappers for `PGobject` _and_ all its registered subtypes.

This way we can also _not_ provide a column mapper for `PGobject` subtypes which have _not_ been registered through `PostgresTypes`.

Also, please make class private

spasa:
@qualidafial Can you check now, you should be able to push

postgres/src/main/java/org/jdbi/v3/postgres/PostgresTypes.java
qualidafial: Looking at this again, I think it _should_ be a JdbiConfig class.

We'll have to add a new method overload `JdbiPlugin.customizeConnection(Connection, ConfigRegistry)` to give `PostgresPlugin.customizeConnection()` access to the config registry.

postgres/src/main/java/org/jdbi/v3/postgres/PostgresTypes.java
```diff
@@ -0,0 +1,62 @@
...
+    public static void addTypesToConnection(PGConnection connection) {
+        TYPES.forEach((clazz, type) -> Unchecked.<String, Class>biConsumer(connection::addDataType).accept(type, clazz));
+    }
+
+    /**
+     * Register array element type for each registered PostgreSQL custom type
+     *
+     * @param jdbi {@link Jdbi} on which to register array element type
+     */
+    public static void registerArrayTypes(Jdbi jdbi) {
```
qualidafial: This method can be inlined into `registerCustomType()` once this is a `JdbiConfig`

postgres/src/test/java/org/jdbi/v3/postgres/TestPostgresTypes.java
qualidafial: Even if the same database is used, down below all the statements are using `CREATE OR REPLACE FUNCTION` so these drop statements are definitely redundant.

postgres/src/test/java/org/jdbi/v3/postgres/TestPostgresTypes.java
```diff
@@ -0,0 +1,232 @@
...
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.postgresql.util.PGobject;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TestPostgresTypes {
```
qualidafial: Great test coverage

postgres/src/test/java/org/jdbi/v3/postgres/TestPostgresTypes.java
```diff
@@ -0,0 +1,232 @@
...
+
+    @After
+    public void after() {
+        handle.close();
+    }
+
+    @Test
+    public void testReadViaFluentAPI() {
+        FooBarPGType result = (FooBarPGType) handle.createQuery("SELECT get_foo_bar(1)")
+                .map(new PGObjectColumnMapper())
```
qualidafial: Would be better if we could use `mapTo(FooBarPGType.class)` instead of having to cast. I think the column mapper needs to be a column mapper factory for that.

postgres/src/main/java/org/jdbi/v3/postgres/PostgresTypes.java
```diff
@@ -0,0 +1,62 @@
...
+    public static String getTypeName(Class<? extends PGobject> clazz) {
+        return TYPES.get(clazz);
+    }
+
+    /**
+     * Add handler for each registered PostgreSQL custom type
+     *
+     * @param connection connection on which to add all registered PostgreSQL custom types
+     */
+    public static void addTypesToConnection(PGConnection connection) {
```
qualidafial: Make package private

docs/src/test/java/jdbi/doc/KotlinPluginTest.kt
```diff
@@ -27,7 +30,10 @@ import kotlin.test.assertEquals
 
 class KotlinPluginTest {
     @Rule @JvmField
-    val db = H2DatabaseRule().withPlugins()
+    val db = H2DatabaseRule()
+        .withPlugin(SqlObjectPlugin())
+        .withPlugin(KotlinPlugin())
+        .withPlugin(KotlinSqlObjectPlugin())
```
qualidafial: `withPlugins()` was picking up postgres plugin, which tried to use Connection.unwrap(), which H2 connection doesn't support

postgres/src/test/java/org/jdbi/v3/postgres/TestPostgresTypes.java
```diff
@@ -0,0 +1,386 @@
...
+
+        assertThat(handle.select("select :lsegs")
+            .bind("lsegs", new PGlseg[] {new PGlseg(1, 2, 3, 4), new PGlseg(5, 6, 7, 8)})
+            .mapTo(PGlseg[].class)
+            .findOnly())
+            .containsExactly(new PGlseg(1, 2, 3, 4), new PGlseg(5, 6, 7, 8));
+    }
+
+    @Test
+    public void testReadWriteMoney() {
```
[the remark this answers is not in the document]
stevenschlansker: I verified this makes the build pass even with `LANG="sr_RS"`.

alankila:
Hello,

this change unfortunately breaks use of jdbi-postgres plugin in a pooled tomcat connection setup.
The issue is with the new method, PostgresPlugin::customizeHandle, which attempts to unwrap the connection to a PGConnection, which is a concrete instance and defined locally via dependency to pg JDBC library 42.2.4. This works fine in a standalone application, because there will be no confusion about which Pg driver has been loaded.

However, if tomcat is pooling the Pg connection on behalf of the application, it will have already loaded its own pg JDBC library, and it will be serving instances of it via the DataSource interface, or some such mechanism. Additionally, tomcat gives a custom classloader to every application it is serving, so when that classloader again loads some PGConnection, the tomcat's object can not be cast to it, because identity-wise, it is not that class even if it were loaded from the same jar file. I think it would be best to only depend on the shared Connection interface, and just use reflection to access any methods of a true PGConnection, unfortunately.

You will see this stack trace due to this issue:

```
Caused by: java.sql.SQLException: Cannot unwrap to org.postgresql.PGConnection
	at org.postgresql.jdbc.PgConnection.unwrap(PgConnection.java:1494)
	at org.apache.tomcat.dbcp.dbcp2.DelegatingConnection.unwrap(DelegatingConnection.java:842)
	at org.apache.tomcat.dbcp.dbcp2.DelegatingConnection.unwrap(DelegatingConnection.java:842)
	at org.jdbi.v3.postgres.PostgresPlugin.lambda$customizeHandle$0(PostgresPlugin.java:150)
	at org.jdbi.v3.core.internal.exceptions.Unchecked.lambda$supplier$3(Unchecked.java:66)
	at org.jdbi.v3.postgres.PostgresPlugin.customizeHandle(PostgresPlugin.java:150)
	at org.jdbi.v3.core.Jdbi.open(Jdbi.java:311)
	... 27 more
```

I'm also not sure if trying to use 3.6.0 pg with 3.7.0 jdbi causes this issue:

```
java.lang.IllegalAccessError: class org.jdbi.v3.sqlobject.statement.internal.SqlUpdateHandler tried to access private method org.jdbi.v3.core.qualifier.Qualifiers.<init>()V (org.jdbi.v3.sqlobject.statement.internal.SqlUpdateHandler and org.jdbi.v3.core.qualifier.Qualifiers are in unnamed module of loader org.apache.catalina.loader.ParallelWebappClassLoader @17b016ac)
```

It does not seem to me that this is the case, but I haven't really investigated it. There may be further java 12 compatibility issues in JDBI 3.7.0, however. I have reverted to 3.6.0 for the time being.

TheRealMarnes:
@alankila could you file a new Issue for this to start a fresh and focused discussion about it, please?

PR #1426, opened 2019-01-21
enum handling @ steven

### LINKED ISSUE #1197: ArgumentFactory for enums by ordinal

### LINKED ISSUE #1273: Let user configure if enums should be mapped by name or ordinal

### LINKED ISSUE #1354: @ByName and @ByOrdinal for enum values

TheRealMarnes:
@stevenschlansker 's approach to fix #1273 fix #1354 fix #1197

Creates an argument+mapper abstraction `EnumStrategy` implemented by the 2 strategies, and a front handler `Unqualified ` that delegates unqualified enums to one of them. Shares a lot of identical code with the other PR (annotations, tests, etc) and has some minor differences like how `Enums` represents its options (enum vs classes) that can be swapped around later.

The point of these PRs is for you two to decide on which architecture you wanna roll with. Finer details can come later. I'm mildly in favor of @qualidafial 's approach but I'm perfectly fine with either one.


core/src/main/java/org/jdbi/v3/core/internal/enums/ByName.java
```diff
@@ -0,0 +1,71 @@
...
+import org.jdbi.v3.core.argument.Arguments;
+import org.jdbi.v3.core.mapper.ColumnMapper;
+import org.jdbi.v3.core.result.UnableToProduceResultException;
+import org.jdbi.v3.core.statement.StatementContext;
+
+@EnumByName
+public class ByName implements EnumStrategy {
+    private static final ByName INSTANCE = new ByName();
+
+    public static ByName singleton() {
```
TheRealMarnes: I left these in from the original rework but I think we can reasonably forget about caching the instances

stevenschlansker:
I left my thoughts on the other PR -- I think we can work from that one.

PR #1427, opened 2019-01-21
enum handling @ qualidafial

### LINKED ISSUE #1197: ArgumentFactory for enums by ordinal

### LINKED ISSUE #1273: Let user configure if enums should be mapped by name or ordinal

### LINKED ISSUE #1354: @ByName and @ByOrdinal for enum values

TheRealMarnes:
@qualidafial 's approach to fix #1273 fix #1354 fix #1197 

Creates 2 qualified factories, `QualifiedEnumArgumentFactory` and `QualifiedEnumMapperFactory`,both handling all 3 strategies. Shares a lot of identical code with the other PR (annotations, tests, etc) and has some minor differences like how `Enums` represents its options (enum vs classes) that can be swapped around later.

The point of these PRs is for you two to decide on which architecture you wanna roll with. Finer details can come later. I'm mildly in favor of @qualidafial 's approach but I'm perfectly fine with either one.


core/src/main/java/org/jdbi/v3/core/Enums.java
```diff
@@ -0,0 +1,100 @@
...
+        return deriveStrategy(givenType)
+            .orElseGet(() -> deriveStrategy(qualifyFromSource(clazz))
+                .orElse(handling));
+    }
+
+    private <E extends Enum<E>> Optional<EnumHandling> deriveStrategy(QualifiedType<E> type) {
+        boolean hasByName = type.hasQualifier(EnumByName.class);
+        boolean hasByOrdinal = type.hasQualifier(EnumByOrdinal.class);
+
+        if (hasByName && hasByOrdinal) {
```
TheRealMarnes: This is an advantage this approach has over @stevenschlansker 's. The latter can check for this conflict when adding qualifiers from source, but not when the user builds the qualifiedtype themself because the factories are matched by qualifiers and registration order.

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
```diff
@@ -53,4 +54,16 @@
             .flatMap(factory -> JdbiOptionals.stream(factory.build(expectedType, value, config)))
             .findFirst();
     }
+
+    private static class EnumByNameArgumentFactory implements ArgumentFactory {
```
TheRealMarnes: this is purely legacy support, `QualifiedEnumArgumentFactory`'s structure doesn't really allow extracting just this `ColumnMapper`

core/src/main/java/org/jdbi/v3/core/internal/QualifiedEnumMapperFactory.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private static <E extends Enum<E>> Optional<ColumnMapper<?>> makeEnumArgument(QualifiedType<E> givenType, Class<E> enumClass, ConfigRegistry config) {
+        boolean byName = Enums.EnumStrategy.BY_NAME == config.get(Enums.class).findStrategy(givenType, enumClass);
+
+        return Optional.of(byName
+            ? new EnumByNameColumnMapper<>(enumClass)
+            : new EnumByOrdinalColumnMapper<>(enumClass));
+    }
+
+    public static <E extends Enum<E>> ColumnMapper<E> byName(Class<E> enumClass) {
```
TheRealMarnes: These two factory methods serve for encapsulation but you could consider them redundant since it's all internal

core/src/main/java/org/jdbi/v3/core/internal/QualifiedEnumMapperFactory.java
```diff
@@ -0,0 +1,112 @@
...
+            return name == null ? null : getValueByName(enumClass, name, ctx);
+
+        }
+
+        private static <E extends Enum<E>> E getValueByName(Class<E> enumClass, String name, StatementContext ctx) {
+            try {
+                return Enum.valueOf(enumClass, name);
+            } catch (IllegalArgumentException ignored) {
+                return Arrays.stream(enumClass.getEnumConstants())
+                    .filter(e -> e.name().equalsIgnoreCase(name))
```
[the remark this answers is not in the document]
stevenschlansker: see #1431

core/src/main/java/org/jdbi/v3/core/Enums.java
```diff
@@ -0,0 +1,109 @@
...
+
+        return QualifiedType.of(type).with(onClass);
+    }
+
+    @Override
+    public Enums createCopy() {
+        return new Enums(this);
+    }
+
+    public enum EnumStrategy {
```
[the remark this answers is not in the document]
TheRealMarnes: Wouldn't we want to expose them to users so they validate how jdbi will handle a certain enum without doing an actual call?

core/src/main/java/org/jdbi/v3/core/mapper/EnumByNameMapperFactory.java
```diff
@@ -17,19 +17,27 @@
...
 
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.internal.QualifiedEnumMapperFactory;
 
 import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
 
 /**
  * Produces enum column mappers, which map enums from varchar columns using {@link Enum#valueOf(Class, String)}.
+ *
+ * @deprecated this class has been superseded by a new implementation
```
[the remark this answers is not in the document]
TheRealMarnes: So people can abandon the deprecated class, sure. But this also raises the question if we shouldn't discourage people from using these at all. They're needed in `resultbearing.mapto` if I'm not mistaken, but outside of that, if these mappers are registered in the chain for example, they'll cause issues.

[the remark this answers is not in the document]
qualidafial: I would point to `Enums.setEnumStrategy(BY_NAME)` as the successor in the deprecated message.

Aside from that, the main reason this class is public was to make it compatible with `@RegisterRowMapperFactory`.

If we add a `@DefaultEnumStrategy(EnumStrategy)` SQL object annotation then we have a viable successor for that scenario as well.

core/src/main/java/org/jdbi/v3/core/mapper/EnumByOrdinalMapperFactory.java
```diff
@@ -17,19 +17,28 @@
...
 
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.internal.QualifiedEnumMapperFactory;
 
 import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
 
 /**
  * Produces enum column mappers, which map enums from numeric columns according to ordinal value.
+ *
+ * @deprecated this class has been superseded by a new implementation
```
qualidafial: Message should direct users to the successor API

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
```diff
@@ -13,87 +13,44 @@
...
-import net.jodah.expiringmap.ExpirationPolicy;
-import net.jodah.expiringmap.ExpiringMap;
-import org.jdbi.v3.core.statement.StatementContext;
+import org.jdbi.v3.core.internal.QualifiedEnumMapperFactory;
 
 /**
  * Column mapper for Java {@code enum} types.
  * @param <E> the enum type mapped
+ *
+ * @deprecated this class has been superseded by a new implementation
```
qualidafial: direct users to successor API

core/src/main/java/org/jdbi/v3/core/mapper/EnumByNameMapperFactory.java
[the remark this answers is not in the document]
qualidafial: > if these mappers are registered in the chain for example, they'll cause issues.

If there is any concrete reason to suspect this will be a problem, we must fix that. We can't deprecate mappers that worked before _and_ break them.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/SqlobjectEnumQualifierTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.jdbi.v3.core.Enums;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+import org.jdbi.v3.testing.JdbiRule;
+import org.junit.Rule;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SqlobjectEnumQualifierTest {
```
qualidafial: Nit: fix class name capitalization

core/src/main/java/org/jdbi/v3/core/internal/QualifiedEnumMapperFactory.java
```diff
@@ -0,0 +1,117 @@
...
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericTypes;
+import org.jdbi.v3.core.mapper.ColumnMapper;
+import org.jdbi.v3.core.mapper.QualifiedColumnMapperFactory;
+import org.jdbi.v3.core.qualifier.QualifiedType;
+import org.jdbi.v3.core.result.UnableToProduceResultException;
+import org.jdbi.v3.core.statement.StatementContext;
+
+public class QualifiedEnumMapperFactory implements QualifiedColumnMapperFactory {
+    private static final Map<Class<? extends Enum<?>>, ColumnMapper<? extends Enum<?>>> BY_NAME_MAPPER_CACHE = new ConcurrentHashMap<>();
```
qualidafial: Use expiring map

core/src/main/java/org/jdbi/v3/core/internal/QualifiedEnumMapperFactory.java
```diff
@@ -0,0 +1,117 @@
...
+    private static <E extends Enum<E>> Optional<ColumnMapper<?>> makeEnumArgument(QualifiedType<E> givenType, Class<E> enumClass, ConfigRegistry config) {
+        boolean byName = Enums.EnumStrategy.BY_NAME == config.get(Enums.class).findStrategy(givenType, enumClass);
+
+        return Optional.of(byName
+            ? byName(enumClass)
+            : byOrdinal(enumClass));
+    }
+
+    @SuppressWarnings("unchecked")
+    public static <E extends Enum<E>> ColumnMapper<E> byName(Class<E> enumClass) {
```
qualidafial: Method name is deceptive. The method call is `QualifiedEnumMapperFactory.byName()`, which gives the impression you'll get back a factory, but you get a mapper instead.

This dissonance does not happen with the old `EnumMapper.byName()` method.

core/src/main/java/org/jdbi/v3/core/internal/QualifiedEnumArgumentFactory.java
```diff
@@ -0,0 +1,70 @@
...
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.argument.ArgumentFactory;
+import org.jdbi.v3.core.argument.Arguments;
+import org.jdbi.v3.core.argument.NullArgument;
+import org.jdbi.v3.core.argument.QualifiedArgumentFactory;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.generic.GenericTypes;
+import org.jdbi.v3.core.qualifier.QualifiedType;
+
+public class QualifiedEnumArgumentFactory implements QualifiedArgumentFactory {
```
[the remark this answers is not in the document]
qualidafial: Also moved to argument package.

core/src/main/java/org/jdbi/v3/core/internal/QualifiedEnumMapperFactory.java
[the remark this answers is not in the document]
qualidafial: Learned that the seed data doesn't change for the duration of the benchmarks, so had to go back to the drawing board.

Split out my benchmarks to see exact match, case insensitive match, and mismatch benchmarks separately:

```
Benchmark                                                         Mode  Cnt    Score    Error  Units
CaseInsensitiveStringEqualsBenchmark.collatorEqualsCopy           avgt    5  965.257 ± 28.639  ns/op
CaseInsensitiveStringEqualsBenchmark.collatorEqualsOther          avgt    5  185.364 ±  3.253  ns/op
CaseInsensitiveStringEqualsBenchmark.collatorEqualsRandomCase     avgt    5  979.489 ± 41.860  ns/op
CaseInsensitiveStringEqualsBenchmark.equalsIgnoreCaseCopy         avgt    5   20.846 ±  0.949  ns/op
CaseInsensitiveStringEqualsBenchmark.equalsIgnoreCaseOther        avgt    5   13.882 ±  0.527  ns/op
CaseInsensitiveStringEqualsBenchmark.equalsIgnoreCaseRandomCase   avgt    5   29.398 ±  0.677  ns/op
CaseInsensitiveStringEqualsBenchmark.stringEquals                 avgt    5    3.307 ±  0.040  ns/op
CaseInsensitiveStringEqualsBenchmark.stringNotEquals              avgt    5    4.562 ±  0.148  ns/op
CaseInsensitiveStringEqualsBenchmark.toLowerCaseEqualsCopy        avgt    5   73.976 ±  3.852  ns/op
CaseInsensitiveStringEqualsBenchmark.toLowerCaseEqualsOther       avgt    5   58.802 ±  2.628  ns/op
CaseInsensitiveStringEqualsBenchmark.toLowerCaseEqualsRandomCase  avgt    5   90.433 ±  3.479  ns/op
```

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
```diff
@@ -53,4 +55,12 @@
             .flatMap(factory -> JdbiOptionals.stream(factory.build(expectedType, value, config)))
             .findFirst();
     }
+
+    private static class LegacyEnumByNameArgumentFactory implements ArgumentFactory {
+        private final EnumArgumentFactory delegate = new EnumArgumentFactory();
+        @Override
+        public Optional<Argument> build(Type expectedType, Object rawValue, ConfigRegistry config) {
+            return delegate.build(QualifiedType.of(expectedType).with(EnumByName.class), rawValue, config);
```
stevenschlansker: :clap:

core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
```diff
@@ -14,21 +14,57 @@
...
-            return Optional.empty();
-        }
+    @SuppressWarnings("unchecked")
+    public Optional<Argument> build(QualifiedType<?> givenType, Object value, ConfigRegistry config) {
+        return ifEnum(givenType.getType())
+            .flatMap(clazz -> makeEnumArgument((QualifiedType<Enum>) givenType, (Enum) value, config));
+    }
+
+    @SuppressWarnings("unchecked")
+    static <E extends Enum<E>> Optional<Class<?>> ifEnum(Type type) {
```
stevenschlansker: can / should this be `Optional<Class<E>>`?

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
```diff
@@ -50,50 +60,61 @@
...
         @Override
-        public E map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
-            String name = r.getString(columnNumber);
-            return name == null ? null : insensitiveLookup.computeIfAbsent(name, this::resolve);
+        public E map(ResultSet rs, int columnNumber, StatementContext ctx) throws SQLException {
+            String name = ctx.findColumnMapperFor(String.class)
+                .orElseThrow(() -> new UnableToProduceResultException("a String column mapper is required to map Enums from names", ctx))
+                .map(rs, columnNumber, ctx);
+
+            return name == null ? null : nameValueCache.computeIfAbsent(name.toLowerCase(), lowercased -> getValueByName(enumClass, name, ctx));
```
stevenschlansker: I wonder if the cost of lowercasing outweighs the cost of duplicate entries.  Previously I'd operated under the assumption that most users, even if the casing is wrong, will tend to be at least consistently wrong - in which case you'd have few if any truly redundant entries.  And even if they aren't, the extra cache entries may well be cheaper than repeatedly lowercasing.  Part of the reason for adding this cache in the first place is to avoid case sensitive operations.
I worry that this may be a performance regression if that is indeed the common case.

core/src/main/java/org/jdbi/v3/core/internal/QualifiedEnumMapperFactory.java
```diff
@@ -0,0 +1,117 @@
...
+            this.enumClass = enumClass;
+        }
+
+        @Override
+        public E map(ResultSet rs, int columnNumber, StatementContext ctx) throws SQLException {
+            String name = ctx.findColumnMapperFor(String.class)
+                .orElseThrow(() -> new UnableToProduceResultException("a String column mapper is required to map Enums from names", ctx))
+                .map(rs, columnNumber, ctx);
+
+            return name == null ? null : nameValueCache.computeIfAbsent(name.toLowerCase(), lowercased -> getValueByName(enumClass, name, ctx));
```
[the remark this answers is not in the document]
qualidafial: Added a benchmark for enum mapping. It seeds 1000 enums into two tables: one with exact names, and one with randomized case names. The same enum values are inserted into both tables so we're comparing apples to apples.

Variables to test:

- with or without cache
- lowercase cache key or not
- case sensitive (Enum.valueOf) vs insensitive

Without cache, Enum.valueOf() only (baseline):

```
Benchmark                         Mode  Cnt       Score      Error  Units
EnumMapperBenchmark.mapExactCase  avgt    5  137059.724 ± 3079.280  ns/op
```

With cache, Enum.valueOf() only:

```
Benchmark                         Mode  Cnt       Score      Error  Units
EnumMapperBenchmark.mapExactCase  avgt    5  140057.354 ± 2108.761  ns/op
```

Enum.valueOf() maintains a constant directory (map) so it makes sense that adding a second cache actually slowed it down.

Without cache, case insensitive:

```
Benchmark                          Mode  Cnt       Score       Error  Units
EnumMapperBenchmark.mapExactCase   avgt    5  447953.782 ± 10698.905  ns/op
EnumMapperBenchmark.mapRandomCase  avgt    5  816709.142 ± 15131.747  ns/op
```

Adding case-insensitivity (using equalsIgnoreCase) slowed things way down. But..

With cache, case insensitive:

```
Benchmark                          Mode  Cnt       Score      Error  Units
EnumMapperBenchmark.mapExactCase   avgt    5  131139.802 ± 3909.448  ns/op
EnumMapperBenchmark.mapRandomCase  avgt    5  160514.845 ± 1140.261  ns/op
```

So this is actually a little faster for case sensitive, but 17% slower for case insensitive. This might be a reasonable trade-off.

With cache, case insensitive, lowercase cache key:

```
Benchmark                          Mode  Cnt       Score      Error  Units
EnumMapperBenchmark.mapExactCase   avgt    5  179181.546 ± 1043.832  ns/op
EnumMapperBenchmark.mapRandomCase  avgt    5  204252.115 ± 2278.226  ns/op
```

Converting the value to lowercase hurt increased case-sensitive time by 37%, and case-insensitive time by 27%.

If we're committing to support case-insensitive lookups, then using a cache (without altering the key) seems like the way to go.

core/src/main/java/org/jdbi/v3/core/internal/QualifiedEnumMapperFactory.java
```diff
@@ -0,0 +1,117 @@
...
+
+        private EnumByNameColumnMapper(Class<E> enumClass) {
+            this.enumClass = enumClass;
+        }
+
+        @Override
+        public E map(ResultSet rs, int columnNumber, StatementContext ctx) throws SQLException {
+            String name = ctx.findColumnMapperFor(String.class)
+                .orElseThrow(() -> new UnableToProduceResultException("a String column mapper is required to map Enums from names", ctx))
+                .map(rs, columnNumber, ctx);
```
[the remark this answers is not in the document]
qualidafial: Same as the other benchmark, except now I'm comparing mapper lookup vs `getString()` / `getInt()`.

Using `getString()` / `getInt()`:

```
Benchmark                                Mode  Cnt      Score     Error  Units
EnumMapperBenchmark.mapByExactName       avgt    5  79775.267 ± 460.790  ns/op
EnumMapperBenchmark.mapByOrdinal         avgt    5  74462.557 ± 546.199  ns/op
EnumMapperBenchmark.mapByRandomCaseName  avgt    5  92999.559 ± 447.022  ns/op
```

Using column mapper lookup from registry:

```
EnumMapperBenchmark.mapByExactName       avgt    5  123372.656 ±  823.765  ns/op
EnumMapperBenchmark.mapByOrdinal         avgt    5  118028.199 ±  450.061  ns/op
EnumMapperBenchmark.mapByRandomCaseName  avgt    5  137851.096 ± 2441.665  ns/op
```

We're paying a hefty performance penalty for the mapper lookup--55%, 59%, and 48%, respectively.

We could EnumByName mapper to check for empty string:

```
Benchmark                                Mode  Cnt      Score      Error  Units
EnumMapperBenchmark.mapByExactName       avgt    5  81420.110 ± 3290.003  ns/op
EnumMapperBenchmark.mapByOrdinal         avgt    5  75045.948 ±  298.406  ns/op
EnumMapperBenchmark.mapByRandomCaseName  avgt    5  91684.408 ±  416.760  ns/op
```

This is roughly on par with straight `getString()`. Full disclosure: I had tweak the benchmark for this last run, so that it would insert nulls randomly into the database.

@TheRealMarnes Would you be OK dropping the mapper lookup, if we check for empty string and treat it as null for enum mapping purposes?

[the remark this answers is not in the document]
TheRealMarnes: I think the factory should throw an exception on empty strings because they can't ever match any enum value we could find

core/src/main/java/org/jdbi/v3/core/internal/QualifiedEnumMapperFactory.java
```diff
@@ -0,0 +1,117 @@
...
+
+            return name == null ? null : nameValueCache.computeIfAbsent(name.toLowerCase(), lowercased -> getValueByName(enumClass, name, ctx));
+        }
+
+        private static <E extends Enum<E>> E getValueByName(Class<E> enumClass, String name, StatementContext ctx) {
+            try {
+                return Enum.valueOf(enumClass, name);
+            } catch (IllegalArgumentException ignored) {
```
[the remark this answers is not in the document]
TheRealMarnes: I'm gonna play the copypasta card here and say this is inherited code 👐

core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
qualidafial: It can. I'll change.

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
TheRealMarnes: nice, hadn't thought of that simple fix for some reason

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
qualidafial: I tried it with and without `toLowerCase()` in benchmarks, and it was way worse. I took it out.

TheRealMarnes:
I've given the git diffs a glance over coffee, I like the changes so far but haven't reviewed in detail.

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
stevenschlansker: Thanks for checking!  String operations are surprisingly expensive.

stevenschlansker:
I itchy-trigger-finger merged this.  Sorry I cut off @TheRealMarnes final review - if there's any followup comments we can open a new pr.

TheRealMarnes:
Github won't let me approve my "own" PR, so have this instead:

PR #1429, opened 2019-01-22
fix javadoc building

stevenschlansker:
javadocs are currently not being built, see https://github.com/jdbi/jdbi.github.io/issues/5
javadocs don't build on jdk11 currently so toolchain it to jdk8 for now, and clean up some transitional profile stuff that isn't needed now that jdk9 is dead.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -571,6 +571,14 @@
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-toolchains-plugin</artifactId>
                     <version>1.1</version>
+                    <configuration>
+                        <toolchains>
+                            <jdk>
+                                <version>1.8</version>
+                                <vendor>oracle</vendor>
+                            </jdk>
+                        </toolchains>
+                    </configuration>
                 </plugin>
                 <plugin>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -702,74 +710,15 @@
...
-                            </configuration>
-                        </plugin>
-                    </plugins>
-                </pluginManagement>
-            </build>
-        </profile>
         <profile>
             <id>jdk9</id>
             <activation>
                 <jdk>[1.9,)</jdk>
```
TheRealMarnes: why leave this in though?

stevenschlansker: These plugins currently don't run on jdk9+.  For example Spotbugs (I think) finds bogus bugs that only exist in java 11 generated code and can't be fixed in source.  :rage:

TheRealMarnes: If you changed the id name to something that says as much, it'd make me happy :p

stevenschlansker: I don't want to just go changing profile names around.  People can refer to them, for example we did so for other profiles with `-Ptoolchains`.  We should focus our efforts on getting the build clean on all supported JDKs.

PR #1431, opened 2019-01-23
benchmark: Enum mapping case sensitivity

stevenschlansker:
I wanted to quantify the penalty for using enums case insensitively, so I wrote a basic benchmark.  I tested it against the code in `master` (roughly 3.6.0) versus the latest version of #1427 

Full results
master: https://gist.github.com/stevenschlansker/afc65d80f895fa1000f6e383d16cc2bc
#1427: https://gist.github.com/stevenschlansker/c9baae77be3775868a88474074e04999

Short version: there's a dramatic performance regression in microbenchmark both in the case-insensitive compare as well as the normal case sensitive compare.

Before:
```
EnumBenchmark.mapEnumCaseInsensitive         thrpt    5   2424.413 ± 108.752  ops/s
EnumBenchmark.mapEnumCaseSensitive thrpt 5 12461.018 ± 302.860 ops/s
```

After:
```
EnumBenchmark.mapEnumCaseInsensitive         thrpt    5   335.997 ±  57.533  ops/s
EnumBenchmark.mapEnumCaseSensitive thrpt 5 6441.229 ± 117.578 ops/s
```

While microbenchmark performance isn't the only criteria we're going by, I think it's worth investigating such a dramatic regression.

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -85,6 +85,19 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration combine.children="append">
+                    <annotationProcessorPaths>
+                        <annotationProcessorPath>
+                            <groupId>org.openjdk.jmh</groupId>
+                            <artifactId>jmh-generator-annprocess</artifactId>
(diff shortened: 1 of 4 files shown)


stevenschlansker:
ok I think I figured it out.  spotbugs was having trouble because of a regex typo.

PR #1432, opened 2019-01-23
Fix enum case insensitive matching to actually work

stevenschlansker:
This simple fix improves enum mapping performance.  Previously, the cache was not effective because the mapper is re-created for every jdbi operation you do.

Before:
https://gist.github.com/stevenschlansker/afc65d80f895fa1000f6e383d16cc2bc
```
EnumBenchmark.mapEnumCaseInsensitive         thrpt    5   2424.413 ± 108.752  ops/s
EnumBenchmark.mapEnumCaseSensitive thrpt 5 12461.018 ± 302.860 ops/s
```

After:
https://gist.github.com/stevenschlansker/b7769735ce3205f017f2ae5cf398f933
```
EnumBenchmark.mapEnumCaseInsensitive         thrpt    5  10364.923 ± 426.301  ops/s
EnumBenchmark.mapEnumCaseSensitive thrpt 5 13975.079 ± 544.881 ops/s
```

So in this microbenchmark, expanding the scope of the cache almost eliminates the performance penalty to using loosely cased enums, and in fact makes the fast case even faster.

Unfortunately this PR will conflict with #1427 but it can probably be folded in fairly easily.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
@@ -17,7 +17,10 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeUnit;
 
+import net.jodah.expiringmap.ExpirationPolicy;
+import net.jodah.expiringmap.ExpiringMap;
 import org.jdbi.v3.core.statement.StatementContext;
 
 /**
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
```diff
@@ -17,23 +17,31 @@
...
 
     /**
      * @param <E> the enum type to map
      * @param type the enum type to map
      * @return an enum mapper that matches on {@link Enum#name()}
      */
+    @SuppressWarnings("unchecked")
     public static <E extends Enum<E>> ColumnMapper<E> byName(Class<E> type) {
-        return new ByName<>(type);
+        return (ColumnMapper<E>) NAME_CACHE.computeIfAbsent(type, e -> new ByName<>(type));
```
TheRealMarnes: This could be `ByName::new` because `e == type` but it's all the same

stevenschlansker: I tried that but it didn't compile.  The `Class<E>` is lost; `NAME_CACHE` only knows that the key is an enum class, not `E` in particular.  `ByName::new` is `new ByName<>(key)` rather than `new ByName<>(type)` and then you get an error that `Class<? extends Enum<?>> isn't a subtype of E` or something like that.

PR #1437, opened 2019-01-25
json: remove use of StatementContext, simplify mapper / argument lambdas

stevenschlansker:
we actually only need the ConfigRegistry instead, and that lets us move
a fair amount of work out of the lambdas in the json handling, which should be a minor performance improvement

### DIFF
--- a/json/src/main/java/org/jdbi/v3/json/internal/JsonArgumentFactory.java
+++ b/json/src/main/java/org/jdbi/v3/json/internal/JsonArgumentFactory.java
@@ -18,6 +18,7 @@ import java.util.Optional;
 
 import org.jdbi.v3.core.argument.Argument;
 import org.jdbi.v3.core.argument.ArgumentFactory;
+import org.jdbi.v3.core.argument.Arguments;
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.internal.JdbiOptionals;
 import org.jdbi.v3.core.qualifier.QualifiedType;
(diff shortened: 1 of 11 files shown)


TheRealMarnes:
I'm not really seeing it, could you explain the issue, and which change lies at its source?

stevenschlansker:
It's just a simplification.  Moved a lot of code out of the lambdas and do the work ahead of time.  This way, we don't e.g. re-lookup the mapper every time we call the lambda.  It also removes our need to get the StatementContext god-object and instead we can just use the ConfigRegistry.

json/src/main/java/org/jdbi/v3/json/internal/JsonArgumentFactory.java
```diff
@@ -40,17 +41,12 @@
         if (String.class.equals(type)) {
             return Optional.empty();
         }
-
-        return Optional.of((pos, stmt, ctx) -> {
```
TheRealMarnes: I believe this change makes the json mapper rely on the String mapper present at the time of registering the json mapper, and ignore any string mapper registered afterward. Not a huge problem I think but maybe users need to be aware registration order will matter between those two?

json/src/main/java/org/jdbi/v3/json/internal/JsonColumnMapperFactory.java
```diff
@@ -40,17 +42,17 @@
...
+        // look for specialized json support first, revert to simple String mapping if absent
+        ColumnMapper<String> jsonStringMapper = JdbiOptionals.findFirstPresent(
+                () -> cm.findFor(QualifiedType.of(String.class).with(Json.class)),
+                () -> cm.findFor(String.class))
+                .orElseThrow(() -> new UnableToProduceResultException(JSON_NOT_RETRIEVABLE));
 
+        final JsonMapper mapper = config.get(JsonConfig.class).getJsonMapper();
         return Optional.of((rs, i, ctx) -> {
-            // look for specialized json support first, revert to simple String mapping if absent
-            ColumnMapper<String> jsonStringMapper = JdbiOptionals.findFirstPresent(
```
TheRealMarnes: idem

json/src/main/java/org/jdbi/v3/json/internal/JsonArgumentFactory.java
stevenschlansker: `ArgumentFactory` is only invoked once we are preparing the statement to execute - you have the (static) argument type at this point, this is far after registration time.

TheRealMarnes: Right, I'm stupid, sorry

qualidafial: I think it's reasonable to expect the `@Json String` argument factory and `JsonMapper` to be configured prior to executing a statement that depends on it.

qualidafial: @stevenschlansker That's not accurate. The `ArgumentFactory.build()` method is invoked by `Arguments.findFor()` which is invoked at the specific chaining method where you bind the value, e.g.

```
handle.createUpdate(sql)
    .bindByType("foo", foo, QualifiedType.of(Foo.class).with(Json.class)) // argument factory invoked here
    .execute();
```

So it's conceivable the if somebody configured the `JsonMapper` or registered the `@Json String` mapper later in the call chain, they wouldn't be used:

```
handle.createUpdate(sql)
    .bindByType("foo", foo, QualifiedType.of(Foo.class).with(Json.class))
    .configure(JsonConfig.class, jc -> jc.setJsonMapper(myMapper)) // too late!
    .execute();
```

I just don't think it's a very likely case, and not worth worrying about unless we hear a bug report on it.

I suspect the vast majority of folks using JSON mapping will just follow the cookbook in the developer guide and configure everything on the `Jdbi` instance.

stevenschlansker: Ah, sorry, I misspoke.  Either way my overall point was that you are already logically building the statement, even if we are not actually building the `PreparedStatement`.

PR #1439, opened 2019-01-25
Keep using String.format instead of MessageFormat

TheRealMarnes:
> I don't know why this was ever added in the first place

— Steven, 1 post below, 10 minutes from now

^^


stevenschlansker:
I don't know why this was ever added in the first place...

PR #1441, opened 2019-01-28
fix #1388 sql locator phoney return value

### LINKED ISSUE #1388: `UseClasspathSqlLocatorImpl` should throw an error on failure
The `UseClasspathSqlLocatorImpl` just returns the name of the expected file when it fails to load. This means the parser is given `methodName.sql` as the "located sql" to parse. If the resource can't be found or loaded, that error should be propagated, not swallowed.

TheRealMarnes:
fixes #1388

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/internal/UseClasspathSqlLocatorImpl.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/internal/UseClasspathSqlLocatorImpl.java
@@ -21,25 +21,20 @@ import org.jdbi.v3.core.locator.ClasspathSqlLocator;
 import org.jdbi.v3.sqlobject.SqlObjects;
 import org.jdbi.v3.sqlobject.config.Configurer;
 import org.jdbi.v3.sqlobject.internal.SqlAnnotations;
-import org.jdbi.v3.sqlobject.locator.SqlLocator;
 
 public class UseClasspathSqlLocatorImpl implements Configurer {
-    private static final SqlLocator SQL_LOCATOR = (sqlObjectType, method, config) -> {
-        return SqlAnnotations.getAnnotationValue(method,
-                name -> ClasspathSqlLocator.findSqlOnClasspath(sqlObjectType, defaultName(name, method))).orElseGet(method::getName);
-    };
-
     @Override
     public void configureForType(ConfigRegistry registry, Annotation annotation, Class<?> sqlObjectType) {
-        registry.get(SqlObjects.class).setSqlLocator(SQL_LOCATOR);
-    }
-
-    private static String defaultName(String name, Method method) {
-        return name.isEmpty() ? method.getName() : name;
+        registry.get(SqlObjects.class).setSqlLocator(UseClasspathSqlLocatorImpl::locate);
     }
 
     @Override
     public void configureForMethod(ConfigRegistry registry, Annotation annotation, Class<?> sqlObjectType, Method method) {
         configureForType(registry, annotation, sqlObjectType);
+    }
+
+    private static String locate(Class<?> sqlObjectType, Method method, @SuppressWarnings("unused") ConfigRegistry config) {
+        return SqlAnnotations.getAnnotationValue(method, key -> ClasspathSqlLocator.findSqlOnClasspath(sqlObjectType, key))
+            .orElseGet(() -> ClasspathSqlLocator.findSqlOnClasspath(sqlObjectType, method.getName()));
     }
 }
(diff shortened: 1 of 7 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/internal/UseClasspathSqlLocatorImpl.java
```diff
@@ -15,31 +15,35 @@
...
-        return name.isEmpty() ? method.getName() : name;
+        registry.get(SqlObjects.class).setSqlLocator(LOCATOR);
     }
 
     @Override
     public void configureForMethod(ConfigRegistry registry, Annotation annotation, Class<?> sqlObjectType, Method method) {
         configureForType(registry, annotation, sqlObjectType);
     }
+
+    private static String locate(Class<?> sqlObjectType, Method method) {
```
stevenschlansker: if you add the `config` parameter here, even though its unused, this method becomes a `SqlLocator` and you can remove the constant and use a method reference instead of a lambda

sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/internal/UseClasspathSqlLocatorImpl.java
```diff
@@ -15,31 +15,35 @@
...
+        Optional<String> resolved = SqlAnnotations.getAnnotationValue(method, value -> {
+            String name = value.isEmpty() ? method.getName() : value;
+            return ClasspathSqlLocator.findSqlOnClasspath(sqlObjectType, name);
+        });
+
+        return resolved.orElseThrow(() -> new UnableToCreateStatementException(String.format("the query returned for %s was empty", method)));
```
stevenschlansker: how about something more like `"unable to locate sql for " method` -- "the query returned" sounds like you've actually executed a query on the database, but we never actually did

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestUseClasspathSqlLocator.java
```diff
@@ -71,4 +73,19 @@ public void testCachedOverride() {
...
+        assertThatThrownBy(handle.attach(Blanks.class)::methodWithEmptyFile)
+            .isInstanceOf(UnableToCreateStatementException.class)
+            .hasMessageContaining("the query returned")
+            .hasMessageContaining("methodWithEmptyFile()")
+            .hasMessageContaining("was empty");
+    }
+
+    @UseClasspathSqlLocator
+    public interface Blanks {
+        @SqlQuery("empty")
```
stevenschlansker: I thought this was testing an error if the file is not located.
This seems to test that the file is located but empty -- it's not as clear to me that this is an error; the user has explicitly configured a valid but empty template.

sqlobject/pom.xml
```diff
@@ -123,6 +123,15 @@
                     </compilerArgs>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>com.mycila</groupId>
+                <artifactId>license-maven-plugin</artifactId>
+                <configuration>
+                    <excludes>
+                        <exclude>src/test/resources/org/jdbi/v3/sqlobject/TestUseClasspathSqlLocator$Blanks/empty.sql</exclude>
```
stevenschlansker: why are we adding this exception?  I'd prefer we just add the license to the files instead.

TheRealMarnes: Adding the license makes the file not empty :)

stevenschlansker: Yeah, I don't think that is right.  a locator / template engine is *not* responsible for ensuring we get valid sql, only that we found all the files / strings / etc we look for.  An empty string is valid in that context, but will cause a problem later when you try to prepare it.

TheRealMarnes:
Well having an empty file is the only way to actually get the behavior he described: the path derived from the method name being returned as "query". A missing file just does what it should do, it fails.

sqlobject/pom.xml
TheRealMarnes: Removing the emptiness check presents a backward compatibility problem though because the code is used in a fair few places

stevenschlansker:
Then I'm definitely confused.  The original report:

> The UseClasspathSqlLocatorImpl just returns the name of the expected file when it fails to load. This means the parser is given methodName.sql as the "located sql" to parse. If the resource can't be found or loaded, that error should be propagated, not swallowed.

so he says if you forget to point at an actual file, it doesn't behave correctly.  You say:

> Well having an empty file is the only way to actually get the behavior he described: the path derived from the method name being returned as "query". A missing file just does what it should do, it fails.

you say that it already correctly fails??

TheRealMarnes:
He might've been on an old version. I think the actual report is moot though, what matters is the current code has a flow that returns a file path as if it were a query and that needs to be changed.

TheRealMarnes:
Withdrawing for rework

PR #1442, opened 2019-01-29
Fix remnant legacy related to #1388

TheRealMarnes:
#1388 is a breadcrumb trail to legacy fallback code that I remove with this. The bad behavior could be triggered by an empty sql file (as opposed to a missing file according to #1388 , probably using an old version of the code).

Part of the fix is a refactor of the lookup logic, which was irregularly speckled with external concerns... I think the result is better structured and easier to understand.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlAnnotations.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlAnnotations.java
@@ -17,7 +17,6 @@ import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Optional;
 import java.util.function.Function;
-import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import org.jdbi.v3.core.internal.JdbiOptionals;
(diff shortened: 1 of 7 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlAnnotations.java
```diff
@@ -35,36 +34,54 @@ private SqlAnnotations() {
...
+     * Returns the <code>value()</code> of the <code>@SqlBatch</code>, <code>@SqlCall</code>, <code>@SqlQuery</code>,
+     * <code>@SqlUpdate</code>, <code>@SqlScripts</code>, or <code>@SqlScript</code> annotation on the given method if declared and non-empty; empty otherwise.
      *
      * @param method the method
      * @return the annotation <code>value()</code>
      */
-    public static Optional<String> getAnnotationValue(Method method, Function<String, String> resolveSql) {
-        Predicate<String> nonEmpty = s -> !s.isEmpty();
+    public static Optional<String> getAnnotationValue(Method method) {
+        return getAnnotationValue(method, value -> value.isEmpty() ? null : value);
```
TheRealMarnes: The `nonEmpty` check moved here to keep filtering concerns out of the value lookup. Callers of the overloaded method sign up for having to do any filtering and replacements themselves.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlAnnotations.java
```diff
@@ -35,36 +34,54 @@ private SqlAnnotations() {
...
+
+            () -> Optional.ofNullable(method.getAnnotation(SqlScripts.class))
+                .map(SqlScripts::value)
+                .map(scripts -> Arrays.stream(scripts)
+                    .map(SqlScript::value)
+                    .map(transformation)
+                    .collect(Collectors.joining(" ; "))),
 
-    private static String scriptValue(SqlScript script, Method method) {
-        return script.value().isEmpty() ? method.getName() : script.value();
```
TheRealMarnes: This is an external concern that belongs in `transformation`

PR #1443, opened 2019-01-30
DefineNamedBindings: use a proxy Statement rather than inspecting for `NullArgument`

### LINKED ISSUE #1416: Improve/fix DefineBinds nullness detection
Current `instanceof NullArgument` is rather fickle. ATM it doesn't work for some arguments, like `OptionalArgumentFactory` and `InputStreamArgument`. @stevenschlansker appears to be taking care of it for `ObjectArgument` in #1409.

Suggestion:
```
@FunctionalInterface
public interface Argument {
    /**
     * @return true if the value of this binding is null, false otherwise
     */
    default boolean isNull() {
        return false;
    }
}
```
Override in `NullArgument` with `return true` and in others with a `== null` on a value or whatever's appropriate. `OptionalArgumentFactory` for example would do `isEmpty()`. This is also more discoverable by argument implementors than a javadoc and manual comment, I think.

stevenschlansker:
Fixes #1416

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/internal/exceptions/Unchecked.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/exceptions/Unchecked.java
@@ -38,6 +38,17 @@ public class Unchecked {
         };
     }
 
+    @SuppressWarnings("PMD.DoNotUseThreads")
+    public static Runnable runnable(CheckedRunnable checkedRunnable) {
+        return () -> {
+            try {
+                checkedRunnable.run();
+            } catch (Throwable t) {
+                throw Sneaky.throwAnyway(t);
+            }
+        };
+    }
+
     public static <T> SneakyCallable<T> callable(CheckedCallable<T> checkedCallable) {
         return () -> {
             try {
@@ -81,5 +92,9 @@ public class Unchecked {
     public interface SneakyCallable<T> extends Callable<T> {
         @Override
         T call(); // no 'throws Exception'
+    }
+
+    public interface CheckedRunnable {
+        void run() throws Exception;
     }
 }
--- a/core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
@@ -13,20 +13,48 @@
  */
 package org.jdbi.v3.core.statement;
 
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.sql.PreparedStatement;
-import java.sql.SQLException;
 import java.util.Set;
 
-import org.jdbi.v3.core.argument.NullArgument;
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.internal.exceptions.Unchecked;
 
 class DefineNamedBindingsStatementCustomizer implements StatementCustomizer {
     @Override
-    public void beforeTemplating(PreparedStatement stmt, StatementContext ctx) throws SQLException {
+    public void beforeTemplating(PreparedStatement stmt, StatementContext ctx) {
         final Set<String> alreadyDefined = ctx.getAttributes().keySet();
         final Binding binding = ctx.getBinding();
+        final SetNullHandler handler = new SetNullHandler();
         binding.getNames().stream()
             .filter(name -> !alreadyDefined.contains(name))
             .forEach(name -> binding.findForName(name, ctx).ifPresent(
-                    a -> ctx.define(name, a instanceof NullArgument ? false : true)));
+                    a -> handler.define(name, a, ctx)));
+    }
+
+    private static class SetNullHandler implements InvocationHandler {
+        private final PreparedStatement fakeStmt = (PreparedStatement)
+                Proxy.newProxyInstance(Thread.currentThread().getContextClassLoader(), new Class<?>[] {PreparedStatement.class}, this);
+        private boolean setNull;
+        private boolean setCalled;
+
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) {
+            setCalled = method.getName().startsWith("set");
+            boolean argNull = args.length > 1 && args[1] == null;
+            setNull = "setNull".equals(method.getName()) || argNull;
+            return null;
+        }
+
+        void define(String name, Argument arg, StatementContext ctx) {
+            setNull = false;
+            setCalled = false;
+            Unchecked.runnable(() -> arg.apply(1, fakeStmt, ctx)).run();
+            if (setCalled) {
+                ctx.define(name, !setNull);
+            }
+        }
     }
 }
--- a/core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
@@ -17,6 +17,7 @@ import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.sql.PreparedStatement;
+import java.sql.SQLException;
 import java.util.Set;
 
 import org.jdbi.v3.core.argument.Argument;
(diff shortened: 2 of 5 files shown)


stevenschlansker:
rebased and squished

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
```diff
@@ -13,20 +13,50 @@
...
+                boolean argNull = args.length > 1 && args[1] == null;
+                setNull = method.getName().equals("setNull") || argNull;
+            }
+            return null;
+        }
+
+        void define(String name, Argument a, StatementContext ctx) {
+            setNull = false;
+            setCalled = false;
+            Unchecked.runnable(() -> a.apply(1, fakeStmt, ctx)).run();
```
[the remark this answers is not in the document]
stevenschlansker: I guess.  I'm not sure that I agree that having two different approaches to passing through the three arguments is clearer, though.  It reads fine either way to me.  But I guess we don't need runnable support at all if we change it.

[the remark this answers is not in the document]
stevenschlansker: And it trips up the compiler too.  Now it's

```java
Unchecked.<PreparedStatement>consumer(stmt -> a.apply(1, stmt, ctx)).accept(fakeStmt);
```

Are you sure that's better?  It's significantly longer and introduces two additional constructs that weren't there in the first place (the `<>` and the threading-through of `fakeStmt -> stmt -> apply`).

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
```diff
@@ -13,20 +13,50 @@
...
         final Binding binding = ctx.getBinding();
+        final Handler handler = new Handler();
         binding.getNames().stream()
             .filter(name -> !alreadyDefined.contains(name))
             .forEach(name -> binding.findForName(name, ctx).ifPresent(
-                    a -> ctx.define(name, a instanceof NullArgument ? false : true)));
+                    a -> handler.define(name, a, ctx)));
+    }
+
+    static class Handler implements InvocationHandler {
```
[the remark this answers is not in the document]
stevenschlansker: Isn't its purpose clear from context?  It's not intended to be used from an external location.  And the term 'handler' is what it is actually doing.

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
```diff
@@ -13,20 +13,50 @@
...
+        private final PreparedStatement fakeStmt = (PreparedStatement)
+                Proxy.newProxyInstance(Thread.currentThread().getContextClassLoader(), new Class<?>[] {PreparedStatement.class}, this);
+        private boolean setNull;
+        private boolean setCalled;
+
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+            if (setCalled = method.getName().startsWith("set")) { // NOPMD
+                boolean argNull = args.length > 1 && args[1] == null;
+                setNull = method.getName().equals("setNull") || argNull;
```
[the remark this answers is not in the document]
stevenschlansker: I guess.  but the method name is never null...

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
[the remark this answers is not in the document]
TheRealMarnes: A rose by any other name is still a rose, but in java we name classes after their purpose anyway :) `SetNullDetectingHandler` sound good to you then?

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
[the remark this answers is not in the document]
TheRealMarnes: I guess thats a java 11 thing cause it didn't happen to me. I agree with you then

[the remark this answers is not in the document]
stevenschlansker: hm, ok, I'll push it and lets see what Travis says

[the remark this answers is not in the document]
stevenschlansker: The spurious failures are super annoying.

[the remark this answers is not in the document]
stevenschlansker: java8 travis fails too:
> [ERROR] /home/travis/build/jdbi/jdbi/core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java:[54,53] incompatible types: java.lang.Object cannot be converted to java.sql.PreparedStatement

core/src/main/java/org/jdbi/v3/core/internal/exceptions/Unchecked.java
```diff
@@ -38,6 +38,17 @@ private Unchecked() {
         };
     }
 
+    @SuppressWarnings("PMD.DoNotUseThreads")
```
TheRealMarnes: lol

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
```diff
@@ -13,20 +13,48 @@
...
+                Proxy.newProxyInstance(Thread.currentThread().getContextClassLoader(), new Class<?>[] {PreparedStatement.class}, this);
+        private boolean setNull;
+        private boolean setCalled;
+
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) {
+            setCalled = method.getName().startsWith("set");
+            boolean argNull = args.length > 1 && args[1] == null;
+            setNull = "setNull".equals(method.getName()) || argNull;
+            return null;
```
qualidafial: The only thing I can think that _might_ be a problem is that any argument implementations that depend on e.g. `statement.isWrapperFor()` or `statement.unwrap()` to unwrap a vendor-specific prepared statement might error with `@DefineNamedBindings` where otherwise it might behave fine.

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
```diff
@@ -13,20 +13,48 @@
...
+
+    private static class SetNullHandler implements InvocationHandler {
+        private final PreparedStatement fakeStmt = (PreparedStatement)
+                Proxy.newProxyInstance(Thread.currentThread().getContextClassLoader(), new Class<?>[] {PreparedStatement.class}, this);
+        private boolean setNull;
+        private boolean setCalled;
+
+        @Override
+        public Object invoke(Object proxy, Method method, Object[] args) {
+            setCalled = method.getName().startsWith("set");
```
qualidafial: Might not be a single method call. Should probably change this to `||=`

qualidafial: Something like this, rather:

```java
if (method.getName().startsWith("set")) {
  setCalled = true;
  ...etc
}
```

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
qualidafial: Should return a zero / false value for primitive return types, otherwise ClassCastException

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
stevenschlansker: I'd personally go for `|=` to save the `if` block but the logic change makes sense.

qualidafial: I think it needs to be in an if-block, otherwise the test for null on the following lines isn't valid.

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
qualidafial: The `statement.unwrap()` idea isn't going to work, since all the types you would want to unwrap are classes, not interfaces.

Best we can do is provide an informative exception message stating that because `@DefineNamedBindings` uses proxies to determine which fields are non-null, it is incompatible with arguments that rely on `Wrapper.unwrap()`

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
qualidafial: Pushed fix

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
qualidafial: Pushed fix

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
```diff
@@ -13,20 +13,89 @@
...
+                setCalled = true;
+                boolean argNull = args.length > 1 && args[1] == null;
+                setNull = argNull || "setNull".equals(method.getName());
+            }
+
+            return defaultValue(method.getReturnType());
+        }
+
+        private Object defaultValue(Class<?> type) {
+            if (type.isPrimitive()) {
```
stevenschlansker: It'd be possible to shorten this significantly: `Array.get(Array.newInstance(type, 1), 0)` but maybe that's too obscure

TheRealMarnes: This smells of a Map<Class, Object> to me

PR #1444, opened 2019-01-30
JdbiPlugin throws SqlException

### LINKED ISSUE #1440: Allow JdbiPlugin::customizeConnection to throw SQLException
For implementors of `JdbiPlugin::customizeConnection` that call methods on the given `Connection` it would be convenient to just pass through `SQLException`s.

For example, instead of 
```java
    jdbi.installPlugin(new JdbiPlugin() {
      @Override
      public Connection customizeConnection(Connection conn) {
        try {
          conn.setHoldability(HOLD_CURSORS_OVER_COMMIT);
        } catch (SQLException exception) {
          throw new RuntimeException(exception);
        }
        return conn;
      }
    });
```
clients could write 
```java
    jdbi.installPlugin(new JdbiPlugin() {
      @Override
      public Connection customizeConnection(Connection conn) {
        conn.setHoldability(HOLD_CURSORS_OVER_COMMIT);
        return conn;
      }
    });
```
And spare the boilerplate catch-and-rethrow code. Furthermore, exception handling would be consistent for all `SQLException`s.

stevenschlansker:
Fixes #1440 
This is a small source breaking change.  It should not be a binary incompatible change -- checked exceptions are compiler fiction -- but if anyone calls `JdbiPlugin` methods themselves (why would they??) they may now have to handle it.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Jdbi.java
+++ b/core/src/main/java/org/jdbi/v3/core/Jdbi.java
@@ -31,6 +31,7 @@ import org.jdbi.v3.core.extension.ExtensionConsumer;
 import org.jdbi.v3.core.extension.ExtensionFactory;
 import org.jdbi.v3.core.extension.Extensions;
 import org.jdbi.v3.core.extension.NoSuchExtensionException;
+import org.jdbi.v3.core.internal.exceptions.Unchecked;
 import org.jdbi.v3.core.spi.JdbiPlugin;
 import org.jdbi.v3.core.statement.DefaultStatementBuilder;
 import org.jdbi.v3.core.statement.StatementBuilder;
@@ -224,7 +225,7 @@ public class Jdbi implements Configurable<Jdbi> {
      * @return this
      */
     public Jdbi installPlugin(JdbiPlugin plugin) {
-        plugin.customizeJdbi(this);
+        Unchecked.consumer(plugin::customizeJdbi).accept(this);
         plugins.add(plugin);
         return this;
     }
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -224,7 +225,7 @@ public Jdbi installPlugins() {
      * @return this
      */
     public Jdbi installPlugin(JdbiPlugin plugin) {
-        plugin.customizeJdbi(this);
+        Unchecked.runnable(() -> plugin.customizeJdbi(this)).run();
```
TheRealMarnes: `Unchecked.consumer(plugin::customizeJdbi).accept(this);` is the right mantra :)

core/src/main/java/org/jdbi/v3/core/internal/exceptions/CheckedRunnable.java
```diff
@@ -0,0 +1,19 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.internal.exceptions;
+
+@FunctionalInterface
+public interface CheckedRunnable {
+    void run() throws Exception;
+}
```
TheRealMarnes: missing newline, checkstyle failure

TheRealMarnes:
@stevenschlansker I'll leave merging to you or @qualidafial since I added a commit.

PR #1445, opened 2019-02-02
Extend qualtyp api for collection<class<>>

TheRealMarnes:
Just an idea I had to "complete" the qualtyp api. We support `with`ing with `Annotation...`, `Class<Annotation>...`, and `Collection<Annotation>`, so we should logically also support `Collection<Class<Annotation>>` for those times when you have a `Set` of qualifier classes and the like. Users shouldn't have to reduce highlevel collections to arrays or know about `AnnotationFactory`.

The problem is the type erasure across method overloads, so if we really wanna be able to support more than 1 thing we need to name methods differently based on their arguments.

I'm open to name changes and other thoughts.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/argument/ObjectFieldArguments.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/ObjectFieldArguments.java
@@ -66,7 +66,7 @@ public class ObjectFieldArguments extends ObjectPropertyNamedArgumentFinder {
 
         try {
             QualifiedType<?> type = QualifiedType.of(field.getGenericType())
-                                    .with(getQualifiers(field));
+                                    .withAnnotations(getQualifiers(field));
             Object value = field.get(obj);
 
             return Optional.of(new TypedValue(type, value));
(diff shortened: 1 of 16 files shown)


stevenschlansker:
I don't love `withClasses` but I'm not coming up with anything better immediately.

TheRealMarnes:
I don't love it either but it's all I could come up with without turning the method name into a novel or adding nonsense terminology :/

TheRealMarnes:
I suppose `withAnnotations` is an option, but that implies the other methods aren't about annotations...

We could remove the need for this by turning it into a chaining method that returns the same qualtyp with added qualifiers (instead of a new instance with replaced ones) and takes a single object per call instead of having to work around the erasure on collections...

qualidafial:
I'd prefer to leave the vararg `with()` methods as is, and rename only the collection methods. They will show up the most in inline code for end users, so keeping it terse will improve readability.

```java
QualifiedType.of(type).withAnnotations(Iterable<Annotation>)
QualifiedType.of(type).withAnnotationClasses(Iterable<Class<? extends Annotation>>)
```

PR #1448, opened 2019-02-05
Post-merge refactor #1443

TheRealMarnes:
image

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
@@ -18,6 +18,8 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
+import java.util.IdentityHashMap;
+import java.util.Map;
 import java.util.Set;
 
 import org.jdbi.v3.core.argument.Argument;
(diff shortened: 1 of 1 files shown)


TheRealMarnes:
Yeah, but that's a little... fucky. Making an array with length 1 because you know it'll fill out the default value at index 0 doesn't really rank high on the obviousness/expressiveness chart, I think.

But I would `collect` that stream to a Map, which would then remove the `static {}` block 😛

stevenschlansker:
It's not obvious, I agree.  But it is specified to work.  The JLS says that all arrays are initialized to the default value, and this creates an array of length 1 (specified to have the default value) and then uses `Array.get` which is specified to wrap primitives appropriately into an `Object`.  I personally think a little bit of non-obvious code that uses standard Java APIs in the way they were intended is better than manually writing out all the constants.

I like using `collect` better too.

core/src/main/java/org/jdbi/v3/core/statement/DefineNamedBindingsStatementCustomizer.java
```diff
@@ -36,11 +42,26 @@ public void beforeTemplating(PreparedStatement stmt, StatementContext ctx) {
     }
 
     private static class SetNullHandler implements InvocationHandler {
+        private static final Map<Class<?>, Object> DEFAULT_VALUES = Stream.of(
+            boolean.class,
+            char.class,
+            byte.class,
+            short.class,
+            int.class,
+            long.class,
```
stevenschlansker: if only they provided this list as `Class.PRIMITIVES` or something like that

TheRealMarnes:
I'd prefer to make it an `IdentityHashMap` to save on lookup time though :/ But toMap demands a merge function if you do that, and `throwingMerger` was made private, so we'd have to add a lot of code for nothing.

Does not specifying default values outweigh having a slower Map?

stevenschlansker:
It's hard to imagine that the difference between `HashMap` on `Class` objects (which use the identity hash code anyway, as far as I can tell) and `IdentityHashMap` will be that large compared to a roundtrip to the database server.

TheRealMarnes:
> a roundtrip to the database server.

?

`(short) 0` vs `Array.get(...)` is the question :p

TheRealMarnes:
I understand where you're going with your answer, but for the record, `a roundtrip to the database server` is nowhere in the question, lol

PR #1449, opened 2019-02-05
getObject/setObject factories for LocalDate et al

### LINKED ISSUE #1446: LocalDate binding doesn't work when MySQL server has a different timezone to the client
I have a MySql server that is configured with a timezone of UTC. My Java app has a timezone of "Australia/Perth". When I save down `LocalDate` fields, they are stored in the DB as the previous day. Eg if I bind `2018-01-10` when I query the DB I see `2018-01-09`, *and* when I retrieve it using JDBI I get `2018-01-09`.

Here is a simplified example:
```
public interface ReportDao {

    @SqlUpdate("insert into report (report_date, title) values (:date, :title")
    void insert(@Bind("date") LocalDate date, @Bind("title") String title);

    @SqlQuery("select report_date from report")
    List<LocalDate> getAllDates();
}
```
And this is used as follows:
```
    // ...
    final ReportDao verdictDao = jdbi.onDemand(ReportDao.class);
    LocalDate d = LocalDate.of(2018, Month.JANUARY, 10);
    reportDao.insert(d, "bar");
    reportDao.getAllDates().forEach(System.out::println);
```
`2018-01-09` is printed, when I expect it to print out `2018-01-10`.

Client set-up:
* Timezone Australia/Perth
* jdbi v3.6.0
* mysql-connector-java v8.0.11  (using driver: com.mysql.cj.jdbc.Driver)

Server set-up:
* Timezone UTC
* MySQL 5.7.25-0ubuntu0.18.04.2
* Column type either DATE, or DATETIME

The problem occurs because the MySQL driver will convert the `java.sql.Date` from the client timezone to the server timezone. This seems like the correct thing to do, generally. Here is where JDBI interacts with the MySQL driver, in `JavaTimeArgumentFactory` line 29:
```
register(LocalDate.class, Types.DATE, (p, i, v) -> p.setDate(i, java.sql.Date.valueOf(v)));
```

And in the MySQL driver, in `ClientPreparedQueryBindings` (called from `ClientPreparedStatement`), we see this:
```
    // JDBI calls the 2-arg setDate()
    @Override
    public void setDate(int parameterIndex, Date x) {
        setDate(parameterIndex, x, this.session.getServerSession().getDefaultTimeZone());
    }

    @Override
    public void setDate(int parameterIndex, Date x, TimeZone tz) {
        if (x == null) {
            setNull(parameterIndex);
        } else {
            if (this.ddf == null) {
                this.ddf = new SimpleDateFormat("''yyyy-MM-dd''", Locale.US);
            }
            this.ddf.setTimeZone(tz);
            setValue(parameterIndex, this.ddf.format(x)); // TODO set MysqlType?
        }
    }
```

I've tested the above scenario with PostgreSQL and it doesn't suffer from the same issue.

I can think of 2 possible fixes for this in the JDBI source:

**1. Pin the time-zone by supplying a calendar to the setDate() call, in `JavaTimeArgumentFactory`**
```
// Calendar.getInstance() pins the client-side timezone thereby avoiding conversion
register(LocalDate.class, Types.DATE, (p, i, v) -> p.setDate(i, java.sql.Date.valueOf(v), Calendar.getInstance()));
```

**2. Supply an ISO date string instead of the date,  in `JavaTimeArgumentFactory`**
```
// Send the String to the DB, and allow the DB to coerce it into a date
register(LocalDate.class, Types.DATE, (p, i, v) -> p.setDate(i, v.toString()));
```

I've tested **1.** on both MySQL and Postgresql and it works. It is a bit horrible to create a calendar though. 

I haven't tested **2.** but it should work. Especially, given that both MySQL and Postgresql drivers format the date as a string before sending to the servers. **2.** is flaky though, as I'm pretty sure Oracle allows a date format to be set per session. So there's a good chance that this will not work in systems not using ISO formats.

Please let me know if you'd like more info!

TheRealMarnes:
POC to discuss around, related to the discussion about LocalDate timezones and generally more vendor-specific support in Jdbi.

fixes #1446

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/argument/SetObjectArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/SetObjectArgumentFactory.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.argument;
+
+import java.lang.reflect.Type;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.meta.Beta;
+
+/**
+ * Factory that uses {@link java.sql.PreparedStatement#setObject(int, Object, int)} to bind values.
+ */
+@Beta
+public class SetObjectArgumentFactory implements ArgumentFactory {
+    private final Map<Class<?>, Integer> supportedTypes;
+
+    private SetObjectArgumentFactory(Map<Class<?>, Integer> types) {
+        supportedTypes = types;
+    }
+
+    /**
+     * @param types the Java {@link Class}es to handle with the {@link java.sql.Types} they bind to
+     * @return an {@link ArgumentFactory} that handles only the given {@link Class}es.
+     */
+    public static ArgumentFactory forClasses(Map<Class<?>, Integer> types) {
+        return new SetObjectArgumentFactory(new HashMap<>(types));
+    }
+
+    @Override
+    public Optional<Argument> build(Type type, Object value, ConfigRegistry config) {
+        return Optional.of(type)
+            .filter(Class.class::isInstance)
+            .map(Class.class::cast)
+            .map(supportedTypes::get)
+            .map(sqlType -> ObjectArgument.of(value, sqlType));
+    }
+}
--- a/core/src/main/java/org/jdbi/v3/core/argument/SetObjectArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/SetObjectArgumentFactory.java
@@ -28,7 +28,7 @@ import org.jdbi.v3.meta.Beta;
 public class SetObjectArgumentFactory implements ArgumentFactory {
     private final Map<Class<?>, Integer> supportedTypes;
 
-    private SetObjectArgumentFactory(Map<Class<?>, Integer> types) {
+    protected SetObjectArgumentFactory(Map<Class<?>, Integer> types) {
         supportedTypes = types;
     }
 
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/jdbi/v3/core/argument/SetObjectArgumentFactory.java
```diff
@@ -0,0 +1,42 @@
...
+        supportedTypes = types;
+    }
+
+    public SetObjectArgumentFactory(Class<?>... types) {
+        supportedTypes = new HashSet<>(Arrays.asList(types));
+    }
+
+    @Override
+    public Optional<Argument> build(Type type, Object value, ConfigRegistry config) {
+        return type instanceof Class && supportedTypes.contains(type)
```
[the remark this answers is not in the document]
TheRealMarnes: yeah but it's a justified warning to `contains(x)` on a collection of a different type

core/src/main/java/org/jdbi/v3/core/argument/SetObjectArgumentFactory.java
```diff
@@ -0,0 +1,42 @@
...
+            ? Optional.of(new LoggableBinderArgument<>(value, (p, i, v) -> p.setObject(i, value)))
```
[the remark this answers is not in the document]
stevenschlansker: also, `null` values won't have any sql type set, which might be a deficiency.

TheRealMarnes:
Do you mind if testing of these factories comes with my next pr, which is a similarly small one to add haqldb integration? That was the usecase that led to me writing these in the first place.

stevenschlansker:
I couldn't convert as many factories as I wanted because they relied on finding subclasses as well as the `Class` itself.

TheRealMarnes:
Looks good to me

core/src/main/java/org/jdbi/v3/core/argument/SetObjectArgumentFactory.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.meta.Beta;
+
+/**
+ * Factory that uses {@link java.sql.PreparedStatement#setObject(int, Object, int)} to bind values.
+ */
+@Beta
+public class SetObjectArgumentFactory implements ArgumentFactory {
+    private final Map<Class<?>, Integer> supportedTypes;
```
qualidafial: I feel like this would be better as a registry in a config class

stevenschlansker: `Arguments` could be a reasonable place to put it.

TheRealMarnes: but it affects mapping too?

PR #1450, opened 2019-02-05
RowView: add support for columns with QualifiedType

stevenschlansker:


core/src/main/java/org/jdbi/v3/core/result/RowView.java
```diff
@@ -36,7 +39,7 @@
     private final ResultSet rs;
 
     private final Map<Type, RowMapper<?>> rowMappers = new ConcurrentHashMap<>();
-    private final Map<Type, ColumnMapper<?>> columnMappers = new ConcurrentHashMap<>();
+    private final Map<Object, ColumnMapper<?>> columnMappers = new ConcurrentHashMap<>();
```
TheRealMarnes: Instead of this ugly change, I think this map can simply be a `Map<QualifiedType, ...>`, and the `Type` and `GenericType` from the other methods can be wrapped in qualtyps. It's beating around the bush, but it beats widening this type to Object IMO.

stevenschlansker: Hm.  That requires wrapping every type just to do the lookup, though -- and it means we would no longer distinguish between `QualifiedType.of(MyType.class)` and `MyType.class`.  I guess that's okay?

TheRealMarnes: It's supposed to be okay. All factory matching passes through QualifiedType too.

PR #1452, opened 2019-02-06
Split SqlObject classpath locator to it's own class

FredDeschenes:
This allows setting it project-wide via the SqlObjects config (like for AnnotationSqlLocator) without having to copy the code in the project (as the locate function was private).

Not sure if some things should be renamed as this can get confusing with the other ClasspathSqlLocator in jdbi3-core.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/ClasspathSqlLocator.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/ClasspathSqlLocator.java
@@ -1,0 +1,23 @@
+package org.jdbi.v3.sqlobject.locator;
+
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.sqlobject.internal.SqlAnnotations;
+
+import java.lang.reflect.Method;
+import java.util.function.Function;
+
+/**
+ * @see org.jdbi.v3.core.locator.ClasspathSqlLocator
+ */
+public class ClasspathSqlLocator implements SqlLocator {
+    @Override
+    public String locate(Class<?> sqlObjectType, Method method, ConfigRegistry config) {
+        Function<String, String> valueOrMethodNameToSql = key -> {
+            String filename = key.isEmpty() ? method.getName() : key;
+            return org.jdbi.v3.core.locator.ClasspathSqlLocator.findSqlOnClasspath(sqlObjectType, filename);
+        };
+
+        return SqlAnnotations.getAnnotationValue(method, valueOrMethodNameToSql)
+            .orElseThrow(() -> new IllegalStateException(String.format("method %s has no query annotations", method)));
+    }
+}
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/ClasspathSqlLocator.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/ClasspathSqlLocator.java
@@ -1,10 +1,23 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/ClasspathSqlLocator.java
```diff
@@ -0,0 +1,23 @@
+package org.jdbi.v3.sqlobject.locator;
+
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.sqlobject.internal.SqlAnnotations;
+
+import java.lang.reflect.Method;
+import java.util.function.Function;
+
+/**
+ * @see org.jdbi.v3.core.locator.ClasspathSqlLocator
```
FredDeschenes: Wasn't sure what more could be said here, I can just copy the doc over if you prefer.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/ClasspathSqlLocator.java
```diff
@@ -0,0 +1,23 @@
...
+ */
+public class ClasspathSqlLocator implements SqlLocator {
+    @Override
+    public String locate(Class<?> sqlObjectType, Method method, ConfigRegistry config) {
+        Function<String, String> valueOrMethodNameToSql = key -> {
+            String filename = key.isEmpty() ? method.getName() : key;
+            return org.jdbi.v3.core.locator.ClasspathSqlLocator.findSqlOnClasspath(sqlObjectType, filename);
```
FredDeschenes: I could see this getting confusing.

stevenschlansker: How about calling it `AnnotatedClasspathSqlLocator` or something so as to disambiguate
Also it can probably be non-public

FredDeschenes: Do you mean renaming the one in core or the one in sqlobject?

stevenschlansker: The new one added right here.  Or maybe `SqlObjectClasspathSqlLocator`.  Having two classes with exactly the same name is unfortunate.

FredDeschenes: Yeah I agree, but at the same time it kinda makes sense to have the same name. I'll switch it to `SqlObjectClasspathSqlLocator`.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/SqlObjectClasspathSqlLocator.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @see org.jdbi.v3.core.locator.ClasspathSqlLocator
+ */
+public class SqlObjectClasspathSqlLocator implements SqlLocator {
+    @Override
+    public String locate(Class<?> sqlObjectType, Method method, ConfigRegistry config) {
+        Function<String, String> valueOrMethodNameToSql = key -> {
+            String filename = key.isEmpty() ? method.getName() : key;
+            return org.jdbi.v3.core.locator.ClasspathSqlLocator.findSqlOnClasspath(sqlObjectType, filename);
```
TheRealMarnes: Can't this be an import and short reference now?

FredDeschenes: Yup good catch, I'll change it back.

PR #1455, opened 2019-02-07
DefineNamedBindings: check getConnction works (e.g. sql array uses it)

stevenschlansker:


TheRealMarnes:
Is it worth it to add a test that actually performs the usecase that made you do this, instead of only a standalone assertion?

stevenschlansker:
It's a pain because it actually works just fine in `h2` since it binds it as `Object[]`.  I could spin up a Postgres test, but it'd end up spinning up a database just to ensure a `SqlArray` gets called... this test seemed like a minimal compromise.

TheRealMarnes:
Well we have other databases than heavy postgres too. Sqlite for one, or we could merge my hsqldb integration soon.

stevenschlansker:
Sqlite doesn't even have types... does it have any array support at all?
Postgres is the only database I have experience using arrays with.

stevenschlansker:
sure, we can definitely improve it later, if hsqldb makes it easy to write a better test.

PR #1456, opened 2019-02-07
Pass generic types to Jackson

kvainer:
This modifies JDBI's `JacksonJsonMapper` so that generic types are passed to `ObjectMapper`.

Allows serializing Jackson polymorphic types when the base type is specified as a generic argument. E.g., `Foo<Bar>` where `Bar` has `JsonTypeInfo` annotation. Otherwise Jackson will see this type as simply `Foo` and will not be aware of the `JsonTypeInfo` on the contained type.

### DIFF
--- a/jackson2/src/main/java/org/jdbi/v3/jackson2/JacksonJsonMapper.java
+++ b/jackson2/src/main/java/org/jdbi/v3/jackson2/JacksonJsonMapper.java
@@ -17,7 +17,6 @@ import java.io.IOException;
 import java.lang.reflect.Type;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.result.UnableToProduceResultException;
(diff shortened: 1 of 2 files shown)


jackson2/src/main/java/org/jdbi/v3/jackson2/JacksonJsonMapper.java
```diff
@@ -27,7 +26,10 @@
     @Override
     public String toJson(Type type, Object value, ConfigRegistry config) {
         try {
-            return getMapper(config).writeValueAsString(value);
+            ObjectMapper mapper = getMapper(config);
+            return mapper
+                    .writerFor(mapper.constructType(type))
```
stevenschlansker: Nice catch, I missed this method!

TheRealMarnes:
If you could just drag
`return config.get(Gson2Config.class).getGson().toJson(value);`
-->
`return config.get(Gson2Config.class).getGson().toJson(value, type);`
in `GsonJsonMapper` into this PR while you're at it, that'd be awesome btw :)

kvainer:
> Cool, thanks for the swift contribution :) Would like to see a test to prove the point though.

Added a test, not sure if its in the right place. It covers only the specific scenario I had trouble with so there might be a simpler way of testing it. Open to suggestions.

TheRealMarnes:
I'll give the test and symmetry with gson a look one of the next days. Could you open up your branch for write access by us so we can push our own commits, please?

kvainer:
> I'll give the test and symmetry with gson a look one of the next days. Could you open up your branch for write access by us so we can push our own commits, please?

If you mean the "Allow edits from maintainers" option, then its already enabled. Do I need to do anything else?

TheRealMarnes:
Oh ok, that will probably do it then

stevenschlansker:
Thanks for the improvement @kvainer !  I fixed up the dependency declarations that needed to change due to the added test.

PR #1457, opened 2019-02-08
pass the type arg to gson for serialization

TheRealMarnes:
If you remove the type arg, you'll see that mapping happens according to the `getClass()` type of the value, regardless of the type specified in our API call. Since gson offers passing the type as an arg, we should do so, as users will expect it.

### DIFF
--- a/gson2/src/test/java/org/jdbi/v3/gson2/TestGson2Plugin.java
+++ b/gson2/src/test/java/org/jdbi/v3/gson2/TestGson2Plugin.java
@@ -13,11 +13,23 @@
  */
 package org.jdbi.v3.gson2;
 
+import java.io.IOException;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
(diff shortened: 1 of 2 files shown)


gson2/src/test/java/org/jdbi/v3/gson2/TestGson2Plugin.java
```diff
@@ -27,4 +39,66 @@
...
+                // declare that the subuser should be mapped as a superuser
+                .bindByType("user", new SubUser(), QualifiedType.of(SuperUser.class).with(Json.class))
+                .execute();
+
+            User subuser = h.createQuery("select usr from users")
+                .mapTo(QualifiedType.of(User.class).with(Json.class))
+                .findOnly();
+
+            assertThat(subuser.name)
+                .describedAs("instead of being mapped via getClass(), the object was mapped according to the type param")
```
TheRealMarnes: *bound

PR #1458, opened 2019-02-08
Isolate H2DatabaseRule legacy table creation

TheRealMarnes:
Of 150–160 tests using `H2DatabaseRule`, only 93 actually use the legacy `something` table it automatically creates.

This PR isolates the creation of the table to a `withSomething()` config method, makes it opt-in so the current and any new tests built on it will explicitly depend on it, and existing tests that don't depend on it no longer cause it to be created.

A `DatabaseRule`'s job is to set up a connection, not to spawn cookie cutter test materials. Half the time it goes unused since a good test is encapsulated and manages its own setup, and it can influence test writers to write their tests to suit the existing table rather than letting the test data takes its own natural form (e.g. different datatypes or more expressive names). For this reason I've marked the method `@Deprecated`, but it's a soft preference so if someone _really_ objects I'm ok with leaving it open to guiltless use.

Scratch 1 `TODO` off the list 😃

### DIFF
--- a/core/src/test/java/org/jdbi/v3/core/TestClosingHandle.java
+++ b/core/src/test/java/org/jdbi/v3/core/TestClosingHandle.java
@@ -30,7 +30,7 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 public class TestClosingHandle {
     @Rule
-    public H2DatabaseRule dbRule = new H2DatabaseRule();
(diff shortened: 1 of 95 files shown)


TheRealMarnes:
I can agree it's a bit spammy... If I just move the annotation to the field, it'll be a single warning so no spam, but still clearly marked as what it is to remind everyone to fix it some day.

PR #1460, opened 2019-02-11
fix #1459 and clean up test

### LINKED ISSUE #1459: vavr argument factory should check type rather than `instanceof`
I think this is wrong:
https://github.com/jdbi/jdbi/blob/master/vavr/src/main/java/org/jdbi/v3/vavr/VavrValueArgumentFactory.java#L41

This ends up giving the wrong type for `null` values.
Ref: https://groups.google.com/forum/#!topic/jdbi/piSOY5yB320

TheRealMarnes:


vavr/src/test/java/org/jdbi/v3/vavr/TestVavrValueArgumentFactory.java
```diff
@@ -17,64 +17,77 @@
...
+    private static final Type EITHER_WILDCARD = new GenericType<Either<?, ?>>() {}.getType();
+    private static final Type VALIDATION_STRING_INT = new GenericType<Validation<String, Integer>>() {}.getType();
 
-    private static final Argument MOCK_ARGUMENT = ((position, statement, ctx) -> new NotImplementedError());
+    @Rule
+    public MockitoRule mockito = MockitoJUnit.rule();
 
-    private VavrValueArgumentFactory unit;
+    @Mock
+    private ConfigRegistry configRegistry;
```
stevenschlansker: Is there a particular reason this is written as a mock rather than ensuring we can e.g. actually insert null values into a h2 test table?

TheRealMarnes: The test was built on a fake Argument. I just replaced it with a proper way to fake the ConfigRegistry, Felt changing the "bigger picture" of the test was out of scope.

TheRealMarnes: And tbh this still rings in my mind

> This PR would have been a slam dunk if it just had the relevant fix at hand...

stevenschlansker: I don't think that's an entirely fair comparison given that this test is introducing `mockito` and two new fixtures.  And part of the reason I was hoping for an actual test is that mocks often make for weak testing -- in this case, notably, we're just checking `assertThat(arg).isNotEmpty();` rather than making sure that an actual typed null gets the right sql type when it goes to the db vendor.  So the mocks are giving us a bit of a false sense of security here.

TheRealMarnes: So you wanna turn this simple bugfix (and collateral encapsulation to fix a gross leak) PR into one that also revises the relevant existing tests? Just to confirm

stevenschlansker: In fact, I think the mocking is trivially unnecessary.  I'll push a commit showing what I mean momentarily.

stevenschlansker: Cmon man, you're being really unfair.  Just because I want to remove mocking doesn't mean it's going to become gross.  Look at the change I made, don't you agree it's an improvement?

TheRealMarnes: ? The "gross leak" referred to the non-`private` private method overridden for unit testing, which is the only collateral change I thought would be ok to include in a purely bugfix pr.

stevenschlansker: whoops, I misread, sorry.  But hopefully you agree that removing the mocking makes the code nicer.

TheRealMarnes:
I thought you wanted to add an h2 instance and do roundtrip assertions, not just `new ConfigRegistry()`...

TheRealMarnes:
Since I can't approve my own PR to approve your changes:

image

stevenschlansker:
That'd be nice but adding the check for the null argument type is sufficient for me for now.

TheRealMarnes:
Do we need release notes?

Alarinth:
Hey, where can I find this as a patch? :) Have only really added libs as artifacts through the pom.xml until now. Sorry, I'm terribly new at this and probably am asking stupid questions...

TheRealMarnes:
You can either wait for the release, or if you want the fix now, just take this class and register it on a `Jdbi` instance:
```
class FixedVavrValueArgumentFactory implements ArgumentFactory {
    private static final Set<Class<?>> VALUE_CLASSES = new HashSet<>(Arrays.asList(Option.class, Lazy.class, Try.class, Either.class, Validation.class));

    @Override
    public Optional<Argument> build(Type type, Object value, ConfigRegistry config) {
        Class<?> rawType = GenericTypes.getErasedType(type);

        if (VALUE_CLASSES.stream().anyMatch(vc -> vc.isAssignableFrom(rawType))) {
            return buildValueArgument(type, config, (Value) value);
        }

        return Optional.empty();
    }

    private static Optional<Argument> buildValueArgument(Type type, ConfigRegistry config, Value<?> value) {
        Type nestedType = findGenericParameter(type, Value.class).orElseGet(() -> extractTypeOfValue(value));
        Object nestedValue = value == null ? null : value.getOrNull();
        return config.get(Arguments.class).findFor(nestedType, nestedValue);
    }

    private static Type extractTypeOfValue(Value<?> value) {
        Value<Class<?>> classOfValue = value.map(Object::getClass);
        return classOfValue.getOrElse(Object.class);
    }
}
```

Alarinth:
Thanks for the quick response. Not sure I fully understood how, does this look right?
this.jdbi.registerArgument(new FixedVavrValueArgumentFactory());

Then that class is just a copypaste of what you pasted above - could you perhaps edit in the imports in that code as well so I'm sure I'm not using any wrong ones? I'm still getting the java.sql.SQLException: Invalid column type: 1111 exception as it is right now, so I believe I may have not done this correctly.

TheRealMarnes:
It might just be `registerArgument`, sorry.

```
import java.util.Arrays;
import java.util.HashSet;
import java.util.Optional;
import java.util.Set;

import io.vavr.Lazy;
import io.vavr.Value;
import io.vavr.control.Either;
import io.vavr.control.Option;
import io.vavr.control.Try;
import io.vavr.control.Validation;
import org.jdbi.v3.core.argument.Argument;
import org.jdbi.v3.core.argument.ArgumentFactory;
import org.jdbi.v3.core.argument.Arguments;
import org.jdbi.v3.core.generic.GenericTypes;

import static org.jdbi.v3.core.generic.GenericTypes.findGenericParameter;
```

Alarinth:
I still get the same errors. Does it work for you?
`Option<String> value = Option.none()` <-- for examples sake only, my actual usecase is parsing a csvlist and in this instance I know some values will be read as returned as Option.none().
`jdbi.useHandle(handle -> handle.execute("insert into schema.table (column) values (value)");`

That is essentially what I am doing, albeit at a much bigger scale.

TheRealMarnes:
1111 is `OTHER` and the default untyped null argument... How do you `bind` your values? Try using `bindByType` with a `new GenericType<Option<Foo>>() {}` to make sure the right nested type is used for empty values.

Alarinth:
I model the info I need into one Person.class, which has `Option<T>` for the fields which I know in advance might have null in the db (let's say the fields are Option<String> age, Option<String> name).

Then I simply do:
`jdbi.useHandle(handle -> handle.execute("insert into table (column, column) values (?,?)", person.getAge(), person.getName()`

silly example aside, this still uses the wrong type of null. I can use your approach if that is the "correct" way of doing it, it's just I find it a bit weird my approach works with postgres etc but not oracle.

TheRealMarnes:
You're not really supposed to use `execute(query, args...)`. I assumed you were pseudocoding in your last examples. The mantra for Jdbi update queries is `handle.createQuery(query).bind(name, value).bindByType(name, nullish, type).execute()`.

Have you read [the developer guide](http://jdbi.org/), used Jdbi before, gotten familiar with it? You're dropping hints that you don't actually know how much of it works.

Are you using oracle db? We don't officially support it anymore. I don't see why it would matter for this case, but there may be some subtlety I don't know or can't think of right now.

`config.get(Arguments.class).findFor(nestedType, nestedValue)` is the magic that should make it work: assuming it correctly resolves `nestedType` to the runtime value of your `T` (String?), and you have a properly `T-ArgumentFactory` that returns a typed `NullArgument` for any values including null (the default String argumentfactory is ok), this should work. Have you registered any other custom argumentfactories, especially for String?

stevenschlansker:
> Are you using oracle db? We don't officially support it anymore. I don't see why it would matter for this case, but there may be some subtlety I don't know or can't think of right now.

That's not quite right.  What happened is that we moved the Oracle *extensions* out of the main project, into a side module.  No supported features changed, the extensions just moved to a new home.  As before, we try to support any JDBC-compliant driver (which Oracle sometimes is).  The Oracle specific extensions may receive less attention, but they are relatively limited to start with.

The problem here is that a Java `null` is untyped.  The `handle.execute` form takes a varargs parameter list, which is convenient for some operations, but has the unfortunate side effect of losing the type.

To preserve the type and insert typed nulls in databases, it's necessary to specify the type statically.  That means using `bindByType` in the fluent form (rather than `bind` or `execute`), or using SqlObject style where you can declare the type as a method parameter.  Another option is to use bean / Immutables mapping.

PR #1461, opened 2019-02-11
add `@Unmappable` to disable property mapping of read-only properties like `@Value.Derived`

stevenschlansker:


stevenschlansker:
I'd appreciate this making it in the next release, if anyone has time to review it.  We're currently running a patched build that includes this.

stevenschlansker:
Rebased against master to fix conflicts.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
```diff
@@ -164,9 +164,12 @@ private MethodHandle findBuilderSetter(final Class<?> builderClass, String name,
                     return MethodHandles.lookup().unreflect(m);
                 }
             }
-            final IllegalArgumentException iae = new IllegalArgumentException("Failed to find builder setter for property " + name + " of " + type);
+            final IllegalArgumentException iae = new IllegalArgumentException("Failed to find builder setter for property " + name + " on " + builderClass);
             failures.forEach(iae::addSuppressed);
-            throw iae;
+            return MethodHandles.dropArguments(
+                    MethodHandles.throwException(Object.class, IllegalArgumentException.class),
```
stevenschlansker: This crazy contraption constructs a MethodHandle that drops two arguments (`this` and the set value) and instead throws the constructed `iae`.  This laziness means if we don't ever try to call a setter, its absence is not fatal.

core/src/main/java/org/jdbi/v3/core/internal/AnnotationFactory.java
```diff
@@ -59,6 +60,10 @@ private AnnotationFactory() {}
                 return "@" + annotationType.getName() + "()";
             }
 
+            if (method.getDeclaringClass() == annotationType) {
+                return method.getDefaultValue();
```
qualidafial: `Annotation` javadocs define an explicit contract for how `equals()` and `hashCode()` should be implemented. Qualifier lookup and comparison will likely see subtle bugs for annotations with default attributes if we fail to follow it--especially considering that `QualifiedType` stores qualifiers in `HashSet`s.

* `Annotation.equals()` returns true if the argument is an annotation, `other.annotationType()` equals `this.annotationType()`, and all attributes are `equals()` to each other.
* `Annotation.hashCode()` returns the sum of attribute hashes, where an attribute hash equals `(127 * attributeName.hashCode()) ^ attributeValue.hashCode()`.
* `Annotation.toString()` doesn't have a strict contract, but we should follow suit with what the JDK does to make troubleshooting more approachable.

The current implementation already complies with the Annotation contract--which was easy since we don't yet allow attributes.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
qualidafial: So we return a method handle that throws when called, instead of throwing eagerly?

core/src/main/java/org/jdbi/v3/core/annotation/JdbiProperty.java
```diff
@@ -0,0 +1,37 @@
...
+import org.jdbi.v3.meta.Beta;
+
+/**
+ * Configure aspects of discovered bean / pojo properties.
+ * Currently only works with reflective bean and pojo mapping.
+ */
+@Beta
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface JdbiProperty {
```
qualidafial: What would you think of an `@Unmappable` annotation instead?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/ImmutablesPropertiesFactory.java
stevenschlansker: Exactly.

core/src/main/java/org/jdbi/v3/core/annotation/JdbiProperty.java
stevenschlansker: So, I was doing this for two reasons:

It gives us a natural place to add more properties later like `@JdbiProperty(name="custom_name")`

Allow you to re-enable mapping in subclasses.  This is a practice Jackson has: https://fasterxml.github.io/jackson-annotations/javadoc/2.0.0/com/fasterxml/jackson/annotation/JsonValue.html#value()

core/src/main/java/org/jdbi/v3/core/internal/AnnotationFactory.java
stevenschlansker: Thanks for catching this.  The simple implementation fooled me into thinking it was not important, not sure how I missed this obvious bit.  Fixed.

stevenschlansker:
Thanks @qualidafial for the review, I've significantly improved the `AnnotationFactory` interface to properly implement the `Annotation` contract and rebased on master.  Any further thoughts?

core/src/test/java/org/jdbi/v3/core/internal/AnnotationFactoryTest.java
```diff
@@ -25,17 +29,9 @@
...
-            .getAnnotation(Foo.class);
-
-        assertThat(real).isEqualTo(synthetic);
-        assertThat(synthetic).isEqualTo(real);
-        assertThat(synthetic.hashCode()).isEqualTo(real.hashCode());
-        assertThat(synthetic.toString()).isEqualTo(real.toString());
-        assertThat(synthetic.annotationType()).isEqualTo(Foo.class);
+    public void createNoProperties() throws Exception {
+        checkImplementation(Foo.class, "createNoProperties", AnnotationFactory.create(Foo.class),
+                "@org.jdbi.v3.core.internal.AnnotationFactoryTest$Foo()");
```
stevenschlansker: While `equals` and `hashCode` are well-specified, `toString` is not, so I check it against expected values rather than "whatever the current JVM does".

core/src/test/java/org/jdbi/v3/core/mapper/ImmutablesTest.java
```diff
@@ -182,4 +185,52 @@ public void testByteArray() {
...
+
+        @Value.Check
+        default void checkFoo() {
+            if (foo() == 999) {
+                throw new Boom();
+            }
+        }
+
+        @Value.Derived
+        @JdbiProperty(mappable = false)
```
stevenschlansker: This is the new feature -- previously, any attempt to use `@Value.Derived` makes your type totally unmappable.

core/src/test/java/org/jdbi/v3/core/mapper/ImmutablesTest.java
```diff
@@ -182,4 +185,52 @@ public void testByteArray() {
...
     }
+
+    @Value.Immutable
+    public interface DerivedProperty {
+        @Value.Default
+        default int foo() {
+            return 1;
+        }
+
+        @Value.Check
```
stevenschlansker: This is just a bonus test that `@Value.Check` methods work as you expect.

core/src/main/java/org/jdbi/v3/core/annotation/JdbiProperty.java
qualidafial: What about `@Mappable(false)` then?

core/src/main/java/org/jdbi/v3/core/internal/AnnotationFactory.java
```diff
@@ -48,18 +82,58 @@ private AnnotationFactory() {}
...
+    }
+
+    private static <A extends Annotation> boolean valuesEqual(Class<A> annotationType, Object a, Object b) {
+        for (Method m : annotationType.getDeclaredMethods()) {
+            Function<Object, Object> invoker = Unchecked.function(m::invoke);
+            Object valueA = invoker.apply(a);
+            Object valueB = invoker.apply(b);
+            if (valueA != null
+               && valueA.getClass().isArray()
+               && valueA.getClass().equals(valueB.getClass())
```
qualidafial: missing null check on `b`

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoMapper.java
```diff
@@ -95,6 +96,12 @@ public T map(ResultSet rs, StatementContext ctx) throws SQLException {
 
         for (PojoProperty<T> property : properties.getProperties().values()) {
             Nested anno = property.getAnnotation(Nested.class).orElse(null);
+            JdbiProperty propertyAnno = property.getAnnotation(JdbiProperty.class).orElse(JdbiProperty.DEFAULT);
+
+            if (!propertyAnno.mappable()) {
+                unmatchedColumns.remove(property.getName());
```
qualidafial: `property.getName()` is a bean property name, but `unmatchedColumn` contains column names. So a column name like `foo_bar` wouldn't be matched up to a `fooBar` Java property.

This check probably needs to be farther down, with the `findColumnIndex()` stuff.

core/src/main/java/org/jdbi/v3/core/annotation/JdbiProperty.java
stevenschlansker: Do you not like the ability to extend the same annotation later with more properties?  `Mappable` means it will only ever control mapping behavior, I was imagining it more like `@JsonProperty` which has a number of knobs.

qualidafial: Maybe I misunderstand your intent, but I don't see the need for a `@JdbiProperty` with a name attribute, since we already have `@ColumnName`.

stevenschlansker: OK, we'll give it a shot.

PR #1465, opened 2019-02-18
Fixed platform-dependent test

### LINKED ISSUE #1464: At least 1 unit test depends on the platform's locale to succeed
The [TestMessageFormatTemplateEngine::deprecationReason](https://github.com/jdbi/jdbi/blob/master/core/src/test/java/org/jdbi/v3/core/statement/TestMessageFormatTemplateEngine.java#L164) will fail for locales that don't insert commas (',') in numbers (ie: Mine is fr-CA which outputs 1 million as '1 000 000'). I had to set my locale through an environments variable to get this test to run correctly. Should this be documented somewhere or should this test just be removed (not sure what it's actually testing tbh).

FredDeschenes:
Fixes #1464

### DIFF
--- a/postgres/src/test/java/org/jdbi/v3/TestStatementsTimeout.java
+++ b/postgres/src/test/java/org/jdbi/v3/TestStatementsTimeout.java
@@ -13,7 +13,6 @@
  */
 package org.jdbi.v3;
 
-import org.apache.commons.lang3.SystemUtils;
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.statement.SqlStatements;
 import org.jdbi.v3.core.statement.UnableToExecuteStatementException;
@@ -22,24 +21,15 @@ import org.jdbi.v3.testing.JdbiRule;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.postgresql.util.PSQLException;
+import org.postgresql.util.PSQLState;
 
 import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 public class TestStatementsTimeout {
     @Rule
-    public JdbiRule dbRule = PostgresDbRule.rule(builder -> {
-        // We need to force the locale for the 'testTimeout' test
-        final String locale;
-
-        if (SystemUtils.IS_OS_WINDOWS) {
-            locale = "English_United States";
-        } else {
-            locale = "en_US";
-        }
-
-        builder.setLocaleConfig("locale", locale);
-    });
+    public JdbiRule dbRule = PostgresDbRule.rule();
 
     private Handle h;
 
@@ -57,6 +47,7 @@ public class TestStatementsTimeout {
 
         assertThatThrownBy(h.createQuery("select pg_sleep(3)").mapTo(String.class)::findOnly)
             .isInstanceOf(UnableToExecuteStatementException.class)
-            .hasMessageContaining("canceling statement due to user request");
+            .hasCauseInstanceOf(PSQLException.class)
+            .matches(ex -> PSQLState.QUERY_CANCELED.getState().equals(((PSQLException)ex.getCause()).getSQLState()));
     }
 }
--- a/postgres/src/test/java/org/jdbi/v3/TestStatementsTimeout.java
+++ b/postgres/src/test/java/org/jdbi/v3/TestStatementsTimeout.java
@@ -48,6 +48,6 @@ public class TestStatementsTimeout {
         assertThatThrownBy(h.createQuery("select pg_sleep(3)").mapTo(String.class)::findOnly)
             .isInstanceOf(UnableToExecuteStatementException.class)
             .hasCauseInstanceOf(PSQLException.class)
-            .matches(ex -> PSQLState.QUERY_CANCELED.getState().equals(((PSQLException)ex.getCause()).getSQLState()));
+            .matches(ex -> PSQLState.QUERY_CANCELED.getState().equals(((PSQLException) ex.getCause()).getSQLState()));
     }
 }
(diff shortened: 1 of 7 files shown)


FredDeschenes:
Btw I'm getting the same error when running the tests locally as Travis just hit, I'm guessing that's "normal"/expected?

TheRealMarnes:
The travis build isn't failing on this test, but on a dependency divergence... which is really strange considering you changed nothing of the sort in this PR.

FredDeschenes:
Yeah I was having this issue previously anyway on my laptop (whose locale is set to en-US), looks like it`s whining about the assertj library requiring an update, I'll look into pushing a PR for this as well.

FredDeschenes:
@TheRealMarnes : FYI I pushed the AssertJ update in #1466

FredDeschenes:
After having done some AspectJ upgrades, I've discovered that [TestStatementsTimeout](https://github.com/jdbi/jdbi/blob/master/postgres/src/test/java/org/jdbi/v3/TestStatementsTimeout.java) also depends on the locale, but this time it's the booted PostgreSQL instance and not the Java locale that's causing the issue (ie: the error message being returned by Postgres is in French on my computer and thus does not contain the expected string). Any advice as what should be done with this?

FredDeschenes:
Alright so I've allowed Postgres' locale to be set when using the JUnit rules that are provided by Jdbi, I've force the locale to "en_US" (and whatever it is for Windows, because of course it's different...) for all the tests that required a specific locale to be set, and I've removed an 'Assume' call in a currency related test to make it run on all locales (since it should now work everywhere).

Travis should still whine about the broken dependency, I can re-branch and re-apply the changes once #1466 is merged to validate that everything is fixed (I did locally before pushing the branches and everything seemed fine).

stevenschlansker:
I'm a little worried about all this platform detection leaking into each test.  If we improved the test harness either to fix the locale to a known value (like `C` or `ROOT`), or improved the test to work no matter what the locale is, that'd probably be preferable to a bunch of `if(WINDOWS)` in various tests.

postgres/src/test/java/org/jdbi/v3/TestStatementsTimeout.java
```diff
@@ -27,7 +28,18 @@
 
 public class TestStatementsTimeout {
     @Rule
-    public JdbiRule dbRule = PostgresDbRule.rule();
+    public JdbiRule dbRule = PostgresDbRule.rule(builder -> {
+        // We need to force the locale for the 'testTimeout' test
+        final String locale;
+
+        if (SystemUtils.IS_OS_WINDOWS) {
+            locale = "English_United States";
```
stevenschlansker: Is there a constant here we could use that doesn't vary by platform?  Like `Locale.ROOT`?

stevenschlansker: We could also fix it to a constant for every test case in the `JdbiRule` factory.

FredDeschenes: Yeah I thought about just changing JdbiRule to do this but realized this would force all users of the library to also have their tests be forced to en_US or whatever default we choose (which should be fine for most people, but will suck for the couple of edge cases where you do need another locale). I could add a `JdbiRule::embeddedEnUsPostgres()` function that takes care of setting this when required but would still let users use whatever locale they want.

stevenschlansker: It looks like the locale is only relevant because we look at the error message.  We could check the SQLSTATE of the exception instead and then we don't depend on locale at all, perhaps?

FredDeschenes: Good call, I've removed the locale dependency on this test.

FredDeschenes:
Believe me I'd much rather have my work system in good ol' en_US and not have to deal with this :P

postgres/src/test/java/org/jdbi/v3/postgres/TestPostgresTypes.java
```diff
@@ -49,7 +45,18 @@
...
     @ClassRule
-    public static JdbiRule postgresDBRule = JdbiRule.embeddedPostgres();
+    public static JdbiRule postgresDBRule = PostgresDbRule.rule(builder -> {
+        // We need to force the locale for the 'testReadWriteMoney' test
+        final String locale;
+
+        if (SystemUtils.IS_OS_WINDOWS) {
+            locale = "English_United States";
+        } else {
+            locale = "en_US";
```
stevenschlansker: This seems to be failing Travis.  Should it be `en_US.UTF-8`?  (That seems to be what it was before).

```
[INFO] Running org.jdbi.v3.postgres.TestPostgresTypes
[log:pid(7708)] INFO init-cd8b89f2-e221-4491-b490-e4b22edbf586:initdb - The files belonging to this database system will be owned by user "travis".
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.067 s <<< FAILURE! - in org.jdbi.v3.postgres.TestPostgresTypes
[ERROR] org.jdbi.v3.postgres.TestPostgresTypes  Time elapsed: 0.067 s  <<< ERROR!
java.lang.IllegalStateException: 
Process [/tmp/embedded-pg/PG-06e3a92a2edb6ddd6dbdf5602d0252ca/bin/initdb, -A, trust, -U, postgres, -D, /tmp/epg8422493978783602566, -E, UTF-8, --locale, en_US] failed
```

FredDeschenes: Yeah that's possible, let me validate this on my laptop.

FredDeschenes: Yup "en_US.UTF-8" fixed it.

PR #1466, opened 2019-02-18
Update to AssertJ 3.12

FredDeschenes:
'TestJavaTime' in the PostgreSQL module fails on Windows but works on
Linux. I think this is caused by the embedded Postgres being used, anyone have an input on this?

Example failing test output:
```
org.junit.ComparisonFailure: 
Expected :2019-02-18T18:06:07.957086900Z
Actual   :2019-02-18T18:06:07.957087Z
```

EDIT: By the way the tests will randomly pass/fail.

### DIFF
--- a/postgres/src/test/java/org/jdbi/v3/postgres/TestJavaTime.java
+++ b/postgres/src/test/java/org/jdbi/v3/postgres/TestJavaTime.java
@@ -20,7 +20,10 @@ import java.time.LocalTime;
 import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
+import java.time.temporal.ChronoUnit;
 
+import org.apache.commons.lang3.SystemUtils;
+import org.assertj.core.data.TemporalUnitOffset;
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.testing.JdbiRule;
 import org.junit.Before;
@@ -28,6 +31,7 @@ import org.junit.ClassRule;
 import org.junit.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.within;
 
 public class TestJavaTime {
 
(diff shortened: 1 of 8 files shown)


TheRealMarnes:
> By the way the tests will randomly pass/fail.

Yeah, travis has been doing that a whole lot lately. We also have to manually start the build almost every pr or git push.

FredDeschenes:
No I'm talking about the failing 'TestJavaTime' tests on Windows (unless you're talking about those tests also failing on Travis?)

TheRealMarnes:
I guess we're not talking about the same thing. We've just been having a lot of random test failures on travis lately, without a code change to blame AFAIK.

FredDeschenes:
Yeah I just saw the second build failed for no reason.

Anyway for the time-related PostgreSQL test I can just disable them on Windows if you want.

TheRealMarnes:
I'd rather widen the assertion than put platform detection in tests and selectively skip them.

postgres/src/test/java/org/jdbi/v3/postgres/TestJavaTime.java
```diff
@@ -45,6 +49,15 @@ public void setUp() {
         });
     }
 
+    private TemporalUnitOffset getAllowableOffset() {
+        // PostgreSQL seems to not have as much precision on Windows as it does on Linux.
+        if (SystemUtils.IS_OS_WINDOWS) {
+            return within(0, ChronoUnit.MICROS);
```
stevenschlansker: Can we just use the more lenient comparison window always, and get rid of the OS detection?  We are not really intending to test the system clock here at all, and OS-specific behavior should be avoided whenever possible.

FredDeschenes: Yup I'll just remove the condition.

qualidafial:
I suspect this had more to do with the JDK version you were running. When we upgraded from JDK 8 to 11, we noted several test failures, due to JDK 9+ Instant having more precision than in JDK 8, and also more precision than Postgres.

FredDeschenes:
Ah yeah that makes sense, I was running the tests with JDK11.

stevenschlansker:
FWIW, I was running jdk11 on Mac and didn't have any troubles.  But it could easily vary both by jdk version as well as OS.

PR #1472, opened 2019-02-21
EnumSetCollectorFactory: ecj compile fix

stevenschlansker:
Eclipse's `ecj` is unhappy with the `EnumSet::add` of a wildcard type and causes an error.
Factoring it out into a named type `E` and making the return type `Collector<E` gives it just enough more information to make it happy.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/collector/EnumSetCollectorFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/collector/EnumSetCollectorFactory.java
@@ -39,10 +39,15 @@ class EnumSetCollectorFactory implements CollectorFactory {
         return findGenericParameter(containerType, EnumSet.class);
     }
 
-    @SuppressWarnings("unchecked")
     @Override
     public Collector<?, ?, ?> build(Type containerType) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jdbi/v3/core/collector/EnumSetCollectorFactory.java
```diff
@@ -39,10 +39,14 @@ public boolean accepts(Type containerType) {
         return findGenericParameter(containerType, EnumSet.class);
     }
 
-    @SuppressWarnings("unchecked")
     @Override
     public Collector<?, ?, ?> build(Type containerType) {
-        Class<? extends Enum> componentType = (Class<? extends Enum>) findGenericParameter(containerType, EnumSet.class)
+        return build0(containerType);
```
TheRealMarnes: Could you give it a more functional name and/or comment why it exists?

qualidafial: I don't mind the name, but a comment about ecj would be good

PR #1480, opened 2019-03-03
Remove shading once and for all.

stevenschlansker:
Upgrade antlr-runtime to 3.5.2, which corrects the dependencies to no longer
drag in `stringtemplate` and `antlr` as `compile` scoped dependencies.

When combined with #1479 and #1335, we have no shade-worthy dependencies left!
The only strong dependencies are on `antlr-runtime`, `geantyref`, and `slf4j-api`.
Fixes #1332


TheRealMarnes:
Updating antlr broke the runtime though

stevenschlansker:
I upgraded to antlr4 since I couldn't figure out WTF antlr3's problem was and it's ancient.

core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/DefineStatementLexer.g4
```diff
@@ -20,9 +8,9 @@ fragment LT: '<' ;
 fragment GT: '>' ;
 fragment NAME: 'a'..'z' | 'A'..'Z' | '0'..'9' | '_';
 
-COMMENT: '/*' .* '*/';
-QUOTED_TEXT: QUOTE (ESCAPE_QUOTE | ~QUOTE)* QUOTE;
-DOUBLE_QUOTED_TEXT: DOUBLE_QUOTE (~DOUBLE_QUOTE)+ DOUBLE_QUOTE;
+COMMENT: '/*' .*? '*/';
+QUOTED_TEXT: QUOTE (ESCAPE_QUOTE | ~'\'')* QUOTE;
```
stevenschlansker: https://stackoverflow.com/questions/16790861/rule-reference-is-not-currently-supported-in-a-set-in-antlr4-grammar

stevenschlansker:
I marked this for re-review since it now also does a major antlr4 upgrade.

core/src/main/java/org/jdbi/v3/core/statement/internal/ErrorListener.java
```diff
@@ -0,0 +1,52 @@
...
+    public void syntaxError(Recognizer<?, ?> recognizer, Object offendingSymbol,
+            int line, int charPositionInLine, String msg,
+            RecognitionException e) {
+        throw new UnableToCreateStatementException(e);
+    }
+    @Override
+    public void reportAmbiguity(Parser recognizer, DFA dfa, int startIndex,
+            int stopIndex, boolean exact, BitSet ambigAlts,
+            ATNConfigSet configs) {
+        // not used by lexers
```
TheRealMarnes: Wouldn't a stub exception anyway be better here and in the other two? If by some miracle these are called regardless of the contract of lexers, I think we'd want to know. Unless these are called all the time but for non-errors. `UnsupportedOperationException` maybe?

PR #1481, opened 2019-03-03
Replace synchronized WeakHashMap with ConcurrentHashMap

### LINKED ISSUE #1103: Experiencing java monitor blocks
When Handle creation is contented, we experience long waits on java monitor 

` at java.util.Collections$SynchronizedMap.forEach(BiConsumer)
 at org.jdbi.v3.core.config.ConfigRegistry.<init>(ConfigRegistry)
 at org.jdbi.v3.core.config.ConfigRegistry.createCopy()
 at org.jdbi.v3.core.Jdbi.open()
`

is it possible to cache already created handles as a workaround?

stevenschlansker:
Fixes #1103

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
+++ b/core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
@@ -14,9 +14,8 @@
 package org.jdbi.v3.core.config;
 
 import java.util.Map;
-import java.util.WeakHashMap;
-
-import static java.util.Collections.synchronizedMap;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * A registry of {@link JdbiConfig} instances by type.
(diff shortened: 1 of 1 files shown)


TheRealMarnes:
I think we might wanna discuss this, or maybe it's just me. We're removing the map weakness, but we're also having all that talk about enabling reloading by not holding references? And the jdk bug, as far as I read it, was about recursive computeIfAbsent calls, but I don't see any recursion here.

stevenschlansker:
The `computeIfAbsent` is definitely possibly recursive -- any JdbiConfig that implements `setRegistry` may try to look up collaborator configs, causing a violent explosion.

The idea is that we tie all cached data to the `Jdbi` instance -- it's stored in the configuration registry, rather than `static` fields.  That way when an application is unloaded and releases its Jdbi we also release all cached data.

stevenschlansker:
See #1479 for some related context

Issue #1382, opened 2018-12-30
Calling @CreateSqlObject method on an on-demand DAO should return another on-demand DAO


qualidafial:
Users intuitively expect that if you call a `@CreateSqlObject` method on an on-demand SQL object, the returned SQL Object should also be on-demand. In reality:

- If you call the method from _outside_ the SQL object, the returned SQL object will be unusable, since its connection is closed.
- If you call it from _inside_ another method of the SQL object, it is attached to the same connection, and remains usable within the scope of that method call.

The fact that it doesn't behave as expected is a usability problem which keeps getting reported over and over.

The root of the problem is that `Jdbi.onDemand()` is implemented in core using proxies. When you call a method on an on-demand object, it is effectively calling `jdbi.useExtension(Dao.class, dao.callMethod())`. The implementation is agnostic to the implementation of SQL objects, let alone the `@CreateSqlObject` annotation.

I have some ideas on how we can improve this:

- Push the `onDemand()` feature into a new default method in the `ExtensionFactory` interface (maintains backward compatibility).
- Override the `onDemand()` implementation in `SqlObjectExtension` and expose the on-demand status through the API somehow. Maybe an `OnDemandSqlObject` interface, or something?
- Modify `@CreateSqlObject` handler to produce an on-demand or attached DAO based on whether the callee is on-demand or attached.
- Throw exceptions for bad usage with informative messaging
  - Calling `getHandle()` from outside an on-demand SQL object. Send them to `useHandle(...)` or `withHandle()` instead.
  - Calling `Transactional.begin()`, `rollback()`, or `commit()` from outside an on-demand SQL object. Send them to `inTxn()` or `withTxn()` instead.
  - Any other bad usage we encounter

@jdbi/contributors thoughts?

stevenschlansker:
Could CreateSqlObjectHandler simply itself return an OnDemand proxy rather than an attached proxy? OnDemand proxies already understand how to be reentrant (via ThreadLocal, ick), so I think it will behave the same in the existing "attach" case (the only one that worked).  This also avoids any of the "bad" cases you outline; they simply behave as an existing onDemand SqlObject would.

qualidafial:
I think that would break transactions for any existing SQL object code that use a `@CreateSqlObject` within a `@Transaction` method

qualidafial:
Today while looking at how Guice's ServletScopes works (request scopes in particular) I got to wondering whether Jdbi couldn't do a similar thing to make everything play nice with each other.

[...]

PR #1484, opened 2019-03-08
Reuse handles in nested withHandle, withExtension, inTransaction, and on-demand extension calls

### LINKED ISSUE #1350: Inconsistency between nested @Transaction and inTransaction(..)
Hello @jdbi team,

I came across some inconsistent mechanism. Documentation for `@Transaction` states:

>  Nested <code>@Transaction</code> annotations (e.g. one method calls another method, where both methods have this annotation) are collapsed into a single transaction. If the outer method annotation specifies an isolation level, then the inner method must either specify the same level, or not specify any level.

Which is very nice since it would be extremely hard to track transaction without having `Handle` available in transactional method scope.

But when I tried transaction nesting with SQL objects:

```java
final int id = dao(TransactionalInterfaceDao.class).inTransaction(dao1 -> {
  return dao(TransactionalInterfaceDao.class).inTransaction(dao2 -> {
    return dao2.get(dao2.insert(msg.getValue()));
  });
});
```

Such as:

```java
interface TransactionalInterfaceDao extends Transactional<TransactionalInterfaceDao> {

  @GetGeneratedKeys
  @SqlUpdate("insert into test (value) values (:value)")
  int insert(@Bind("value") int value);

  @SqlQuery("select value from test where id = :id")
  int get(@Bind("id") int id);
}
```

I got:

```plain
Caused by: java.lang.IllegalStateException: Already in transaction
	at org.jdbi.v3.core.transaction.LocalTransactionHandler.inTransaction(LocalTransactionHandler.java:129)
	at org.jdbi.v3.core.Handle.inTransaction(Handle.java:393)
	at org.jdbi.v3.sqlobject.transaction.Transactional.inTransaction(Transactional.java:96)
	at java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:627)
	at org.jdbi.v3.sqlobject.DefaultMethodHandler.invoke(DefaultMethodHandler.java:102)
	at org.jdbi.v3.sqlobject.SqlObjectFactory.lambda$null$13(SqlObjectFactory.java:163)
	at org.jdbi.v3.core.ConstantHandleSupplier.invokeInContext(ConstantHandleSupplier.java:52)
	at org.jdbi.v3.sqlobject.SqlObjectFactory.lambda$createInvocationHandler$14(SqlObjectFactory.java:162)
	at com.sun.proxy.$Proxy51.inTransaction(Unknown Source)
	***
```

I was under impression that declarative approach given by `@Transaction` annotation and imperative approach given by `Transactional` interface should work very similar, but they are very different when it comes to nesting.

I noticed #937, which is a nice approach to translate nested transaction as a sub-transaction, but this is applicable only in case when user is working on a RDBMS which supports sub-transactions specifically.

Right now I'm in the middle of changing `inTransaction()` to `@Transaction` because there is no need to track nested transactions, but I wanted to point this out and suggest that maybe, just maybe, both approaches should handle transactions nesting the same way. Shouldn't default behaviour for this be the same for both mechanisms? Use not-yet-implemented sub-transactions if they are supported by RDBMS, or collapse transactions to the uppermost one otherwise?

qualidafial:
Partial solution for #1382.

Nested calls on the same thread to any combination of these methods will share the same handle:

* `Jdbi.withHandle()`
* `Jdbi.useHandle()`
* `Jdbi.withExtension()`
* `Jdbi.useExtension()`
* `Jdbi.inTransaction()`
* `Jdbi.useTransaction()`
* Any method of an object obtained from `Jdbi.onDemand()`

Fixes #1350 by allowing nested `inTransaction` / `useTransaction` calls (on `Jdbi` or `Handle`) under the same transaction isolation rules currently in force for the `@Transaction` annotation.

TODO

* [x] Update release notes
* [x] Refactor threadlocal in Jdbi to store HandleSupplier instead of Handle, to play nice with extensions
* [x] Add tests for on-demand as the outer method call
* [x] Add tests for other combinations of nested calls from above list
* [x] Move nested transaction handling from SQL Object's `TransactionDecorator` into `Handle.inTransaction` in core.


qualidafial:
It occurred to me I missed one case: if the on-demand method call is the outer call, the handle won't be allocated until the on-demand object needs it, which means it might not be shared with a nested e.g. `useHandle()` call

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -327,8 +329,15 @@ public Handle open() {
      * @throws X any exception thrown by the callback
      */
     public <R, X extends Exception> R withHandle(HandleCallback<R, X> callback) throws X {
+        if (threadHandle.get() != null) {
```
stevenschlansker: I'm slightly worried some user will have code that expects to be able to nest `withHandle` to get two separate database connections.

qualidafial: I agree that this could be seen as breaking behavior by some people. There is actually a test that depended on this assumption, and had to be modified under the new behavior.

With that said, I think this is a step forward. I've seen multiple users who were confused by the existing behavior of opening separate handles per invocation. Even some of my coworkers.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -444,15 +454,28 @@ public Handle open() {
...
-            return callback.withExtension(extension);
+        try (LazyHandleSupplier handleSupplier = new LazyHandleSupplier(this, config)) {
+            threadHandleSupplier.set(handleSupplier);
+            return tryWithExtension(extensionType, callback, handleSupplier);
+        } finally {
+            threadHandleSupplier.remove();
         }
     }
 
+    private <R, E, X extends Exception> R tryWithExtension(Class<E> extensionType,
```
stevenschlansker: Maybe this is `callWithExtension`?  I don't see a `try` block or other failure handling here.

core/src/test/java/org/jdbi/v3/core/TestJdbiNestedCallBehavior.java
```diff
@@ -0,0 +1,437 @@
...
+                assertThat(e.getHandle()).isSameAs(onDemand.getHandle()));
+            return null;
+        });
+
+        TestExtension onDemand2 = jdbi.onDemand(TestExtension.class);
+        onDemand.run(() ->
+            onDemand2.run(() ->
+                assertThat(onDemand.getHandle()).isSameAs(onDemand2.getHandle())));
+    }
+}
```
stevenschlansker: Holy test cases Batman!

core/src/main/java/org/jdbi/v3/core/Jdbi.java
stevenschlansker: Yeah, my coworker was similarly confused.  I'm convinced.

PR #1485, opened 2019-03-09
Document Kotlin compiler target version needed for ResultSet mapping

rigoleto:
This PR simply documents that for Kotlin ResultSet mapping helpers to work, the Kotlin compiler's JVM target version must be 1.8 instead of the default 1.6.

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -4094,6 +4094,13 @@ To use this plugin, add a Maven dependency:
 </dependency>
 ----
 
+Ensure the Kotlin compiler's https://kotlinlang.org/docs/reference/using-maven.html#attributes-specific-for-jvm[JVM target version] is set to 1.8:
+[source,xml,subs="specialchars"]
+----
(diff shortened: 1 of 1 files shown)


docs/src/adoc/index.adoc
```diff
@@ -4094,6 +4094,13 @@ To use this plugin, add a Maven dependency:
 </dependency>
 ----
 
+Ensure the Kotlin compiler's https://kotlinlang.org/docs/reference/using-maven.html#attributes-specific-for-jvm[JVM target version] is set to 1.8:
```
stevenschlansker: Does this advice apply even if you use a later jdk, say 11?

rigoleto: I was seeing this problem on OpenJDK 10, the latest available in the Ubuntu repos. Installing 11 shows the same issue, though with a useful error message I wasn't seeing previously:
```
Cannot inline bytecode built with JVM target 1.8 into bytecode that is being built with JVM target 1.6. Please specify proper '-jvm-target' option
```

stevenschlansker: How about let's say "at least 1.8" so we don't drift out of date.

PR #1488, opened 2019-03-14
Loosen grammar definitions to treat any non-comment or -parameter as a literal

### LINKED ISSUE #1487: jdbi not support non alpahbet charaset in sql
Hi~

I meet some error using jdbi3(v3.6.0)
that's ORA-00936 error.

```
sql : SELECT nonAlphabetCharacterColumnName FROM nonAlphabetCharacter_TableName
rewritten: "SELECT FROM _"
```
oracle throw error.
database can using locale encoding database object name.

how to resolve this error???

qualidafial:
Fixes #1487 

@jdbi/contributors This will require a discussion.

In a nutshell, the use of a whitelist in our literal definition (rather than just `.` to match any character) causes a `NoViableAltException` for any SQL using non-latin object names.

This PR opens up the definition of a literal to include any text that is not part of a comment, quoted text, or parameter marker.

I had to disable several tests to get the build to pass, and I think we should consider removing them or porting them to a higher level:

* `TestColonGrammar` is over-testing the grammar definition. The existing tests expect a block of literal characters to be parsed into a single "literal" string, but the new grammar treats each character as a separate "literal." It doesn't actually matter, provided the parser doesn't mess with enquoted blocks, and successfully identifies parameter markers. I suggest we take the SQL strings in this test case, and use them to create higher-level tests in `TestColonPrefixSqlParser`, and only test that the right SQL got generated with the expected parameters.
* `TestColonPrefixSqlParser` and `TestHashPrefixSqlParser` both have a test named `testBailsOutOnInvalidInput` which expects an exception for SQL strings where the parameter prefix character (`:` or `#`) is followed by a non-identifier character. I would suggest that this doesn't matter. We should just treat `:` or `#` as a literal if it is not followed by identifier characters.

This should get rid of most (if not all) `NoViableAltException`s in Jdbi.


.gitignore
```diff
@@ -22,3 +22,4 @@ travis-settings-security.xml
 .build-oracle
 .factorypath
 docs/src/main/resources/jdbi2/apidocs
+*.tokens
```
qualidafial: antlr-generated alongside .g4 files in the src directory :unamused:

stevenschlansker:
I think I'm ok with making a change like this.  The less parsing of SQL we do the better.  I agree with your assessment of the tests -- we should make them test "end user behavior" rather than implementation details, and it is not our job to eagerly fail on unusual SQL syntax (and every time we do, we have to keep adding exceptions -- I'm looking at you `::`)

qualidafial:
Ok. I will refactor the tests and get back to you for a final review

qualidafial:
@stevenschlansker @TheRealMarnes I've updated the tests, so this is ready for review

core/src/test/java/org/jdbi/v3/core/statement/TestColonPrefixSqlParser.java
```diff
@@ -33,77 +32,146 @@ public void setUp() {
...
+        assertThat(parser.parse("SELECT '{\"a\":1, \"b\":2}'::jsonb ?? :key", ctx))
+            .isEqualTo(ParsedSql.builder()
+                .append("SELECT '{\"a\":1, \"b\":2}'::jsonb ?? ")
+                .appendNamedParameter("key")
+                .build());
+    }
+
+    @Test
+    public void testKoreanIdentifiers() {
+        String sql = "SELECT 제목 FROM 업무_게시물";
```
stevenschlansker: Should we additionally have a test for named parameters like `:제목`?

core/src/test/java/org/jdbi/v3/core/statement/TestColonPrefixSqlParser.java
```diff
@@ -33,77 +32,146 @@ public void setUp() {
...
+            .isEqualTo(ParsedSql.builder().append(doubleColon).build());
     }
 
     @Test
-    public void testSubstitutesDefinedAttributes() {
-        String sql = "select foo from bar where foo = :someValue";
-        ParsedSql parsed = parser.parse(sql, ctx);
-        assertThat(parsed.getSql()).isEqualTo("select foo from bar where foo = ?");
+    public void testColonWithoutNameNotAParameter() {
+        String sql = "select * from something\n where id = :\u0087\u008e\u0092\u0097\u009c";
```
TheRealMarnes: could we have this magic unicode string explained with a comment or `describedAs`? I don't see how this is testing what the method name says because it looks like a named parameter to me.

core/src/test/java/org/jdbi/v3/core/statement/TestColonPrefixSqlParser.java
TheRealMarnes: seconded

core/src/test/java/org/jdbi/v3/core/statement/TestDefinedAttributeTemplateEngine.java
```diff
@@ -89,4 +89,10 @@ public void testColonInComment() {
         String sql = "/* comment with : colons :: inside it */ select 1";
```
TheRealMarnes: let's add a valid `:abc` parameter in here too, not only invalid bait

core/src/test/java/org/jdbi/v3/core/statement/TestHashPrefixSqlParser.java
```diff
@@ -67,9 +66,10 @@ public void testBacktickOkay() {
     }
 
     @Test
-    public void testBailsOutOnInvalidInput() {
-        assertThatThrownBy(() -> parser.parse("select * from something\n where id = #\u0087\u008e\u0092\u0097\u009c", ctx))
-            .isInstanceOf(UnableToCreateStatementException.class);
+    public void testHashWithoutNameNotAParameter() {
+        String sql = "select * from something\n where id = #\u0087\u008e\u0092\u0097\u009c";
```
TheRealMarnes: same as before, I don't see how whatever this unicode string means is not a named parameter

core/src/test/java/org/jdbi/v3/core/statement/TestDefinedAttributeTemplateEngine.java
qualidafial: DefineStatementLexer.g is totally agnostic to colons, so I think this test is a leftover of when we split apart StatementRewriter into TemplateEngine and SqlParser

core/src/test/java/org/jdbi/v3/core/statement/TestHashPrefixSqlParser.java
qualidafial: Parameter names are restricted to Java-ish parameter names:

```antlr
fragment NAME: 'a'..'z' | 'A'..'Z' | '0'..'9' | '_' | '.' | '?.' | '#';
```

qualidafial:
@stevenschlansker @TheRealMarnes 

I've expanded the scope of this PR a little. A little digging uncovered some Unicode ranges we can use to recognize any "java letter" per the JLS spec. This means that this PR now expands the SQL parsers to recognize _any_ valid java identifier in a named parameter. I've also updated the defined attribute template engine to recognize any valid java identifier as an attribute name.

Could you review this PR again?

PR #1500, opened 2019-04-09
Detect ambiguous interface multiple inheritance

### LINKED ISSUE #1497: Undefined behavior when inheriting from multiple interfaces
I had a use case in which I had two interfaces, used in an on-demand manner.  Each had a `@Transaction`-annotated method, and I wanted one to be able to call the other reentrantly.  So, I created a new interface that inherited from both and simply used a single on-demand interface.  But, this had unintended consequences.  Java happily allows multiple interface inheritance as long as the signatures match.  And indeed, I had some overlapping, and matching, signatures.  However, the functions from the parent interfaces had _different_ Jdbi annotations!

```
interface A {
  @SqlUpdate("delete from a")
  void do();
}
interface B {
  @SqlUpdate("delete from b")
  void do();
}
interface C extends A, B {
}
```

The multiple inheritance hack was nice from a reentrant `@Transaction` point of view, but it permitted this hyper-subtle unintended consequence.

Could Jdbi even meaningfully detect the duplication of, e.g., `@SqlUpdate`s, and error out when you try to construct the on-demand instance?

stevenschlansker:
Fixes #1497

cc @pennello

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
@@ -27,6 +27,7 @@ import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.jdbi.v3.core.config.ConfigRegistry;
@@ -126,6 +127,17 @@ public class SqlObjectFactory implements ExtensionFactory {
                 }
                 handlers.computeIfAbsent(method, m -> buildMethodHandler(sqlObjectType, m, registry, decorators));
             }
+
+            handlers.keySet().stream()
+                .filter(m -> !m.isSynthetic())
+                .collect(Collectors.groupingBy(m -> Arrays.asList(m.getName(), Arrays.asList(m.getParameterTypes()))))
+                .values()
+                .stream()
+                .filter(l -> l.size() > 1)
+                .findAny()
+                .ifPresent(ms -> {
+                    throw new UnableToCreateSqlObjectException(sqlObjectType + " has ambiguous methods " + ms + ", please resolve with an explicit override");
+                });
 
             return handlers;
         });
--- a/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestVariousOddities.java
+++ b/sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestVariousOddities.java
@@ -31,6 +31,7 @@ import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 public class TestVariousOddities {
     @Rule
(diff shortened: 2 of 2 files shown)


sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestVariousOddities.java
```diff
@@ -121,4 +122,30 @@ public void testNullQueryReturn() {
...
+    public void testInterfaceMIResolution() {
+        dbRule.getSharedHandle().attach(SpiffyUniverseSaved.class);
+    }
+
+    public interface SpiffyWithMustache {
+        @SqlQuery("select id, name from something where id = :id")
+        Something byId(@Bind("id") long id);
+    }
+
+    public interface SpiffyParallelUniverse extends Spiffy, SpiffyWithMustache {}
```
TheRealMarnes: can we get less arbitrary names here? `ParentFoo extends SubFoo1, SubFoo2` or something

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestVariousOddities.java
```diff
@@ -121,4 +122,30 @@ public void testNullQueryReturn() {
      * This interface should not be loaded by any test other than {@link TestVariousOddities#testConcurrentHashCode()}.
      */
     public interface SpiffyConcurrent extends SqlObject {}
+
+    @Test
+    public void testInterfaceMIAbiguity() {
```
TheRealMarnes: would like to see more expressive/clearer test names

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -127,6 +128,17 @@ private boolean looksLikeSqlObject(Class<?> extensionType) {
...
 
+            handlers.keySet().stream()
+                .filter(m -> !m.isSynthetic())
+                .collect(Collectors.groupingBy(m -> Arrays.asList(m.getName(), Arrays.asList(m.getParameterTypes()))))
+                .values()
+                .stream()
+                .filter(l -> l.size() > 1)
+                .findAny()
+                .ifPresent(ms -> {
+                    throw new UnableToCreateSqlObjectException(sqlObjectType + " has ambiguous methods " + ms);
```
qualidafial: This error message would be more helpful if it included a suggestion to override the method and annotate it / provide a `default` implementaiton in `sqlObjectType`

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestVariousOddities.java
```diff
@@ -121,4 +122,30 @@ public void testNullQueryReturn() {
...
+            .hasMessageContaining("SpiffyParallelUniverse has ambiguous methods");
+    }
+
+    @Test
+    public void testInterfaceMIResolution() {
+        dbRule.getSharedHandle().attach(SpiffyUniverseSaved.class);
+    }
+
+    public interface SpiffyWithMustache {
+        @SqlQuery("select id, name from something where id = :id")
```
TheRealMarnes: I think we can simplify the queries to just `select 1` and `select 2` or something instead of reusing this crazy fantasy domain, and to stick closer to the given example code (`delete from a` and `delete from b`), which made a more succinctly understandable conflict

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -127,6 +128,17 @@ private boolean looksLikeSqlObject(Class<?> extensionType) {
                 handlers.computeIfAbsent(method, m -> buildMethodHandler(sqlObjectType, m, registry, decorators));
             }
 
+            handlers.keySet().stream()
+                .filter(m -> !m.isSynthetic())
+                .collect(Collectors.groupingBy(m -> Arrays.asList(m.getName(), Arrays.asList(m.getParameterTypes()))))
+                .values()
```
TheRealMarnes: I wonder if this could be made more lazy, so we don't group everything needlessly when a single size>2 entry is enough. Then again, micro-optimization so let's not spend too much thought on it.

stevenschlansker: This is only done once per class during initialization, so I don't think we should spend time micro-optimizing it.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestVariousOddities.java
stevenschlansker: Is the joke not funny at all? :(  Coming up with silly test names is one of the only fun parts of writing tests, and I was trying to re-use what was already there (the whole `Spiffy`) thing.  I guess I can make a separate hierarchy.

qualidafial: I laughed

TheRealMarnes: Not much of a hierarchy, just 1 interface extending two others and that's it... It's not like we need to use the made-up domain here

PR #1502, opened 2019-04-09
oraclejdk9 EOL -> openjdk13

stevenschlansker:


qualidafial:
NPE inside the jdk12 compiler :facepalm:

stevenschlansker:
Yeah, wtf??

brianm:
I would strongly consider sticking to LTS releases (so 11, not 12)

qualidafial:
I'd like to see us testing both, actually. That keeps us compatible with LTS releases, but also gives us notice when something breaks on a bleeding edge JDK.

Practically speaking, we're probably going to stick to JDK 8 as the baseline for several years yet, notwithstanding it being EOL'd. Adoption of JDK 9+ has been slow.

stevenschlansker:
Yeah, note that we already test jdk11, this just replaces testing against 9 with 12.  No change to baseline requirements or source code compatibility level.

I opened this because I was having trouble with my local jdk12 and sure enough Travis has trouble too...

stevenschlansker:
Created minimal reproducer, looks like it's the new nested Jdbi tests:

https://gist.github.com/stevenschlansker/a23cc97fb6d15fad6ae04bf54511ce62

Filed as JBS review id 9060134, will update when we have a public JBS id.

stevenschlansker:
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8222251

stevenschlansker:
Woohoo, finally!

PR #1509, opened 2019-04-16
Convenience methods for registering many Immutables types

stevenschlansker:


core/src/main/java/org/jdbi/v3/core/mapper/immutables/JdbiImmutables.java
```diff
@@ -54,6 +55,23 @@ public void setRegistry(ConfigRegistry registry) {
...
+    public JdbiImmutables registerImmutable(Class<?>... specs) {
+        return registerImmutable(Arrays.asList(specs));
+    }
+
+    /**
+     * Convenience method for registering many immutable types.
+     * @see #registerImmutable(Class)
+     */
+    public JdbiImmutables registerImmutable(Iterable<Class<?>> specs) {
+        specs.iterator().forEachRemaining(this::registerImmutable);
```
qualidafial: Iterable already has a forEach IIRC

core/src/main/java/org/jdbi/v3/core/mapper/immutables/JdbiImmutables.java
```diff
@@ -54,6 +55,23 @@ public void setRegistry(ConfigRegistry registry) {
                     .orElseThrow(() -> new IllegalArgumentException("Neither " + spec + " nor " + impl + " have a 'builder' method")));
     }
 
+    /**
+     * Convenience method for registering many immutable types.
+     * @see #registerImmutable(Class)
+     */
+    public JdbiImmutables registerImmutable(Class<?>... specs) {
```
qualidafial: Do we want to likewise overload `registerModifiable` too?

stevenschlansker: I actually considered dropping Modifiables completely -- it's got some hairy limitations that made it unsuitable at least for my use case.  So I wasn't investing in it as much as the main Immutables section.  But you're entirely right; I'll add it for completeness.

PR #1511, opened 2019-04-23
Add new ResultIterable<T> methods for single row results

### LINKED ISSUE #1498: How to get the only row if present?
When loading scalar result one can  conveniently use `findOnly`:
```java
handle.createQuery("SELECT name FROM user WHERE user_id = :id")
  .bind("id", ...)
  .mapTo(String.class)
  .findOnly()
```
This approach has a problem though -- it fails when the record is not found, for example has just been deleted.

Alternative approach, with `findFirst` works very well, however, it can mask bugs, should my query unexpectedly return more than 1 result row.

What is the recommended way to get the only result row if present?

qualidafial:
- `one() : T` - throws if zero or multiple rows returned
- `findOne() : Optional<T>` - throws if multiple rows returned
- `first() : T` - throws if zero rows returned
- Deprecate `findOnly()` in favor of `one()`

Fixes #1498


core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
```diff
@@ -89,27 +89,79 @@ default void forEach(Consumer<? super T> action) {
...
+    default Optional<T> findOne() {
+        try (ResultIterator<T> iter = iterator()) {
+            if (!iter.hasNext()) {
+                return Optional.empty();
+            }
+
+            final T r = iter.next();
+
+            if (iter.hasNext()) {
+                throw new IllegalStateException("Expected zero to one elements, but found multiple");
```
findepi: would it be useful to include values in err msg like this:
```suggestion
                throw new IllegalStateException(format(
                    "Expected zero to one elements, but found multiple. First two are: %s, %s",
                    r,
                    iter.next()));
```

?

core/src/test/java/org/jdbi/v3/core/statement/TestQueries.java
```diff
@@ -275,8 +274,8 @@ public void testFirstWithNoResult() {
     @Test
     public void testNullValueInColumn() {
         h.execute("insert into something (id, name) values (?, ?)", 1, null);
-        Optional<String> s = h.createQuery("select name from something where id=1").mapTo(String.class).findFirst();
-        assertThat(s.isPresent()).isFalse();
+        String s = h.createQuery("select name from something where id=1").mapTo(String.class).first();
+        assertThat(s).isNull();
```
findepi: This changes test semantics. 

BTW What about testing all cases:
```
h.createQuery("....").mapTo(String.class).first(); // .. should be null
h.createQuery("....").mapTo(String.class).one(); // .. should be null
h.createQuery("....").mapTo(String.class).findOne(); // .. should be Optional.empty
```

?

docs/src/adoc/index.adoc
```diff
@@ -553,22 +553,44 @@ that returns a result set from the database.
 include::{exampledir}/StatementsTest.java[tags=query]
 ----
 
-For single rows, you can use `findOnly()`, which expects exactly one row (or
-throws an exception):
+To get a single row from a query, there are a few possible methods to use,
+depending on the number of rows that might be in the result set.
+
+Call `one()` when the result contains exactly one row. Throws an exception if
```
findepi: Caller doesn't know the result cardinality, but they can have their expectations about that:

```suggestion
Call `one()` when you expect the result to contain exactly one row. Throws an exception if
```

(same below in other cases)

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
```diff
@@ -89,27 +89,79 @@ default void forEach(Consumer<? super T> action) {
...
+     * Returns the only row in the result set. Returns {@code null} if the row itself is
+     * {@code null}.
+     * @throws IllegalStateException if the result set contains zero or multiple rows
+     * @return the only row in the result set.
      */
-    default T findOnly() {
+    default T one() {
         try (ResultIterator<T> iter = iterator()) {
             if (!iter.hasNext()) {
-                throw new IllegalStateException("No element found in 'only'");
```
stevenschlansker: In theory this might be a `NoSuchElementException`

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
stevenschlansker: It would be useful, but I think it would also risk inadvertent disclosure.  For example if you are reading a user row that contains a password, you would not want that to accidentally get emitted in an error message.

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
```diff
@@ -89,27 +89,79 @@ default void forEach(Consumer<? super T> action) {
...
-     * Get the only row in the result set.
-     * @throws IllegalStateException if zero or multiple rows are returned
-     * @return the object mapped from the singular row in the results
+     * Returns the only row in the result set. Returns {@code null} if the row itself is
+     * {@code null}.
+     * @throws IllegalStateException if the result set contains zero or multiple rows
+     * @return the only row in the result set.
      */
-    default T findOnly() {
+    default T one() {
```
stevenschlansker: This could be a simple `findOne().orElseThrow(() -> new ...)`

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
stevenschlansker: I guess we probably don't want to change that.

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
qualidafial: I agree with Steven. It would be useful, but the security concern would discourage users with PCI / PHI uses from using Jdbi for fear of exposing sensitive data in logs.

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
qualidafial: This is an important distinction and I'm glad you called it out.

`one()` and `first()` can return null, if the row itself maps to null. Both throw if there are no rows.

`findOne()` and `findFirst()`, because they return `Optional<T>`, conflate "no rows" with "row that contained null" into the single result value `Optional.empty()`. So with the `find_()` methods there's no way to tell the difference.

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
qualidafial: I left it as `IllegalStateException` since we had several tests expecting that exception type.

core/src/test/java/org/jdbi/v3/core/statement/TestQueries.java
qualidafial: It does change semantics--by switching from `findFirst()` to `first()` we're getting more specific, and requiring the result to contain a single row that is null. The old test could conflate no rows with a single null row.

docs/src/adoc/index.adoc
qualidafial: Good idea. I incorporated this rewording as well as some clarifications about return types in the latest commit.

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
qualidafial: I added clarifications to the developer guide in the latest commit.

stevenschlansker: Ah, indeed!

Issue #1516, opened 2019-04-29
Custom StatementCustomizer does not work with @SqlBatch anymore


christopheblin:
I just updated from JDBI 3.4.0 to 3.8.0 and when I use a custom StatementCustomizer on a SqlBatch (that worked on 3.4.0) it now fails with a NullPointerException

The method beforeExecution is receiving first parameter "stmt" which is null !

What did I do wrong ? Is that a bug in JDBI ?

    java.lang.NullPointerException
	at org.jdbi.v3.sqlobject.customizer.internal.ReadOnlyCheckCustomizer.beforeExecution(ReadOnlyCheckCustomizer.java:15)
	at org.jdbi.v3.core.statement.SqlStatement.lambda$beforeExecution$9(SqlStatement.java:1600)
	at org.jdbi.v3.core.statement.BaseStatement.callCustomizers(BaseStatement.java:80)
	at org.jdbi.v3.core.statement.SqlStatement.beforeExecution(SqlStatement.java:1600)
	at org.jdbi.v3.core.statement.PreparedBatch.internalBatchExecute(PreparedBatch.java:171)
	at org.jdbi.v3.core.statement.PreparedBatch.execute(PreparedBatch.java:66)
	at org.jdbi.v3.core.statement.PreparedBatch.executeAndGetModCount(PreparedBatch.java:71)
	at org.jdbi.v3.sqlobject.statement.internal.SqlBatchHandler.executeBatch(SqlBatchHandler.java:310)
	at org.jdbi.v3.sqlobject.statement.internal.SqlBatchHandler.access$000(SqlBatchHandler.java:46)
	at org.jdbi.v3.sqlobject.statement.internal.SqlBatchHandler$2.hasNext(SqlBatchHandler.java:206)
	at org.jdbi.v3.sqlobject.statement.internal.SqlBatchHandler.invoke(SqlBatchHandler.java:232)
	at org.jdbi.v3.sqlobject.SqlObjectFactory.lambda$null$18(SqlObjectFactory.java:184)
	at org.jdbi.v3.core.ConstantHandleSupplier.invokeInContext(ConstantHandleSupplier.java:52)
	at org.jdbi.v3.sqlobject.SqlObjectFactory.lambda$createInvocationHandler$19(SqlObjectFactory.java:181)
	at com.sun.proxy.$Proxy46.insertCommunityRatings(Unknown Source)
[...]

PR #1518, opened 2019-04-29
forbid hidden fields

TheRealMarnes:
#1516 inspired me to do this. I think forbidding shadowing reduces the possibility of confusion. Having 2 distinct variables with essentially the same name within any given context is a dangerous enough concept on its own.

The cleanup revealed a few cases of odd design that I fixed collaterally, like some objects passing their own fields to their own instance methods for no reason. The "special" changes are in separate commits, the rest are basically renames. I'm very much open to ideas for better names for any of them, I've lacked the contextual knowledge for a lot of them.

I checked and found only a single backward-incompatible change, see below.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -85,8 +85,8 @@ public class Handle implements Closeable, Configurable<Handle> {
         this.config.set(config);
     }
 
-    void setConfigThreadLocal(ThreadLocal<ConfigRegistry> config) {
-        this.config = config;
+    void setConfigThreadLocal(ThreadLocal<ConfigRegistry> configThreadLocal) {
+        this.config = configThreadLocal;
     }
 
     /**
@@ -550,8 +550,8 @@ public class Handle implements Closeable, Configurable<Handle> {
         this.extensionMethod.set(extensionMethod);
     }
 
-    void setExtensionMethodThreadLocal(ThreadLocal<ExtensionMethod> extensionMethod) {
-        this.extensionMethod = requireNonNull(extensionMethod);
+    void setExtensionMethodThreadLocal(ThreadLocal<ExtensionMethod> extensionMethodThreadLocal) {
+        this.extensionMethod = requireNonNull(extensionMethodThreadLocal);
     }
 
     interface ConnectionCloser {
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -55,13 +55,13 @@ public class Handle implements Closeable, Configurable<Handle> {
     private final Connection connection;
     private final boolean forceEndTransactions;
 
-    private ThreadLocal<ConfigRegistry> config;
-    private ThreadLocal<ExtensionMethod> extensionMethod;
+    private ThreadLocal<ConfigRegistry> localConfig;
+    private ThreadLocal<ExtensionMethod> localExtensionMethod;
     private StatementBuilder statementBuilder;
 
     private boolean closed = false;
 
-    Handle(ConfigRegistry config,
+    Handle(ConfigRegistry localConfig,
            ConnectionCloser closer,
            TransactionHandler transactions,
            StatementBuilder statementBuilder,
@@ -70,23 +70,23 @@ public class Handle implements Closeable, Configurable<Handle> {
         this.transactions = transactions;
         this.connection = connection;
 
-        this.config = ThreadLocal.withInitial(() -> config);
-        this.extensionMethod = new ThreadLocal<>();
+        this.localConfig = ThreadLocal.withInitial(() -> localConfig);
+        this.localExtensionMethod = new ThreadLocal<>();
         this.statementBuilder = statementBuilder;
         this.forceEndTransactions = !transactions.isInTransaction(this);
     }
 
     @Override
     public ConfigRegistry getConfig() {
-        return config.get();
+        return localConfig.get();
     }
 
     void setConfig(ConfigRegistry config) {
-        this.config.set(config);
+        this.localConfig.set(config);
     }
 
-    void setConfigThreadLocal(ThreadLocal<ConfigRegistry> configThreadLocal) {
-        this.config = configThreadLocal;
+    void setLocalConfig(ThreadLocal<ConfigRegistry> configThreadLocal) {
+        this.localConfig = configThreadLocal;
     }
 
     /**
@@ -130,7 +130,7 @@ public class Handle implements Closeable, Configurable<Handle> {
         }
 
         boolean wasInTransaction = false;
-        if (forceEndTransactions && config.get().get(Handles.class).isForceEndTransactions()) {
+        if (forceEndTransactions && localConfig.get().get(Handles.class).isForceEndTransactions()) {
             try {
                 wasInTransaction = isInTransaction();
             } catch (Exception e) {
@@ -138,8 +138,8 @@ public class Handle implements Closeable, Configurable<Handle> {
             }
         }
 
-        extensionMethod.remove();
-        config.remove();
+        localExtensionMethod.remove();
+        localConfig.remove();
 
         if (wasInTransaction) {
             try {
@@ -543,15 +543,15 @@ public class Handle implements Closeable, Configurable<Handle> {
      * @return the extension method currently bound to the handle's context
      */
     public ExtensionMethod getExtensionMethod() {
-        return extensionMethod.get();
+        return localExtensionMethod.get();
     }
 
     void setExtensionMethod(ExtensionMethod extensionMethod) {
-        this.extensionMethod.set(extensionMethod);
+        this.localExtensionMethod.set(extensionMethod);
     }
 
-    void setExtensionMethodThreadLocal(ThreadLocal<ExtensionMethod> extensionMethodThreadLocal) {
-        this.extensionMethod = requireNonNull(extensionMethodThreadLocal);
+    void setLocalExtensionMethod(ThreadLocal<ExtensionMethod> extensionMethodThreadLocal) {
+        this.localExtensionMethod = requireNonNull(extensionMethodThreadLocal);
     }
 
     interface ConnectionCloser {
--- a/core/src/main/java/org/jdbi/v3/core/argument/ObjectFieldArguments.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/ObjectFieldArguments.java
@@ -71,8 +71,8 @@ public class ObjectFieldArguments extends ObjectPropertyNamedArgumentFinder {
     }
 
     @Override
-    protected NamedArgumentFinder getNestedArgumentFinder(Object bean) {
-        return new ObjectFieldArguments(null, bean);
+    protected NamedArgumentFinder getNestedArgumentFinder(Object source) {
+        return new ObjectFieldArguments(null, source);
     }
 
     @Override
--- a/core/src/main/java/org/jdbi/v3/core/argument/ObjectFieldArguments.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/ObjectFieldArguments.java
@@ -71,8 +71,8 @@ public class ObjectFieldArguments extends ObjectPropertyNamedArgumentFinder {
     }
 
     @Override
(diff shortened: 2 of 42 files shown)


core/src/main/java/org/jdbi/v3/core/collector/OptionalBuilder.java
```diff
@@ -20,21 +20,21 @@
     private final Supplier<O> empty;
     private final Function<T, O> factory;
 
-    boolean set;
-    T value;
+    private boolean set;
```
TheRealMarnes: technically not backward-compatible, but this isn't really API, so what do we do?

qualidafial: Demoting from package-private to private is fine. You just can't demote from API to non-API

TheRealMarnes: slipped my mind that the `jdbi` package tree is personal space :)

core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -550,8 +550,8 @@ void setExtensionMethod(ExtensionMethod extensionMethod) {
         this.extensionMethod.set(extensionMethod);
     }
 
-    void setExtensionMethodThreadLocal(ThreadLocal<ExtensionMethod> extensionMethod) {
-        this.extensionMethod = requireNonNull(extensionMethod);
+    void setExtensionMethodThreadLocal(ThreadLocal<ExtensionMethod> extensionMethodThreadLocal) {
+        this.extensionMethod = requireNonNull(extensionMethodThreadLocal);
```
qualidafial: Can we disable this rule on setters? This is a pretty common idiom.

core/src/main/java/org/jdbi/v3/core/argument/ObjectFieldArguments.java
```diff
@@ -71,8 +71,8 @@ public ObjectFieldArguments(String prefix, Object bean) {
     }
 
     @Override
-    protected NamedArgumentFinder getNestedArgumentFinder(Object obj) {
-        return new ObjectFieldArguments(null, obj);
+    protected NamedArgumentFinder getNestedArgumentFinder(Object bean) {
+        return new ObjectFieldArguments(null, bean);
```
qualidafial: Not strictly a bean

core/src/main/java/org/jdbi/v3/core/argument/ObjectMethodArguments.java
```diff
@@ -73,8 +73,8 @@ public ObjectMethodArguments(String prefix, Object object) {
     }
 
     @Override
-    protected NamedArgumentFinder getNestedArgumentFinder(Object obj) {
-        return new ObjectMethodArguments(null, obj);
+    protected NamedArgumentFinder getNestedArgumentFinder(Object bean) {
+        return new ObjectMethodArguments(null, bean);
```
qualidafial: Ditto--not strictly a bean

core/src/main/java/org/jdbi/v3/core/Handle.java
TheRealMarnes: the rule is disabled on setters, but the param name must match the setter name

TheRealMarnes: http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/HiddenFieldCheck.html

core/src/main/java/org/jdbi/v3/core/argument/ObjectMethodArguments.java
TheRealMarnes: changed

core/src/main/java/org/jdbi/v3/core/argument/ObjectFieldArguments.java
TheRealMarnes: changed

core/src/main/java/org/jdbi/v3/core/Handle.java
qualidafial: This is package private so we can rename either the field or the setter so they match

core/src/test/java/org/jdbi/v3/core/statement/TestBindList.java
```diff
@@ -61,25 +60,25 @@ public void testBindList() {
...
+        handle.registerRowMapper(FieldMapper.factory(Thing.class));
+
+        handle.createUpdate("insert into thing (<columns>) values (<values>)")
+            .setSqlParser(new HashPrefixSqlParser())
+            .defineList("columns", "id", "foo")
+            .bindList("values", 3, "abc")
+            .execute();
+
+        List<Thing> list = handle.createQuery("select id, foo from thing where id in (<ids>)")
+            .setSqlParser(new HashPrefixSqlParser())
```
qualidafial: Nit: Could just set the SQL parser once on the handle

core/src/main/java/org/jdbi/v3/core/statement/Call.java
```diff
@@ -117,15 +117,15 @@ public OutParameters invoke() {
         }
 
         @Override
-        public void apply(int position, PreparedStatement statement, StatementContext ctx) throws SQLException {
-            ((CallableStatement) statement).registerOutParameter(position, sqlType);
-            this.position = position;
+        public void apply(int newPosition, PreparedStatement statement, StatementContext ctx) throws SQLException {
```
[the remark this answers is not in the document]
TheRealMarnes: yeah this may have been one of those where I couldn't tell the meaning of the param

core/src/test/java/org/jdbi/v3/core/mapper/reflect/BeanMapperTest.java
```diff
@@ -42,27 +42,26 @@
...
-    ResultSetMetaData resultSetMetaData;
+    private ResultSetMetaData resultSetMetaData;
 
-    Handle handle = HandleAccess.createHandle();
-    StatementContext ctx = StatementContextAccess.createContext(handle);
-
-    RowMapper<SampleBean> mapper = BeanMapper.of(SampleBean.class);
+    private Handle sharedHandle;
+    private Handle handle = HandleAccess.createHandle();
+    private StatementContext ctx = StatementContextAccess.createContext(handle);
```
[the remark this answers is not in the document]
TheRealMarnes: I don't think that belongs in this PR?

core/src/test/java/org/jdbi/v3/core/mapper/reflect/FieldMapperTest.java
```diff
@@ -42,27 +42,26 @@
...
-    ResultSetMetaData resultSetMetaData;
+    private ResultSetMetaData resultSetMetaData;
 
-    Handle handle = HandleAccess.createHandle();
-    StatementContext ctx = StatementContextAccess.createContext(handle);
-
-    RowMapper<SampleBean> mapper = FieldMapper.of(SampleBean.class);
+    private Handle sharedHandle;
+    private Handle handle = HandleAccess.createHandle();
+    private StatementContext ctx = StatementContextAccess.createContext(handle);
```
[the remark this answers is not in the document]
TheRealMarnes: idem

TheRealMarnes:
Commits rebased and remarks handled

core/src/main/java/org/jdbi/v3/core/array/InferredSqlArrayTypeFactory.java
```diff
@@ -26,18 +26,18 @@
...
-        this.elementType = findGenericParameter(arrayType.getClass(), SqlArrayType.class)
+        this.inferredElementType = findGenericParameter(arrayType.getClass(), SqlArrayType.class)
                 .orElseThrow(() -> new UnsupportedOperationException("Must use a concretely typed SqlArrayType here"));
         this.arrayType = arrayType;
     }
 
     @Override
     public Optional<SqlArrayType<?>> build(Type elementType, ConfigRegistry config) {
-        return this.elementType.equals(elementType)
+        return this.inferredElementType.equals(elementType)
```
stevenschlansker: redundant `this`

core/src/main/java/org/jdbi/v3/core/statement/Call.java
```diff
@@ -83,7 +83,7 @@ public Call registerOutParameter(String name, int sqlType, CallableStatementMapp
      */
     public OutParameters invoke() {
         try {
-            final PreparedStatement stmt = this.internalExecute();
+            this.internalExecute();
```
stevenschlansker: Another redundant `this`

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -222,11 +222,11 @@ public boolean isUnusedBindingAllowed() {
...
      *
      * @see org.jdbi.v3.core.argument.Argument
-     * @param allowUnusedBindings the new setting
+     * @param unusedBindingAllowed the new setting
      * @return this
      */
-    public SqlStatements setUnusedBindingAllowed(boolean allowUnusedBindings) {
-        this.allowUnusedBindings = allowUnusedBindings;
+    public SqlStatements setUnusedBindingAllowed(boolean unusedBindingAllowed) {
+        this.allowUnusedBindings = unusedBindingAllowed;
```
stevenschlansker: This is a good example of why I'm worried this kind of rule might be over-applied.  We're now permuting components of our terms just so that they are different and unique, but really they are the naming the same concept.

core/src/test/java/org/jdbi/v3/core/TestJdbi.java
```diff
@@ -104,10 +95,10 @@ public void testUseHandle() {
     }
 
     @Test
-    public void testGlobalStatementCustomizers() throws Exception {
+    public void testGlobalStatementCustomizers() {
         dbRule.getJdbi().addCustomizer(StatementCustomizers.maxRows(1));
 
-        handle = dbRule.openHandle();
+        Handle handle = dbRule.openHandle();
```
stevenschlansker: This change introduces a resource leak.

guava/src/test/java/org/jdbi/v3/guava/TestGuavaCollectors.java
```diff
@@ -56,12 +58,10 @@
...
     @Before
     public void addData() {
-        ImmutableList.Builder<Integer> expected = ImmutableList.builder();
-        for (int i = 0; i < 10; i++) {
-            dbRule.getSharedHandle().execute("insert into something(name, intValue) values (?, ?)", Integer.toString(i), i);
-            expected.add(i);
-        }
-        this.expected = expected.build();
+        expected = IntStream.range(0, 10)
+            .peek(i -> dbRule.getSharedHandle().execute("insert into something(name, intValue) values (?, ?)", Integer.toString(i), i))
```
stevenschlansker: Strange use of `peek` here

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
TheRealMarnes: Huh, I kinda wonder why this was renamed. It should pass through the `setter` exclusion.

TheRealMarnes:
So I'm guessing you handled some of your remarks manually in the merge commit?

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
TheRealMarnes: Ah, wait, of course. This parameter had to be changed because it didn't match the setter method name. The real problem here is that the setter doesn't match the field name, and this rule just forces the parameter to stick to the method instead of the field. I agree with your remark but it's applying to the symptom of a whole other problem here, not to the cause.

guava/src/test/java/org/jdbi/v3/guava/TestGuavaCollectors.java
TheRealMarnes: How so? `peek` is for performing an action on each value in the stream without altering the stream content or terminating it, like logging... or running a query.

stevenschlansker:
Yup, I fixed it on merge.

PR #1520, opened 2019-04-30
Support postgres JSON operators

### LINKED ISSUE #1519: Sql Script Parser treats #>> postgres operator as a comment
Sample:

```sql
SELECT b.json
FROM bar b
WHERE b.json #>> '{path,inside,the,file}' = 'some_value'
```

Yields a query like:

```sql
SELECT b.json
FROM bar b
WHERE b.json
```

Because the parser treats anything starting with `#` as a comment. In Postgres, this is a valid json operator.

haykinson:
Fixes #1519 by requiring `#comment` token to have a non-`>` character after the `#` character. Adds a test to make sure that comments continue to work while the JSON operator now starts to function.

Postgres JSON operators are described at https://www.postgresql.org/docs/11/functions-json.html

### DIFF
--- a/core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/SqlScriptLexer.g4
+++ b/core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/SqlScriptLexer.g4
@@ -17,8 +17,7 @@ lexer grammar SqlScriptLexer;
 COMMENT
     : '--' ~('\n'|'\r')* |
       '//' ~('\n'|'\r')* |
-      '#' ~('>'|'\r'|'\n') ~('\n'|'\r')* |
-      '#' ('\n'|'\r')
+      {_input.LA(2) != '>'}? '#' ~('\n'|'\r')* // Exception for Postgres #> and #>> JSON operators
      { skip(); }
     ;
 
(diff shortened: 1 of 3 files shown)


core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/SqlScriptLexer.g4
```diff
@@ -17,7 +17,8 @@ lexer grammar SqlScriptLexer;
 COMMENT
     : '--' ~('\n'|'\r')* |
       '//' ~('\n'|'\r')* |
-      '#'  ~('\n'|'\r')*
+      '#' ~('>'|'\r'|'\n') ~('\n'|'\r')* |
+      '#' ('\n'|'\r')
```
qualidafial: It's cleaner to have a single expression for `#` comments, and make it conditional on whether the next character is `>`:

```antlr
COMMENT
    : '--' ~('\n'|'\r')* |
      '//' ~('\n'|'\r')* |
      {_input.LA(2) != '>'}? '#' ~('\n'|'\r')* // Exception for Postgres #> and #>> JSON operators
     { skip(); }
    ;
```

haykinson: Thanks, that's a much more elegant way of doing it. Changed.

Issue #1523, opened 2019-05-01
ColonPrefixSqlParser not reused across queries


bezmax:
I'm investigating a heap dump of a very SQL intensive application (JDBI based) I have, and I see that there are just too many **unreachable** `org.antlr.v4.runtime.atn.LexerATNConfig` (and other antlr objects) sitting in the heap. They are eventually collected, but they are constantly polluting the memory. 

So I went into the source code, and it seems that this happens because the Lexer is executed every single time a query is executed. 

I tracked it down to `ColonPrefixSqlParser.java` never using it's cache and always calling the `internalParse` method. This happens because for each query I make a new ColonPrefixSqlParser is created. Which happens because the `SqlStatements` object is being created in a context of each `BaseStatement`, instead of being created in a global context.

A hotfix I'm using right now, is to call `jdbi.getConfig(SqlStatements.class)` immediately after creating the `Jdbi` object. This seems to be working, both the parser and it's cache is being reused across Jdbi instead of calling Lexer every single time. 

The long-term fix would be probably adding this inside of ConfigRegistry constructor:

```
    public ConfigRegistry() {
        configs.put(JdbiCaches.class, new JdbiCaches());
        configs.put(SqlStatements.class, new SqlStatements()); // <<< Add this
    }
```

I can create a PR for this, but I am not sure if this is actually a bug (maybe I'm not using Jdbi correctly). Also, I'm not too familiar with JDBI code to be sure that this change will not break anything.

Does this change make sense and would a PR like I am proposing make sense?

TheRealMarnes:
Does setting a single `ColonPrefixSqlParser` (or other parser) on the highest scope possible (your `Jdbi` instance) stop all the child scope instantiations?

bezmax:
@TheRealMarnes Just verified and yes it does.

TheRealMarnes:
True, but then I wonder if it wasn't built to work this way for a reason, and implementing this fix would re-break something that was fixed in the past.

qualidafial:
This result makes sense when you know the underlying implementation. Config objects aren't instantiated until they are first needed.

Looks like ColonPrefixSqlParser might benefit from using the new JdbiCache mechanism.

bezmax:
Yes, I thought about using JdbiCache, but given that it limits to 1 cache per class, it might become a problem when a parser itself is configurable.

For example:
```
jdbi.open().setSqlParser(new MyParser(config));
jdbi.open().setSqlParser(new DifferentParser());
[...]

PR #1525, opened 2019-05-02
Parsers use JdbiCaches instead of internal cache

bezmax:
Instead of relying on internal cache, the in-built parsers are able to utilize the global JdbiCaches to cache their parsed queries.

See discussion at https://github.com/jdbi/jdbi/issues/1523

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/ColonPrefixSqlParser.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/ColonPrefixSqlParser.java
@@ -37,10 +37,8 @@ import static org.jdbi.v3.core.internal.lexer.ColonStatementLexer.QUOTED_TEXT;
  * </p>
  */
 public class ColonPrefixSqlParser implements SqlParser {
-    private static final String CACHE_PREFIX = "ColonPrefixSqlParser:";
-
     private static final JdbiCache<String, ParsedSql> PARSED_SQL_CACHE =
-        JdbiCaches.declare(rawSql -> CACHE_PREFIX + rawSql, ColonPrefixSqlParser::internalParse);
+        JdbiCaches.declare(ColonPrefixSqlParser::internalParse);
 
     @Override
     public ParsedSql parse(String sql, StatementContext ctx) {
--- a/core/src/main/java/org/jdbi/v3/core/statement/ColonPrefixSqlParser.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/ColonPrefixSqlParser.java
@@ -13,12 +13,10 @@
  */
 package org.jdbi.v3.core.statement;
 
-import java.util.Collections;
-import java.util.Map;
-import java.util.WeakHashMap;
-
 import org.antlr.v4.runtime.CharStreams;
 import org.antlr.v4.runtime.Token;
+import org.jdbi.v3.core.config.JdbiCache;
+import org.jdbi.v3.core.config.JdbiCaches;
 import org.jdbi.v3.core.internal.lexer.ColonStatementLexer;
 import org.jdbi.v3.core.statement.internal.ErrorListener;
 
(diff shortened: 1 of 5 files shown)


bezmax:
I still wonder if this can't be optimized even further somehow. Even though the parsed queries will be properly cached now, I still do not like the fact that by default every query issued will trigger creation of a separate SqlStatements and separate ColonPrefixSqlParser. Although, maybe that's by design? The overhead should not be large.

qualidafial:
Let's wait on optimizing `SqlStatements` and `ColonPrefixSqlParser` until we have performance data that indicates they are a problem.

bezmax:
Here's also a screenshot to give you an idea of memory impact this fix resolves:

image

Apparently instances of Lexers are VERY memory intensive. Look at the unreachable objects in the heap:

image

And it seems like instantiation of the lexer is what's very memory intensive, not the parsing itself:

image

Which gives me another idea: **Should we also put the instance of Lexer itself into a separate cache?**

core/src/main/java/org/jdbi/v3/core/statement/ColonPrefixSqlParser.java
```diff
@@ -39,23 +37,22 @@
  * </p>
  */
 public class ColonPrefixSqlParser implements SqlParser {
-    private final Map<String, ParsedSql> cache = Collections.synchronizedMap(new WeakHashMap<>());
+    private static final String CACHE_PREFIX = "ColonPrefixSqlParser:";
+
+    private static final JdbiCache<String, ParsedSql> PARSED_SQL_CACHE =
+        JdbiCaches.declare(rawSql -> CACHE_PREFIX + rawSql, ColonPrefixSqlParser::internalParse);
```
qualidafial: You don't need this prefix. Each `JdbiCache` obtained through `JdbiCaches.declare()` is distinct and independent.

qualidafial:
Looks like `TestEscapedCharacters` needs a real `StatementContext` too

core/src/main/java/org/jdbi/v3/core/statement/ColonPrefixSqlParser.java
bezmax: You are right, I did not notice that the instance of JdbiCache itself is also used as a key in the global cache. Will fix in a second.

qualidafial:
Lexers are created with the input string as a parameter, so there's not much benefit to caching them for reuse

bezmax:
> Lexers are created with the input string as a parameter, so there's not much benefit to caching them for reuse

Rrright, I was kind of hoping that lexer can be reused, but now I see it can't.

> Looks like TestEscapedCharacters needs a real StatementContext too.

Yeah, fixing that too.

core/src/main/java/org/jdbi/v3/core/statement/HashPrefixSqlParser.java
```diff
@@ -36,23 +34,20 @@
...
+        JdbiCaches.declare(HashPrefixSqlParser::internalParse);
 
     @Override
     public ParsedSql parse(String sql, StatementContext ctx) {
-        try {
-            return cache.computeIfAbsent(sql, this::internalParse);
-        } catch (IllegalArgumentException e) {
-            throw new UnableToCreateStatementException("Exception parsing for named parameter replacement", e, ctx);
-        }
+        return PARSED_SQL_CACHE.get(sql, ctx);
```
qualidafial: The exception thrown by the ANTLR parser is unintuitive for most Jdbi users. Let's continue wrapping any parse exception in a Jdbi exception so parsing exceptions are easier to recognize.

bezmax:
@qualidafial Fixed. Although I tried writing a test for the wrapping of lexer exceptions but could not reproduce the `IllegalArgumentException` being thrown from inside the lexer. If you know the case when that is thrown - I can add that as a test.

qualidafial:
I think with the recent changes to the grammar definition, it should hopefully be much harder to produce parsing errors. :crossed_fingers:

qualidafial:
Thanks for the PR!

PR #1529, opened 2019-05-07
Add ResultIterable<T>.map(Function<T, U>) method

### LINKED ISSUE #1527: ResultIterable<T> to ResultIterable<V>
First off, I love jdbi3. I have been using it a while now because it is such a pleasure to use, so thank you very much for the library!

I am wondering if it has been considered to provide a method which maps a `ResultIterable<T>` to `ResultIterable<V>` perhaps given a `Function<T, V>`.

My use case is that I have a "low level" database interface that ideally would return a `ResultIterable<T>`.  Callers of that interface may want to transform that `T` into a different "higher level" type `V` without having to deal with `Stream<T>` (somewhat verbose via `withStream`, and alternatives have their flaws too).

I could go into the details about why it seems helpful, or maybe what the actual implementation would look like, but I first wanted to know if this had been considered or is infeasible for some reason.

qualidafial:
Fixes #1527

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
+++ b/core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
@@ -91,31 +91,28 @@ public interface ResultIterable<T> extends Iterable<T> {
      * @return the new ResultIterable
      */
     default <U> ResultIterable<U> map(Function<T, U> mapper) {
-        return () -> {
+        return () -> new ResultIterator<U>() {
             ResultIterator<T> delegate = iterator();
 
-            return new ResultIterator<U>() {
+            @Override
+            public boolean hasNext() {
+                return delegate.hasNext();
+            }
 
-                @Override
-                public boolean hasNext() {
-                    return delegate.hasNext();
-                }
+            @Override
+            public U next() {
+                return mapper.apply(delegate.next());
+            }
 
-                @Override
-                public U next() {
-                    return mapper.apply(delegate.next());
-                }
+            @Override
+            public StatementContext getContext() {
+                return delegate.getContext();
+            }
 
-                @Override
-                public StatementContext getContext() {
-                    return delegate.getContext();
-                }
-
-                @Override
-                public void close() {
-                    delegate.close();
-                }
-            };
+            @Override
+            public void close() {
+                delegate.close();
+            }
         };
     }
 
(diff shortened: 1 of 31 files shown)


qualidafial:
Build failing from bogus PMD error--wants me to name an instance field like a constant.

qualidafial:
Opened an issue with PMD: pmd/pmd#1812

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
```diff
@@ -81,6 +82,40 @@
...
 
+    /**
+     * Returns a {@code ResultIterable<U>} derived from this {@code ResultIterable<T>}, by
+     * transforming elements using the given mapper function.
+     *
+     * @param mapper function to apply to elements of this ResultIterable
+     * @param <U>    Element type of the returned ResultIterable
+     * @return the new ResultIterable
+     */
+    default <U> ResultIterable<U> map(Function<T, U> mapper) {
```
thekeenant: Would `Function<? super T, ? extends U>` be more appropriate? This is convenient for passing `Function`'s, or specifically method references, which are compatible with the `T` input and `U` output but are not strictly those types, no casting needed ([Stream#map](https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/java/util/stream/Stream.java#L184)).

qualidafial: Done

qualidafial:
Closed PMD issue--upgrading to latest PMD version resolved it.

qualidafial:
I think I may have inadvertently introduced an infinite recursion. Will resolve once I'm off the road

qualidafial:
@stevenschlansker Yep, the only thing we needed to change was where (and when) the delegate Iterable gets created.

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
```diff
@@ -81,6 +82,40 @@
...
+     * Returns a {@code ResultIterable<U>} derived from this {@code ResultIterable<T>}, by
+     * transforming elements using the given mapper function.
+     *
+     * @param mapper function to apply to elements of this ResultIterable
+     * @param <U>    Element type of the returned ResultIterable
+     * @return the new ResultIterable
+     */
+    default <U> ResultIterable<U> map(Function<? super T, ? extends U> mapper) {
+        return () -> new ResultIterator<U>() {
+            ResultIterator<T> delegate = iterator();
```
TheRealMarnes: should be `private final`, no? The return signature will prevent it from being accessed, but still, if it was a class in its own right we wouldn't allow it.

qualidafial: Yep. I'll push another change

qualidafial: Addressed in #1530

PR #1531, opened 2019-05-09
Immutables: respect `@Value.Default` when mapping nulled primitives

stevenschlansker:
also avoids NPE when trying to call a builder's setter when mapping reference types that are nullable in the db but nonnull with a default in Immutables-land.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoMapper.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoMapper.java
@@ -22,6 +22,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 
+import io.leangen.geantyref.GenericTypeReflector;
 import org.jdbi.v3.core.annotation.Unmappable;
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.generic.GenericTypes;
@@ -35,6 +36,7 @@ import org.jdbi.v3.core.mapper.reflect.ColumnNameMatcher;
 import org.jdbi.v3.core.mapper.reflect.ReflectionMappers;
 import org.jdbi.v3.core.mapper.reflect.internal.PojoProperties.PojoBuilder;
 import org.jdbi.v3.core.mapper.reflect.internal.PojoProperties.PojoProperty;
+import org.jdbi.v3.core.qualifier.QualifiedType;
 import org.jdbi.v3.core.result.UnableToProduceResultException;
 import org.jdbi.v3.core.statement.StatementContext;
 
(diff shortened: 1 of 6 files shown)


qualidafial:
3.9.0 isn't released yet, so this can go in 3.9.0 instead of 3.9.1

qualidafial:
Lol quick work

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoMapper.java
```diff
@@ -139,13 +141,19 @@ public T map(ResultSet rs, StatementContext ctx) throws SQLException {
 
                 Object value = mapper.map(r, ctx);
 
-                pojo.set(property, value);
+                if (value != null) {
+                    pojo.set(property, value);
+                }
```
qualidafial: This makes me a little nervous. A SQL `NULL` might be an intentional override of the default value.

This is probably safe assuming the qualified type is a primitive and the value is null. Sometimes it really stinks that Immutables annotations don't have runtime retention.

I think I can live with this if we explicitly document this behavior in the developer guide.

stevenschlansker: Generally, I agree with your concern.  Specifically, though, Immutables eagerly disallows calling `set` with a null value.  An example generated builder method:

```java
    @CanIgnoreReturnValue 
    @JsonProperty("name")
    public final Builder name(String name) {
      this.name = Objects.requireNonNull(name, "name");
      initBits &= ~INIT_BIT_NAME;
      return this;
    }
```

If you use `Optional<T>`, the value will be `empty()` not `null`, so the builder setter will be called.
I will see if we can detect the `@Nullable` annotation to handle that case as well.

qualidafial: Does Immutables even let you combine `@Nullable` with `@Value.Default`? If not then we don't have to bother with that extra check.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoMapper.java
```diff
@@ -139,13 +141,19 @@ public T map(ResultSet rs, StatementContext ctx) throws SQLException {
...
             }
 
             return pojo.build();
         });
     }
 
+    private static QualifiedType<?> box(QualifiedType<?> qualifiedType) {
+        return qualifiedType.mapType(t -> Optional.of(GenericTypeReflector.box(t))).get();
```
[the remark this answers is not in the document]
stevenschlansker: This highligted a slightly mis-named `QualifiedType.mapType`.  I'm going to take the liberty of fixing it...

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoMapper.java
stevenschlansker: I believe it does, but I'm going to KISS and not add the check until someone complains about the behavior.

Issue #1404, opened 2019-01-13
Optional (nullable) @Nested one-to-one mappings


xak2000:
Is there a way to make `BeanMapper` to map `@Nested` bean to `null` if there is `LEFT JOIN` present in the query, but related (joined) row is absent in the DB?

I found this in [release notes](https://github.com/jdbi/jdbi/blob/master/RELEASE_NOTES)

```
    - Optional mapping of @Nested objects when using BeanMapper, ConstructorMapper, FieldMapper, or
      KotlinMapper. @Nested objects are only mapped when the result set contains columns that match
      the nested object.
```

But what does mean `the result set contains columns that match the nested object`? I think this means that if `LEFT JOIN` with at least one suitable column is present in the query, then nested object will be mapped anyway, even if actually no related row exist in the DB?

This leads to empty nested objects filled with nulls.

For example, there are two beans:

```
class Account {
  Integer id;
  User user;

  @Nested("u")
  void setUser(User user) {
    this.user = user;
  }
}

class User {
  String name;
}
```

DB structure:
```
CREATE TABLE account (
  id INTEGER PRIMARY KEY
)

CREATE TABLE user (
  account_id INTEGER PRIMARY KEY REFERENCES account(id),
  name VARCHAR(100),
)
```

And the query:

```
SELECT a.id, u.name u_name
FROM account a
LEFT JOIN user u ON u.account_id = a.id
WHERE a.id = :accountId
```

So, the `Account` to `User` is optional `one-to-one` relation, where `Account` can have a `User` or not have a `User`.

Using this mappings and query, `BeanMapper` will always create a `User` bean, even if there is no `user` record found in the DB for this `account`. So we end up with `Account[id=123, user=User[name=null]]`.

What I want to achieve in this case is to end up with `Account[id=123, user=null]`.

**Is there are way to achive it?**

If not, maybe it can be a feature-request then. I don't see any reliable way to make sure that related row doesn't exist in the DB when using `LEFT JOIN` like this, but we can add new parameter to `@Nested` annotation, like `presenceColumn`, which sets a column name in result set, that must have `non-null` value to make `BeanMapper` consider the nested bean as "present". For example, it can be `user.account_id` column from example above. If `user.account_id` is null, then assuredly there is no related `User` record exist (because `account_id` is FK).

So, our query becomes:

```
SELECT a.id, u.name u_name, u.account_id u_account_id
FROM account a
LEFT JOIN user u ON u.account_id = a.id
[...]

PR #1535, opened 2019-05-23
PropagateNull annotation for BeanMapper, ConstructorMapper, and FieldMapper

jpmeyer:
Adds a new annotation called PropagateNull as a possible solution to nested null objects as in #1404. I initially attempted to use a wrapper that maps the underlying mapper, but this was cumbersome because of the weird result for the column names in any annotations that operated on the parent context. Moving the annotation to the child allows access to parameters in specialize0 and significantly simplifies the code changes. 

I also added the ugly wasNull check and primitive checks to allow propagating null for an entire object if a primitive mapped value is null. Happy to see any changes from contributors if the approach is generally agreed upon.


stevenschlansker:
Released in `3.9.0`

PR #1541, opened 2019-06-06
PreparedBatch: allow empty batches

stevenschlansker:
This mirrors the behavior of normal `Batch`, and eliminates an error case
that is already otherwise correctly handled (by returning a no-element ResultBearing.)


stevenschlansker:
Build failure seems to be related to kotlin docs needing an unavailable network dependency :/

PR #1544, opened 2019-06-06
Allow configuring StatementException message generation

### LINKED ISSUE #1532: Truncated long Exception messages
It would be nice to have an option to truncate exception messages. Currently, it seems the entire statement is in there, which on huge batch operations leads to multi-megabyte log lines, which is obviously problematic for a variety of reasons.

I could manually wrap all exception handling in my own exception that truncates the message, but it seems like this would something people would want to be configurable on a library level.

stevenschlansker:
Fixes #1532

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/StatementExceptions.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/StatementExceptions.java
@@ -127,7 +127,7 @@ public class StatementExceptions implements JdbiConfig<StatementExceptions> {
         public String apply(StatementException exc) {
             final StatementContext ctx = exc.getStatementContext();
             if (ctx == null) {
-                return NONE.render(exc, ctx);
+                return NONE.render(exc, null);
             }
             return render(exc, ctx);
         }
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/jdbi/v3/core/statement/StatementExceptions.java
```diff
@@ -0,0 +1,142 @@
...
+        @Override
+        public String apply(StatementException exc) {
+            final StatementContext ctx = exc.getStatementContext();
+            if (ctx == null) {
+                return NONE.apply0(exc, ctx);
+            }
+            return apply0(exc, ctx);
+        }
+
+        protected abstract String apply0(StatementException exc, StatementContext ctx);
```
TheRealMarnes: not a fan of naming it something abstract like `apply0`

TheRealMarnes:
Hang on. Is the new default behavior equivalent to the original behavior? I think the shortening should be opt-in, as in motivated by "our logs are getting too big", to avoid "we logged an exception but apparently jdbi cuts the text off now so we don't know anything".

stevenschlansker:
I think the new behavior is a more sensible default.  The old behavior prints the statement three times over when usually once is enough, and 1024 bytes is a pretty generous starting limitation.

PR #1548, opened 2019-06-23
POC to discuss/refine: @Nonnull mapper

TheRealMarnes:
Light-hearted idea I had as a fun concept, but I could see it having value for users who like declarative approaches, and to serve as an example of qualified factories.

If the `NonnullMapper` idea is shot down, we should at least salvage the API changes so users can actually start using qualified mappers and arguments...

One of the new `register` methods created an API ambiguity:
image
Changing this to what it should be (`registerArgument(new HStoreArgumentFactory())`) causes test failures. Smells like a fishy design to me but I'm tapped out to tackle it for the moment.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/NonnullColumnMapperFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/NonnullColumnMapperFactory.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.mapper;
+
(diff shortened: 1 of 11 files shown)


TheRealMarnes:
I decided to write an sqlobject test to illustrate its usefulness there, and realized the way qualifiers work there undermines the practicality of this feature... We'd need to have our own `@Nonnull`-like annotation to be able to mark it as `@Qualifier` :/

This kinda feels like a logical gap in our qualifiers system: the fluent api can apply any annotation as a qualifier, but sqlobject can only do that for `@Qualifier` annotations. Maybe we should add an annotation like `@Qualified({Nonnull.class})` so that sqlobject can assign any (library) annotation as a qualifier just like the fliuent api can?

core/src/main/java/org/jdbi/v3/core/mapper/NonnullColumnMapperFactory.java
```diff
@@ -0,0 +1,65 @@
...
+ * and the mapped value is checked, throwing {@link NullPointerException} if null.
+ *
+ * This allows you to query for any column type and make {@code null} explicitly forbidden as a result,
+ * for when declarative programming is your cup of tea.
+ */
+@Beta
+public class NonnullColumnMapperFactory implements QualifiedColumnMapperFactory {
+    @Override
+    public Optional<ColumnMapper<?>> build(QualifiedType<?> type, ConfigRegistry config) {
+        return Stream.of(Nonnull.class)
```
TheRealMarnes: this construction was meant to support multiple similar annotations, not sure to what extent that goal should be pursued

TheRealMarnes:
withdrawing to break it up

PR #1549, opened 2019-06-23
Let user configure if db null => java primitive default value or exception

### LINKED ISSUE #1270: Allow user to configure if db nulls should translate to exceptions or defaults when mapped to java primitives
From #1268

TheRealMarnes:
fixes #1270 

I've been wanting to patch this since day 1 of using Jdbi because I don't like stray/default values sneaking into my resultsets as a result of a bug producing nulls where there shouldn't be any... It doesn't really make sense to just assume one wants a null to become an arbitrary-ish value when explicitly working with types that don't support nulls.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
@@ -34,6 +34,7 @@ import org.jdbi.v3.meta.Beta;
 public class ColumnMappers implements JdbiConfig<ColumnMappers> {
     private final List<QualifiedColumnMapperFactory> factories = new CopyOnWriteArrayList<>();
     private final ConcurrentHashMap<QualifiedType<?>, ColumnMapper<?>> cache = new ConcurrentHashMap<>();
+    private boolean nullPrimitivesToDefaults = true;
     private ConfigRegistry registry;
 
     public ColumnMappers() {
@@ -58,6 +59,7 @@ public class ColumnMappers implements JdbiConfig<ColumnMappers> {
     private ColumnMappers(ColumnMappers that) {
         factories.addAll(that.factories);
         cache.putAll(that.cache);
+        nullPrimitivesToDefaults = that.nullPrimitivesToDefaults;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


vavr/src/main/java/org/jdbi/v3/vavr/TupleMappers.java
```diff
@@ -30,12 +33,13 @@
...
     private ConfigRegistry registry;
 
-    private String[] columns = new String[Tuple.MAX_ARITY];
+    private List<String> columns = Arrays.asList(new String[Tuple.MAX_ARITY]);
 
     public TupleMappers() {}
 
     private TupleMappers(TupleMappers that) {
-        System.arraycopy(that.columns, 0, this.columns, 0, Tuple.MAX_ARITY);
+        columns = Arrays.asList(that.columns.toArray(new String[0]));
```
TheRealMarnes: the `List` from `asList` is fixed-size

core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
```diff
@@ -195,6 +197,17 @@ private ColumnMappers register(QualifiedColumnMapperFactory factory) {
         return mapper;
     }
 
+    /**
+     * @return {@code true} if database {@code null}s should translate to the Java defaults for primitives, or throw an exception otherwise
+     */
```
qualidafial: Can we add a sentence that the default setting is `true`?

core/src/main/java/org/jdbi/v3/core/mapper/PrimitiveMapperFactory.java
```diff
@@ -59,7 +60,14 @@
     }
 
     private static <T> ColumnMapper<T> primitiveMapper(ColumnGetter<T> getter) {
-        return (r, i, ctx) -> getter.get(r, i);
+        return (r, i, ctx) -> {
+            T value = getter.get(r, i);
+            if (r.wasNull() && !ctx.getConfig(ColumnMappers.class).getNullPrimitivesToDefaults()) {
+                throw new UnableToProduceResultException("a Java primitive value was attempted to be mapped from a database null value");
```
qualidafial: If there are multiple primitives columns, users won't be able to tell which column failed to map from this error message. We should provide more information about which column failed to be mapped. Maybe include the column number (and column label from the result set metadata)?

We may also want to include a suggestion in the error message to change the mapped type from a primitive to a wrapper.

TheRealMarnes:
remarks handled

stevenschlansker:
Someday maybe this will be the default

PR #1551, opened 2019-07-03
Boyscouting

TheRealMarnes:
We have a rule at work that boils down to collaterally refactoring/cleaning up any code one comes across during work on a feature. This PR contains minor fixups I've gathered left and right over time, of beauty flaws that bothered me every time I saw them.

None of these are a functional change, and they don't set any kind of precedent regarding code style or something, I can't think of any reasonable objections. These are all just basically pulling some deviating code in line with the rest of our code: declaration order, assignment location, consistent structure of classes of the same kind, that kind of thing.

Micro spring cleaning :)


core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -32,24 +32,19 @@
...
 public final class SqlStatements implements JdbiConfig<SqlStatements> {
 
-    private final Map<String, Object> attributes;
-    private TemplateEngine templateEngine;
-    private SqlParser sqlParser;
-    private SqlLogger sqlLogger;
-    private Integer queryTimeout;
-    private boolean allowUnusedBindings;
+    // cannot be a ConcurrentHashMap because of null keys
+    private final Map<String, Object> attributes = Collections.synchronizedMap(new HashMap<>());
```
qualidafial: Exhibit A for thread safety of config being a poor API choice. Make everything slower "just in case."

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -32,24 +32,19 @@
...
+    private TemplateEngine templateEngine = new DefinedAttributeTemplateEngine();
+    private SqlParser sqlParser = new ColonPrefixSqlParser();
+    private SqlLogger sqlLogger = SqlLogger.NOP_SQL_LOGGER;
+    private Integer queryTimeout = null;
+    private boolean allowUnusedBindings = false;
```
qualidafial: Initializing these fields here creates a bunch of garbage objects that get immediately clobbered when the copy constructor is used.

I actually prefer the symmetry of having statements in the default and copy constructors that you can compare. Makes it easy to see when you've forgotten to copy a field in the copy constructor.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/BindTime.java
```diff
@@ -34,18 +32,7 @@
     class Factory implements SqlStatementCustomizerFactory {
         @Override
         public SqlStatementCustomizer createForType(Annotation annotation, Class<?> sqlObjectType) {
-            return stmt -> stmt.bind("now", OffsetDateTime.now(stmt.getConfig(Config.class).clock));
-        }
-    }
-
-    class Config implements JdbiConfig<Config> {
```
qualidafial: This is a breaking API change--inner classes of interfaces are implicitly public.

core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
```diff
@@ -66,6 +66,7 @@ public void setRegistry(ConfigRegistry registry) {
     private Arguments(Arguments that) {
         factories.addAll(that.factories);
         untypedNullArgument = that.untypedNullArgument;
+        registry = null;
```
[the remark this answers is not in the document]
TheRealMarnes: What you do manually/on sight, I let my IDE do for me:
image

Don't set `registry` to null -> you learn to ignore the "missing a field" warning, so when you add a fiield and really do forget to copy it, you won't notice. Shooting yourself in the foot twice.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
```diff
@@ -56,7 +54,10 @@ private ReflectionMappers(ReflectionMappers that) {
      * @return this
      */
     public ReflectionMappers setColumnNameMatchers(List<ColumnNameMatcher> columnNameMatchers) {
-        this.columnNameMatchers = new ArrayList<>(columnNameMatchers);
+        synchronized (lock) {
+            this.columnNameMatchers.clear();
+            this.columnNameMatchers.addAll(columnNameMatchers);
+        }
```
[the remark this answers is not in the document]
TheRealMarnes: I recall thread-safety being a concern in Config classes, as shown by the use of concurrent collections everywhere. That's why I went for this.

But I agree it's probably safe to leave out.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoTypes.java
```diff
@@ -14,22 +14,23 @@
...
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.config.JdbiConfig;
 import org.jdbi.v3.core.generic.GenericTypes;
 
 public class PojoTypes implements JdbiConfig<PojoTypes> {
-    private final Map<Class<?>, ImmutablesPropertiesFactory> factories = new HashMap<>();
+    private final Map<Class<?>, ImmutablesPropertiesFactory> factories = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
TheRealMarnes: Until that day though, I'd rather we consistently do this half-assed solution than half-assing a half-ass solution.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
TheRealMarnes: Agreeable. Compactness vs constructor symmetry, either way works for me.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/BindTime.java
TheRealMarnes: This is in `test` of `sqlobject`

core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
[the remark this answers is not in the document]
TheRealMarnes: Functionally pointless as it may be, forcing a copy constructor to deal with all fields to live up to its design concept adds clarity to code, and allows IDEs to be more helpful.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoTypes.java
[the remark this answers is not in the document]
qualidafial: so no quarter-assing?

sqlobject/src/test/java/org/jdbi/v3/sqlobject/BindTime.java
qualidafial: whoops! carry on

core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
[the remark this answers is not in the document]
qualidafial: Since registry should only be set in the `setRegistry()` method, I think it's ok to explicitly set it to null on the field initializer.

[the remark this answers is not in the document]
TheRealMarnes: I can leave it as defult (null), but the copy constructor should assign null to avoid a false-positive "copy constructor misses a field", which is a helpful warning I'd prefer to enable.

TheRealMarnes:
remarks handled

TheRealMarnes:
Included a collateral fix: 
image

core/src/main/java/org/jdbi/v3/core/statement/StatementExceptions.java
```diff
@@ -91,7 +95,7 @@ public String render(StatementException exc, StatementContext ctx) {
         PARAMETERS {
             @Override
             public String render(StatementException exc, StatementContext ctx) {
-                return String.format("%s [arguments:%s]", ctx.getBinding());
+                return String.format("%s [arguments:%s]", exc.getShortMessage(), ctx.getBinding());
```
TheRealMarnes: image

core/src/main/java/org/jdbi/v3/core/Handles.java
```diff
@@ -19,12 +19,14 @@
  * Configuration class for handles.
  */
 public class Handles implements JdbiConfig<Handles> {
-    private boolean forceEndTransactions = true;
```
stevenschlansker: I'm not sure why this is better.  Why are we moving initializers away from the declaration?  It's nice to know the initial value right away, and means you don't have to repeat it in every constructor.

core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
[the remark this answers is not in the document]
stevenschlansker: I'm rather -1 to pointless null initializations.  It's hard to see it as anything but noise in the code.

core/src/main/java/org/jdbi/v3/core/extension/Extensions.java
```diff
@@ -24,19 +24,21 @@
...
 public class Extensions implements JdbiConfig<Extensions> {
-    private final List<ExtensionFactory> factories = new CopyOnWriteArrayList<>();
+    private final List<ExtensionFactory> factories;
 
     /**
      * Create an empty {@link ExtensionFactory} configuration.
      */
-    public Extensions() {}
+    public Extensions() {
+        factories = new CopyOnWriteArrayList<>();
```
stevenschlansker: Again, now this initializer will have to be repeated in any additional constructors added.

core/src/main/java/org/jdbi/v3/core/mapper/MapEntryMappers.java
```diff
@@ -19,16 +19,19 @@
  * Configuration class for MapEntryMapper.
  */
 public class MapEntryMappers implements JdbiConfig<MapEntryMappers>, MapEntryConfig<MapEntryMappers> {
-    public MapEntryMappers() {}
+    private String keyColumn;
+    private String valueColumn;
+
+    public MapEntryMappers() {
```
stevenschlansker: This constructor does nothing.

core/src/main/java/org/jdbi/v3/core/mapper/MapMappers.java
```diff
@@ -24,6 +24,7 @@
     private UnaryOperator<String> caseChange;
 
     public MapMappers() {
+        // TODO move to DefaultsPlugin because this is opinionated behavior
```
stevenschlansker: I'm not sure what this comment means.  There has to be some default behavior without any plugins installed, and we're not likely to change it.  What would the default behavior be, if not the currently existing "lowercase"?

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -32,30 +32,33 @@
...
         templateEngine = new DefinedAttributeTemplateEngine();
         sqlParser = new ColonPrefixSqlParser();
         sqlLogger = SqlLogger.NOP_SQL_LOGGER;
         queryTimeout = null;
+        allowUnusedBindings = false;
+        customizers = new CopyOnWriteArrayList<>();
     }
 
     private SqlStatements(SqlStatements that) {
-        this.attributes = Collections.synchronizedMap(that.getAttributes()); // already copied
```
stevenschlansker: Being more defensive here does make sense.

core/src/main/java/org/jdbi/v3/core/statement/StatementExceptions.java
stevenschlansker: Nice catch

core/src/main/java/org/jdbi/v3/core/argument/Arguments.java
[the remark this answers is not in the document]
TheRealMarnes: So I'm getting a vibe from both of you that you don't really care about copy constructors violating the contract of a copy constructor due to implementation details then

core/src/main/java/org/jdbi/v3/core/Handles.java
TheRealMarnes: I was with you on this (initialize-to-field for clarity), but @qualidafial made an argument for symmetric constructors

core/src/main/java/org/jdbi/v3/core/extension/Extensions.java
TheRealMarnes: see symmetric constructors argument. There isn't really any prospect of having more than a default and copy constructor to Config classes

core/src/main/java/org/jdbi/v3/core/mapper/MapEntryMappers.java
TheRealMarnes: also symmetric constructors

core/src/main/java/org/jdbi/v3/core/mapper/MapMappers.java
TheRealMarnes: The code-default ought to be NOP, no case-change. That stuff gets lowercased when put into a Map is a jdbi opinion. Sure surprised me when I used this feature for the first time, I was expecting the map to contain my columns with the name I had given them.

TheRealMarnes:
You really want even less content to a PR than this? Seems an awful lot of noise in PRs for almost no code...

So the disagreement I feel is between you and @qualidafial about symmetric constructors, and neither of you seems interested in strictly satisfying the contract of a copy constructor to silence the warning noise. Can't fix the latter without agreement on the former. I'm personally more of a fan of initialize-to-field for brevity, leaving only an empty default and a (partial, I guess) copy constructor. Can we agree on going with those then?

As for bikeshedding, well, cleanups gotta happen, and you don't want irrelevant refactoring included with a meaningful PR about a feature, so a dedicated cleanup PR seems the only option then.

TheRealMarnes:
That would prevent one of the mishaps I fixed in one of the commits, where a field was assigned different types of collections depending on which constructor or setter was called... I'm open to subverting the "declare interfaces" rule for this case, until we drop the pretense of thread safety in config classes.

leaumar:
Controversial changes removed so we can at least get the agreeable fixes on board

PR #1553, opened 2019-07-05
Fix NPE

28Smiles:
fix: NPE if @DefineList annotated parameter is supplied with an ImmutableList, see https://docs.oracle.com/javase/8/docs/api/java/util/List.html#contains-java.lang.Object-


TheRealMarnes:
Do you feel like you could add a test to verify future changes don't break this again?

Also, man, that's nasty coming from the JDK ._.

28Smiles:
There is no list in Java 8 I know implementing this behavior and since the language level is 8, it maybe should be concidered using language level 11 or 12 for tests

qualidafial:
The spec on contains allows an NPE if the collection is null intolerant.

The spec is fucking dumb in this case, but we should fix it on our end.

PR #1554, opened 2019-07-08
Prettify truncated StatementExceptionS to signal modifications

Gaibhne:
I think it's important to make it clear to a casual reader of the logs that a logged statement was truncated. Due to the nature of SQL statemenets, this may not immediately be obvious, for example a statement cut off (at the `|`) like this: "`SELECT ... FROM ...| WHERE ...`" would still be a valid SQL statement.

Adding a "`[...]`" to the end of the Statement if anything is cut off makes this more evident.

The following test case statements demonstrate the result. I didn't think those were worth adding to the PR as code:

```
String exception = "This is a very very long exception that I use for testing";

assertEquals("[...]",
		StatementExceptions.limit(exception, 0));
assertEquals("This is a [...]",
		StatementExceptions.limit(exception, 10));
assertEquals("This is a very very [...]",
		StatementExceptions.limit(exception, 20));
assertEquals("This is a very very long exception that I use for testin[...]",
		StatementExceptions.limit(exception, 56));
assertEquals("This is a very very long exception that I use for testing",
		StatementExceptions.limit(exception, 57));
assertEquals("This is a very very long exception that I use for testing",
		StatementExceptions.limit(exception, 60));
```

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/StatementExceptions.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/StatementExceptions.java
@@ -135,7 +135,13 @@ public class StatementExceptions implements JdbiConfig<StatementExceptions> {
         protected abstract String render(StatementException exc, StatementContext ctx);
     }
 
-    private static String limit(String s, int len) {
-        return s.substring(0, Math.min(len, s.length()));
+    @SuppressWarnings("PMD.UseStringBufferForStringAppends")
+    protected static String limit(String s, int len) {
+        String truncated = s.substring(0, Math.min(len, s.length()));
+        boolean isTruncated = len < s.length();
+        if (isTruncated) {
+            truncated += "[...]";
+        }
+        return truncated;
     }
 }
--- a/core/src/test/java/org/jdbi/v3/core/statement/TestStatementExceptions.java
+++ b/core/src/test/java/org/jdbi/v3/core/statement/TestStatementExceptions.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


TheRealMarnes:
> I didn't think those were worth adding to the PR

i believe we disagree, please include them :) We strive for maximum test coverage, even in soft requirements. What you might do is relax the constraints, just specify that the message must (not) `contain [...]`.

TheRealMarnes:
Oh and also, please use assertj with semantic assertions: `assertThat(msg).contains(str)`

core/src/main/java/org/jdbi/v3/core/statement/StatementExceptions.java
```diff
@@ -136,6 +136,6 @@ public String apply(StatementException exc) {
     }
 
     private static String limit(String s, int len) {
-        return s.substring(0, Math.min(len, s.length()));
+        return s.substring(0, Math.min(len, s.length())) + (s.length() > len ? "[...]" : "");
```
TheRealMarnes: Call me a nitpick, but I'm wondering if it might be worth doing this more verbosely, to stay in line with the clarity of code we generally try to achieve by keeping the number of operations per line low?

```
limited = limit(...);
isTruncated = limited.length < str.length;
if (isTruncated) {limit = limit + "[...]"};
```

Gaibhne: Splitting it into multiple lines I agree with, but adding another method seems overkill and overly verbose, especially considering that both parts contextually are part of the whole idea of 'truncating/limiting'.

TheRealMarnes: oh yes, sorry, the `limit` part should be `substring`. No new method of course, just splitting it into lines and a variable/condition :)

Gaibhne:
It seemed cleaner to collapse the two code commits into one and force push it while it's still a PR. I added the `@SuppressWarnings` because it really didn't seem conductive to readability to convert that teensy bit of code to using a `StringBuilder` as PMD requested.

TheRealMarnes:
I'm gonna go ahead and merge this

Thanks @Gaibhne

qualidafial:
@Gaibhne Thanks for contributing!

Issue #1558, opened 2019-07-15
Customizable enum mapping (similar to Jackson’s `@JsonValue`)


maffe:
My database stores values I’d like to map to `enum`, however they differ slightly from the `enum` `name()`s. Currently I have to write a `ColumnMapper` to map them correctly. It would be nice to use annotations instead.

Example:
Database contains values `CREATE` and `UPD`
Java `enum` consists of `CREATE` and `UPDATE`

Suggestion: Provide annotations handled by the default enum mappers. For example, the `enum` could look like this:
```
enum Action {
  CREATE,
  @DatabaseValue("UPD")
  UPDATE
}
```
This is similar to Jackson which provides `@JsonValue` to map `enum`s from and to JSON.

TheRealMarnes:
I'm gonna assume you would have this take effect only when your `EnumStrategy` is `BY_NAME`?

TheRealMarnes:
I'm neutral on adding this as a feature... Could give it a quick go if someone else feels this would be added value.

TheRealMarnes:
You could always PR it with some tests by the way :) Might convince more people.

maffe:
Pull request: #1561

PR #1561, opened 2019-07-23
Customizable enum mapping (similar to Jackson’s @JsonValue)

maffe:
As requested in #1558

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
@@ -59,9 +59,9 @@ class EnumArgumentFactory implements QualifiedArgumentFactory {
                 throw new RuntimeException(ex);
             }
             final DatabaseValue databaseValue = field.getAnnotation(DatabaseValue.class);
-            return databaseValue != null
-                    ? databaseValue.value()
-                    : e.name();
+            return databaseValue == null
+                    ? e.name()
+                    : databaseValue.value();
         }, config);
     }
 
--- a/core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
@@ -24,6 +24,7 @@ import org.jdbi.v3.core.enums.DatabaseValue;
 import org.jdbi.v3.core.enums.EnumStrategy;
 import org.jdbi.v3.core.generic.GenericTypes;
 import org.jdbi.v3.core.internal.EnumStrategies;
+import org.jdbi.v3.core.internal.exceptions.Unchecked;
 import org.jdbi.v3.core.qualifier.QualifiedType;
 
 class EnumArgumentFactory implements QualifiedArgumentFactory {
@@ -51,18 +52,14 @@ class EnumArgumentFactory implements QualifiedArgumentFactory {
     }
 
     private static <E extends Enum<E>> Optional<Argument> byName(E value, ConfigRegistry config) {
-        return makeArgument(Types.VARCHAR, String.class, value, e -> {
-            final Field field;
-            try {
-                field = e.getDeclaringClass().getField(e.name());
-            } catch (final NoSuchFieldException ex) {
-                throw new RuntimeException(ex);
-            }
+        final Function<E, String> transform = e -> {
+            final Field field = Unchecked.function(e.getDeclaringClass()::getField).apply(e.name());
             final DatabaseValue databaseValue = field.getAnnotation(DatabaseValue.class);
             return databaseValue == null
                     ? e.name()
                     : databaseValue.value();
-        }, config);
+        };
+        return makeArgument(Types.VARCHAR, String.class, value, transform, config);
     }
 
     private static <E extends Enum<E>> Optional<Argument> byOrdinal(E value, ConfigRegistry config) {
--- a/core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
@@ -13,12 +13,14 @@
  */
 package org.jdbi.v3.core.argument;
 
+import java.lang.reflect.Field;
 import java.lang.reflect.Type;
 import java.sql.Types;
 import java.util.Optional;
 import java.util.function.Function;
 
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.enums.DatabaseValue;
 import org.jdbi.v3.core.enums.EnumStrategy;
 import org.jdbi.v3.core.generic.GenericTypes;
 import org.jdbi.v3.core.internal.EnumStrategies;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
```diff
@@ -49,7 +51,18 @@
     }
 
     private static <E extends Enum<E>> Optional<Argument> byName(E value, ConfigRegistry config) {
-        return makeArgument(Types.VARCHAR, String.class, value, E::name, config);
+        return makeArgument(Types.VARCHAR, String.class, value, e -> {
+            final Field field;
+            try {
+                field = e.getDeclaringClass().getField(e.name());
+            } catch (final NoSuchFieldException ex) {
+                throw new RuntimeException(ex);
```
TheRealMarnes: please use our `Unchecked.function` for this sort of thing (inspired by jool, if you're familiar with that). I assume both methods `throw`, so you'll have to `uncheck` a lambda rather than the preferred method ref.

core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
```diff
@@ -49,7 +51,18 @@
     }
 
     private static <E extends Enum<E>> Optional<Argument> byName(E value, ConfigRegistry config) {
-        return makeArgument(Types.VARCHAR, String.class, value, E::name, config);
+        return makeArgument(Types.VARCHAR, String.class, value, e -> {
```
TheRealMarnes: this may be more suited to a private method, being such a big lambda in an already long line of code

core/src/main/java/org/jdbi/v3/core/enums/DatabaseValue.java
```diff
@@ -0,0 +1,30 @@
...
+import java.lang.annotation.Target;
+import org.jdbi.v3.core.qualifier.Qualifier;
+import org.jdbi.v3.meta.Beta;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+@Qualifier
+@Beta
+public @interface DatabaseValue {
+
```
TheRealMarnes: absolute nitpick: empty line below class header. Ignore this comment if you feel it's a common thing to do, I'm just assuming my personal preference of not leaving a blank is the more common style. But then I think there should be an empty line before the class end as well.

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
```diff
@@ -78,6 +80,16 @@ public E map(ResultSet rs, int columnNumber, StatementContext ctx) throws SQLExc
         private static Object getValueByName(Class<? extends Enum<?>> enumClass, String name) {
             final Enum<?>[] enumConstants = enumClass.getEnumConstants();
             return JdbiOptionals.findFirstPresent(
+                    () -> Arrays.stream(enumConstants).filter(e -> {
+                        final Field field;
+                        try {
+                            field = enumClass.getField(e.name());
+                        } catch (final NoSuchFieldException ex) {
+                            throw new RuntimeException(ex);
```
TheRealMarnes: same as before: please use `Unchecked` and consider a private method ref instead of a lambda

core/src/test/java/org/jdbi/v3/core/EnumsConfigTest.java
```diff
@@ -70,6 +71,39 @@ public void namesAreMappedCorrectly() {
...
+                    .bind("name", Foobar.CUSTOM)
+                    .execute();
+
+            String ordinal = h.createQuery("select name from enums")
+                    .mapTo(String.class)
+                    .one();
+
+            assertThat(ordinal)
+                    .isEqualTo(Foobar.class.getField(Foobar.CUSTOM.name())
+                            .getAnnotation(DatabaseValue.class).value());
```
TheRealMarnes: I think it's better for a test to be explicit about expected values rather than using calculated values where possible. This test reads more like `assert impl() == impl()` than it should.

core/src/test/java/org/jdbi/v3/core/EnumsConfigTest.java
```diff
@@ -70,6 +71,39 @@ public void namesAreMappedCorrectly() {
...
+        });
+    }
+
+    @Test
+    public void customizedNamesAreMappedCorrectly() throws NoSuchFieldException {
+        db.getJdbi().useHandle(h -> {
+            Foobar name = h.createQuery("select :name")
+                    .bind("name", Foobar.class.getField(Foobar.CUSTOM.name())
```
TheRealMarnes: idem

core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
maffe: That’s what I did first, but then the extra `.get()` on the Supplier annoyed me. Is it okay to extend `Unchecked` so it directly returns the `CheckedCallable`’s result?

TheRealMarnes: superfluous-looking `get` calls on `Unchecked` usually means you should actually be using `function` rather than `supplier`

core/src/test/java/org/jdbi/v3/core/EnumsConfigTest.java
maffe: Right. I did it like this because the two methods the new tests are based on looked quite dynamic as well.

TheRealMarnes: I don't really follow. Do you object to changing it to `isEqualTo("CUST")`?

core/src/main/java/org/jdbi/v3/core/enums/DatabaseValue.java
maffe: This is what my NetBeans does when using autoformat ¯\_(ツ)_/¯ I’ll remove it to match the rest of Jdbi.

core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
maffe: I moved the lambda expression out of the method call.

core/src/main/java/org/jdbi/v3/core/argument/EnumArgumentFactory.java
```diff
@@ -49,7 +52,14 @@
     }
 
     private static <E extends Enum<E>> Optional<Argument> byName(E value, ConfigRegistry config) {
-        return makeArgument(Types.VARCHAR, String.class, value, E::name, config);
+        final Function<E, String> transform = e -> {
```
TheRealMarnes: Not what I meant :) This is best suited to a private static method, maybe `String enumToName(E e)`, used by `makeArgument` as `EnumArgumentFactory::enumToName`

maffe: I prefer to keep the enum-to-name conversion in the in the `byName` method as is it not used anywhere else. This way it’s not needed to look for the `enumToName` method when trying to understand what `byName` does and one does not have to look for other uses of the `enumToName` method when it’s required to modify it somehow. Not that hard (especially using an IDE), but would still require some effort.

TheRealMarnes: TLDR: I'll not argue this further after this post so that this feature doesn't get hung up on nitpicks, but I would like another reviewer to cut the knot and either accept the multi-line pure lambda or request a method.

--

I disagree with some of your concepts. `as is it not used anywhere else` is what `private` methods are for. `not needed to look for the enumToName method`, sure, only for the `enumToName` variable instead, which is really no difference as both blocks are code are declared one or two lines away. `look for other uses` really not a concern with `private` members of such a small class.

My main objection is just that multi-line lambdas that don't rely on external state (pure functions) have no business being lambdas. A lambda declared as a variable inside a method (different from anonymous lambdas used with streams) suggests the lambda's content relies on some state of the declaring method. If they don't (such as here), what you're really doing is writing a method inside of another method and falsely suggesting it has a mutable nature, which is just confusing.

stevenschlansker: I did end up flattening this method out, by using `Optional` the logic became clearer.

PR #1563, opened 2019-07-24
Enum database value customization

stevenschlansker:
Fixes #1561

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/enums/DatabaseValue.java
+++ b/core/src/main/java/org/jdbi/v3/core/enums/DatabaseValue.java
@@ -17,6 +17,7 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+
 import org.jdbi.v3.core.qualifier.Qualifier;
 import org.jdbi.v3.meta.Beta;
 
--- a/core/src/main/java/org/jdbi/v3/core/enums/DatabaseValue.java
+++ b/core/src/main/java/org/jdbi/v3/core/enums/DatabaseValue.java
@@ -1,0 +1,29 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.enums;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/jdbi/v3/core/enums/DatabaseValue.java
```diff
@@ -0,0 +1,30 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.jdbi.v3.core.qualifier.Qualifier;
+import org.jdbi.v3.meta.Beta;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+@Qualifier
```
qualidafial: This shouldn't be a qualifier annotation, should it?

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
```diff
@@ -78,6 +81,11 @@ public E map(ResultSet rs, int columnNumber, StatementContext ctx) throws SQLExc
         private static Object getValueByName(Class<? extends Enum<?>> enumClass, String name) {
             final Enum<?>[] enumConstants = enumClass.getEnumConstants();
             return JdbiOptionals.findFirstPresent(
+                    () -> Arrays.stream(enumConstants).filter(e -> {
```
qualidafial: If somebody annotates an enum with `@DatabaseValue` then I don't think we should be falling back to matching the enum name.

core/src/main/java/org/jdbi/v3/core/enums/DatabaseValue.java
maffe: I don’t know what `@Qualifier` does, it’s an oversight from copy & paste.

qualidafial: Qualifying annotations are a recent addition to Jdbi, typically used to distinguish when multiple SQL data types resolve to the same Java type. e.g. both `VARCHAR` and `NVARCHAR` SQL columns are represented as `String` in Java.

http://jdbi.org/#_qualified_types

stevenschlansker: https://github.com/jdbi/jdbi/commit/c6e243cc4c374fe3510b3b3a967bc4c8f6e31cdc

core/src/main/java/org/jdbi/v3/core/mapper/EnumMapper.java
stevenschlansker: https://github.com/jdbi/jdbi/pull/1566

maffe: > If somebody annotates an enum with `@DatabaseValue` then I don't think we should be falling back to matching the enum name.

If I understand you correctly, given the already included enum
```
private enum Foobar {
    BAR, FOO, @DatabaseValue("CUST") CUSTOM
}
```
this test should pass (currently it doesn’t) and be included:
```
@Test
public void customizedNamesAreNotMappedToEnumName() throws NoSuchFieldException {
    db.getJdbi().useHandle(h -> {
        Optional<Foobar> mappedEnum = h.createQuery("select :name")
                .bind("name", "CUSTOM")
                .mapTo(Foobar.class)
                .findOne();

        assertThat(mappedEnum)
                .isEmpty();
    });
}
```

PR #1568, opened 2019-07-31
StatementExceptions: Deal with null RenderedSql

spmason:
I bumped into this when using `@SqlScript` with a resources file - `getRenderedSql()` returns null which causes an NPE and masks the proper exception (very confusing)


stevenschlansker:
Thanks!

PR #1570, opened 2019-08-01
OptionalMapperFactory: specialize created OptionalMappers

stevenschlansker:
so they don't lookup the mapper every row
also, since the factory's configuration is constant, we stash it in an enum to only have it once.


core/src/main/java/org/jdbi/v3/core/mapper/OptionalMapper.java
```diff
@@ -1,45 +0,0 @@
...
-
-    public static ColumnMapper<?> of(Type type) {
-        return new OptionalMapper<>(type);
-    }
-
-    @SuppressWarnings("unchecked")
-    @Override
-    public Optional<T> map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
-        final ColumnMapper<?> mapper = ctx.findColumnMapperFor(
-                GenericTypes.findGenericParameter(type, Optional.class)
```
stevenschlansker: Previously, this code was executed for every row -- but we can look the mapper up ahead of time.

core/src/main/java/org/jdbi/v3/core/mapper/OptionalMapperFactory.java
```diff
@@ -38,28 +42,59 @@
...
+                .map(strategy -> strategy.create(type, config));
+    }
+
+    private enum OptionalStrategy {
+        OBJ(Optional.class) {
+            @Override
+            ColumnMapper<?> create(Type type, ConfigRegistry config) {
+                final ColumnMapper<?> mapper = config.get(ColumnMappers.class).findFor(
+                        GenericTypes.findGenericParameter(type, Optional.class)
+                            .orElseThrow(() -> new NoSuchMapperException("No mapper for raw Optional type")))
```
stevenschlansker: This also means that configuration errors (mapping `Optional<T>` when `T` isn't mappable) are checked eagerly.

core/src/main/java/org/jdbi/v3/core/mapper/OptionalMapperFactory.java
```diff
@@ -38,28 +42,59 @@
...
+
+        final Class<?> optClass;
+        final ColumnMapper<?> impl;
 
-        if (rawType == Optional.class) {
-            return Optional.of(OptionalMapper.of(type));
+        OptionalStrategy(Class<?> optClass) {
+            this(optClass, null);
+        }
+        OptionalStrategy(Class<?> optClass, ColumnMapper<?> impl) {
```
TheRealMarnes: call me a twit, but I would like to see the conventional spacing between methods respected

core/src/main/java/org/jdbi/v3/core/mapper/OptionalMapperFactory.java
```diff
@@ -38,28 +42,59 @@
...
     public Optional<ColumnMapper<?>> build(Type type, ConfigRegistry config) {
-        Class<?> rawType = getErasedType(type);
+        return Optional.ofNullable(STRATEGIES.get(getErasedType(type)))
+                .map(strategy -> strategy.create(type, config));
+    }
+
+    private enum OptionalStrategy {
+        OBJ(Optional.class) {
+            @Override
+            ColumnMapper<?> create(Type type, ConfigRegistry config) {
```
TheRealMarnes: Passing a null to a functionally not null-tolerant method and overriding the method to get around that... Kinda feels to me like the field should store a `Function`, there should be a constructor that takes a constant value and stores a constant-supplying function, and a second constructor that takes this dynamic function. That feels like less logical branching to me. This is a private impl class though so if you feel that's bikeshedding, I can live with it.

core/src/main/java/org/jdbi/v3/core/mapper/OptionalMapperFactory.java
```diff
@@ -38,28 +42,59 @@
...
-        mappers.put(OptionalInt.class, optionalMapper(ResultSet::getInt, OptionalInt::empty, OptionalInt::of));
-        mappers.put(OptionalLong.class, optionalMapper(ResultSet::getLong, OptionalLong::empty, OptionalLong::of));
-        mappers.put(OptionalDouble.class, optionalMapper(ResultSet::getDouble, OptionalDouble::empty, OptionalDouble::of));
-    }
+    private static final Map<Class<?>, OptionalStrategy> STRATEGIES =
+            Collections.unmodifiableMap(Arrays.stream(OptionalStrategy.values())
+                .collect(Collectors.toMap(
+                        OptionalStrategy::optClass,
+                        Function.identity(),
+                        (u, v) -> {
```
TheRealMarnes: I think `collect(collectingAndThen(toMap(OptionalStrategy::optClass, identity()), IdentityHashMap::new))` is more suitable here, so you don't have to redundantly specify the throwingMerger

core/src/main/java/org/jdbi/v3/core/mapper/OptionalMapperFactory.java
```diff
@@ -38,28 +42,59 @@
...
+                        IdentityHashMap::new)));
 
     @Override
     public Optional<ColumnMapper<?>> build(Type type, ConfigRegistry config) {
-        Class<?> rawType = getErasedType(type);
+        return Optional.ofNullable(STRATEGIES.get(getErasedType(type)))
+                .map(strategy -> strategy.create(type, config));
+    }
+
+    private enum OptionalStrategy {
```
qualidafial: Feels like we're jumping through hoops to shoehorn the primitive mappers and reference mapper into a single type.

Is there anything wrong with putting the reference mapper special case right in the `build()` method? Then `OptionalMapper` could just take a delegate mapper as a constructor parameter.

core/src/main/java/org/jdbi/v3/core/mapper/OptionalMapperFactory.java
stevenschlansker: It's a little nicer to not have to specify the collector arg, but it does result in a redundant map copy, which is a little sad.

stevenschlansker: I'm actually not sure why this uses `IdentityHashMap` in the first place.  `Class` doesn't even override `equals` so there shouldn't be any behavioral differences.

core/src/main/java/org/jdbi/v3/core/mapper/OptionalMapperFactory.java
stevenschlansker: I made it nicer.  Like it better now?

core/src/main/java/org/jdbi/v3/core/mapper/OptionalMapperFactory.java
```diff
@@ -38,28 +41,40 @@
...
-        mappers.put(OptionalInt.class, optionalMapper(ResultSet::getInt, OptionalInt::empty, OptionalInt::of));
-        mappers.put(OptionalLong.class, optionalMapper(ResultSet::getLong, OptionalLong::empty, OptionalLong::of));
-        mappers.put(OptionalDouble.class, optionalMapper(ResultSet::getDouble, OptionalDouble::empty, OptionalDouble::of));
+    static {
+        Map<Class<?>, BiFunction<Type, ConfigRegistry, ColumnMapper<?>>> s = new HashMap<>();
+
+        s.put(Optional.class, OptionalMapperFactory::create);
+        s.put(OptionalInt.class, singleton(create(ResultSet::getInt, OptionalInt::empty, OptionalInt::of)));
+        s.put(OptionalLong.class, singleton(create(ResultSet::getLong, OptionalLong::empty, OptionalLong::of)));
+        s.put(OptionalDouble.class, singleton(create(ResultSet::getDouble, OptionalDouble::empty, OptionalDouble::of)));
```
qualidafial: Does `ResultSet::get<Primitive>` even work the way we're using it? It doesn't look like we're checking `wasNull()`. Do we have tests for optional primitives where we expect the result to be empty?

core/src/main/java/org/jdbi/v3/core/mapper/OptionalMapperFactory.java
qualidafial: :+1:

PR #1573, opened 2019-08-05
fix #1572: backport EmptyHandling

### LINKED ISSUE #1572: Do not throw an IllegalArgumentException when calling SqlStatement.bindList with an empty list
I have a query such as this:

```
    public List<String> load(List<String> namesToFind) {
        return db.withHandle(h -> h.createQuery("SELECT id FROM users WHERE name IN (<namesToFind>)")
                .bind("namesToFind", namesToFind)
                .mapTo(String.class))
                .list();
    }
```

Currently, when I call this method when the `namesToFind` parameter is an empty list, I get the following exception:

```
java.lang.IllegalArgumentException: Query.bindList was called with an empty list.
	at org.jdbi.v3.core.statement.SqlStatement.bindList(SqlStatement.java:1299) ~[jdbi3-core-3.6.0.jar:?]
```

I understand that it is part of the API specification that `bindList` and related methods throw this exception. However, when using the `BindList` annotation, there is an `onEmpty` attribute which allows the user to change the behaviour when the input list is null. Would it be possible to add a similar `onEmpty` configuration for the fluent API? I am not really experienced enough with JDBI to be able to suggest a solution so I'd be happy to hear suggestions.

*Affects JDBI-core 3.9.0*

TheRealMarnes:
Implemented the EmptyHandling enum in core, leaving out the deprecated value and redesigning the new API to accept an interface (making the enum a convenience, not a restriction).

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
@@ -30,6 +30,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
 
@@ -47,6 +48,7 @@ import org.jdbi.v3.core.argument.ObjectFieldArguments;
 import org.jdbi.v3.core.argument.ObjectMethodArguments;
 import org.jdbi.v3.core.argument.internal.PojoPropertyArguments;
 import org.jdbi.v3.core.generic.GenericType;
+import org.jdbi.v3.core.internal.IterableLike;
 import org.jdbi.v3.core.mapper.Mappers;
 import org.jdbi.v3.core.mapper.RowMapper;
 import org.jdbi.v3.core.mapper.immutables.JdbiImmutables;
@@ -1234,21 +1236,7 @@ public abstract class SqlStatement<This extends SqlStatement<This>> extends Base
     }
 
     /**
-     * Bind a parameter for each value in the given vararg array, and defines an attribute as the comma-separated list
-     * of parameter references (using colon prefix).
-     * <p>
-     * Examples:
-     * <pre>
-     * handle.createUpdate("insert into things (&lt;columnNames&gt;) values (&lt;values&gt;)")
-     *     .defineList("columnNames", "id", "name", "created_on")
-     *     .bindList("values", 1, "Alice", LocalDate.now())
-     *     .execute();
-     *
-     * List&lt;Thing&gt; things = handle.createQuery("select * from things where id in (&lt;ids&gt;)")
-     *     .bindList("ids", 1, 2, 3)
-     *     .mapTo(Contact.class)
-     *     .list();
-     * </pre>
+     * see {@link #bindList(BiConsumer, String, List)}
      *
      * @param key    attribute name
      * @param values vararg values that will be comma-spliced into the defined attribute value.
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -1250,20 +1263,34 @@ public final This bindBySqlType(int position, Object value, int sqlType) {
...
+     * @param values  vararg values that will be comma-spliced into the defined attribute value.
      * @return this
      * @throws IllegalArgumentException if the vararg array is empty.
+     * @see BindListEmptyHandling
      */
-    public final This bindList(String key, Object... values) {
-        if (values.length == 0) {
-            throw new IllegalArgumentException(
-                    getClass().getSimpleName() + ".bindList was called with no vararg values.");
+    public final This bindList(String key, BiConsumer<SqlStatement, String> onEmpty, Object... values) {
```
TheRealMarnes: the odds of someone legit passing a `BiConsumer<SqlStatement, String>` as a vararg value to be bound as a query param, which would cause this method to falsely match and break their query, are ridiculous...

stevenschlansker: Would it be better to move `onEmpty` to the first argument?  That both avoids this caveat (even though I agree it'd be a ridiculous situation) and it also avoids splitting `key` from `values` which are the "most related" arguments.

core/src/main/java/org/jdbi/v3/core/statement/BindListEmptyHandling.java
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.statement;
+
+import java.util.function.BiConsumer;
+
+/**
+ * describes what should be done if the value passed to {@link SqlStatement#bindList} is null or empty
+ */
+public enum BindListEmptyHandling implements BiConsumer<SqlStatement, String> {
```
stevenschlansker: Should this really have `BindList` in the name?  What if we want to use it from another feature?  Seems like this might be overly specific.

core/src/main/java/org/jdbi/v3/core/statement/BindListEmptyHandling.java
```diff
@@ -0,0 +1,64 @@
...
+     * {@code select * from things where x in ()}
+     */
+    VOID {
+        @Override
+        public void accept(SqlStatement stmt, String key) {
+            stmt.define(key, "");
+        }
+    },
+    /**
+     * <p>Output "null" (without quotes, as keyword), useful e.g. in postgresql where "in ()" is invalid syntax.</p>
```
stevenschlansker: The usual way to indicate a keyword is `{@code null}`

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -1284,15 +1311,17 @@ public final This bindList(String key, Object... values) {
...
      * @return this
      * @throws IllegalArgumentException if the list is empty.
+     * @see BindListEmptyHandling
      */
-    public final This bindList(String key, List<?> values) {
-        if (values.isEmpty()) {
-            throw new IllegalArgumentException(
-                    getClass().getSimpleName() + ".bindList was called with an empty list.");
+    public final This bindList(String key, BiConsumer<SqlStatement, String> onEmpty, List<?> values) {
+        if (values == null || values.isEmpty()) {
```
stevenschlansker: Hm, this is a sneaky change.  I guess that's fine to catch `null` here too.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
TheRealMarnes: Hmm why not

core/src/main/java/org/jdbi/v3/core/statement/BindListEmptyHandling.java
TheRealMarnes: I named it so specifically so that it would be discoverable via IDE hints to people typing "bindlist", and because it really doesn't feel like something that can be reused for anything. But if you insist

core/src/main/java/org/jdbi/v3/core/statement/BindListEmptyHandling.java
TheRealMarnes: The problem here is the perspective to document from: do we document from the perspective of what we are binding, or from what we will output in the query? Both are familiar to users and neither seems like the less important one to me, though if I absolutely had to choose I guess I would choose the query POV.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
TheRealMarnes: Yeah I figured this would be ok to change since nobody is eager for NPEs from libraries

core/src/main/java/org/jdbi/v3/core/statement/BindListEmptyHandling.java
stevenschlansker: I don't insist.  Just a comment.  I don't have a strong preference.

core/src/main/java/org/jdbi/v3/core/statement/BindListEmptyHandling.java
stevenschlansker: I just think it's strange to put a quotation mark and then immediately undo it in a parenthetical, when there's an out-of-the-box style for `keyword`.  I don't feel too strongly about it either way.

TheRealMarnes: Yeah but that's not the issue... We _define_ a _string_ `"null"`, which _renders_ to the _keyword_ `null`. It does both. The question is which POV do we document from? If I had to guess I'd document the query content since that's what people will mostly need to think about.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -1284,15 +1311,17 @@ public final This bindList(String key, Object... values) {
...
+     * @param values  list of values that will be comma-spliced into the defined attribute value.
      * @return this
      * @throws IllegalArgumentException if the list is empty.
+     * @see EmptyHandling
      */
-    public final This bindList(String key, List<?> values) {
-        if (values.isEmpty()) {
-            throw new IllegalArgumentException(
-                    getClass().getSimpleName() + ".bindList was called with an empty list.");
+    public final This bindList(String key, BiConsumer<SqlStatement, String> onEmpty, List<?> values) {
```
TheRealMarnes: should I actually maybe declare this as `Iterable` and add an overload for `Iterator` so that it aligns with `@BindList`?

TheRealMarnes: also, parameter order is not correct yet

TheRealMarnes:
Commits rebased and overloads for `Iterable` and `Iterator` added for similar flexibility to `@BindList`

PR #1578, opened 2019-08-09
jackson bom 2.9.9.20190807

stevenschlansker:
fixes multiple CVEs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 
 sudo: false
-dist: trusty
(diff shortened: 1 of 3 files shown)


stevenschlansker:
The build hang seems to be due to Aether's PartialFile locking never returning.  Might be an artifact of running in a container.

stevenschlansker:
Upgrading to Xenial (which required switching from `oraclejdk8` to `openjdk8`) seems to fix the problem :confused:

PR #1579, opened 2019-08-09
SqlObject: fix CreateSqlObject and withHandle outer scope

### LINKED ISSUE #1011: Make @CreateSqlObject less confusing.
**Problem**
The JDBI official documentation leads developers to believe the @CreateSqlObject annotation is the mechanism for adding transactional support to disparate DAOs.

http://jdbi.org/#__createsqlobject

Which seems a bit misleading to the actual behavior which can lead to state errors when used w/ the preferred `jdbi.onDemand` creation mechanism.

There's been a few messages in the forum that illustrate this confusion:

- https://groups.google.com/forum/#!searchin/jdbi/CreateSqlObject%7Csort:date/jdbi/FPI1l1OBLU0/slXj_8hJBwAJ

- https://groups.google.com/forum/#!searchin/jdbi/CreateSqlObject%7Csort:date/jdbi/b1rWVQ7JNAg/2R2c9okAAgAJ

**Request**
Not sure what the best solution(s) to this problem would be.  Some that come to mind:

- Deprecation
- Better documentation as to the short-comings
- Another solution to splitting queries in multiple DAOs, but allow logical transactional semantics to tie them together.  Specifically, in supporting a mix of read-only / write transactions.

stevenschlansker:
This fixes up two long-standing problems:
* onDemand does not play well with CreateSqlObject
* SqlObject.withHandle (and Transactional.withTransaction) has surprisingly useless behavior when called outside an onDemand scope since it immediately operates on a closed handle.

Fixes #1011


sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestVariousOddities.java
```diff
@@ -151,4 +154,49 @@ public void testAbiguityResolved() {
...
+        onDemand.withHandle(h ->
+            h.inTransaction(txn ->
+                assertThat(onDemand.versionA().value()).isEqualTo("intriguing")));
+    }
+
+    @Test
+    public void decoratedOnDemandInTransaction() throws Exception {
+        OnDemandOddities onDemand = new DecoratedOnDemandOddities();
+        onDemand.inTransaction(txn ->
+                assertThat(onDemand.versionA().value()).isEqualTo("intriguing"));
```
TheRealMarnes: So this fails because onDemand opens and closes a connection around method invocations, but this `versionA()` returns the created object, which of course cannot be called outside the onDemand method because its handle is closed upon returning.

TheRealMarnes: And the idea for the fix is that since this invocation happens inside the scope of the same onDemand object's `Transactional<>` method, it should share the outer Handle instead of opening+closing a redundant onDemand one.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestVariousOddities.java
```diff
@@ -151,4 +154,49 @@ public void testAbiguityResolved() {
...
+
+        @Override
+        public VersionA versionA() {
+            return onDemand.versionA();
+        }
+    }
+
+    @Test
+    public void onDemandCreateSqlObject() throws Exception {
+        assertThat(dbRule.getJdbi().onDemand(OnDemandOddities.class).versionA().value())
```
TheRealMarnes: Why should this work? You create an onDemand object and call a query method outside its demand scope with no outer scope to piggyback on.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestVariousOddities.java
stevenschlansker: Exactly!  I think it's a `LazyHandleSupplier` that ends up getting shared, but that's the right idea.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestVariousOddities.java
stevenschlansker: The short answer is, "because users expect it to".  `@CreateSqlObject` is a bit of a wart on the API.  Imagine that you have two types of SqlObject, "Type A" and "Type B".  (ondemand vs attached).  It's surprising to have the created sub-sql-object not be of the same type as the parent, and plenty of people have gotten tripped up.  Since there's a clearer semantic that is also actually more useful (a closed SqlObject is useless) we can remove the surprise.

TheRealMarnes: You want CreateSqlObject instances to inherit the parent's onDemand aspect?

stevenschlansker: Yep.  See #1011 for background info.

TheRealMarnes: I find a fix for this a bit of anti-feature tbh. That CreateSqlObject subobjects are visible to callers is a wart indeed. Users should ignore them because they're _obviously_ not intended to be used independently of the parent onDemand object, they're essentially private state. If you want to use one of the CreateSqlObjected classes, you should create your own instance properly instead of piggybacking on "accidentally" exposed factory methods.

Making this work is an API-subverting convenience that sends a signal to users that violating the internal kitchen boundary of a class is an acceptable design to deliberately go with, rather than an unfortunate wart (interface components forcibly being public) that should be ignored.

stevenschlansker: I don't disagree that `@CreateSqlObject` is a bit of a wart -- but there's no fewer than three different users that have been bit by the confusing and broken behavior (personally, we had a production outage due to the onDemand scope not working as expected) so I don't think "not fixing it" is really an option.  Perhaps we drop `@CreateSqlObject` entirely in v4 but as long as it lives in our tree we should try to make it behave as nicely as we can.  And the current behavior does nothing useful, and throws a confusing and difficult to diagnose error message.  This PR doesn't have to be the fix, but we must ship *some* fix for this issue.

TheRealMarnes: It's really only as confusing as hibernate entities being detached outside transactional methods... I find CreateSqlObject as it works now ok, you just have to refrain from thinking you can pull a subcomponent out of the owning class' scope and use it. Be it 4 or 40 users making that mistake, it's just not what the feature is for and there's no functional need to use it that way.

I won't turn this change down, just know I find it an anti-feature that encourages violating the internal kitchen boundary for a minor convenience, which is bad in its own right.

stevenschlansker: I agree with your sentiment, but let's take this conversation to a larger "What should CreateSqlObject be (or not be) in 4.0" at a later time.  For now let's make the existing feature as least-warty as it can be.

TheRealMarnes: Well yeah, but that's just it... I find this change even more warty because it isn't even consistent with itself anymore.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/WithHandleMethodHandlerFactory.java
```diff
@@ -0,0 +1,37 @@
...
+    private static final Method WITH_HANDLE = Handlers.methodLookup(SqlObject.class, "withHandle", HandleCallback.class);
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public Optional<Handler> buildHandler(Class<?> sqlObjectType, Method method) {
+        if (!WITH_HANDLE.equals(method)) {
+            return Optional.empty();
+        }
+        return Optional.of((t, a, h) ->
+            h.invokeInContext(
```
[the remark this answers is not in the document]
TheRealMarnes: `Transactional.commit` and such still use `getHandle()`, so they'll keep throwing the "connection closed" exception. Should we maybe do something about that? `onDemand.commit` makes no sense to write in the first place, but it could maybe detect being invoked on an ondemand and throw a more meaningful exception since that's what this PR is about?

Maybe we need an OnDemand marker interface, since there was also that `instanceof` check on the handlesupplier to check for ondemand instances.

core/src/main/java/org/jdbi/v3/core/internal/OnDemandHandleSupplier.java
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.internal;
+
+public interface OnDemandHandleSupplier {}
```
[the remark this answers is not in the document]
TheRealMarnes: Oh god, does this only exist because LazyHandleSupplier is package-private and not visible to CreateSqlObjectHandler, and can't be made public because it can't move to `internal` because it needs to call package-private methods on `Handle` that can't be made public?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/WithHandleMethodHandlerFactory.java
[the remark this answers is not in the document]
TheRealMarnes: So I tried using an OnDemand marker interface, it can't be done because the proxy with the interface on it isn't passed to the handler.

TheRealMarnes:
Curious thought: what happens now with `onDemand.getA().getB()` before my #1580 to implement SqlObject, since the code relies on reflectively resolving some SqlObject methods? @stevenschlansker could you maybe come up with a test for that?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/WithHandleMethodHandlerFactory.java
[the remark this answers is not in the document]
stevenschlansker: I made it implement *all* declared interfaces in addition to your hard-coded `SqlObject`.

stevenschlansker:
I spruced the docs up a fair bit.

stevenschlansker:
I'm not sure what problem you're seeing with adding the declared interfaces to the created `Proxy` object.  I don't expect it to affect method resolution in any way, since we always inspect the extension type `Class` (your `interface`) and never the `Proxy` object class.

PR #1583, opened 2019-08-17
Expose Qualified(Argument/Mapper)Factories registration API

TheRealMarnes:
We forgot to expose this to users, they can't currently register their own qualified factories.

Apparently this introduces some ambiguity on existing calls (separate commit), despite `Type` and `QualifiedType` being pretty distinct from each other...

### DIFF
--- a/postgres/src/main/java/org/jdbi/v3/postgres/PostgresPlugin.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/PostgresPlugin.java
@@ -18,6 +18,7 @@ import java.util.UUID;
 
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.argument.ArgumentFactory;
 import org.jdbi.v3.core.generic.GenericType;
 import org.jdbi.v3.core.internal.JdbiClassUtils;
 import org.jdbi.v3.core.internal.exceptions.Unchecked;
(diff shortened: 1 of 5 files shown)


stevenschlansker:
I don't think this was an oversight.  I think the intent was to start with simply annotation-driven qualified factory binding, and then maybe expose this later if needed.  Is there a specific use case that needs to be able to inspect all qualifiers, that drives exposing this as API?

TheRealMarnes:
I wanted to register a NonNull qualifier to force non-null columns through jdbi

stevenschlansker:
Right -- can't you just annotate the argument factory itself as `@NonNull`, and used the qualifier inferrer like the existing factories do?

TheRealMarnes:
That doesn't let you take the qualified value type, remove the single qualifier, and search a match against the remaining qualifiers

postgres/src/main/java/org/jdbi/v3/postgres/PostgresPlugin.java
```diff
@@ -136,7 +137,7 @@ public void customizeJdbi(Jdbi jdbi) {
         jdbi.registerColumnMapper(new PGobjectColumnMapperFactory());
 
         // legacy unqualified HSTORE
-        jdbi.registerArgument(new HStoreArgumentFactory()::build);
+        jdbi.registerArgument((ArgumentFactory) new HStoreArgumentFactory()::build);
```
qualidafial: I know I'm coming to this weeks after the fact, but why does this require a cast?

TheRealMarnes: I wondered too, it didn't make any sense to me at the time since the different possibilities are incompatible with each other

PR #1588, opened 2019-08-30
SqlParsers: use `caffeine` cache

### LINKED ISSUE #1528: JdbiCache needs some kind of expiry mechanism
With #1525 merged in (which I love) we are now unfortunately possibly caching an unlimited number of parsed sql statements, which in principle may be dynamically generated.  See #93 / #125 for evidence that this does bite people.

stevenschlansker:
Fixes #1528

We can worry about minimizing the size later, rather than continue to have this problem persist.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/ColonPrefixSqlParser.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/ColonPrefixSqlParser.java
@@ -13,12 +13,12 @@
  */
 package org.jdbi.v3.core.statement;
 
+import com.github.benmanes.caffeine.cache.Caffeine;
 import org.antlr.v4.runtime.CharStreams;
 import org.antlr.v4.runtime.Token;
-import org.jdbi.v3.core.config.JdbiCache;
-import org.jdbi.v3.core.config.JdbiCaches;
 import org.jdbi.v3.core.internal.lexer.ColonStatementLexer;
 import org.jdbi.v3.core.statement.internal.ErrorListener;
+import org.jdbi.v3.meta.Beta;
 
 import static org.antlr.v4.runtime.Recognizer.EOF;
 import static org.jdbi.v3.core.internal.lexer.ColonStatementLexer.COMMENT;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/jdbi/v3/core/statement/ColonPrefixSqlParser.java
```diff
@@ -36,25 +36,24 @@
...
-            return PARSED_SQL_CACHE.get(sql, ctx);
-        } catch (IllegalArgumentException e) {
-            throw new UnableToCreateStatementException("Exception parsing for named parameter replacement", e, ctx);
-        }
+    public ColonPrefixSqlParser() {
+        this(Caffeine.newBuilder());
+    }
+
+    @Beta
+    public ColonPrefixSqlParser(Caffeine<Object, Object> cache) {
```
qualidafial: This adds Caffeine to our public API. I'm ok with this as `@Beta` but we should be very confident about this decision before making it permanent (i.e. removing the Beta annotation).

core/src/main/java/org/jdbi/v3/core/statement/HashPrefixSqlParser.java
```diff
@@ -33,25 +33,24 @@
...
-            return PARSED_SQL_CACHE.get(sql, ctx);
-        } catch (IllegalArgumentException e) {
-            throw new UnableToCreateStatementException("Exception parsing for named parameter replacement", e, ctx);
-        }
+    public HashPrefixSqlParser() {
+        this(Caffeine.newBuilder());
+    }
+
+    @Beta
+    public HashPrefixSqlParser(final Caffeine<Object, Object> cache) {
```
qualidafial: Ditto

core/src/main/java/org/jdbi/v3/core/statement/ColonPrefixSqlParser.java
stevenschlansker: Yup, agreed.  I wasn't sure of a good way to expose Caffeine configuration otherwise, and I did not want to create a "Jdbi cache config" interface.

ben-manes: Perhaps all you need is a loading function, e.g. `Function<K, V>`, as settable in the public API. Then `cache::get` can be passed in to delegate to caffeine, for instance. This of course reduces the flexibility of how you can use the cache (e.g. no `asMap` view) but decouples with minimal work.

qualidafial: I like the idea of putting a facade API in front of the cache provider. Then somebody could configure e.g. a provider that performs no caching if their hit rate is too low.

qualidafial: Similar to how we have TemplateEngine as an abstraction in front of FreeMarker and StringTemplate.

PR #1591, opened 2019-08-31
Change most JdbiPlugins to be single-install

### LINKED ISSUE #1495: Kotlin SqlObject plugin should either include or check for the presence of normal SqlObject plugin
From the mailing list, if you install the Kotlin SqlObject plugin but not the vanilla one, you get "extension not found" errors.  We could add a more specific check, "The vanilla SqlObject plugin is not installed", or perhaps just have the Kotlin version install it for you if it is missing.

stevenschlansker:
KotlinSqlObjectPlugin installs SqlObjectPlugin

Fixes #1495

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,8 @@
   - Handle.getJdbi gets owning Jdbi instance
   - SqlStatement (like Query) has new bindArray helper methods
   - sqlobject's `EmptyHandling` enum backported to core for invocations of `SqlStatement.bindList`
+  - installPlugin skips duplicate JdbiPlugins (according to Object.equals)
+  - KotlinSqlObjectPlugin will install forgotten SqlObjectPlugin for you
 - New Beta Features
   - added `register` methods for qualified factories on `Configurable`,
     `ColumnMappers`, and `ArgumentFactories`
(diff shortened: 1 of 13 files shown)


RELEASE_NOTES.md
```diff
@@ -2,6 +2,8 @@
 - New Features
   - Handle.getJdbi gets owning Jdbi instance
   - sqlobject's `EmptyHandling` enum backported to core for invocations of `SqlStatement.bindList`
+  - installPlugin skips duplicate JdbiPlugins (accoring to Object.equals)
```
TheRealMarnes: typo `accoring`

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -229,8 +229,9 @@ public Jdbi installPlugins() {
      * @return this
      */
     public Jdbi installPlugin(JdbiPlugin plugin) {
-        Unchecked.consumer(plugin::customizeJdbi).accept(this);
-        plugins.add(plugin);
+        if (plugins.addIfAbsent(plugin)) {
```
TheRealMarnes: I can see this causing issues for some people since behavior in jdbi can depend a lot on registration order. If someone does duplicate plugin registrations, this change will break things for them. Is it acceptable in this case?

stevenschlansker: I think that's okay.  Repeatedly installing a plugin just to change the ordering is pretty broken to start with.  Also, plugins do not currently implement `equals` -- so it would have to be the same *instance* to trigger this check.  I think the risk is small.

PR #1592, opened 2019-09-01
SqlCall: OutParameters processing in Consumer or Function callback

### LINKED ISSUE #76: New feature Request:  Add supporting Stored Procedure which returns ResultSets (one and multiple)
JDBC callable statements can return multiple resultsets. 
Currently in the jDBI doesn't seam to support retrieving result data from callable statements
(only output parameters!).

### LINKED ISSUE #1392: Allow Oracle refcursor to be processed by standard Jdbi mechanisms
@qualidafial [requested a new issue be created](https://github.com/jdbi/jdbi/issues/76#issuecomment-451038215) showing a ref cursor being returned from a stored procedure and read using JDBC.

```
PROCEDURE read_example_data (
  o_error_string  OUT VARCHAR2,
  o_output_cursor OUT SYS_REFCURSOR,
  i_input_key     IN  VARCHAR2
)
IS
  -- local variables here
BEGIN

  -- pretend we used i_input_key in this query
  OPEN o_output_cursor FOR
    select 1111 as key from dual union all select 2222 as key from dual;

EXCEPTION
  WHEN OTHERS THEN
    o_error_string := TO_CHAR(SQLCODE) || '::' || 'READ failure'|| '::' || substr(1,1000,SQLERRM)
END read_example_data;
```

In reality the stored procedure could be considerably more complex and do multiple operations before opening the cursor.

The Java to read from the cursor would look something like this...

```
  private static final String EXAMPLE_READ_STORED_PROCEDURE_CALL =
    "BEGIN                                                          \n" +
    "  read_example_data(                            \n" +
    "    o_error_string                => ?,                        \n" +
    "    o_output_cursor               => ?,                        \n" +
    "    i_input_key                   => ?);                       \n" +
    "END;";
...
      java.sql.Connection  jdbcConnection = ....
      oracle.jdbc.OracleCallableStatement cs = (OracleCallableStatement) jdbcConnection.prepareCall( EXAMPLE_READ_STORED_PROCEDURE_CALL );

      cs.registerOutParameter(1, oracle.jdbc.OracleTypes.VARCHAR);
      cs.registerOutParameter(2, oracle.jdbc.OracleTypes.CURSOR);
      cs.setString(3, aParameter);
      cs.execute();
        
      String errorText = cs.getString(1);
      if (errorText != null && !errorText.isEmpty()) {
        throw new Exception( "Error! " + errorText );
      }
      
      oracle.jdbc.OracleResultSet exampleCursor = (OracleResultSet) cs.getCursor( 2 );

      while ( exampleCursor != null && cursor.next() ) {
        String key = exampleCursor .getString( "KEY" );
      ...
      }
```

Having a way to use an OracleResultSet (returned as an out parameter of a stored procedure) as a [ResultBearing](http://jdbi.org/apidocs/org/jdbi/v3/core/result/ResultBearing.html) perhaps?  Basically some way of being able to use the various Mapping and other great functionality of JDBI with such a cursor.

Thanks.

stevenschlansker:
Add test returning multiple result sets from a Postgres stored function
Closes our oldest outstanding issue!!! :cake: 

Fixes #1392
Fixes #76

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/OutParameters.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/OutParameters.java
@@ -14,16 +14,24 @@
 package org.jdbi.v3.core.statement;
 
 import java.sql.Date;
+import java.sql.ResultSet;
 import java.sql.Timestamp;
 import java.util.HashMap;
 import java.util.Map;
+
+import org.jdbi.v3.core.result.ResultBearing;
 
 /**
  * Represents output from a Call (CallableStatement).
  * @see Call
  */
 public class OutParameters {
+    private final StatementContext ctx;
     private final Map<Object, Object> map = new HashMap<>();
+
+    OutParameters(StatementContext ctx) {
+        this.ctx = ctx;
+    }
 
     /**
      * Type-casting convenience method which obtains an object from the map, the
@@ -211,6 +219,14 @@ public class OutParameters {
 
     public Float getFloat(int pos) {
         return getNumber(pos).floatValue();
+    }
+
+    public ResultBearing getRowSet(String name) {
+        return ResultBearing.of(() -> getObject(name, ResultSet.class), ctx);
+    }
+
+    public ResultBearing getRowSet(int pos) {
+        return ResultBearing.of(() -> getObject(pos, ResultSet.class), ctx);
     }
 
     private Number getNumber(String name) {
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/jdbi/v3/core/statement/OutParameters.java
```diff
@@ -14,17 +14,25 @@
...
+
 /**
  * Represents output from a Call (CallableStatement).
  * @see Call
  */
 public class OutParameters {
+    private final StatementContext ctx;
     private final Map<Object, Object> map = new HashMap<>();
 
+    OutParameters(StatementContext ctx) {
```
qualidafial: Adding this private constructor removes the existing default constructor, which technically makes this a breaking change. However, the constructor should never have been visible in the first place.

Can we add a compatibility comment to the release notes?

PR #1597, opened 2019-09-01
@Qualified

leaumar:
`core` can qualify a call with any annotation, but `sqlobject` requires `@Qualifier`, which cannot be added to external annotations. `@Qualified` fills that functional gap:
`@Qualified(NonNull.class) String getNonNullString();`

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
+++ b/core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
@@ -64,17 +64,15 @@ public class Qualifiers implements JdbiConfig<Qualifiers> {
      * which searches for annotations themselves annotated with {@link Qualifier}.
      *
      * Resolvers should return static results that don't vary over time; therefore their results are cached.
-     * Adding a resolver will clear the local qualifier resolution cache, but is currently not guaranteed to have any effect on {@code SqlObjects}.
-     * Add any resolvers you may need <em>before</em> using any qualifiers-related features!
+     * Adding a resolver will clear the local qualifier resolution cache, but {@code SqlObjects} and other features
+     * may have their own caching mechanisms. Add any resolvers you may need <strong>before</strong> using any qualifiers-related features!
      *
      * @param resolver the resolver to be included in qualifier resolution
      * @return this
      */
     public Qualifiers addResolver(Function<AnnotatedElement, Set<Annotation>> resolver) {
         resolvers.add(resolver);
-        if (registry != null) {
-            QUALIFIER_CACHE.clear(registry);
-        }
+        QUALIFIER_CACHE.clear(registry);
         return this;
     }
 
@@ -114,7 +112,7 @@ public class Qualifiers implements JdbiConfig<Qualifiers> {
 
     private static Set<Annotation> inspectQualifiedCarrier(AnnotatedElement element) {
         return Arrays.stream(element.getAnnotations())
-            .filter(anno -> anno instanceof Qualified)
+            .filter(Qualified.class::isInstance)
             .map(Qualified.class::cast)
             .map(Qualified::value)
             .flatMap(Arrays::stream)
--- a/core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
+++ b/core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
@@ -16,10 +16,10 @@ package org.jdbi.v3.core.qualifier;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
 import java.util.Arrays;
-import java.util.Collections;
+import java.util.Collection;
 import java.util.HashSet;
-import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.function.Function;
 
 import org.jdbi.v3.core.config.ConfigRegistry;
@@ -37,14 +37,42 @@ import static java.util.stream.Collectors.toSet;
 public class Qualifiers implements JdbiConfig<Qualifiers> {
     private static final JdbiCache<AnnotatedElement[], Set<Annotation>> QUALIFIER_CACHE = JdbiCaches.declare(
         elements -> elements.length == 1 ? elements[0] : new HashSet<>(Arrays.asList(elements)),
-        (Function<AnnotatedElement[], Set<Annotation>>) Qualifiers::getQualifiers);
+        (registry, elements) -> registry.get(Qualifiers.class).resolveQualifiers(elements));
+    private final Set<Function<AnnotatedElement, Set<Annotation>>> resolvers = new CopyOnWriteArraySet<>();
     private ConfigRegistry registry;
 
-    public Qualifiers() {}
+    public Qualifiers() {
+        resolvers.add(Qualifiers::getQualifierAnnotations);
+    }
+
+    private Qualifiers(Qualifiers other) {
+        this.resolvers.addAll(other.resolvers);
+        this.registry = null;
+    }
 
     @Override
     public void setRegistry(ConfigRegistry registry) {
         this.registry = registry;
+    }
+
+    /**
+     * Adds a qualifier resolver, which inspects annotated elements (e.g. methods or parameters)
+     * and returns a set of qualifiers that are found to belong on it. An example resolver is {@link #getQualifierAnnotations},
+     * which searches for annotations themselves annotated with {@link Qualifier}.
+     *
+     * Resolvers should return static results that don't vary over time; therefore their results are cached.
+     * Adding a resolver will clear the local qualifier resolution cache, but is currently not guaranteed to have any effect on {@code SqlObjects}.
+     * Add any resolvers you may need <em>before</em> using any qualifiers-related features!
+     *
+     * @param resolver the resolver to be included in qualifier resolution
+     * @return this
+     */
+    public Qualifiers addResolver(Function<AnnotatedElement, Set<Annotation>> resolver) {
+        resolvers.add(resolver);
+        if (registry != null) {
+            QUALIFIER_CACHE.clear(registry);
+        }
+        return this;
     }
 
     /**
@@ -54,23 +82,30 @@ public class Qualifiers implements JdbiConfig<Qualifiers> {
      * @return the set of qualifying annotations on the given elements.
      */
     public Set<Annotation> findFor(AnnotatedElement... elements) {
-        if (registry == null) {
-            return getQualifiers(elements);
-        }
-        return QUALIFIER_CACHE.get(elements, registry);
+        return registry == null
+            ? resolveQualifiers(elements)
+            : QUALIFIER_CACHE.get(elements, registry);
     }
 
-    private static Set<Annotation> getQualifiers(AnnotatedElement... elements) {
-        return Collections.unmodifiableSet(Arrays.stream(elements)
+    private Set<Annotation> resolveQualifiers(AnnotatedElement... elements) {
+        Set<AnnotatedElement> nonNullElements = Arrays.stream(elements)
             .filter(Objects::nonNull)
-            .map(AnnotatedElement::getAnnotations)
-            .flatMap(Arrays::stream)
-            .filter(anno -> anno.annotationType().isAnnotationPresent(Qualifier.class))
-            .collect(toSet()));
+            .collect(toSet());
+
+        return resolvers.stream()
+            .flatMap(resolver -> nonNullElements.stream().map(resolver))
+            .flatMap(Collection::stream)
+            .collect(toSet());
     }
 
     @Override
     public Qualifiers createCopy() {
-        return this;
+        return new Qualifiers(this);
+    }
+
+    private static Set<Annotation> getQualifierAnnotations(AnnotatedElement element) {
+        return Arrays.stream(element.getAnnotations())
+            .filter(anno -> anno.annotationType().isAnnotationPresent(Qualifier.class))
+            .collect(toSet());
     }
 }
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
```diff
@@ -36,40 +36,54 @@
...
-    public void setRegistry(ConfigRegistry registry) {
-        this.registry = registry;
+    public Qualifiers(Qualifiers other) {
+        this.resolvers.addAll(other.resolvers);
+        this.registry = null;
+    }
+
+    public Qualifiers addResolver(Function<AnnotatedElement, Set<Annotation>> resolver) {
+        resolvers.add(resolver);
+        QUALIFIER_CACHE.clear(registry);
```
leaumar: you guys seemed pretty wary of doing this but I'm not seeing an issue tbh

qualidafial: This is standard procedure in our existing registries. :+1:

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
```diff
@@ -36,40 +36,54 @@
...
 
-    public Qualifiers() {}
+    public Qualifiers() {
+        resolvers.add(Qualifiers::getQualifierAnnotations);
+    }
 
-    @Override
-    public void setRegistry(ConfigRegistry registry) {
-        this.registry = registry;
+    public Qualifiers(Qualifiers other) {
```
qualidafial: Should be package private

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
```diff
@@ -36,40 +36,54 @@
...
+        this.resolvers.addAll(other.resolvers);
+        this.registry = null;
+    }
+
+    public Qualifiers addResolver(Function<AnnotatedElement, Set<Annotation>> resolver) {
```
qualidafial: Needs javadoc

sqlobject/src/main/java/org/jdbi/v3/sqlobject/RegisterQualifier.java
```diff
@@ -0,0 +1,43 @@
...
+ * This annotation is recognized by Jdbi as a "container" of qualifying annotations. Use this when you
+ * want to use an annotation as a qualifier on SQL Objects but are unable to add @{@link Qualifier} to the annotation's source.
+ *
+ * An example use case is putting @{@link javax.annotation.Nonnull} as a qualifier on an SQL Object's fields/methods.
+ */
+@Beta
+@Retention(RUNTIME)
+@Target({FIELD, PARAMETER, METHOD, CONSTRUCTOR, TYPE})
+@Repeatable(RegisterQualifiers.class)
+public @interface RegisterQualifier {
```
qualidafial: Let's rename this back to `@Qualified`, since there isn't a corresponding `Qualifiers.register` method

sqlobject/src/main/java/org/jdbi/v3/sqlobject/Qualified.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * This annotation is recognized by Jdbi as a "container" of qualifying annotations. Use this when you
+ * want to use an annotation as a qualifier on SQL Objects but are unable to add @{@link Qualifier} to the annotation's source.
+ *
+ * An example use case is putting @{@link javax.annotation.Nonnull} as a qualifier on an SQL Object's fields/methods.
+ */
+@Beta
+@Retention(RUNTIME)
+@Target({FIELD, PARAMETER, METHOD, CONSTRUCTOR, TYPE})
+public @interface Qualified {
```
qualidafial: I'd actually like to have this in the `qualifier` package in core. There's nothing SQL-object specific going on here.

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
leaumar: `clear` did not exist on jdbicache until now though, how can it be SOP?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/Qualified.java
leaumar: Except that core has no knowledge of being able to annotate anything since sqlobject is where the whole class-oriented approach is created

leaumar: For stuff like bean mapping, right?

leaumar:
Everything handled and fixed:
- release notes
- dev guide about qualified factories and custom qualifiers
- new methods on qualtyp and jdbicache
- Qualified in core
- tests tests tests

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
```diff
@@ -36,40 +38,87 @@
...
+
+    private static Set<Annotation> getQualifierAnnotations(AnnotatedElement element) {
+        return Arrays.stream(element.getAnnotations())
+            .filter(anno -> anno.annotationType().isAnnotationPresent(Qualifier.class))
+            .collect(toSet());
+    }
+
+    private static Set<Annotation> inspectQualifiedCarrier(AnnotatedElement element) {
+        return Arrays.stream(element.getAnnotations())
+            .filter(anno -> anno instanceof Qualified)
```
qualidafial: Alternatively: `Qualified.class::isInstance`

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
```diff
@@ -36,40 +38,87 @@
...
         this.registry = registry;
     }
 
+    /**
+     * Adds a qualifier resolver, which inspects annotated elements (e.g. methods or parameters)
+     * and returns a set of qualifiers that are found to belong on it. An example resolver is {@link #getQualifierAnnotations},
+     * which searches for annotations themselves annotated with {@link Qualifier}.
+     *
+     * Resolvers should return static results that don't vary over time; therefore their results are cached.
+     * Adding a resolver will clear the local qualifier resolution cache, but is currently not guaranteed to have any effect on {@code SqlObjects}.
```
stevenschlansker: I think this is a little misleading: as long as you add the resolvers before you initialize your SqlObjects, it should work just fine with SqlObject.

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
```diff
@@ -36,40 +38,87 @@
...
+            .collect(toSet());
+    }
+
+    private static Set<Annotation> inspectQualifiedCarrier(AnnotatedElement element) {
+        return Arrays.stream(element.getAnnotations())
+            .filter(anno -> anno instanceof Qualified)
+            .map(Qualified.class::cast)
+            .map(Qualified::value)
+            .flatMap(Arrays::stream)
+            .map(AnnotationFactory::create)
```
stevenschlansker: I worry that this approach does not work for any annotations that themselves have properties, as you can only specify their `Class`.

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
```diff
@@ -36,40 +38,87 @@
...
 
-    public Qualifiers() {}
+    public Qualifiers() {
+        resolvers.add(Qualifiers::getQualifierAnnotations);
+        resolvers.add(Qualifiers::inspectQualifiedCarrier);
+    }
+
+    private Qualifiers(Qualifiers other) {
+        this.resolvers.addAll(other.resolvers);
+        this.registry = null;
```
stevenschlansker: It's a little weird to have an optional registry here.
If you instead do `new ConfigRegistry().get(Qualifiers.class)` for testing does that remove the need for `Qualifiers` to work without the registry?

docs/src/adoc/index.adoc
```diff
@@ -2603,6 +2603,47 @@ and setter parameters.
 - `@BindJpa` and `JpaMapper` respect qualifiers on getters and setters.
 - `@BindKotlin`, `bindKotlin()`, and `KotlinMapper` respect qualifiers on constructor parameters, getters, setters, and setter parameters.
 
+==== Custom qualifiers
+
+There are always 2 parties involved in using qualifiers:
+the target (method, parameter, etc) and the handler (`ArgumentFactory`, etc). Generally, both need to be _qualified_ identically to interact with each other.
```
stevenschlansker: I think it's clearer to associate the qualifiers exclusively with the *type* rather than the executable element it's attached to.

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
qualidafial: Correct, but this is the most we can express within the `Qualified` meta-annotation approach. For annotations with attributes you'll have to register your own resolver.

docs/src/adoc/index.adoc
```diff
@@ -2603,6 +2603,47 @@ and setter parameters.
...
+will receive _any_ target along with its qualifier data, and must use its own logic
+to decide to interact with a given value.
+
+Consider this transparent `@javax.annotation.NonNull` enforcer example:
+
+[source,java]
+----
+class NonnullColumnMapperFactory implements QualifiedColumnMapperFactory {
+    @Override
+    public Optional<ColumnMapper<?>> build(QualifiedType<?> type, ConfigRegistry config) {
```
stevenschlansker: This could be a reference to the test code, to ensure it doesn't drift out of sync with the actual implementation.

docs/src/adoc/index.adoc
qualidafial: * Qualifier on method: qualifies return type
* Qualifier on type: qualifiers the type the class operates on, usually.
  * On `ArgumentFactory`, qualifies input argument type
  * On `ColumnMapper<T>`, qualifies `T`
  * On `ColumnMapperFactory`, qualifies the mapped type
* Qualifier on parameter: qualifies parameter type

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
stevenschlansker: I guess I don't have a better suggestion at the moment anyway.

core/src/test/java/org/jdbi/v3/core/qualifier/TestQualifiers.java
```diff
@@ -13,56 +13,115 @@
...
+
     @Test
     public void getQualifiers() throws Exception {
-        Qualifiers qualifiers = new Qualifiers();
-
         assertThat(foo(1))
             .isEqualTo(foo(1))
             .isNotEqualTo(foo(2));
 
-        assertThat(qualifiers.findFor(getClass().getDeclaredField("qualifiedField")))
```
stevenschlansker: I think the fact that the tests have to use reflection to find methods is an indication that the API isn't quite right.  Qualifiers are part of `core`, yes, but determining the types of methods is not.

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
leaumar: Guess I left out the word "_this_ is not guaranteed to _retroactively_ affect sqlobjects"

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
leaumar: I tried with `new` in the `before` and jdbicache still gave me nullpointers

leaumar: Also, the other block has to be null-safe anyway for the illegitimate uses of this class, so this way at least it's consistent

docs/src/adoc/index.adoc
leaumar: This isn't actually the impl used in tests because that one is more complex for the sake of thorough testing (different exception because an NPE can be anything)

core/src/test/java/org/jdbi/v3/core/qualifier/TestQualifiers.java
leaumar: I'm not sure how else you would do this in unit test form? You gotta get a hold of a `method` or `field` instance somehow...

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
leaumar: This is what happens with a `new` registry:
image

I'm not quite immediately seeing a logical explanation for it

leaumar:
some remarks handled

core/src/main/java/org/jdbi/v3/core/qualifier/Qualifiers.java
leaumar: Huh... adding the `new` registry makes `addResolver` not work...

leaumar: The resolver is added to `resolvers` (size = 3), and then in the next invocation on `qualifiers`, it's gone (size = 2)??

leaumar: Ugh, it causes some sort of problem with jdbicache's loader creating a new instance of Qualifiers that doesn't inherit the resolvers list or something :/

leaumar: I overlooked that the solution was in your post all along 😅

hgschmie:
so I read this PR and I am not really sure what it accomplishes. It is stalled since 2019 and I am inclined to close it. I will mark it as stale and close 30 days from now.

PR #1598, opened 2019-09-01
KotlinMapper: handle @PropagateNull

### LINKED ISSUE #1584: @PropagateNull is not supported for KotlinMapper
The `@PropagateNull` support added in #1543 doesn't seems like work for kotlin classes using KotlinMapper. Can someone confirm?

stevenschlansker:
Fixes #1584


kotlin/src/test/kotlin/org/jdbi/v3/core/kotlin/KotlinMapperTest.kt
```diff
@@ -554,4 +555,30 @@ class KotlinMapperTest {
...
+        var name:String? = null
+    }
+
+    @Test
+    fun memberPropagateNull() {
+        assertThat(handle.select("select 1 as id, null as name")
+                .mapTo<MemberWithPropagateNull>()
+                .one())
+            .isNull()
+    }
```
qualidafial: Can we also get a test for nullable constructor parameters with `@PropagateNull`? Or is that not a valid use case?

PR #1603, opened 2019-09-07
Postgres: support EnumSet->varbit column binding/mapping

leaumar:
Old feature written by @arteam , updated where needed and refactored. Turns EnumSets into bitstrings like `01001` (meaning enum constants 1 and 4 are present, the rest are absent). I presume postgres has special support for such strings, otherwise this might as well go in core, but I personally don't think it's useful enough for that.

I'm not sure if we maybe want a Qualifier on this or anything, for backward compatibility. We don't currently have any kind of EnumSet integration though (aside from the Collector, so you need `@SingleValue` for this), and user-written EnumSet features will still be registed later and probably take precedence anyway.

Not gonna push to merge this, I was just going through dead git branches to clean up and found a feature I wanted to give a fair chance at being merged.

### DIFF
--- a/postgres/src/main/java/org/jdbi/v3/postgres/internal/BitStringEnumSetColumnMapper.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/internal/BitStringEnumSetColumnMapper.java
@@ -11,7 +11,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jdbi.v3.postgres;
+package org.jdbi.v3.postgres.internal;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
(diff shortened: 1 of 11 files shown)


leaumar:
@arteam do you wanna add anything yourself since you started the work on this?

stevenschlansker:
How come we're using bitstrings rather than `byte[]`?

postgres/src/main/java/org/jdbi/v3/postgres/internal/BitStringEnumSetColumnMapper.java
```diff
@@ -0,0 +1,54 @@
...
+    public EnumSet<E> map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
+        String bits = r.getString(columnNumber);
+
+        if (bits == null) {
+            return null;
+        }
+        if (bits.length() != enumConstants.length) {
+            throw new IllegalArgumentException("bit string \"" + bits + "\" for " + enumType + " should not contain " + bits.length() + " characters");
+        }
+        if (!bits.matches("^[01]+$")) {
```
stevenschlansker: This is probably a very slow operation involving a regex computation for every cell.  Any such check should probably be done when we iterate `charAt` below.

leaumar: #1634

PR #1626, opened 2019-11-25
DurationColumnMapperFactory uses BigDecimal for math operations to av…

ScottFechser:
I found an issue where certain interval values would be off by a microsecond when converted to a duration by jdbi. This PR fixes the issue.

Also, this is my first time contributing to this project, please let me know if I need to do anything differently, Thanks

### DIFF
--- a/postgres/src/main/java/org/jdbi/v3/postgres/DurationArgumentFactory.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/DurationArgumentFactory.java
@@ -13,6 +13,7 @@
  */
 package org.jdbi.v3.postgres;
 
+import java.math.BigDecimal;
 import java.sql.Types;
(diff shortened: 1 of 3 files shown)


stevenschlansker:
Thank you for the contribution!  I note that the `DurationArgumentFactory` has similar code that didn't receive the fix:
```java
        double seconds = d.getSeconds() + d.getNano() / 1e9;
        final PGInterval interval = new PGInterval(0, 0, (int) days, hours, minutes, seconds);
```

Do you think that code is safe as is?  Otherwise I'd apply the same fix to the argument factory as well, to ensure all calculations are symmetric.

ScottFechser:
@stevenschlansker `DurationArgumentFactory` appeared to work as it was, but to be safe, I added the recommended change. Thanks

PR #1630, opened 2019-12-05
Use the new fluent configuration for Flyway

roaraya8:
Move to the new fluent configuration for Flyway as the classic configuration is being deprecated in `Flyway 6.x.x` (see https://github.com/flyway/flyway/issues/2218) and makes `jdbi3-testing` fail if projects use Flyway `6.0.0+`.


stevenschlansker:
The dependency version is only for tests and is optional, so it shouldn't cause too many problems.

PR #1634, opened 2019-12-14
restore validation of bitstring content

leaumar:
1fdb5a32a4044e33f6f20a32ddd78887f2796b4f removed the validation entirely, this PR restores the validation (but more performantly) and adds the test for it.

### DIFF
--- a/postgres/src/main/java/org/jdbi/v3/postgres/internal/BitStringEnumSetColumnMapper.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/internal/BitStringEnumSetColumnMapper.java
@@ -16,7 +16,6 @@ package org.jdbi.v3.postgres.internal;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.EnumSet;
-import java.util.function.IntConsumer;
 import java.util.stream.IntStream;
 
 import org.jdbi.v3.core.mapper.ColumnMapper;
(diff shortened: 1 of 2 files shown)


leaumar:
> PSQLException: ERROR: cached plan must not change result type

postgres/src/main/java/org/jdbi/v3/postgres/internal/BitStringEnumSetColumnMapper.java
```diff
@@ -44,8 +45,18 @@
...
         return IntStream.range(0, bits.length())
+            .peek(checkIfBitIn(bits))
             .filter(i -> bits.charAt(i) == '1')
             .mapToObj(i -> enumConstants[i])
             .collect(toCollection(() -> EnumSet.noneOf(enumType)));
     }
+
+    private static IntConsumer checkIfBitIn(String bits) {
+        return i -> {
+            char bit = bits.charAt(i);
```
stevenschlansker: You could just fold this check into the `map`, then you save a stream pipeline and one call to `charAt`.  Not that it matters all that much...

postgres/src/test/java/org/jdbi/v3/postgres/TestEnumSets.java
```diff
@@ -148,6 +149,26 @@ public void testAmountPlatforms() {
         assertThat(amount).isEqualTo(3);
     }
 
+    @Test
+    public void throwsOnNonBitChars() {
+        db.getHandle().useTransaction(handle -> {
+            // redefine column to varchar type
+            handle.execute("drop table if exists videos");
```
stevenschlansker: The test error seems to be due to redefining the table in a transaction while the client has cached data.  Consider using a new table instead of redefining an existing table.

PR #1657, opened 2020-02-01
WIP concept of nicer json mapping DSL for fluent API

alwins0n:
Hey folks,

since we work a lot with postgres and are experimenting with json, I revisited the plugins after using jdbi a lot with the "stringtype=unspecified" hack.

I am happy to see that there are json plugins which seem to work nicely with postgres.

One caveat I found reading the docs is that the SQLObject API reads a lot nicer than the fluent API. Since I work mostly with the latter, I was pondering if there is a way to achieve a similarly nice user experience as with the `ConstructorMapper` class. 

```
h.registerRowMapper(ConstructorMapper.of(MyType.class))
  .createQuery(...)
  ....
```

I came up with the following draft, which would probably require moving the existing Argumentfactory of the postgres plugin to the json plugin (which may be a "hack" atm anyway, see `PostgresPlugin` class.

I am willing to work on this, if of interest, but I could use some hint on how to best implement the RowMapper(-factory?) for the jsonmapper

### DIFF
--- a/json/src/main/java/org/jdbi/v3/json/JsonArgument.java
+++ b/json/src/main/java/org/jdbi/v3/json/JsonArgument.java
@@ -1,0 +1,22 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.json;
+
+import org.jdbi.v3.core.argument.ArgumentFactory;
+
+public class JsonArgument {
+    public static ArgumentFactory of(Class<?> cls) {
+        throw new UnsupportedOperationException("TODO move/use JsonArgumentFactory from pg plugin?");
+    }
+}
--- a/json/src/main/java/org/jdbi/v3/json/JsonMapper.java
+++ b/json/src/main/java/org/jdbi/v3/json/JsonMapper.java
@@ -16,6 +16,8 @@ package org.jdbi.v3.json;
 import java.lang.reflect.Type;
 
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.mapper.ColumnMapper;
+import org.jdbi.v3.core.mapper.RowMapper;
 import org.jdbi.v3.meta.Beta;
 
 /**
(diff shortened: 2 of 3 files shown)


stevenschlansker:
Hi @alwins0n thanks for looking into improving the api.  I read through the test case and if I understand correctly, you don't like the syntax where you have to write out the qualified type explicitly?
```java
static final var MY_POJO_JSON = QualifiedType.of(MyPojo.class).with(Json.class);
q
    .bindByType("myPojo", myPojo, MY_POJO_JSON)
    ...
```

Your current proposal fixes this for one case, the `@Json` annotation.
But it looks like this might be a more general problem: you want to bind a bunch of objects, but assume their type is qualified.

What if we extend the `Binding` (and perhaps `ResultBearing` as well) to allow you to declare default qualifiers for types -- something like:

```java
jdbi.getConfig(Qualifiers.class).defaultQualify(MyBean.class).with(Json.class);
MyBean myBean = ... ;
jdbi.createQuery(...)
      .bind("myBean", myBean)  // automatically marked as Json
      .mapTo(MyBean.class);     // here too
```
What do you think?

alwins0n:
yeah you got my intention right. 

in essence I think it goes well with the separation of concerns which is otherwise nicely implemented - being:

`registerRowMapper` or the annotation `@RegisterRowMapper` is separated from the actual call which would be `createQuery(..).bind(..)` or the sql object implementation.

so it registers some metadata which causes the "normal" jdbi code to work. Which has a nice advantage. Usually in my repositories I centralize the setup code (mappers, arguments, etc) in the constructor on the injected jdbi instance and just use jdbi as if its has always been there.

`bindByType` breaks this a little bit. now i have to adapt the code to make it work, not the configuration.

`@Json` on the other hand does this nicely. another metadata annotation which causes the existing code to work.

So - in that respect, I think your solution is a more general one and allows jdbi users to separate configuration from actual invocation as I layed out. If you agree this is where I will take this

alwins0n:
ping @stevenschlansker : if you read my comment - your solution captures my intent: to separate config from actual querying in the fluent DSL.

how should we proceed?

alwins0n:
@qualidafial @stevenschlansker can I get some feedback on this one?

alwins0n:
last response was almost a year ago. closed

stevenschlansker:
Sorry @alwins0n, didn't mean to leave you hanging here.  I'll re-open this for now and take a look at it - I can't right this moment but I will get to it soon.

stevenschlansker:
Hi again @alwins0n, I do think the idea for making binding easier is good.  Ideally it will work independent of the actual qualifying annotation - it should work with `@UserDefinedAnno` just as well as `@Json` if possible.  If we can get it to that point, I think we can accept the improvement.  Seem reasonable to you?  I also understand if it's been too long and you still want to abandon this work.  Sorry again for the delay in reply.

alwins0n:
I may pick work up on this - I will get back to you soon

hgschmie:
This is now hanging again for > 2 years. marking as stale, will close in 30 days.

hgschmie:
Closing as stale

Issue #1454, opened 2019-02-07
Support R2DBC


VaughnVernon:
Implement support for R2DBC. This may be something such as `jdbi-r2dbc`. FYI:

https://r2dbc.io/
https://github.com/r2dbc

Also, we support Jdbi in the `vlingo-symbio` which is part of our reactive platform. Check it out. We'd love to get your team up to speed on our full platform.

https://github.com/vlingo/vlingo-symbio-jdbc/tree/master/src/main/java/io/vlingo/symbio/store/object/jdbc/jdbi

stevenschlansker:
Hi Vaughn, thanks for the pointer.  Reactive is very new and exciting but my understanding is that the current implementation of JDBC, and in fact the database server software itself, makes it a poor candidate for reactive concepts at the lower layers.  The database server has worker processes that take queries and produce results, and each process will work on its query until completion.  There's no ability to asynchronously submit 10 queries and await their results.

This means that as you implement JDBC reactive, you eventually find that you must implement an old style thread pool that holds your connections / handles, services queries, and then emits results into the reactive framework.  This means that you can have some of the benefits of reactive, in particular you can write reactive code, but it seriously limits the scalability benefits that reactive attempts to achieve.

So we'd definitely welcome API integrations, but unless the state of the JDBC world has moved on since I last investigated, it'll be a somewhat limited wrapper that just hides the fact that the code underneath is still old school threaded code.

VaughnVernon:
Hey, Steven! Thanks for your comments. Currently we solve the reactive-over-JDBC by serializing requests within actors in our `vlingo-symbio` implementations. You can create as many `JDBCObjectStoreActor` as makes sense for your connection limitations, but the actors themselves access JDBC synchronously. It's more that the client of the `ObjectStore` implementations running on actors don't block while queries are being fulfilled.

I don't have a thorough understanding of R2DBC, but it is touted as being a full JDBC replacement and works completely asynchronously as a database driver. There are currently implementations for Postgres and MS SQL Server only, but probably will grow in time. I was wondering what it would be like for Jdbi to implement over R2DBC. I understand if this doesn't make sense for you now, but the reference is here in case it does at some future time. No stress :)

stevenschlansker:
Interesting, they actually do provide their own drivers.  So maybe there is some benefit to be had here!

[...]

PR #1660, opened 2020-02-10
Support R2DBC

mcgin:
This is a fiirst attempt at bringing the r2dbc-client into the jdbi project as per https://github.com/jdbi/jdbi/issues/1454

This is still a WIP and the intent is to get a discussion going around how this might look rather than merge it in it's current form.

For now it is more or less a copy of the existing r2dbc-client with minimal changes, git history should be pulled over from the r2dbc-client project too.

The main changes from the existing client are:

- Package renamed from io.r2dbc.client -> org.jdbi.v3.r2dbc
- Various changes to styling etc. to satisfy the checks enforcing the jdbi conventions
- dependency updates
- Removal of Changelog generator file in the test/util

I've tried to keep the changes as separate from the jdbi project as much as possible, with only miinor changes to the parent pom.

The r2dbc pom.xml file will need some cleaning up, I've simply commented out most of the old r2dbc project specific stuff for now, but likely that will all be deleted


mcgin:
I've done some additional cleaning and this is probably ready for discussion.

The core library seems pretty lean and fit for purpose to me although I wonder whether it should depend explicitly on a particular implementation of reactive streams as it does now.  I think the tests perhaps need a bit of work, I don't think we really need spring boot as a dependency here.

Points I think need discussion/agreement:

- Does the overall API/integration of this module make sense?
- Should the exposed interface be reliant on Project Reactor as it is now, or should it be refactored to be based on the reactive streams spec - i.e. `Publisher<?>` vs `Mono<?>/Flux<?>`.  This probably requires splitting the module in two, an api (r2dbc-core) and an implementation (r2dbc-reactor).
- Do we spend the time now to remove the spring boot dependency or a refactor later to remove it.  As it's a test dependency either approach is sensible imho.
- Does it make sense to include examples of all the different drivers in the test project (e.g. H2Example, PostgresqlExample etc.). Should we push those tests to the driver projects themselves?

@mp911de @stevenschlansker as you both were involved in the original issue what are your thoughts on the above?

mp911de:
Awesome work! Here are my thoughts on your questions.

> Should the exposed interface be reliant on Project Reactor

I have no strong opinion. In any case, a raw `Publisher` is unusable from a user perspective. The API should expose a higher-level reactive type to make it usable. As of now, there are two options:

* Project Reactor
* RxJava 3 (RxJava 2 is being phased out as RxJava 3 is getting closer to its release)

> Do we spend the time now to remove the spring boot dependency

I think it makes a lot of sense to remove the Spring Boot dependency. Luckily we have a JDBC client with JDBI available here. Spring `JdbcTemplate` was used to avoid all the fiddling with `SQLException` and resource management.

> Does it make sense to include examples 

No, we should remove these. The examples were mostly intended for compatibility/verification. Drivers should not depend on a client that actually uses the driver itself to avoid cycles. If we feel that we need the examples, then we can get them back from the original r2dbc-client repository.

mcgin:
> > Do we spend the time now to remove the spring boot dependency
> > Does it make sense to include examples

I've removed these now and it significantly reduces the dependency chain of the tests.  
Although I'm wondering should there be a replacement test of some kind or not.

I'll await input from the JDBI team before making any further adjustments.

stevenschlansker:
Hi guys, thanks for preparing this.
I'm concerned that it feels like it duplicates a lot of existing Jdbi API surface but in a very slightly different way.
Even core concepts like Handle now have two different classes.

Do you think there's any possibility to make the additional functionality reuse the existing API surface as much as possible?  I was imagining an extension more along the lines of `.mapTo(new GenericType<Flow<MyPojo>>(){})` as opposed to an entirely parallel set of APIs to support

mp911de:
Hey Steven, I did a pass over the objects that are provided by JDBI and how they compare to the types provided by this PR:

* `ResultBearing` returns `ResultIterable`, `ResultIterable` is blocking API
* `Handle` returns `StatementBuilder` that works with JDBC `Statement` (JDBC functionality is not available from R2DBC)
* `Batch` -> `execute()` method is blocking

JDBI is what it is, it's a JDBC based library and from that perspective, it makes a lot of sense to build on top of JDBC. R2DBC does not use nor provide JDBC API, rather it's a specification on its own. Each bit of JDBC is blocking. R2DBC is non-blocking. Hiding R2DBC behind a blocking API does not make any sense at all.

Over at Spring Framework, a lot of lower-level components could be reused when WebFlux and reactive Spring Data projects were built. The resource interaction and even some mapping components required a parallel set of API which comes with a lot of code.

mcgin:
I think a good analogy is to look at how Spring adopted the Reactive Streams spec on the web tier.  They provide a blocking web api through the `spring-mvc` module which builds on top of the servlet API - a blocking API.  In addition they provide a non blocking module `spring-webflux` which builds on the Reactive Streams API to provide a non blocking alternative.

In a similar way you could think jdbi might provide the blocking database access through the combination of jdbi-core and jdbc (JDBC is a blocking API), and a non blocking version with this module (which could do with a new name) combined with r2dbc.

Spring's web model:

Spring Web Tier

stevenschlansker:
Hi guys thanks again for all your work here.  I appreciate the similarity to Spring's approach with traditional servlet vs async vs Reactive, and one decision they made that always disappoints me is that each essentially has their own parallel and incompatible hierarchies.  It feels like we are duplicating that approach here, with parallel `Handle` and other implementations, that do not reuse the core Jdbi classes.

As a counterexample, consider the approach taken by JAX-RS: you use one core api to declare what the request looks like, and then it has different invocation strategies (async, reactive, etc).  But, critically: the declaration api is independent of the invocation api, you can freely switch without updating the request declaration.

I am not sure how or if it's possible to retrofit the existing API to serve the new need in the way that I want.  But, I do not think I want to merge such a large new API surface like you propose into Jdbi so quickly.  New APIs will have to be maintained forever, and I am not personally familiar with successful uses of the new reactive feature.

What would you say to starting this project off as an extension module that you host yourself - and we can consider it for possible inclusion in core later?  I believe this will give you much more flexibility in terms of release schedule to grow the community using this "reactive" jdbi and be more responsive to their needs.  We would be happy to consider any needed improvements into jdbi core to make hosting this as an out-of-tree module straightforward.

mp911de:
Thanks Steven for your guidance. It's always hard to retrofit existing API to cater for multiple usages with a different API behind. Most projects end up with separate hierarchies (such as Spring `RestTemplate` vs. `WebClient`, Apache Cassandra with their sync/async and reactive support, MongoDB shipping multiple API facades).

JAX-RS consists of mostly annotations. At the time the client API was built, folks already knew about various executor requirements. TBH, the reactive client API is a mess with `request().rx(…).get(…)`.

Having something like

```java
Publisher<Integer> result = handle.createUpdate("INSERT INTO user(id, name) VALUES (?, ?)")
            .bind(0, 1)
            .bind(1, "Bob")
            .execute(ReactiveInvoker.class).getPublisher();
```

in the long run could be desirable but I'm not sure how much it takes to get there.

I'm happy to guide whoever takes a spike on a reactive JDBI on how to use R2DBC and to learn from use-cases that we can support in R2DBC.

rstoyanchev:
> I appreciate the similarity to Spring's approach with traditional servlet vs async vs Reactive, and one decision they made that always disappoints me is that each essentially has their own parallel and incompatible hierarchies.

The `@Controller` annotated programming model is the same in Spring MVC and WebFlux. The core APIs however (request, response, handler mapping, adapter, etc) could not be the same because they have sync/blocking semantics, and a hard dependency on the Servlet API that can't work for runtimes like Netty. 

JAX-RS was created to be independent of the Servlet API which does helps but in terms of async API's and reactive support it's far from ideal.

Servlet 3.1 added non-blocking I/O in place. The same `ServletInputStream` and `ServletOutputStream` but instead of reading them as `InputStream` and `OutputStream`, you set a ReadListener` and `WriteListener` on them. Minimal API surface change, but for users it's an API with many otherwise innocent-looking parts of the API (e.g. request parameters) that are landmines to be avoided, and ultimately require re-thinking and parallel solutions when used in non-blocking mode.

pom.xml
```diff
@@ -84,20 +85,27 @@
...
         <dep.jetbrainsAnnotations.version>13.0</dep.jetbrainsAnnotations.version>
+        <dep.junit5.version>5.5.2</dep.junit5.version>
         <dep.kotlin.version>1.3.61</dep.kotlin.version>
         <dep.plugin.checkstyle.version>3.0.0</dep.plugin.checkstyle.version>
         <dep.plugin.pmd.version>3.9.0</dep.plugin.pmd.version>
         <dep.plugin.sortpom.version>2.8.0</dep.plugin.sortpom.version>
         <dep.plugin.spotbugs.version>3.1.10</dep.plugin.spotbugs.version>
         <dep.pmd.version>6.14.0</dep.pmd.version>
+        <dep.r2dbc.version>Arabba-SR2</dep.r2dbc.version>
+        <dep.reactor-netty.version>0.9.4.RELEASE</dep.reactor-netty.version>
```
x80486, line 104: Did you consider [SmallRye Mutiny](https://smallrye.io/smallrye-mutiny/index.html) also?

At this stage (in my opinion) it's comparable with Reactor and/or RxJava — plus it's already part of MicroProfile.

hgschmie:
as much as I would love to see R2JDBC support in jdbi, there is not a lot of bandwidth to work on this. I will mark it as stale and close 30 days from now.

mp911de:
Then it would make sense to close this one directly as there is no intent to merge the PR in the first place.

stevenschlansker:
Thank you for your interest and hard work everyone.

Issue #1652, opened 2020-01-21
Add a useful toString for KotlinPropertyArguments


sharedprophet:
For error messages, it would be helpful to have KotlinPropertyArguments print out the toString of the obj parameter.

PR #1667, opened 2020-02-20
#1652 useful toString for KotlinPropertyArguments

sharedprophet:


stevenschlansker:
Thanks for the improvement :)

PR #1675, opened 2020-02-28
feature: enable forbidding binding null as a primitive

leaumar:
I.e. `.bindByType("foo", null, int.class)` becomes illegal if opted in. I recommend reviewing commit per commit since a method rename is involved.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/argument/PrimitivesArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/PrimitivesArgumentFactory.java
@@ -13,10 +13,27 @@
  */
 package org.jdbi.v3.core.argument;
 
+import java.lang.reflect.Type;
 import java.sql.PreparedStatement;
 import java.sql.Types;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.BiFunction;
+import java.util.function.Function;
 
+import org.jdbi.v3.core.argument.internal.StatementBinder;
+import org.jdbi.v3.core.argument.internal.strategies.LoggableBinderArgument;
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
+// TODO this factory isn't actually preparable anymore
 class PrimitivesArgumentFactory extends DelegatingArgumentFactory {
+    private final Map<Class<?>, BiFunction<Object, ConfigRegistry, Argument>> builders = new IdentityHashMap<>();
+
     PrimitivesArgumentFactory() {
         register(boolean.class, Types.BOOLEAN, PreparedStatement::setBoolean);
         register(byte.class, Types.TINYINT, PreparedStatement::setByte);
@@ -26,5 +43,44 @@ class PrimitivesArgumentFactory extends DelegatingArgumentFactory {
         register(long.class, Types.INTEGER, PreparedStatement::setLong);
         register(float.class, Types.FLOAT, PreparedStatement::setFloat);
         register(double.class, Types.DOUBLE, PreparedStatement::setDouble);
+    }
+
+    @Override
+    public Optional<Argument> build(Type expectedType, Object value, ConfigRegistry config) {
+        Class<?> expectedClass = getErasedType(expectedType);
+
+        if (value != null && expectedClass == Object.class) {
+            expectedClass = value.getClass();
+        }
+
+        return Optional.ofNullable(builders.get(expectedClass)).map(r -> r.apply(value, config));
+    }
+
+    @Override
+    public Optional<Function<Object, Argument>> prepare(Type type, ConfigRegistry config) {
+        return Optional.empty();
+    }
+
+    @Override
+    public Collection<? extends Type> prePreparedTypes() {
+        return Collections.emptySet();
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    <T> void register(Class<T> klass, int sqlType, StatementBinder<T> binder) {
+        builders.put(klass, (value, config) -> {
+            if (value != null) {
+                return new LoggableBinderArgument<>((T) value, binder);
+            }
+
+            if (config.get(Arguments.class).isBindingNullToPrimitivesPermitted()) {
+                return new NullArgument(sqlType);
+            }
+
+            throw new IllegalArgumentException(String.format(
+                "binding null to a primitive %s is forbidden by configuration, declare a boxed type instead", klass.getSimpleName()
+            ));
+        });
     }
 }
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/jdbi/v3/core/argument/DelegatingArgumentFactory.java
```diff
@@ -42,16 +44,37 @@
...
+        preparableBuilders.put(klass, value -> value == null ? new NullArgument(sqlType) : new LoggableBinderArgument<>((T) value, binder));
     }
 
     @SuppressWarnings("unchecked")
-    <T> void register(Class<T> klass, int sqlType, StatementBinder<T> binder) {
-        builders.put(klass, value -> value == null ? new NullArgument(sqlType) : new LoggableBinderArgument<>((T) value, binder));
+    <T> void registerHot(Class<T> klass, int sqlType, StatementBinder<T> binder) {
+        hotBuilders.put(klass, (value, config) -> {
+            if (value == null) {
+                if (klass.isPrimitive() && !config.get(Arguments.class).isBindingNullToPrimitivesPermitted()) {
```
leaumar: `klass.isPrimitive()` is redundant in the current impl (no non-primitive uses), but I'm not sure removing the check because of mere circumstances would be wise

leaumar: actually, hang on, I'll refactor this tomorrow to make more structural sense

leaumar:
It works in 529b067da3bf6c3c4d81c2cf07d9a5650f0fa19d, but then the better implementation in dea687bc83895449a81ba7c428d9bcb927ba9599 doesn't receive the change in config. @stevenschlansker is this because of `Preparable`  somehow? Do the builders get curried persistently for an open handle instead of per execution?

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
```diff
@@ -70,9 +70,9 @@
     @Override
     public Collection<? extends Type> prePreparedTypes() {
         return FACTORIES.stream()
-                .map(ArgumentFactory.Preparable::prePreparedTypes)
```
stevenschlansker: This looks like a spacing-only change, should it be removed from the PR?

core/src/main/java/org/jdbi/v3/core/argument/PrimitivesArgumentFactory.java
```diff
@@ -13,10 +13,27 @@
...
 
+import org.jdbi.v3.core.argument.internal.StatementBinder;
+import org.jdbi.v3.core.argument.internal.strategies.LoggableBinderArgument;
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
+// TODO this factory isn't actually preparable anymore
 class PrimitivesArgumentFactory extends DelegatingArgumentFactory {
+    private final Map<Class<?>, BiFunction<Object, ConfigRegistry, Argument>> builders = new IdentityHashMap<>();
```
stevenschlansker: It's weird to extend `DelegatingArgumentFactory` but not use its functionality, but instead replicate it here.  Feels like a design smell.

core/src/main/java/org/jdbi/v3/core/argument/PrimitivesArgumentFactory.java
```diff
@@ -13,10 +13,27 @@
...
+import java.util.function.BiFunction;
+import java.util.function.Function;
 
+import org.jdbi.v3.core.argument.internal.StatementBinder;
+import org.jdbi.v3.core.argument.internal.strategies.LoggableBinderArgument;
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+import static org.jdbi.v3.core.generic.GenericTypes.getErasedType;
+
+// TODO this factory isn't actually preparable anymore
```
stevenschlansker: We should fix this concern before merging.

core/src/main/java/org/jdbi/v3/core/argument/PrimitivesArgumentFactory.java
```diff
@@ -27,4 +44,43 @@
...
+    @Override
+    public Collection<? extends Type> prePreparedTypes() {
+        return Collections.emptySet();
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    <T> void register(Class<T> klass, int sqlType, StatementBinder<T> binder) {
+        builders.put(klass, (value, config) -> {
+            if (value != null) {
```
stevenschlansker: Most of this lambda's functionality is duplicated from the super class.  Can we wrap the `binder` argument with the new null check and call `super.register` to handle the existing functionality?  That would be much more in line with the expected use of the abstract superclass.

core/src/main/java/org/jdbi/v3/core/argument/PrimitivesArgumentFactory.java
leaumar: Removing the `extends` would technically be breaking API, though for this case I'm hoping you'll say we can shrug that off

leaumar:
How can it be a `Preparable` if the arguments need to check the live context when invoked though? I thought the point of Preparable was to be sort of fixed, immutable. I tried to keep it Preparable but then I actually had to change the contract of Preparable to include the context.

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
leaumar: collateral no-consequences formatting correction, may as well merge it? Or are 8-space indents not accidents?

core/src/main/java/org/jdbi/v3/core/argument/PrimitivesArgumentFactory.java
leaumar: Same as other comment, how can it be preparable if the arguments are dynamic?

core/src/main/java/org/jdbi/v3/core/argument/BuiltInArgumentFactory.java
stevenschlansker: Well, that's what my IDE wants to do.  As soon as you change it to 4, if I edit the file and update the indentation, Eclipse changes it right back :/ so we end up in format edit wars.  That's part of why I don't like seeing format changes included in otherwise unrelated MRs, it tends to introduce unnecessary conflicts.

stevenschlansker: The file's inconsistent to start with though, so maybe it's worth fixing it here.

leaumar: I agree on avoiding formatting wars, but yeah, this and most files use 4-space indents even in these structures, so this (and your eclipse's behavior) seems like a fluke to me. I'll take this up separately from this PR then.

stevenschlansker:
Merged as #1691

PR #1690, opened 2020-04-14
Use kotlin default for null columns

### LINKED ISSUE #1689: Map to Kotlin default value if column value is null and property is not nullable
When creating data objects in Kotlin, the intuitive expectation is that having a default value for a non-null property will map null columns to the default value. However currently `KotlinMapper` throws `InvocationTargetException` because it attempts to set the value to null.

It seems `KotlinMapper` currently only uses the default value if the column is not mapped in the query at all, but it should also use the default value for non-null properties when encountering a null value from the db.

sharedprophet:
fix #1689


stevenschlansker:
Hm, the build failed.  Might be missing an import statement?

sharedprophet:
I couldn't figure out how to see what happened on the CI server, but I was able to build and pass all tests locally...

stevenschlansker:
https://travis-ci.org/github/jdbi/jdbi/jobs/674955929

sharedprophet:
I fixed it and it didn't kick off the build again...

stevenschlansker:
Are you sure you pushed your change?  I don't see new commits on this PR.

sharedprophet:
Hmmmm...

sharedprophet:
Well, it is correct that that import is missing, and I don't seem to have a new commit locally. But now I am getting an error saying a folder already exists when it tries to unpack embedded postgres on a local build?

sharedprophet:
Ok, I deleted the folder in question and it ran through the build successfully. Hopefull the CI build will work now...

PR #1702, opened 2020-05-10
Add Kotlin sqlobject extensions taking a KClass

conleym:
It's not always possible to use the reified type versions of these extensions (because you're passed a `KClass` instance, for example). In such a case, one must ordinarily use the standard java versions of `onDemand` and `attach`, like so:

```
val sqlObject = jdbi.onDemand(kclass.java)
```

As a small convenience, I've added the obvious extensions, which can be used like this:

```
val sqlObject = jdbi.onDemand(kclass)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 dependency-reduced-pom.xml
+.flattened-pom.xml
 GPATH
 GRTAGS
 GTAGS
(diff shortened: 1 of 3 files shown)


fabianedl777:
Hello

I can uses val sqlObject: kclass = jdbi.onDemand() for my is more simple.

PR #1705, opened 2020-05-24
Proof of Concept for FreeBuilder Support

wjdix:
This is a proof of concept for adding FreeBuilder support to JDBI3 in line with the existing Immutables support. In fact, for the most part, it is implemented with a very slightly modified version of the Immutables code.

Obviously, this is in need of a great deal more polish, but I wanted to open this proof of concept PR to verify if this is a feature the maintainers are interested in adding to the library.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+# 3.14.0
+  - immutables: support getter with @ColumnName, #1704
+
 # 3.13.0
   - Kotlin: respect default values in methods when passed null, #1690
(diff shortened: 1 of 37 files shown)


stevenschlansker:
Thanks, it is exciting to have more integrations of this type!
It would be nice if we could share the majority of the code, as this is a fairly substantial addition that looks very similar to existing code.

stevenschlansker:
Oh, I see, you did actually already factor out some common logic into a new utility class.  Nice!  Yes, I think this is a reasonable feature to merge in once it's ready, polished, tested, and documented :)

wjdix:
Closing this in favor of #1713

Issue #1706, opened 2020-05-27
postgres: Read blob/bytea as stream via LargeObject api


gpsfl:
How do you read a blob (or bytea in postgres to be exact) as a stream using jdbi?

With normal jdbc it would be `ResultSet#getBinaryStream`, however I couldn't find any equivalent in jdbi. I tried using `mapTo(InputStream.class)` and `mapTo(Blob.class)`, but both failed with a `NoSuchMapperException`.

It seems to work with `mapTo(byte[].class)` however I'd like to avoid loading all the data at once (due to memory/performance).

leaumar:
I think supporting this would be a oneliner for us. I may be wrong though.
```java
mappers.put(InputStream.class, new GetterMapper<>(ResultSet::getBinaryStream));
```
Meanwhile, you could just register this on `ColumnMappers` in your config:
```java
public class InputStreamColumnMapper implements ColumnMapper<InputStream> {
    @Override
    public InputStream map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
        return r.getBinaryStream(columnNumber);
    }
}
```

stevenschlansker:
Do note that if we do add such a mapper, there's a good chance the data is not properly streamed at all.  I think it will be transferred as a `byte[]` and then read through some analogue of `ByteArrayInputStream`.  Maybe that's fine for your use case.  But if you do actually need to stream, the setup is a bit more complicated: https://jdbc.postgresql.org/documentation/head/binary-data.html

stevenschlansker:
@gpsfl , your problem statement is a bit contradictory: do you use a `bytea` or a `blob` column type?
If it's the former, I think the answer is that the driver cannot stream the data and you are best just retrieving as a `byte[]`.  If it's actually stored in a blob column, then this could be a feature request to support a facade over Postgres large object storage to make it easy to use in Jdbi.

gpsfl:
As far as I know postgres does not have a blob type (large objects are `oid`s). To clarify: We are currently using a `bytea` in postgres. We now switched to using a byte[] in jdbi based on what you said about it the binary stream just being a ByteArrayInputStream. However it would be great to see some support for postgres large object as streaming should provide a better performance.

stevenschlansker:
You're right, it seems Postgres doesn't have a blob type - but per jdbc docs, using the `oid` based large object storage is how you do streaming.  I agree, easy large object streaming is a good feature request.  Thanks for suggesting it!

[...]

PR #1711, opened 2020-06-04
Postgres: add simple LargeObject support

stevenschlansker:
https://github.com/jdbi/jdbi/issues/1706

### DIFF
--- a/postgres/src/main/java/org/jdbi/v3/postgres/BlobInputStreamArgumentFactory.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/BlobInputStreamArgumentFactory.java
@@ -1,0 +1,38 @@
+package org.jdbi.v3.postgres;
+
+import java.io.InputStream;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Types;
+
+import org.jdbi.v3.core.argument.AbstractArgumentFactory;
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.statement.StatementContext;
+
+class BlobInputStreamArgumentFactory extends AbstractArgumentFactory<InputStream> {
+    BlobInputStreamArgumentFactory() {
+        super(Types.BLOB);
+    }
+
+    @Override
+    protected Argument build(InputStream value, ConfigRegistry config) {
+        return new LobInputStreamArgument(value);
+    }
+
+    static class LobInputStreamArgument implements Argument {
+        private final InputStream value;
+
+        LobInputStreamArgument(InputStream value) {
+            this.value = value;
+        }
+
+        @Override
+        public void apply(int pos, PreparedStatement stmt, StatementContext ctx) throws SQLException {
+            PgLobApi lob = ctx.getConfig(PostgresTypes.class).getLobApi();
+            long oid = lob.createLob();
+            lob.writeLob(oid, value);
+            stmt.setLong(pos, oid);
+        }
+    }
+}
--- a/postgres/src/main/java/org/jdbi/v3/postgres/ClobReaderColumnMapperFactory.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/ClobReaderColumnMapperFactory.java
@@ -1,0 +1,23 @@
+package org.jdbi.v3.postgres;
+
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.mapper.ColumnMapper;
+import org.jdbi.v3.core.mapper.ColumnMapperFactory;
+import org.jdbi.v3.postgres.BlobInputStreamColumnMapperFactory.LobColumnMapper;
+
+class ClobReaderColumnMapperFactory implements ColumnMapperFactory {
+    @Override
+    public Optional<ColumnMapper<?>> build(Type type, ConfigRegistry config) {
+        if (Reader.class != type) {
+            return Optional.empty();
+        }
+        LobColumnMapper inner = new LobColumnMapper();
+        return Optional.of((rs, col, ctx) -> new InputStreamReader(inner.map(rs, col, ctx), StandardCharsets.UTF_8));
+    }
+}
--- a/postgres/src/main/java/org/jdbi/v3/postgres/PostgresTypes.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/PostgresTypes.java
@@ -20,6 +20,7 @@ import org.jdbi.v3.core.array.SqlArrayTypes;
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.config.JdbiConfig;
 import org.jdbi.v3.core.internal.exceptions.Unchecked;
+import org.jdbi.v3.meta.Beta;
 import org.postgresql.PGConnection;
 import org.postgresql.util.PGobject;
 
@@ -29,12 +30,14 @@ import org.postgresql.util.PGobject;
 public class PostgresTypes implements JdbiConfig<PostgresTypes> {
     private final Map<Class<? extends PGobject>, String> types = new ConcurrentHashMap<>();
     private ConfigRegistry registry;
+    private PgLobApi lob;
 
     @SuppressWarnings("unused")
     public PostgresTypes() {}
 
     private PostgresTypes(PostgresTypes that) {
         this.types.putAll(that.types);
+        this.lob = that.lob;
     }
 
     @Override
@@ -54,6 +57,20 @@ public class PostgresTypes implements JdbiConfig<PostgresTypes> {
         types.put(clazz, typeName);
 
         return this;
+    }
+
+    PostgresTypes setLobApi(PgLobApi lob) {
+        this.lob = lob;
+        return this;
+    }
+
+    /**
+     * Provide access to Large Object streaming via Postgres specific API.
+     * @return the postgres large object api
+     */
+    @Beta
+    public PgLobApi getLobApi() {
+        return lob;
     }
 
     /**
--- a/postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
+++ b/postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
@@ -1,0 +1,167 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.postgres;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.Reader;
+import java.sql.SQLException;
+import java.util.function.Supplier;
+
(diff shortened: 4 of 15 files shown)


postgres/src/main/java/org/jdbi/v3/postgres/BlobInputStreamArgumentFactory.java
```diff
@@ -0,0 +1,38 @@
...
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Types;
+
+import org.jdbi.v3.core.argument.AbstractArgumentFactory;
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.statement.StatementContext;
+
+class BlobInputStreamArgumentFactory extends AbstractArgumentFactory<InputStream> {
```
leaumar, line 13: Are there no other potential/reasonable use cases for an InputStream in queries, like are these postgres lobs equivalent to general lobs for all dbs? If there are uses that don't mix with this, this may need a qualifier

postgres/src/main/java/org/jdbi/v3/postgres/ClobReaderColumnMapperFactory.java
```diff
@@ -0,0 +1,23 @@
...
+import org.jdbi.v3.postgres.BlobInputStreamColumnMapperFactory.LobColumnMapper;
+
+class ClobReaderColumnMapperFactory implements ColumnMapperFactory {
+    @Override
+    public Optional<ColumnMapper<?>> build(Type type, ConfigRegistry config) {
+        if (Reader.class != type) {
+            return Optional.empty();
+        }
+        LobColumnMapper inner = new LobColumnMapper();
+        return Optional.of((rs, col, ctx) -> new InputStreamReader(inner.map(rs, col, ctx), StandardCharsets.UTF_8));
```
leaumar, line 21: I'm assuming this is similarly hardcoded in postgres then? ctrl-f "utf" on the doc page shows nothing.

postgres/src/main/java/org/jdbi/v3/postgres/PgLobImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import java.io.InputStream;
+import java.sql.Connection;
+import java.sql.SQLException;
+
+import org.postgresql.PGConnection;
+import org.postgresql.largeobject.LargeObject;
+import org.postgresql.largeobject.LargeObjectManager;
+
+class PgLobImpl implements PgLob {
+    private static final int BUF_SIZE = 1024 * 64;
```
leaumar, line 13: Any particular reason for this exact number? Wouldn't this make a good Config value?

postgres/src/main/java/org/jdbi/v3/postgres/PostgresTypes.java
```diff
@@ -56,6 +59,20 @@ public PostgresTypes registerCustomType(Class<? extends PGobject> clazz, String
         return this;
     }
 
+    PostgresTypes setLobApi(PgLob lob) {
```
leaumar, line 62: Is the name `PgLob` appropriate? It reads like an instance of a lob, not as the handling implementation for lobs.

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
```diff
@@ -0,0 +1,167 @@
...
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.postgresql.PGConnection;
+import org.postgresql.largeobject.LargeObjectManager;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TestLobStream {
+    private static final int BIGLY = 1024 * 1024 * 64;
```
leaumar, line 35: I'm missing an expression of what this is for and why it's the value that it is, e.g. `SAMPLE_SIZE_GREATLY_EXCEEDING_LIMIT` or a short comment?

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
```diff
@@ -0,0 +1,167 @@
...
+            assertSameChars(lob.findClob(2), expectedData.get());
+        });
+    }
+
+    private void assertSameBytes(InputStream a, InputStream b) throws IOException {
+        int pos = 0;
+        int read;
+        while ((read = a.read()) != -1) {
+            assertThat(b.read())
+                    .describedAs("byte at position %s", pos)
```
leaumar, line 77: is it a byte though? An int is 4 bytes

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
```diff
@@ -0,0 +1,167 @@
...
+            lob.insert(2, expectedData.get());
+            assertSameChars(lob.findClob(2), expectedData.get());
+        });
+    }
+
+    private void assertSameBytes(InputStream a, InputStream b) throws IOException {
+        int pos = 0;
+        int read;
+        while ((read = a.read()) != -1) {
+            assertThat(b.read())
```
leaumar, line 76: if I understand right, this is gonna be called for 1024*1024*64 ints, so isn't this test... really slow, asserting one int at a time?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
```diff
@@ -252,7 +275,11 @@ protected Object reducedResult(Stream<?> stream, StatementContext ctx) {
         }
     }
 
-    static class CollectedResultReturner<T> extends ResultReturner {
+    public static ResultReturner forCollecetedType(QualifiedType<?> returnType) {
```
leaumar, line 278: "colleceted"

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
```diff
@@ -109,7 +109,11 @@ private static Object checkResult(Object result, QualifiedType<?> type) {
         return result;
     }
 
-    static class VoidReturner extends ResultReturner {
+    public static ResultReturner forVoid() {
```
leaumar, line 112: what's with the changes in this file? Doesn't seem related, this method for example isn't called in this diff.

leaumar:
Reviewed with a relatively critical mindset, but I'm not a postgres guy so I'm unaware of some conventions that may or may not exist among postgresers.

postgres/src/main/java/org/jdbi/v3/postgres/ClobReaderColumnMapperFactory.java
stevenschlansker: Yeah, supporting other encodings is left as a future enhancement.

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
leaumar: I mean, the test limits the bounds to [0, 255], but everything about the impl says ints, so...

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
stevenschlansker: Yeah, this was part of a refactoring that turned out not to be necessary.

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
stevenschlansker: It's buffered in the driver so the test only takes a few seconds to run.

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
stevenschlansker: Yeah, `InputStream` works on bytes but the type is `int` since its *unsigned* bytes.

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
stevenschlansker: It's just a big amount of data that would cause problems to use inline.  Nothing more.

postgres/src/main/java/org/jdbi/v3/postgres/BlobInputStreamArgumentFactory.java
stevenschlansker: It only does that for now.  We can always change the name later if the class ends up doing more, it's not API.

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
leaumar: Sure, but from a functional perspective this impl reads like two different books. Might as well use the `byte` type and name everything as such

postgres/src/main/java/org/jdbi/v3/postgres/PgLobImpl.java
stevenschlansker: No, it's just picked as a "big enough" value.  I aligned it to the Postgres driver which uses 4K.
It's hard coded in the driver so we don't really have the ability to tune it properly, so we shouldn't expose configuring as API at this time.

stevenschlansker:
Thanks for the review.  I think I covered most of the important bits.

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
leaumar: I meant the assertions themselves, all the assertj internals churning. Wouldn't it make sense to shave off some seconds by asserting larger chunks?

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
leaumar: Can we give it a serious name expressing that then instead of a cheeky one?

postgres/src/main/java/org/jdbi/v3/postgres/BlobInputStreamArgumentFactory.java
leaumar: I meant if InputStreams couldn't serve purposes not implemented by this class, like different kinds of lobs or something.

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
stevenschlansker: It's essentially nothing, the expensive resource is server round-trips and those are fixed by the driver.

postgres/src/main/java/org/jdbi/v3/postgres/BlobInputStreamArgumentFactory.java
stevenschlansker: If / when that happens we might have to differentiate by a qualifier annotation?

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
stevenschlansker: You can't use `byte` because it's a signed byte -- `-128..127` -- `InputStream` requires you to work in `0..255` which cannot fit in a Java `byte`.  This isn't a Jdbi-ism, this is a `java.io`-ism that we can't change.

stevenschlansker: Also, the return value has a special flag value `-1` for EOF which could not also fit in a byte

postgres/src/test/java/org/jdbi/v3/postgres/TestLobStream.java
stevenschlansker: Coming up with fun names is part of the otherwise not-fun of writing tests... I'll change this I guess but it really is just a "big" number for testing purposes.  Consider focusing your reviewing time on the meat of the issue at hand.  If you're not comfortable with the api, maybe checkout the upstream docs and learn a bit about it!  I didn't know anything about the Postgres LargeObject stuff before starting this PR.

PR #1713, opened 2020-06-05
Implement mapping and binding for FreeBuilder

wjdix:
This supersedes #1705 and is intended to be a production ready implementation.

Some interesting things to note:

1. FreeBuilder and Immutables use the same properties factory. They only differ in the lookups for value and builder classes. Modifiable uses its own properties factory now.
2. This test https://github.com/jdbi/jdbi/blob/948dbed26f588ad2edf70064f073bf295d3105dd/core/src/test/java/org/jdbi/v3/core/mapper/ImmutablesTest.java#L298-L313 appears to only be working on master b/c Immutables defines the `answer(int)` method even though the spec only has the getter method listed. The lookup of builder setter methods didn't correctly find `setAnswer(int)`. Because FreeBuilder does not create the `answer` method when the spec includes the getter form, the builderSetter lookup failed.

Please let me know if there are other things you'd like to see in this PR. I've got some time carved out at work to invest in this, so I'm very motivated to get it into master.


stevenschlansker:
Thank you for this significant contribution!  I will try to find time to do an in-depth review today or this weekend.

PR #1716, opened 2020-06-05
Update: add `one` helper method

### LINKED ISSUE #1677: Consider adding Update.one() just like ResultIterable.one()
`ResultIterable.one()` throws  `IllegalStateException` if the result set contains zero or multiple rows. It would be useful to add a `void one()` method to the `Update` class to be used instead of `execute()` to work the same way, that is to throw `IllegalStateException` for updates (and inserts) with a count that is not one.

stevenschlansker:
Fixes #1677

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/Update.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/Update.java
@@ -32,6 +32,13 @@ public class Update extends SqlStatement<Update> {
         super(handle, sql);
     }
 
+    public void one() {
+        int count = execute();
+        if (count != 1) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jdbi/v3/core/statement/Update.java
```diff
@@ -32,6 +32,13 @@ public Update(Handle handle, String sql) {
         super(handle, sql);
     }
 
+    public void one() {
+        int count = execute();
+        if (count != 1) {
+            throw new IllegalStateException("Expected one modified row, got: " + count);
```
leaumar, line 38: grammar nit: 1, not one, and no `:`

PR #1722, opened 2020-06-17
Properly look up Builder classes for FreeBuilder classes

wjdix:
Found this bug while trying use this new API released in 3.14.0 in our production app. My implementation did not correctly lookup the builder class in FreeBuilder classes that are not inner classes. This was missed because all the classes in the tests were defined within the test class itself, mea culpa.


stevenschlansker:
Released as `3.14.1`

PR #1724, opened 2020-06-19
Update README.md

Dominik3011:


leaumar:
If you have legit criticisms, we'd be glad to hear them explained. :)

Dominik3011:
Both the declarative and the fluent APIs of JDBI are unmitigated disasters, compounded by the lack of string interpolation in Java. 
There are no options available to make the binding of the SQL queries type-safe.

If you want to use the object mapping (BeanMapper) of JDBI with any database without losing your sanity you have to abandon snake_case naming in your database and switch to Java `camelCase`. 

I have to use this library at my work every day and I am getting very frustrated by the lack of expressiveness and any assistance to catch the errors at compile-time.

The documentation is very incomplete, especially for JDBI 2. 
Can you recommend me a better way to build type-safe SQL queries for Java?

stevenschlansker:
Hi @Dominik3011 I'm sorry you had such a poor experience.

I do agree that lack of documentation was a serious issue in Jdbi 2 days.  You may notice that with Jdbi 3 (the mainline for 3 years now) one of the main features we added was documentation.
I am sorry that Jdbi 2 never got the documentation it deserved, but as a volunteer-only project, we have long ended "support" for the old major version.

Additionally, jdbi3 introduced support for `snake_case` vs `camelCase` mapping, and in fact using annotations to drive your own arbitrary names if desired.

Regarding the lack of type safety, Jdbi is "type safe" in some respects - with a cooperating database, like Postgres, you could never accidentally store a `CustomObject` into a `int` field.  But you are correct that the SQL syntax is not checked at compile time.  My go-to solution for this is extensive unit testing: if you have a quick test that runs each statement at least once, it is a lot harder to accidentally break things and frustrate yourself and your team.  I would love to have a "type safe SQL" construction but it's not entirely clear to me how we could implement that in the constraints of the Java language as it stands today.

I am sure you would have a more positive experience if you moved on to Jdbi 3, in which we've spent a lot of effort improving documentation, error messages, and otherwise "sawing off the edges".  We would love to help you and your team get upgraded if you want to engage constructively, but venting in this manner isn't a good use of anyone's time.

PR #1727, opened 2020-06-27
enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"


leaumar:
Personally, I want to. I've started doing it in private projects and I've been pushing the topic at work. Evergreening is good in the long run because it reduces dependency divergence and pulls people off legacy (even my evergreened projects still transitively get saddled up with unsupported legacy via nevergreen deps and their build warnings). The alternative I see everywhere of everyone picking $current_version at project init and then never updating certainly doesn't sit well with me.

Problem is, we have downstream users and the ones that are nevergreen/legacy or mostly use nevergreen deps (like current jdbi) would then likely be dismayed by the increased friction. Of course, the opposite is true for downstreams that evergreen, and dep divergence will always exist anyway so one might as well aim for divergence in a good direction at least.

It may just be the push that finally gets nevergreens on the evergreen train too: when the burden of evolving along becomes smaller than clinging to stability/staleness. That's pretty much how all innovation happens, something has to make way for a new thing, just like the new generation of audio/video/image codecs trying to displace last gen.

Perhaps we can roll with this on jdbi4? There's no better time than a major release for policy changes.

PS: just my opinion

stevenschlansker:
Thank you for the suggestion.  Automatic dependency tracking makes great sense for assembled software that ships an executable.
However, `jdbi` is a library meant to be consumed by application distributors.  Our version declarations are "this is the minimum API jdbi can link to" not "this is the version you should deploy your app with".
Unfortunately upgrading without due cause can cause extreme difficulty for users: Guava in particular has been a thorn in the past.

We won't be adopting automatic updates at this time, but if there's specific issues in our dependencies that need upgrading, we're always happy to consider specific PRs.

PR #1746, opened 2020-08-28
Fix typo in JDBI2 sql object docs

theholla:
Fixes typo in JDBI2 sql object docs


stevenschlansker:
Thanks!  Surprised anyone still uses v2 : ]

Issue #1732, opened 2020-07-16
Poor DAO performance in jdbi3


gtaborcrossword:
Because I spotted the degraded performance of SqlObject (DAO) calls between jdbi2 and jdbi3, I did some quick and simple comparison tests.

The test code itself is very simple - it is just two database operations in the loop. The first operation inserts the record into the database, second operation just reads it using the primary key.

**Test conditions**
- jdbi3 - v3.13.0
- jdbi2 - v2.78
- database: local Postgresql 11.7
- JDBC driver - postgresql:42.2.12

**Results**

1. Execution time in milliseconds, `num` is a number of iterations.
```
  num     3-DAO    2-DAO    3-API    2-API
   10       480      165      167      159
  100     3 769      690      690      643
 1000    31 259    4 771    4 800    4 889
10000   291 862   39 221   40 308   42 069
```

2. Iteration average execution time in milliseconds.
```
  num   3-DAO   2-DAO   3-API   2-API
   10   48.00   16.50   16.70   15.90
  100   37.69    6.90    6.90    6.43
 1000   31.25    4.77    4.80    4.88
10000   29.18    3.92    4.03    4.20
```

**Conclusions/Comments**

 * The difference between "3-DAO" column and the others is obvious and significant. I'm conscious of switch from cglib to dynamic proxies between jdbi2 and jdbi3 (#341, #456), but that level of degradation is really worrying.
 * I didn't notice any other symptoms that could justify such difference - no connection leakage, no increasing memory consumption.
 
Any ideas what could be wrong? Maybe some other cglib alternative could be considered/verified (e.g Byte Buddy)?

stevenschlansker:
Thanks for measuring this @gtaborcrossword .  Could you please share your benchmarks, preferably as run by JMH?  I would like to try to reproduce your results.  Usually we find that the overhead of Proxy objects is very low compared to the cost of a remote call to the database, but of course we would always love to find improvements :)

stevenschlansker:
Also, you could try using the new `jdbi3-generator` code generator artifact to replace the Proxy based implementation with a generated source file.  That should highlight the possibility of Proxy being the root cause or not.

gtaborcrossword:
Thanks @stevenschlansker. I'll surely will do both.

natan-abolafya:
Hi, a dropwizard user here.

[...]

PR #1749, opened 2020-09-12
Initial V2 vs V3 benchmark for both fluent and sqlobject

stevenschlansker:
jmh 1.25

Inspired by #1732 

Currently, it looks like there's a significant bottleneck in creating copies of `ConfigRegistry` for every statement made.

### DIFF
--- a/benchmark/src/main/java/org/jdbi/v3/benchmark/sqlobject/TestContent.java
+++ b/benchmark/src/main/java/org/jdbi/v3/benchmark/sqlobject/TestContent.java
@@ -1,0 +1,69 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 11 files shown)


benchmark/src/main/java/org/jdbi/v3/benchmark/sqlobject/TestContent.java
```diff
@@ -0,0 +1,69 @@
...
+        return result;
+    }
+
+    @Override
+    public String toString() {
+        return new StringJoiner(", ", TestContent.class.getSimpleName() + "[", "]")
+                .add("name='" + name + "'")
+                .add("description='" + description + "'")
+                .toString();
+    }
```
leaumar, line 68: wish we had somewbere to delegate this cruft to with one-liners, it's non-functional noise

leaumar (approved):
Makes sense. A v2 and v3 base scenario, with pluggable database connections provided by subclasses. 👍

PR #1750, opened 2020-09-12
ConfigRegistry: use new MethodHandle lookup and memoize supplier

stevenschlansker:
~ 20% increase in H2.fluentSelectOne benchmark

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
+++ b/core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
@@ -13,12 +13,17 @@
  */
 package org.jdbi.v3.core.config;
 
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 
 import org.jdbi.v3.core.argument.Arguments;
 import org.jdbi.v3.core.collector.JdbiCollectors;
+import org.jdbi.v3.core.internal.exceptions.Unchecked;
 import org.jdbi.v3.core.mapper.ColumnMappers;
 import org.jdbi.v3.core.mapper.Mappers;
 import org.jdbi.v3.core.mapper.RowMappers;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
```diff
@@ -68,21 +75,39 @@ private ConfigRegistry(ConfigRegistry that) {
...
-                } catch (NoSuchMethodException e) {
-                    config = configClass.getDeclaredConstructor().newInstance();
-                    config.setRegistry(this);
-                }
-                return Optional.ofNullable(configClass.cast(configs.putIfAbsent(configClass, config))).orElse(config);
+                MethodHandle mh = MethodHandles.publicLookup().findConstructor(klass,
+                        MethodType.methodType(void.class, ConfigRegistry.class))
+                    .asType(MethodType.methodType(JdbiConfig.class, ConfigRegistry.class));
+                return Unchecked.function(registry -> (JdbiConfig<?>) mh.invokeExact(registry));
+            } catch (NoSuchMethodException ign) {
```
leaumar, line 90: I tend to name the exception for what it signifies rather than what to do with it later, so that the _next_ line reads like "ignore <this thing>". `noRegistryConstructor` or some such in this case. Like how an event handler should call what it does, not what it handles.

core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
```diff
@@ -68,21 +75,39 @@ private ConfigRegistry(ConfigRegistry that) {
...
+                ignored = ign;
             } catch (ReflectiveOperationException e) {
```
leaumar, line 92: failing a meaningful name, I like to at least see an exception abbreviated to more than `e`, `roe` in this case

core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
```diff
@@ -68,21 +75,39 @@ private ConfigRegistry(ConfigRegistry that) {
...
+            try {
+                MethodHandle mh = MethodHandles.publicLookup().findConstructor(klass,
+                        MethodType.methodType(void.class))
+                    .asType(MethodType.methodType(JdbiConfig.class));
+                return Unchecked.function(registry -> {
+                    JdbiConfig<?> result = (JdbiConfig<?>) mh.invokeExact();
+                    result.setRegistry(registry);
+                    return result;
+                });
+            } catch (ReflectiveOperationException e) {
```
leaumar, line 104: idem

leaumar (approved):
just semantic nitpicks

PR #1751, opened 2020-09-12
Qualifiers: optimize common case of zero or one qualifier

stevenschlansker:
~15% optimization on H2.fluentSelectOne

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
+++ b/core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
@@ -16,6 +16,9 @@ package org.jdbi.v3.core.qualifier;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
@@ -26,10 +29,6 @@ import org.jdbi.v3.core.generic.GenericType;
 import org.jdbi.v3.core.internal.AnnotationFactory;
 import org.jdbi.v3.meta.Beta;
 
-import static java.util.Collections.emptySet;
-import static java.util.Collections.unmodifiableSet;
-import static java.util.stream.Collectors.toSet;
-
 /**
  * A {@link java.lang.reflect.Type} qualified by a set of qualifier annotations. Two qualified types are equal to each other
  * if their {@link #getType()} and {@link #getQualifiers()} properties are equal.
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
```diff
@@ -73,9 +97,19 @@
         return (QualifiedType<T>) of(type.getType());
     }
 
-    private QualifiedType(Type type, Set<? extends Annotation> qualifiers) {
+    private QualifiedType(Type type, Object qualifiers) {
```
leaumar, line 100: could it be interesting to guard type safety against mistakes by having 3 overloaded constructors for the specific cases instead?

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
```diff
@@ -59,7 +59,31 @@
      * @see #with(Annotation...) to then qualify your type
      */
     public static QualifiedType<?> of(Type type) {
-        return new QualifiedType<>(type, emptySet());
+        return new QualifiedType<>(type, null);
+    }
+
+    /**
+     * Creates a wildcard {@code QualifiedType<?>} for a {@link Type} with the given qualifiers.
```
leaumar, line 66: how is this a wildcard? Doesn't seem to leave much to chance

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
```diff
@@ -134,8 +167,29 @@ public Type getType() {
...
+        if (qualifiers == null) {
+            return Collections.emptySet();
+        } else if (qualifiers instanceof Set<?>) {
+            return (Set<Annotation>) qualifiers;
+        } else {
+            return Collections.singleton((Annotation) qualifiers);
+        }
+    }
+
+    public boolean sameQualifiersAs(Set<Annotation> annotations) {
```
leaumar, line 181: "same qualifiers as" makes me expect a qualified type as argument. Perhaps `qualifiersEqualTo` or some such?

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
```diff
@@ -39,7 +38,8 @@
 @Beta
 public final class QualifiedType<T> {
     private final Type type;
-    private final Set<Annotation> qualifiers;
+    // null for no qualifiers, Annotation itself for one qualifier, Set<Annotation> for many
+    private final Object qualifiers;
```
leaumar, line 42: Could it be interesting to encapsulate this variable type into its own class, to separate that concern from the rest of QualifiedType? I'd have to give it a try to be sure but I have a feeling it would make the current spaghetti of constructors and factories calling each other simpler.

leaumar (changes requested):
Just some things I'd like discussed

core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
leaumar: see my other comment on the variable type

leaumar:
Strange, I'm getting a compile error locally on this current branch...
image

stevenschlansker:
Replaced by #1766

PR #1752, opened 2020-09-13
Arguments: don't prePrepare types, it's too expensive in the common case

stevenschlansker:
also, reduce some Streams usage in the hot path
~20% improvement on H2.fluentSelectOne


leaumar (approved):
PR comments fixed before review was posted ;)

PR #1753, opened 2020-09-13
Bean / Pojo Properties: level up our MethodHandle usage by fixing asType

stevenschlansker:
and then using invokeExact

smaller but still measurable ~5% improvement


leaumar (approved):
Can't vouch for the changes, but they don't look bad to me either

PR #1754, opened 2020-09-13
AnnotationSqlLocator: computation is surprisingly expensive due to

stevenschlansker:
annotation and stream usage, so cache it

~20% improvement on H2 sqlobject insert bean benchmark

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/AnnotationSqlLocator.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/AnnotationSqlLocator.java
@@ -14,6 +14,8 @@
 package org.jdbi.v3.sqlobject.locator;
 
 import java.lang.reflect.Method;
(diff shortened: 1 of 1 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/locator/AnnotationSqlLocator.java
```diff
@@ -22,9 +24,10 @@
  * Locates SQL on the SQL method annotations like <code>@SqlQuery("foo")</code>. This is the default SqlLocator.
  */
 public class AnnotationSqlLocator implements SqlLocator {
+    private final ConcurrentMap<Method, String> located = new ConcurrentHashMap<>();
```
leaumar, line 27: should be an empty line around methods/after fields

leaumar (approved):
I know they're nitpicks but I see no reason to neglect consistent formatting 🙂

PR #1755, opened 2020-09-13
Qualifiers #1751: refactoring suggestion

leaumar:
suggestion for #1751 

IMO separating these concerns feels less tangly, easier on the mind, even if it is a bit more code. I think it's also quite well optimized for performance as originally intended.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
+++ b/core/src/main/java/org/jdbi/v3/core/qualifier/QualifiedType.java
@@ -15,18 +15,12 @@ package org.jdbi.v3.core.qualifier;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Iterator;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
-import java.util.stream.StreamSupport;
 
 import org.jdbi.v3.core.generic.GenericType;
-import org.jdbi.v3.core.internal.AnnotationFactory;
 import org.jdbi.v3.meta.Beta;
 
 /**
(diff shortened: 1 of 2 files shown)


leaumar:
I don't know what the kotlin error is about...

stevenschlansker:
Did you have a chance to compare performance before / after / vs my change?

leaumar:
No, sorry. I can try to do that tomorrow if you don't do it. My focus was on code simplicity, I just tried to avoid doing anything that would be an obvious step backward in performance.

leaumar:
Before refactor:
```
Benchmark                                     Mode  Cnt      Score      Error  Units
QualifiersBenchmark.getQualifiersAnnotated    avgt    5    225.286 ±   17.535  ns/op
QualifiersBenchmark.getQualifiersUnannotated  avgt    5    135.407 ±    6.985  ns/op
QualifiersBenchmark.mapQualifiedBean          avgt    5  32375.018 ± 1512.807  ns/op
QualifiersBenchmark.mapUnqualifiedBean        avgt    5  34898.519 ± 7136.024  ns/op
```

After refactor:
```
Benchmark                                     Mode  Cnt      Score      Error  Units
QualifiersBenchmark.getQualifiersAnnotated    avgt    5    222.257 ±   10.369  ns/op
QualifiersBenchmark.getQualifiersUnannotated  avgt    5    140.961 ±   10.855  ns/op
QualifiersBenchmark.mapQualifiedBean          avgt    5  32894.482 ± 2735.528  ns/op
QualifiersBenchmark.mapUnqualifiedBean        avgt    5  32898.203 ± 2078.309  ns/op
```

Seems to me the difference is within the margin of error, so no loss of performance?

stevenschlansker:
Unfortunately I don't think the QualifiersBenchmark actually covers this refactoring - you'll notice it changes argument factory code, while the qualifier benchmark only tests the Qualifiers code itself.

leaumar:
The only way it touches argument factory code is renaming the method you added in your original PR (which was one of my other review remarks), so I don't think it matters for the concern of refactoring just a part of your original PR?

stevenschlansker:
I tried to make some comparisons but even running for 2h the results were inconclusive.
I guess I'll save this for last, get the rest merged in, then re-test and re-profile.

stevenschlansker:
replaced by #1766

PR #1757, opened 2020-09-15
*Mappers: cache misses as well as hits

stevenschlansker:
In particular, the `Mappers` class relies on first checking row mappers then column mappers - it has to exhaust every row mapper to not find any before trying a column mapper.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
@@ -33,7 +33,7 @@ import org.jdbi.v3.meta.Beta;
  */
 public class ColumnMappers implements JdbiConfig<ColumnMappers> {
     private final List<QualifiedColumnMapperFactory> factories = new CopyOnWriteArrayList<>();
-    private final ConcurrentHashMap<QualifiedType<?>, ColumnMapper<?>> cache = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<QualifiedType<?>, Optional<? extends ColumnMapper<?>>> cache = new ConcurrentHashMap<>();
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
```diff
@@ -184,23 +184,22 @@ public ColumnMappers register(QualifiedColumnMapperFactory factory) {
...
-        ColumnMapper<T> cached = (ColumnMapper<T>) cache.get(type);
+        Optional<ColumnMapper<T>> cached = (Optional) cache.get(type);
 
         if (cached != null) {
-            return Optional.of(cached);
+            return cached;
         }
 
-        Optional<ColumnMapper<T>> mapper = factories.stream()
+        Optional<ColumnMapper<T>> mapper = (Optional) factories.stream()
```
leaumar, line 198: I'd rename this to `mapperHit` or something, since it's not just an instance of a mapper anymore

leaumar (approved):
A note on semantics, but that's a surprisingly elegant solution :p

leaumar:
Looks like the same weird build error as my PR

core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
stevenschlansker: This isn't actually changing the type - this was always how it was.  `mapperHit` is a little bit misleading because this is *not* a hit case -- it's actually the "miss" case where we look it up in the factories.

leaumar: Myeah I guess I was using `hit` a bit dubiously in my mind, having it also mean "miss" :p I figured you'd come up with a better word that means "hit or miss"

PR #1763, opened 2020-09-24
openjdk 15

stevenschlansker:


core/src/test/java/org/jdbi/v3/core/TestJsr310.java
```diff
@@ -35,6 +36,8 @@
 
     Handle h;
 
+    Clock fixed = Clock.fixed(Instant.ofEpochSecond(123456789), ZoneOffset.UTC);
```
leaumar, line 39: I'd like to see the reason for this clock being here (incompatible precisions) documented, either in a comment or even better through its name, like `withPgPrecision`

PR #1766, opened 2020-10-01
QualifiedType performance

stevenschlansker:
Once we're done constructing it, hold it in a JdbiCache for later use to save re-computation effort
When on Java 10+, use `Collectors.toUnmodifiableSet()` as it's much more performant than `Collections.unmodifiableSet(HashSet(...))`

Significantly faster and replaces both #1751 and #1755 as they're no longer needed.

### DIFF
--- a/benchmark/src/main/java/org/jdbi/v3/benchmark/QualifiersBenchmark.java
+++ b/benchmark/src/main/java/org/jdbi/v3/benchmark/QualifiersBenchmark.java
@@ -13,13 +13,16 @@
  */
 package org.jdbi.v3.benchmark;
 
-import java.lang.annotation.Annotation;
-import java.util.Set;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 import java.util.concurrent.TimeUnit;
 
 import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.mapper.reflect.BeanMapper;
 import org.jdbi.v3.core.qualifier.NVarchar;
+import org.jdbi.v3.core.qualifier.QualifiedType;
+import org.jdbi.v3.core.qualifier.Qualifier;
 import org.jdbi.v3.core.qualifier.Qualifiers;
 import org.jdbi.v3.testing.JdbiRule;
 import org.openjdk.jmh.annotations.Benchmark;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/jdbi/v3/core/internal/CollectionFactory.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Collections;
+import java.util.Set;
+import java.util.function.Supplier;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+
+import org.jdbi.v3.core.internal.exceptions.Unchecked;
+
+@SuppressWarnings("rawtypes")
+public final class CollectionFactory {
```
leaumar, line 28: `CollectorFactory`?

core/src/main/java/org/jdbi/v3/core/internal/CollectionFactory.java
```diff
@@ -0,0 +1,54 @@
...
+        Supplier<Collector> factory = null;
+        try {
+            MethodHandle mh = MethodHandles.publicLookup()
+                    .findStatic(Collectors.class, "toUnmodifiableSet", MethodType.methodType(Collector.class));
+            factory = Unchecked.supplier(() -> (Collector) mh.invokeExact());
+        } catch (final ReflectiveOperationException e) {
+            // ignored
+            e.fillInStackTrace();
+        }
+        if (factory == null) {
```
leaumar, line 40: I think most of this method would be neater as a static factory method that returns the instance and sets it statically, i.e. `private static final Supplier<Collector> SET_COLLECTOR = getOptimalInstance();`. That would avoid this ugly temp var logic and conditional (the logic of which can go into the `catch` clause).

leaumar (approved):
As usual, it all _works_, so only remarks are superficial things 🙂

core/src/main/java/org/jdbi/v3/core/internal/CollectionFactory.java
leaumar:
```java
    private static Supplier<Collector> getOptimalInstance() {
        try {
            MethodHandle mh = MethodHandles.publicLookup().findStatic(Collectors.class, "toUnmodifiableSet", MethodType.methodType(Collector.class));
            return Unchecked.supplier(() -> (Collector) mh.invokeExact());
        } catch (final ReflectiveOperationException ignored) {
            return () -> Collectors.collectingAndThen(
                Collectors.toSet(),
                Collections::unmodifiableSet
            );
        }
    }
```

core/src/main/java/org/jdbi/v3/core/internal/CollectionFactory.java
stevenschlansker: That name's already used :/

PR #1775, opened 2020-10-09
 SqlObject decorators, Transactional: support annotating types 

### LINKED ISSUE #1774: Creating a retry annotation for DAO objects.
I was following along with this post as it requires almost the exact same requirements as my issue, but I did not have access to reply there:
https://groups.google.com/g/jdbi/c/ib2tybgf0RM

And I came across an issue when trying to implement the suggestion solution:
Currently, I am implementing a new annotation:
```
@SqlMethodDecoratingAnnotation(RetryDecorator.class)
@Retention(RetentionPolicy.RUNTIME)
@Target({ElementType.TYPE, ElementType.METHOD})
public @interface Retry {}
```

With a RetryDecorator class:
```
public class RetryDecorator implements HandlerDecorator {
  @Override
  public Handler decorateHandler(
      final Handler base, final Class<?> sqlObjectType, final Method method) {
    return (target, args, handle) ->
        {
          System.out.println("RetryDecorator called.");
          return base.invoke(obj, args, handle);
        };
  }
}
```

Then inside my DAO object I would like to annotate at the interface level to apply the retry to all methods inside:

```
@Retry
public interface MyDAO {
	@SqlUpdate("...")
	void add();
	
	@SqlUpdate("...")
	void update();
	
	@SqlUpdate("...")
	void delete();
}
```
However, when I call my methods (add, update, delete) nothing happens. I am only seeing the "RetryDecorator called." message when I place my @Retry annotation at the method level for each method (add, update, delete). 
Ideally I could add the @Retry annotation to the top level of each DAO interface and worry about the decorator order at this point as well.

I have looked through the documentation and am trying to piece my gap in understanding, if any advice can be offered. Thanks!

stevenschlansker:
Fixes #1774

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/transaction/internal/TransactionDecorator.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/transaction/internal/TransactionDecorator.java
@@ -14,6 +14,7 @@
 package org.jdbi.v3.sqlobject.transaction.internal;
 
 import java.lang.reflect.Method;
+import java.util.stream.Stream;
 
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.HandleCallback;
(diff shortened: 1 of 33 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/transaction/internal/TransactionDecorator.java
```diff
@@ -26,7 +27,11 @@
 public class TransactionDecorator implements HandlerDecorator {
     @Override
     public Handler decorateHandler(Handler base, Class<?> sqlObjectType, Method method) {
-        final Transaction txnAnnotation = method.getAnnotation(Transaction.class);
+        final Transaction txnAnnotation = Stream.of(method, sqlObjectType)
```
qualidafial, line 30: allow overriding class-level txn isolation on the method. nice

qualidafial (approved):

PR #1779, opened 2020-10-15
TemplateEngine: use size-limited Caffeine cache so we don't keep an unbounded number of templates

### LINKED ISSUE #1776: DefinedAttributeTemplateEngine max cache size
Right now it's unbounded, reported via mailing list

stevenschlansker:
Fixes #1776

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
@@ -1762,15 +1762,7 @@ public abstract class SqlStatement<This extends SqlStatement<This>> extends Base
 
         beforeTemplating();
 
-        String renderedSql = getConfig(SqlStatements.class)
-                .getTemplateEngine()
-                .render(sql, ctx);
-        ctx.setRenderedSql(renderedSql);
-
-        ParsedSql parsedSql = getConfig(SqlStatements.class)
-                .getSqlParser()
-                .parse(renderedSql, ctx);
-        ctx.setParsedSql(parsedSql);
+        ParsedSql parsedSql = parseSql();
 
         try {
             stmt = createStatement(ctx, parsedSql);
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -244,4 +264,18 @@ public SqlStatements createCopy() {
...
+
+    String preparedRender(String template, StatementContext ctx) {
+        try {
+            return Optional.ofNullable(
+                    templateCache.get(
+                            new AbstractMap.SimpleEntry<>(templateEngine, template),
+                            e -> e.getKey().parse(e.getValue(), ctx.getConfig())
+                                               .orElse(null)))
+                .orElse(cx -> templateEngine.render(template, cx))
+                .apply(ctx);
```
qualidafial, line 276: This is difficult to understand at first glance. Could we add some comments or explanatory variables?

core/src/main/java/org/jdbi/v3/core/statement/TemplateEngine.java
```diff
@@ -37,4 +43,29 @@
...
+     * @return a parsed representation, if available
+     */
+    @Beta
+    default Optional<Function<StatementContext, String>> parse(String template, ConfigRegistry config) {
+        return Optional.empty();
+    }
+
+    @FunctionalInterface
+    @Beta
+    interface Caching extends TemplateEngine {
```
qualidafial: This feels like a misnomer. There's no caching happening here, even though the parse method is making caching possible elsewhere.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -1811,6 +1803,19 @@ PreparedStatement createStatement(final StatementContext ctx, ParsedSql parsedSq
         return handle.getStatementBuilder().create(handle.getConnection(), parsedSql.getSql(), ctx);
     }
 
+    ParsedSql templateSql() {
```
qualidafial, line 1806: This is a minor nit since it's not public API, but this method is rendering a template and parsing SQL, not just templating. And the return type is `ParsedSql`. If we want a terse name, `parseSql` is probably better IMO.

core/src/main/java/org/jdbi/v3/core/statement/TemplateEngine.java
```diff
@@ -37,4 +43,29 @@
...
     String render(String template, StatementContext ctx);
+
+    /**
+     * Parse a SQL template and return a parsed representation ready to apply to a statement.
+     * @param template the sql template to parse
+     * @param config the Jdbi configuration at prepare time
+     * @return a parsed representation, if available
+     */
+    @Beta
+    default Optional<Function<StatementContext, String>> parse(String template, ConfigRegistry config) {
```
qualidafial, line 54: It would be good to point out that implementing this method (or `TemplateEngine.Parsing`) allows parsed templates to be cached, maybe with a `@see` reference

PR #1793, opened 2020-12-03
Update to ANTLR 4.9

electrum:


stevenschlansker:
Is there a particular reason to upgrade, or just keeping things up to date?

electrum:
I was about to file an issue to consider shading `antlr-runtime`. We get this spam message printed multiple times in Presto since we use a newer version:
```
ANTLR Tool version 4.7.2 used for code generation does not match the current runtime version 4.8
```
Given that the generator version needs to exactly match the runtime version, shading seems best, otherwise all usages of ANTLR in a project's dependency closure would need to be the same.

At least for now, I can solve this by updating both projects to use the latest version of ANTLR. Given the current release frequency of ANTLR being every 6-12 months, maybe this is good enough.

stevenschlansker:
Is there a reason you don't just manage both the generator and runtime in presto's pom?  Jdbi is happy to run against 4.7 or 4.9, it just needs to be consistent, as far as I know.  This should only happen if you depend on a specific version of the generator, but let the runtime version "float", which is inherently risky to taking a version you didn't expect from a transitive dependency.  (One of Maven's sharper edges...)

Jdbi upgrading to 4.9 is nice but it would break anyone who expects to continue running against 4.7 or 4.8, so we try not to just bump version numbers for no reason.  I doubt too many people care to keep the old version in this case, though, so we can probably just accept this change.

To me, not shading is a feature - if you shade, it's impossible to bring in bugfixes using `dependencyManagement`.  For example the shaded `cglib` broke horribly when Java 9 came out, and there was no fix except to rebuild Jdbi.

But if you really want shaded jars, maybe we can distribute both regular non-shaded jars and shaded jars... we can follow up on a separate ticket regarding that.

stevenschlansker (approved):

electrum:
For Presto’s own usage of ANTLR, we use the same version of the generator and runtime.

The problem is the generated code in Jdbi — that version needs to match our runtime version but is out of our control (since Jdbi was already built using the older version). I don’t see a good solution given the tight coupling in ANTLR.

stevenschlansker:
Oh, I see.  I didn't realize the generated code depended on the runtime version, that sucks.  Maybe shading in *just* the antlr library is the right solution...

electrum:
FYI, the check for that is here: https://github.com/antlr/antlr4/blob/master/runtime/Java/src/org/antlr/v4/runtime/RuntimeMetaData.java

electrum:
The generated `org.jdbi.v3.core.internal.lexer.ColonStatementLexer` has this call:
```java
static { RuntimeMetaData.checkVersion("4.7.2", RuntimeMetaData.VERSION); }

stevenschlansker:
Opened #1798 to track this

Issue #1712, opened 2020-06-04
Commented out named arguments are not ignored


KanunnikovPavel:
Hi there!
We have some SQL with commented out lines. One of this commented out lines contains a named argument. 
We could not execute this query without providing this (commented out) argument, it throws `"UnableToCreateStatementException: Missing named parameter '...' in binding..."` exception.
Such behavior is not correct because commented out parts of query should not affect it.  

JDBI version: checked in 3.10.1 and 3.13.0. DB: PostgreSQL.

Simple reproducing example:
```SQL
SELECT 
  --:param as value1,
  'some value' as value2
```

P.S. Even if we provide value for this commented out argument query still will not be executed. Another exception will be thrown: `"UnableToCreateStatementException: Exception while binding named parameter..."` (actual reason is: `"PSQLException: The column index is out of range..."`)

pearceeverydaylabs:
Saw this bug on `3.18.0`. Updated to `3.18.1` for this fix (mentioned above): https://github.com/jdbi/jdbi/pull/1796
Worked like a charm.

@KanunnikovPavel any reason to keep this open?

PR #1796, opened 2020-12-08
Skip parsing parameters in single-line comments

rherrmann:
Extend the lexer grammars used to parse statements for colon
parameters, hash parameters, and attribute definitions to
recognize single-line comments starting with `--` and `//`.
As a result, placeholders within comments will remain
untouched when pre-processing statements.

https://github.com/jdbi/jdbi/issues/1712

### DIFF
--- a/core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g4
+++ b/core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g4
@@ -27,7 +27,7 @@ fragment NAME: JAVA_LETTER | [0-9] | '.' | '?.';
 /* Lovingly lifted from https://github.com/antlr/grammars-v4/blob/master/java/JavaLexer.g4 */
 fragment JAVA_LETTER : [a-zA-Z$_] | ~[\u0000-\u007F\uD800-\uDBFF] | [\uD800-\uDBFF] [\uDC00-\uDFFF];
 
-COMMENT: '/*' .*? '*/';
+COMMENT: '/*' .*? '*/' | '--' ~('\r' | '\n')* | '//' ~('\r' | '\n')*;
 QUOTED_TEXT: QUOTE (ESCAPE_QUOTE | ~'\'')* QUOTE;
 DOUBLE_QUOTED_TEXT: DOUBLE_QUOTE (~'"')+ DOUBLE_QUOTE;
 ESCAPED_TEXT : ESCAPE . ;
--- a/core/src/test/java/org/jdbi/v3/core/statement/TestColonPrefixSqlParser.java
+++ b/core/src/test/java/org/jdbi/v3/core/statement/TestColonPrefixSqlParser.java
@@ -53,6 +53,13 @@ public class TestColonPrefixSqlParser {
     }
 
     @Test
+    public void testCommentCharsInQuotes() {
+        String sql = "select '-- // /* */'";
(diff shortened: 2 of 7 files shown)


core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g4
```diff
@@ -27,7 +27,7 @@ fragment NAME: JAVA_LETTER | [0-9] | '.' | '?.';
 /* Lovingly lifted from https://github.com/antlr/grammars-v4/blob/master/java/JavaLexer.g4 */
 fragment JAVA_LETTER : [a-zA-Z$_] | ~[\u0000-\u007F\uD800-\uDBFF] | [\uD800-\uDBFF] [\uDC00-\uDFFF];
 
-COMMENT: '/*' .*? '*/';
+COMMENT: '/*' .*? '*/' | '--' ~('\r' | '\n')* | '//' ~('\n' | '\r')*;
```
leaumar, line 30: Is there any particular reasoning behind the different order of \\r and \\n in the two sides here, or just coincidence? If no reason, I'd prefer both in \\r\\n order (and the same in the other files of course)

leaumar (approved):
Looks neat to me!

leaumar:
There's no realistic and supportable situation where something that looks like an inline comment in a query is actually not an inline comment, right guys?

core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g4
rherrmann: Hell no, thanks for catching this. I've also aligned newline patterns in `SqlScriptLexer.g4` to use \r\n order, which is otherwise unchanged.

rherrmann:
> There's no realistic and supportable situation where something that looks like an inline comment in a query is actually not an inline comment, right guys?
> 
> Edit: perhaps add a test that e.g. string literals with comment markers inside them are still parsed as strings?

I've added a `testCommentCharsInQuotes` to the parameter parsing tests. For the `DefinedAttributeTemplateEngine`, I've added a test to ensure that comment chars within angle brackets aren't recognized as such. Not entirely sure though, if that makes sense. What do you think?

BTW: `TestColonPrefixSqlParser` contains two tests with similar names that are apparently testing the same thing: `testHashInColumnNameOkay` and `testHashInColumnNameWorks`. Shall I remove one of them - within this change or in a separate one?

core/src/test/java/org/jdbi/v3/core/statement/TestColonPrefixSqlParser.java
```diff
@@ -52,6 +52,13 @@ public void testEscapedQuoteInQuotes() {
             .isEqualTo(ParsedSql.builder().append(sql).build());
     }
 
+    @Test
+    public void testCommentCharsInQuotes() {
+        String sql = "select '-- // /* */'";
+        assertThat(parser.parse(sql, ctx))
+            .isEqualTo(ParsedSql.builder().append(sql).build());
```
leaumar, line 59: nice, thanks

qualidafial (approved):

qualidafial:
Great test cases!

stevenschlansker (approved):
This is great, thank you all

stevenschlansker:
Duplicate test removed in 8648c1c56

PR #1800, opened 2020-12-26
enable GitHub Actions CI

sullis:


stevenschlansker:
Hi @sullis, thanks for the interest - can you explain a bit more about why we should make this change?  Currently Jdbi uses Travis which seems to work ok -- I don't think we want to have two separate CI systems to maintain.  But maybe I'm missing something.  Thanks!

shipp02:
Travis-ci.org is shutting down. You now have to request build minutes for open source projects which has a bit of a long process from what I have heard. GitHub Actions still have unlimited builds for open source. Also travis-ci was acquired by a private equity firm so it is uncertain how much build time they will continue giving. 

This is the reason you might want to move to GitHub Actions and maybe drop travis-ci.

stevenschlansker:
Indeed, I see the impending warning on Travis's site.  Thanks for proposing this, we might need to switch indeed.

stevenschlansker:
This looks like a good starting point, but there is some work to be done to complete it.  We have a number of parts of the build that only execute conditionally and we should port or redesign that logic to work in the new GitHub Actions setup.

stevenschlansker:
For the moment Travis has extended us build credits. Just for simplicity's sake we'll keep that running for a bit, but if they start trying to extract money from us or make us talk to them repeatedly then I think this is still the next thing to try.

stevenschlansker:
The Travis security f-up is another nail in the coffin... thankfully we don't have any secrets!

sullis:
Reference:  TravisCI

https://arstechnica.com/information-technology/2021/09/travis-ci-flaw-exposed-secrets-for-thousands-of-open-source-projects/

hgschmie:
Superseded by #1924

PR #1809, opened 2021-02-08
This plugin provides JSON backing through Moshi

unoexperto:
Hi folks!

I'm not sure if this PR is big enough to justify merge so I'll understand if it's rejected. Having said that I felt it would be great to have [Moshi](https://github.com/square/moshi), one of the fastest json parsers for JVM, support in jdbi and in maven central. I mimicked what's done in Gson2 plugin since their API is fairly similar.

Thanks!

### DIFF
--- a/moshi/src/test/java/org/jdbi/v3/moshi/TestMoshiPlugin.java
+++ b/moshi/src/test/java/org/jdbi/v3/moshi/TestMoshiPlugin.java
@@ -65,7 +65,7 @@ public class TestMoshiPlugin extends AbstractJsonMapperTest {
                 .one();
 
             assertThat(subuser.name)
-                .describedAs("instead of being bound via getClass(), the object was bound according to the type param")
+                .describedAs("instead of being bound via getClass(), the object was bound according to the qualified type param")
                 .isEqualTo("super");
         });
     }
--- a/moshi/src/test/java/org/jdbi/v3/moshi/TestMoshiPlugin.java
+++ b/moshi/src/test/java/org/jdbi/v3/moshi/TestMoshiPlugin.java
@@ -14,11 +14,13 @@
 package org.jdbi.v3.moshi;
 
 import java.io.IOException;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.Optional;
+import java.util.Set;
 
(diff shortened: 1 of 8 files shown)


moshi/src/test/java/org/jdbi/v3/moshi/TestMoshiPlugin.java
```diff
@@ -0,0 +1,105 @@
...
+
+public class TestMoshiPlugin extends AbstractJsonMapperTest {
+    @Rule
+    public JdbiRule db = PostgresDbRule.rule();
+
+    @Before
+    public void before() {
+        jdbi = db.getJdbi().installPlugin(new MoshiPlugin())
+                .configure(MoshiConfig.class, c -> c.setMoshi(
+                        new Moshi.Builder().build()));
```
leaumar, line 41: I take it this `configure()` is included just to cover the API?

moshi/src/test/java/org/jdbi/v3/moshi/TestMoshiPlugin.java
```diff
@@ -0,0 +1,105 @@
...
+                // declare that the subuser should be mapped as a superuser
+                .bindByType("user", new SubUser(), QualifiedType.of(SuperUser.class).with(Json.class))
+                .execute();
+
+            User subuser = h.createQuery("select usr from users")
+                .mapTo(QualifiedType.of(User.class).with(Json.class))
+                .one();
+
+            assertThat(subuser.name)
+                .describedAs("instead of being bound via getClass(), the object was bound according to the type param")
```
leaumar, line 65: could you expand this to "the qualified type param"? That's a bit clearer

leaumar (approved):
Approved with a nitpick code-wise, but I leave the decision on whether we want to support moshi upstream to the others. @stevenschlansker @qualidafial ?

moshi/src/test/java/org/jdbi/v3/moshi/TestMoshiPlugin.java
unoexperto: Frankly I'm not sure. I practically copy-pasted this test from gson2 test written by @stevenschlansker. But I think you're right, the rationale was to test .configure().

stevenschlansker (approved):
Thanks for contributing this!

PR #1816, opened 2021-03-25
Register boolean array types

### LINKED ISSUE #1802: Postgres Plugin does not configure boolean array type
The Postgres plugin does not currently support boolean array types, it will be really nice if it was configured like the other primitive types.

stevenschlansker:
Pull common array type configuration with SQL standard names up to base SqlArraytypes
They can always be overridden anyway

Fixes #1802

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/array/SqlArrayTypes.java
+++ b/core/src/main/java/org/jdbi/v3/core/array/SqlArrayTypes.java
@@ -16,6 +16,7 @@ package org.jdbi.v3.core.array;
 import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.function.Function;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/jdbi/v3/core/array/SqlArrayTypes.java
```diff
@@ -35,10 +36,24 @@
...
+        register(Float.class, "float4");
+        register(double.class, "float8");
+        register(Double.class, "float8");
+        register(String.class, "varchar");
+        register(UUID.class, "uuid");
         register(new EnumSqlArrayTypeFactory());
     }
 
-    private SqlArrayTypes(SqlArrayTypes that) {
+    private SqlArrayTypes(final SqlArrayTypes that) {
```
leaumar, line 56: Is this something we want to gradually make our standard code style? If not, I'd rather avoid all this churn since it's of rather little value.

leaumar (approved):
Essence approved, question on the collateral changes

PR #1817, opened 2021-03-25
Add some documentation about escaping ST templates

### LINKED ISSUE #1814: Document the need for escaping the < operator
Refs https://github.com/jdbi/jdbi/issues/1156

I had a query that was trying to use the <= operator for date checking. I kept getting stringtemplate null errors that would say that the query is not an expression (hinting that it had a syntax error). I still don't know why I need to escape the operator either: https://stackoverflow.com/questions/59236448/jdbi-version-3-stringtemplate-when-to-escape-characters is the closest I could find and it doesn't have an answer.

Where would be a good place to add documentation that you may need to escape the `<` operator if you're getting stringtemplate exceptions? Ideally it would be somewhere in the stringtemplate docs. I'm happy to add the documentation in a PR if you know where's the best place to put it.

stevenschlansker:
Fixes #1814

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -4719,6 +4719,10 @@ List<Account> accounts = handle.createQuery(sql)
       .list();
 ----
 
+[WARNING]
+Since StringTemplate by default uses the `<` character to mark ST expressions,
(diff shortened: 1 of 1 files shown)


leaumar:
I guess the placement answers the question if this only happens with the StringTemplate module installed.

leaumar (approved):

PR #1820, opened 2021-04-02
Don't use requireNonNull that needs a supplier

hgschmie:


core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -301,7 +301,7 @@ public Handle open() {
             final long start = System.nanoTime();
             @SuppressWarnings("PMD.CloseResource")
             Connection conn = Objects.requireNonNull(connectionFactory.openConnection(),
-                    () -> "Connection factory " + connectionFactory + " returned a null connection");
```
leaumar, line 304: I recall this was done as a micro-optimization to prevent all the string concatenation work when the error isn't thrown

leaumar (approved):

stevenschlansker:
Yeah, we don't want to assemble the string on connection open every time, just in case there's a null that never should be.

PR #1823, opened 2021-04-02
New toolchain for building JDBI

hgschmie:
This change requires a JDK 9+ to build JDBI. With JDK8 being obsolete
for a long time and JDK11 being replaced by 17 in the fall as the LTS,
it would be good to update the toolchain to use a more current JDK.

This still compiles the jar to be JDK8 compatible (this is actually
easier with newer compilers than with the old toolchains thing).

Also updates Kotlin to use the 1.5.0-M2 release


docs/pom.xml
```diff
@@ -260,7 +260,6 @@
                         </goals>
                         <configuration>
                             <outputDirectory>${project.build.directory}/generated-docs/apidocs</outputDirectory>
-                            <failOnError>true</failOnError>
```
stevenschlansker, line 263: I'm a little concerned removing this line means syntax errors in the docs won't cause a compile error, but instead generate bad docs

src/build/ci-code-style.sh
```diff
@@ -1,8 +1,6 @@
 #!/bin/sh
 set -xe
 
-OPTS="-DskipTests=true -Dmaven.javadoc.skip=true -B"
```
stevenschlansker, line 4: I think it was intentional that we skip building docs here, since that's checked in the `ci-doc` step

stevenschlansker (approved):
Looks good, just a couple small questions

docs/pom.xml
hgschmie: removing this line makes it fall back to `<failOnError>${basepom.check.fail-javadoc}</failOnError>` in the base pom. That is enforced for the release (in the oss profile), so the build will still fail.

We can add this switch to the doc build as well so that the CI doc build would fail.

src/build/ci-code-style.sh
hgschmie: that is the default in the basepom `<basepom.javadoc.skip>true</basepom.javadoc.skip>`

docs/pom.xml
stevenschlansker: Ah, great.

stevenschlansker: Yeah, let's make sure the CI doc build fails if the docs don't build - that way we cannot merge it until we fix it, rather than only figuring it out at release.

PR #1828, opened 2021-04-02
Add a simple Slf4J SQL statement logger

### LINKED ISSUE #1815: [Feature Request] Provide a way to log interactions with database
I'm fairly new to Jdbi. I've been trying to integrate with Spring Boot and everything is going fine so far in my (very basic) tests.

What I haven't been able to find is a way to log the interactions with the database — namely the same result one can achieve when turning Hibernate's `show-sql` to `true`.

I haven't find a way to visualize this, so I was expecting a (set of) configuration setting(s), or something like increasing the log level on certain namespace(s), etc.

> I'm aware Jdbi is using almost raw SQL, but it's always nice to be able to see what is it what the framework/library _says_ it's doing for debugging purposes, etc.

Obligatory StackOverflow question: https://stackoverflow.com/questions/66431905/log-jdbi-interactions-with-database

stevenschlansker:
Fixes #1815

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
@@ -1,0 +1,68 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.statement;
+
+import java.sql.SQLException;
+import java.time.Duration;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Simple {@link SqlLogger} that emits some diagnostic information about
+ * Jdbi usage. Usually you shouldn't log every database statement made but
+ * it can be helpful for debugging.
+ */
+public class Slf4JSqlLogger implements SqlLogger {
+    private final Logger log;
+
+    public Slf4JSqlLogger() {
+        this(LoggerFactory.getLogger("org.jdbi.sql"));
+    }
+
+    public Slf4JSqlLogger(Logger log) {
+        this.log = log;
+    }
+
+    @Override
+    public void logAfterExecution(StatementContext context) {
+        if (log.isDebugEnabled()) {
+            log.debug("Executed in {} '{}' with parameters '{}'",
+                    format(Duration.between(context.getExecutionMoment(), context.getCompletionMoment())),
+                    context.getParsedSql().getSql(),
+                    context.getBinding());
+        }
+    }
+
+    @Override
+    public void logException(StatementContext context, SQLException ex) {
+        if (log.isErrorEnabled()) {
+            log.error("Exception while executing '{}' with parameters '{}'",
+                    context.getParsedSql().getSql(),
+                    context.getBinding(),
+                    ex);
+        }
+    }
+
+    private static String format(Duration duration) {
+        final long totalSeconds = duration.getSeconds();
+        final long h = totalSeconds / 3600;
+        final long m = (totalSeconds % 3600) / 60;
+        final long s = totalSeconds % 60;
+        final long ms = duration.toMillis() % 1000;
+        return String.format(
+                "%d:%02d:%02d.%03d",
+                h, m, s, ms);
+    }
+}
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+# 3.19.0
+  - Simple slf4j SqlLogger implementation to debug all statements executed
+
 # 3.18.1
   - Comments like -- and // now recognized and discarded from SQL, thanks @rherrmann!
   - ANTLR 4.9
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
```diff
@@ -0,0 +1,69 @@
+package org.jdbi.v3.core.statement;
+
+import java.time.Duration;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Simple {@link SqlLogger} that emits some diagnostic information about
+ * Jdbi usage.  Usually you shouldn't log every database statement made but
```
leaumar, line 10: I spy a double space, sorry :p

core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
```diff
@@ -0,0 +1,69 @@
...
+    public void logAfterExecution(StatementContext context) {
+        if (log.isDebugEnabled()) {
+            log.debug("Executed in {} '{}' with parameters '{}'",
+                    fmt(Duration.between(context.getExecutionMoment(), context.getCompletionMoment())),
+                    context.getParsedSql().getSql(),
+                    context.getBinding());
+        }
+    }
+
+    private String fmt(Duration dur) {
```
leaumar, line 34: I'd rather not abbreviate method names and params or variables (except the h/m/s, those are ubiquitously known), especially ts is cryptical

core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
```diff
@@ -0,0 +1,69 @@
...
+    public Slf4JSqlLogger() {
+        this(LoggerFactory.getLogger("org.jdbi.sql"));
+    }
+
+    public Slf4JSqlLogger(Logger log) {
+        this.log = log;
+    }
+
+    @Override
+    public void logAfterExecution(StatementContext context) {
```
leaumar, line 25: should probably do `logException` as well

leaumar (changes requested):
some minor things

core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
```diff
@@ -0,0 +1,80 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Simple {@link SqlLogger} that emits some diagnostic information about
+ * Jdbi usage. Usually you shouldn't log every database statement made but
+ * it can be helpful for debugging.
+ */
+public class Slf4JSqlLogger implements SqlLogger {
```
hgschmie, line 14: consider final

core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
```diff
@@ -0,0 +1,80 @@
...
+    public void logException(StatementContext context, SQLException ex) {
+        if (log.isErrorEnabled()) {
+            log.error("Exception while executing '{}' with parameters '{}'",
+                    context.getParsedSql().getSql(),
+                    context.getBinding(),
+                    ex);
+        }
+    }
+
+    private String format(Duration dur) {
```
hgschmie: `private static formatDuration(Duration duration) { ...`

core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
leaumar: Why would you want to avoid people extending it?

core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
leaumar: upvoted for `static` :p

core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
```diff
@@ -0,0 +1,80 @@
...
+        result.append('.');
+        final long millis = duration.toMillis() % 1000;
+        if (millis < 100) {
+            result.append('0');
+        }
+        if (millis < 10) {
+            result.append('0');
+        }
+        result.append(millis);
+        return result.toString();
```
x80486, line 78: I think a `String.format` here will be simpler. Something like this: `return String.format("%d:%02d:%02d.%03d", h, m, s, millis);` — I'm sure that's not 100% accurate, but you get the idea.

stevenschlansker: Yeah, it is much much slower to call format (20x in a benchmark for this) but I suppose the performance here is not too critical.

core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
```diff
@@ -0,0 +1,56 @@
...
+        if (log.isErrorEnabled()) {
+            log.error("Exception while executing '{}' with parameters '{}'",
+                    context.getParsedSql().getSql(),
+                    context.getBinding(),
+                    ex);
+        }
+    }
+
+    private static String format(Duration duration) {
+        final long totalSeconds = duration.getSeconds();
```
x80486, line 46: Actually, you don't need to do these computations since `Duration` already provides that: `duration.toHours()`, `duration.toMinutesPart()`, `duration.toSecondsPart()`, and `duration.toMillisPart()` 😉

stevenschlansker: Unfortunately those methods are not available in Java 8, so making that improvement will need to wait until we move past supporting 8.

core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
stevenschlansker: I don't think anyone should / will extend this, but there's also no particular reason to prevent it, and there's nothing more frustrating than absolutely needing to subclass a final class :)

PR #1840, opened 2021-04-08
Add Codecs for easy type mapping to columns.

hgschmie:
A codec is a replacement for registering an argument and a column
mapper for a type. It is responsible for serializing a typed value
into a database column and creating a type from a database column.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/codec/Codec.java
+++ b/core/src/main/java/org/jdbi/v3/core/codec/Codec.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.codec;
+
+import java.util.function.Function;
+
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.mapper.ColumnMapper;
+
+/**
+ * A Codec provides a convenient way for a bidirectional mapping of an attribute to a database column.
+ * <p>
+ * Groups a {@link ColumnMapper} and an {@link org.jdbi.v3.core.argument.Argument} mapping function for a given type.
+ */
+public interface Codec<T> {
+
+    /**
+     * Returns a {@link ColumnMapper<T>} that creates an attribute value from a database column.
+     */
+    default ColumnMapper<T> getColumnMapper() {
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/jdbi/v3/core/codec/Codec.java
```diff
@@ -0,0 +1,41 @@
...
+     * Returns a {@link ColumnMapper} that creates an attribute value from a database column.
+     */
+    default ColumnMapper<T> getColumnMapper() {
+        throw new UnsupportedOperationException("getColumnMapper");
+    }
+
+    /**
+     * Returns a {@link Function} that creates an {@link Argument} to map an attribute value onto the database column.
+     */
+    default Function<T, Argument> getArgumentFunction() {
```
leaumar, line 38: it's been a long time for me so maybe a stupid question, but why not `ArgumentFactory`?

core/src/main/java/org/jdbi/v3/core/codec/Codec.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.jdbi.v3.core.argument.Argument;
+import org.jdbi.v3.core.mapper.ColumnMapper;
+
+/**
+ * A Codec provides a convenient way for a bidirectional mapping of an attribute to a database column.
+ * <p>
+ * Groups a {@link ColumnMapper} and an {@link org.jdbi.v3.core.argument.Argument} mapping function for a given type.
+ */
+public interface Codec<T> {
```
leaumar, line 26: since a class can implement multiple interfaces and be a ColumnMapper and ArgumentFactory at the same time IIRC, can we skip the getter methods and let Codec be basically an alias for both interfaces? You would `jdbi.registerColumnMapper(codec); jdbi.registerArgumentFactory(codec);` behind the screen then,

core/src/main/java/org/jdbi/v3/core/codec/Codec.java
hgschmie: argument factory and preparable require you to implement build and prepare. This is what the codec factory does for you. This is the minimal API required to implement two-way mapping.

core/src/main/java/org/jdbi/v3/core/codec/Codec.java
hgschmie: that is what the codec factory is about. This is about code convenience, not new functionality. We had tons of proliferation (and skew) with ArgumentFactory and ColumnMapper classes and the codecs allow you to implement all the functionality in a single place.

core/src/main/java/org/jdbi/v3/core/codec/Codec.java
leaumar: But I don't need to implement `prepare`?

```java
@Component
public class CleanStringArgumentFactory extends AbstractArgumentFactory<String> {
    protected CleanStringArgumentFactory() {
        super(Types.VARCHAR);
    }

    @Override
    protected Argument build(String value, ConfigRegistry config) {
        return (position, statement, ctx) -> statement.setString(position, StringUtils.trimToNull(value));
    }
}
```

stevenschlansker (approved):
I left some notes as comments in the code itself: I'm not 100% sold on the API as it stands currently but it's good enough for alpha and we can continue to iterate over time :)

PR #1852, opened 2021-04-11
add mvnw for builds without maven installed

hgschmie:


stevenschlansker:
Hm.  We will then be responsible for managing the Maven version -- for example, the version `3.6.3` you propose to embed is vulnerable to CVE-2021-26291.

I never really understood the appeal of the wrapper programs, it requires checking in a random jar and incomprehensible shell, and everyone who develops on Jdbi will need to have Java + Maven set up anyway.  I'm not necessarily opposed to it but was there some reason to push it through?

x80486:
Basically you don't pay attention to that JAR file and the shell scripts 🤣

I think this is a good to have, so you don't really need to know the version of Maven you need to work with. The wrapper script takes care of all of that.

By the way, the wrapper is going to be part of the official Maven distribution any time soon, so I guess the setup is going to be like that anyway.

hgschmie:
This is an answer to the "ugh, I need to install build tool <X>" to try this out. Other technologies (and other build tools such as gradle) popularized the approach of "I bring a command line tool, nothing needs to be installed" approach.

As @x80486 mentioned, this is the direction into which maven will go anyway; it may require maven 4 to be released first, though.

The "cryptic shell scripts and jar file" are auto-generated by the takari wrapper plugin.

PR #1863, opened 2021-04-22
Add one-time init methods to row and column mappers

hgschmie:
Allow creation-time initialization of row and column mappers. Also
make the config registry available to the factory methods on the
codec interface.


stevenschlansker (approved):
If you have time I recommend we add a test that init is called, one each for row and column mapper.

PR #1868, opened 2021-05-03
caffeine 3.0.2 with complete Unsafe removal

stevenschlansker:


qualidafial (approved):

GeorgeSalu:
hello, caffeine is used exactly for what? already seen some answers in some closed issues that he is not using to cache queries? caffeine is our main lib cache, my question is will there be any problems using this same version that comes with jdbi to cache queries? we use spring boot

ben-manes:
Caffeine strictly follows semver and we try to be very conscientious about backwards compatibility. The changes visible between 2.x and 3.x are quite minor and both are maintained. In jdbi's case either version works. If you use the Spring Cache abstraction then that is fully backwards compatible. Very few users should be impacted by version changes and those fixes are relatively trivial.

This bump is a patch version so there are no API changes, only implementation improvements.

stevenschlansker:
Hi @GeorgeSalu , there's a lot of work Jdbi does behind the scenes based on the types and SQL you use:
parsing SQL text, re-using created argument and column mappers for the same Type over and over again, things like that.

Previously we used built in JDK Maps to do this, but if you generate new SQL or types dynamically (or expect class unloading to work), a map with a hard reference is essentially a memory leak.  We wanted to stay with built in JDK types if we could, but solutions like WeakHashMap ended up not being a good fit, so we fell back to using Caffeine as a relatively lightweight cache with an expiry mechanism.  In "normal" usage you should never expire any data, but large or specialty applications might have to choose between caching for performance and cache memory usage.

PR #1882, opened 2021-06-09
Fix dependency problems found by muse

hgschmie:


pom.xml
```diff
@@ -225,7 +225,7 @@
             <dependency>
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
-                <version>21.0</version>
```
stevenschlansker, line 228: Guava versioning and backwards compatibility is a mess. I'm somewhat concerned we'll get some random downstream breakage because we changed this version.

stevenschlansker (approved):

hgschmie:
closed in favor of #1894

PR #1885, opened 2021-06-13
Fix typo 'requied'

Dalje-et:
Getting started with Jdbi and found a typo in the docs. Hope this is the correct way to create PRs.


hgschmie:
Thank you!

PR #1888, opened 2021-06-22
Guice integration

hgschmie:
Provide integration for JDBI with the Guice DI framework.


stevenschlansker (approved):
Needs docs before merge.

PR #1889, opened 2021-06-23
Add Database metadata access through the handle

hgschmie:
Allows accessing any method in the Database metadata either through a ResultBearing or simple values.


stevenschlansker (approved):
Nifty. The code turned out really nice. It'd be nice to have some docs

PR #1892, opened 2021-07-03
fix doc build, deploy kotlin docs again

hgschmie:


stevenschlansker:
Thanks for fixing this.  I do think there was a reason `publish-docs` was a `-D` property and not switched on with `-P` profile selection, which is that doing `-Ppublish-docs` inhibits all other profiles. Without toolchains, maybe this is not a big deal, but could be surprising if any new profiles are added later.

hgschmie:
the profile activation is tricky and the inhibiting happens if you have multiple profiles next to each other and one has "activeByDefault".  The docs read somewhat opaque: "This profile will automatically be active for all builds unless another profile in the same POM is activated using one of the previously described methods. All profiles that are active by default are automatically deactivated when a profile in the POM is activated on the command line or through its activation config."

What it means is that this applies to all profiles in the same pom only and as soon as another profile is activated, all profiles that have "activeByDefault" in it get disabled. 

Another weird mavenism.

I tested this with the code in #1894, which introduces a profile that is only activated when running on JDK8:

```
❯ mvn -pl :jdbi3-docs help:active-profiles
[INFO] Scanning for projects...
[INFO]
[INFO] ------------------------< org.jdbi:jdbi3-docs >-------------------------
[INFO] Building jdbi3 documentation 3.20.2-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO]
[INFO] --- maven-help-plugin:3.2.0:active-profiles (default-cli) @ jdbi3-docs ---
[INFO]
Active Profiles for Project 'org.jdbi:jdbi3-docs:jar:3.20.2-SNAPSHOT':

The following profiles are active:

 - jdk8 (source: org.jdbi:jdbi3-parent:3.20.2-SNAPSHOT)
```

```
❯ mvn -Ppublish-docs -pl :jdbi3-docs help:active-profiles
[INFO] Scanning for projects...
[INFO]
[INFO] ------------------------< org.jdbi:jdbi3-docs >-------------------------
[INFO] Building jdbi3 documentation 3.20.2-SNAPSHOT
[INFO] --------------------------------[ jar ]---------------------------------
[INFO]
[INFO] --- maven-help-plugin:3.2.0:active-profiles (default-cli) @ jdbi3-docs ---
[INFO]
Active Profiles for Project 'org.jdbi:jdbi3-docs:jar:3.20.2-SNAPSHOT':

The following profiles are active:

 - publish-docs (source: org.jdbi:jdbi3-docs:3.20.2-SNAPSHOT)
 - jdk8 (source: org.jdbi:jdbi3-parent:3.20.2-SNAPSHOT)
```

It works as expected (and does not deactivate any other active profile). I can see how there are some interactions with the toolchain but I am pretty confident that this works.

PR #1902, opened 2021-07-15
Map `long` and `Long` arguments to `BIGINT`

joschi:
The SQL type `INTEGER` often (always?) maps to integer types which are only 32 bits (4 bytes) long while `BIGINT` maps to integer types which are 64 bits (8 bytes) long.

Considering this, mapping the Java types `long` and `Long` to the SQL type `BIGINT` instead of `INTEGER` seems to make more sense.

This is also consistent with the handling of `OptionalLong` in Jdbi:
https://github.com/jdbi/jdbi/blob/ee22bb85a6971737181a4b231cce4dbd55de8734/core/src/main/java/org/jdbi/v3/core/argument/OptionalArgumentFactory.java#L38-L44

FWIW, this is motivated by a bug we discovered in CockroachDB when sending `NULL` values in a prepared statement which were sent as `NULL::INT4` by Jdbi 3.20.1 instead of `NULL::INT8` for a column declared as `INT8`:
https://github.com/cockroachdb/cockroach/issues/67605


stevenschlansker (approved):
Thanks for catching this!

PR #1905, opened 2021-07-16
fix weird documentation timestamp

hgschmie:


stevenschlansker (approved):
Not quite seeing where the crazy timestamp came from, but the fix looks reasonable.

PR #1907, opened 2021-08-02
Update Travis CI badge URL from `travis-ci.org` to `travis-ci.com`

drunderscore:
The old badge image URL redirects to `api.travis-ci.com`, so I've used that instead.
The old badge URL redirects to `/github/jdbi/jdbi`, instead of just `/jdbi/jdbi`, so I've used the former instead.


stevenschlansker:
Thanks!

stevenschlansker (approved):

PR #1920, opened 2021-10-07
Fix annotation inheritance for non-direct supertypes

electrum:


hgschmie:
Hi @electrum! Thank you for the PR!

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/CustomizingStatementHandler.java
```diff
@@ -74,6 +75,13 @@
             .collect(Collectors.toList());
     }
 
+    private static Stream<Class<?>> superTypes(Class<?> type) {
```
hgschmie, line 78: this seems to be the same as the code in SqlObjectFactory. We tend to use the qualified methods (`Array.stream`) not static imports.

It would be good to note that the same code is also in SqlObjectFactory so that we could consider pulling it together if we need to use it in more than two places.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/TestInheritedAnnotations.java
```diff
@@ -85,7 +85,9 @@ public void testCrud() {
     }
 
     @RegisterConstructorMapper(Character.class)
-    public interface CharacterDao extends CrudDao<Character, Integer> {}
```
hgschmie, line 88: This modifies the existing test. Can we add a second test that explicitly looks for the non-direct supertypes?

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/CustomizingStatementHandler.java
electrum: I was copying the already statically imported `Stream.concat`. Removed static import.

I also added comments to each that mention the other copy. We could move this to somewhere in `org.jdbi.v3.sqlobject.statement.internal`, maybe `SqlAnnotations`, but it doesn't exactly fit.

PR #1929, opened 2021-10-15
Use compiled regex for replacing snake case column names

hkupty:
First of all, thanks for the library.

The idea of this PR comes from analyzing an internal system with JFR/JMC. I can see that the `Pattern` object that results from `removeUnderscores` seems to be created frequently. Below is a 5 minute sample of the system:
image

This should reduce the `Pattern` object creation, reducing the overall impact in GC.


stevenschlansker:
I think this is because `String.replace(String, String)` is actually doing literal char replacement, not Pattern replacement. When I look at the source code for JDK17 I don't see a Pattern call at all. What JDK do you use to test? Maybe upgrading your JDK would fix this performance issue better.

hkupty:
That makes sense... The service is running in jdk8 in prod, while the JMH test was run using the latest jdk (found out now that you pointed out that the local jdk version is newer).

PR #1939, opened 2021-10-24
Resolve generic types on bean property arguments when the bean itself is generic

### LINKED ISSUE #1914: Regression: Invalid type resolution of properties for types with generics
This was caused by #1665.

When binding beans of a type with generics that has some property with generics, the property type is erased.
E.g.
```
public class MyEntity<T> {
    public List<String> getAbc();
}
```
Binding this type as a bean will resolve "abc" property as "List" type, losing the generic type. This is because `GenericTypeReflector.getExactReturnType` will erase the type if the bean type is "raw" (i.e. Class, not ParameterizedType), which is always the case in bound beans.
This might cause problems in custom ArgumentFactories that rely on this information.

stevenschlansker:
Fixes #1914

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,4 +1,5 @@
 # Unreleased
+  - fix Bean property arguments type being erased on generic beans
 
 # 3.24.0
   - Fix batch binding with mixed null and non-null primitives (#1901)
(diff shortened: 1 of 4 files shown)


hgschmie (approved):
LGTM

Issue #1944, opened 2021-10-28
Registering KotlinMapper with a prefix does not seems to be returning correct results


mahesh-chotrani:
While using `KotlinMapper`, when data is being fetched from more than one tables, and few column names are same from both the tables (like, `id`), the result is not populated correctly in the data classes.

Here is how a sample query looks like (`id` column is same in both the tables)
```
SELECT 
  t.id, t.name, 
  p.id pid, p.tagId ptagId, p.primaryName pprimaryName 
FROM Tag t JOIN Product P ON t.id = p.TagId
```
**Tag** and **Product** data classes 
```
data class Tag(
    val id: Int?,
    val name: String?
)
data class Product(
    val id: Int?,
    val primaryName: String?,
    val tagId: Int?,
)
```
And the code snippet
```
val rows = JdbiManager.jdbi.withHandleSync { handle ->
    handle.registerRowMapper(KotlinMapper(Tag::class.java))
    handle.registerRowMapper(KotlinMapper(Product::class.java, "p"))
    handle.registerRowMapper(JoinRowMapper.forTypes(Tag::class.java, Product::class.java))
    handle.createQuery(sql).mapTo<JoinRow>().list()
}
```

`Tag` instance has proper `id` populated where as the `id` in `Product` instances is that of `Tag` instances.

**NOTE** If `FieldMapper` is used in place of `KotlinMapper` (_and including no-arg constructors in both the data classes_), the above code works properly and correct ids are populated in `Product` instances.

hgschmie:
This turned out to be a really interesting problem. See #1945 for the full fix. Your previous code should now work unchanged. 

Fun fact: You did not have to do `handle.registerRowMapper(KotlinMapper(Tag::class.java))` as the kotlin plugin registers a factory to do that for you. That was the root of the problem (that and the fact the java inferring mapper could not detect the types for the kotlin mapper). 

Fix will be in the next release. Until then, you can use the workaround as decribed above (register with explicit types).

mahesh-chotrani:
Thanks @hgschmie for the fix. And the detailed explanation of the issue. 

I'll use the workaround for now.

PR #1945, opened 2021-10-28
Fix the Kotlin mapper lookup problem

hgschmie:
This fixes the problems described in #1944 by changing the inference for mappers without a type information. By default, this uses inference to  find a generic type parameter and use that. This does not work for the Kotlin mappers as they all implement `RowMapper<Object>`. The fix registers an interceptor that uses information in the KotlinMapper to register them correctly.


core/src/main/java/org/jdbi/v3/core/inference/JdbiInterceptorChain.java
```diff
@@ -0,0 +1,119 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * An interception chain to process transformation operations.
+ *
+ * @param <S> The source type.
+ * @param <T> The target type.
+ */
+public final class JdbiInterceptorChain<S, T> {
```
stevenschlansker, line 30: Are we sure we want to make all this interceptor stuff public? We'll be committing to supporting all of it going forward. If we do expose this interface, we should probably document it.

stevenschlansker:
I am not entirely clear why we need this new interceptor chain concept. I had thought this could be implemented as a classic RowMapperFactory et al - they take a Type and are given an opportunity to make a Mapper, which looks very close to the signature of the new interceptor chain concept.

core/src/main/java/org/jdbi/v3/core/inference/JdbiInterceptorChain.java
hgschmie: well, it is in a separate package. And the sql array and the mappers are in separate packages. IMHO, just because it is public, it does not have to be part of the public API. I definitely see your point, what would you suggest to do?

core/src/main/java/org/jdbi/v3/core/inference/JdbiInterceptorChain.java
```diff
@@ -0,0 +1,119 @@
...
+
+    private final List<JdbiInterceptor<S, T>> interceptors = new CopyOnWriteArrayList<>();
+    private JdbiInterceptor<S, T> defaultInterceptor;
```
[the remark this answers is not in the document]
hgschmie: I wanted to make sure that the default interceptor is always run last. With the "addFirst/addLast", one could otherwise not add an interceptor "to the end, right before the last one" unless we want to do length()-1 arithmetics when adding an interceptor. I am open to this; it would just be an implementation detail.

core/src/main/java/org/jdbi/v3/core/inference/JdbiInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+    @CheckForNull
+    T intercept(@Nullable S source);
+
+    /**
+     * Returns true if this interceptor can handle the source object.
+     *
+     * @param source A source object.
+     * @return True when the interceptor can handle the source object.
+     */
+    default boolean accept(@Nullable S source) {
```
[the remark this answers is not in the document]
hgschmie: java.util.Iterator has a very distinct functionality. I could not see how to fit that functionality (test, then do) into that API, can you make a suggestion? 

IMHO, Servlet Filter would be a better example; maybe we could revamp the API to be more of that shape.

hgschmie:
The challenge is that in the current `RowMappers` etc. there is this method hardcoded: (https://github.com/jdbi/jdbi/blob/master/core/src/main/java/org/jdbi/v3/core/mapper/RowMappers.java#L63-L65)

```java
public RowMappers register(RowMapper<?> mapper) {
        return this.register(new InferredRowMapperFactory(mapper));
    }
```

so any caller of `registerRowMappers(RowMapper<?> mapper)` end up here. What the code does is some dark-grey magic to infer the type of the type parameter in `... implements RowMapper<Stuff>` to be stuff.

Now, for the Kotlin code, every time someone registers `KotlinMapper(KotlinType::class.java)`, they assume that this magic works and yields `KotlinType`. It does not. KotlinMapper implements `RowMapper<Object>`, so the InferredRowMapperFactory always finds `Object` for the inferred type. So it never matches when the actual type is requested and the mapper is never used. That is the core of the problem.

For Kotlin (or any other special case where the mappers implement another type but the one they are supposed to be used for) to work, this mapping from type -> `InferredRowMapperFactory` needs to be ... well ... intercepted and handled by a different factory. Hence the interceptor code. The `KotlinPlugin` now registers a new RowMapperFactory (that determines whether it should return KotlinMapper in a different way) and every call to `registerRowMapper(KotlinMapper)` now ends up on that new factory. Hence existing code works unchanged and the bug is fixed. 

We could make the mapper for inference pluggable, but I would bet dollars for donuts that the next bug report will be "module "a" that registers an inference mapper and module "b" that register an inference mapper do not work nicely together. So I just changed the thing wholesale. :-)

core/src/main/java/org/jdbi/v3/core/inference/JdbiInterceptorChain.java
stevenschlansker: We reserve all `.internal` packages as not-API. Maybe it could move into an internal package, at least until we are confident to commit to users extending it as well. Or we could mark it as `@Beta` in case we want to reserve the right to change it a bit.

stevenschlansker:
Another choice would be to instead encourage users to register a `KotlinMapperFactory` with a list of types, which then returns a `KotlinMapper`. This would change the public api slightly, but would remove the need to add a new nontrivial feature to the mapper logic, which is already ridiculously hard to follow...

If you think being compatible with old code is worth introducing this complexity, I have no objection, just wanted to push back a bit on making a new interface that works almost exactly the same as the old interface but a bit different.

hgschmie:
I hate breaking backwards compatibility. If we don't support this, we force kotlin users to use Jdbi different than java users. Whether the interceptor framework (three small classes) is overkill, I can not really judge but I feel that being able to open up that hard dependency (hardcoded class in the RowMappers etc. classes) is important. The alternative (IMHO) would be to deprecate the "register and detect the type by magic" method calls.

PR #1957, opened 2021-11-17
Json: rework inner @Json String binding to be @EncodedJson String

### LINKED ISSUE #1953: @Json String behavior question
I'm pretty sure this isn't a bug, but don't quite understand the behavior of `@Json String param` described here in the docs.
[https://jdbi.org/#_operation]()
 
 
 When using postgres, if I want to update a string field of a jsonb column I seem to have to use the postgres builtin `to_jsonb` in my query:

 ```
     @SqlUpdate("update table "
            + " set detail = jsonb_set(detail, '{fieldKey}', to_jsonb(:fieldValue)) "
            + " where id = :id")
    void setFieldValue(UUID id, String stateId);
```
(using `-parameters` compile flag)

Intuitively, I thought the following would work but it doesn't:
```
    @SqlUpdate("update table "
            + " set detail = jsonb_set(detail, '{fieldKey}', :fieldValue) "
            + " where id = :id")
    void setFieldValue(UUID id, @Json String fieldValue);
```
This is called out as an exception in the documentation but I guess I'd like to understand why or if I'm doing something wrong.

stevenschlansker:
This allows you to actually bind a @Json String which users intuitively expect to be able to do.
Fixes #1953

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,4 +1,5 @@
 # Unreleased
+  - *SPI change* `@Json String` database type mappers now use `@EncodedJson String` instead (#1953)
 
 # 3.24.1
   - fix Bean property arguments type being erased on generic beans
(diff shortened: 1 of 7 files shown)


hgschmie:
the "changed 35 files" scared me quite a bit until I realized that it includes all the pom 3.24.1 -> 3.25.0 changes. It would be good to do that separately.

PR #1958, opened 2021-11-17
Batch: clarify execute() int[] result might be weird values like -2

### LINKED ISSUE #1956: PreparedBatch.execute() return wrong values
(jdbi3-core version 3.21.0)
I am using PreparedBatch, to execute INSERT query with multiple rows.
I am binding 5 maps, and executing the batch.
All 5 rows are written as expected, but the response from the execute method is:
[-2, -2, -2, -2, -2] 
Am I doing something wrong?

```
int result = jdbi.withHandle(handle -> {
    PreparedBatch batch = handle.prepareBatch(query);
    for (Entity entity : entities) {
        batch.add(entity.getMap());
    }
    int[] counts = batch.execute();
    return ?
});
```

stevenschlansker:
Fixes #1956

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/Batch.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/Batch.java
@@ -46,9 +46,12 @@ public class Batch extends BaseStatement<Batch> {
     }
 
     /**
-     * Execute all the queued up statements
+     * Execute the batch and return the number of rows affected for each batch part.
+     * Note that some database drivers might return special values like {@link Statement#SUCCESS_NO_INFO}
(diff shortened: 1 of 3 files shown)


hgschmie:
maybe add a link to https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/Statement.html#executeBatch() and highlight that each return value matches what is written there (>0 means rows but there are some special values).

hgschmie (approved):
shippit

stevenschlansker:
Yeah, I have a link to that Statement as an `@see` which should be automatically xref'ed in Javadoc or IDE.

PR #1959, opened 2021-11-17
ConstructorMapper: defer throwing exception about missing columns

### LINKED ISSUE #1955: PrepareBatch.executeAndReturnGeneratedKeys().mapTo() fails on empty results on Postgres
Jdbi fails when executing a prepared batch using `executeAndReturnGenereratdKeys` and a globally configured constructor mapper and mapTo(Class) when the result set is empty. 

Here is a sample program to reproduce the problem. As you can see we are batching `insert on conflict ignore`. In case of conflicts on all inserts an empty result is produced causing the logic to specialize the constructor mapper to throw an exception.

```
class Test {
    public record User(String userId) {
    }
    
    public static List<User> insert(Jdbi jdbi, List<User> users) {
        return jdbi.withHandle(h -> {
            PreparedBatch batch = h.prepareBatch("""
                    insert into user (user_id)
                    values (:user_id)
                    on conflict (user_id) do nothing
                    returning user_id
                    """);

            for (User user : users) {
                batch.bind("user_id", user.userId());
                batch.add();
            }

            return batch.executeAndReturnGeneratedKeys()
                    .mapTo(User.class) // logic to resolve mapper fails when result set is empty
                    .list();
        });
    }
    
    public static void foo() {
        Jdbi jdbi = Jdbi.create("some-url-to-postgres");
        jdbi.registerRowMapper((RowMapperFactory) (t, r) ->
                t instanceof Class<?> c && c.isRecord() ? Optional.of(ConstructorMapper.of(c)) : Optional.empty());

        List<User> users = List.of(new User("bob"));
        insert(jdbi, users); // -> bob
        insert(jdbi, users); // fails but should return -> empty list
    }
}
```

A workaround is to replace `mapTo` with an explicit `map`.

```
            return batch.executeAndReturnGeneratedKeys()
                    .map(r -> r.getRow(User.class))
                    .list();
```

stevenschlansker:
until we actually see rows 
fixes #1955

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,7 @@
 # Unreleased
   - *SPI change* `@Json String` database type mappers now use `@EncodedJson String` instead (#1953)
   - Sql4JSqlLogger: fix NPE when using Script
+  - fix using ConstructorMapper on empty generated keys results (#1955)
 
 # 3.24.1
   - fix Bean property arguments type being erased on generic beans
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -170,6 +170,7 @@
                                 <exclude>**/TestPreparedBatchGenerateKeysPostgres.java</exclude>
                                 <exclude>**/TestPreparedBatchPG.java</exclude>
                                 <exclude>**/TestScript.java</exclude>
+                                <exclude>**/ConstructorMapperPgTest.java</exclude>
                             </excludes>
                         </configuration>
                     </plugin>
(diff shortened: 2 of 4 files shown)


stevenschlansker:
@hgschmie , the jdk8 tests are failing:
```
[INFO] Running org.jdbi.v3.core.mapper.reflect.ConstructorMapperPgTest
Error:  Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0 s <<< FAILURE! - in org.jdbi.v3.core.mapper.reflect.ConstructorMapperPgTest
Error:  org.jdbi.v3.core.mapper.reflect.ConstructorMapperPgTest  Time elapsed: 0 s  <<< ERROR!
java.lang.UnsupportedClassVersionError: de/softwareforge/testing/postgres/junit5/EmbeddedPgExtension has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

Any ideas?

hgschmie:
yes, you need an exclusion to the pom in the jdk8 profile

Issue #1971, opened 2021-12-18
Regression: @RegisterKotlinMappers cannot be applied to interface


usulkies:
Using the declarative style with Dao required an interface.
Since the merge of PR #1930 - the RegisterKotlinMappers annotation cannot be applied to interfaces, since AnnotationTarget.TYPE does not include interfaces.
Annotating a Dao interface with this annotation is giving a compilation error:
`This annotation is not applicable to target 'interface'`

Seems like there's no test for this anywhere.

hgschmie:
Fixed by cba804c159d2, will be in next release.

PR #1977, opened 2021-12-19
Fix issue #1971

usulkies:
This change is aimed to fix the issue https://github.com/jdbi/jdbi/issues/1971

`AnnotationTarget.CLASS` includes interfaces, while `AnnotationTarget.TYPE` does not.

Added a test that would fail if the RegisterKotlinMappers for Dao interfaces will break.


hgschmie:
Thank you for the PR!

hgschmie:
Merged as #1978 (also fix the other annotation)

PR #1993, opened 2022-01-18
PrepareBatch for every attempt in SqlBatchHandler

### LINKED ISSUE #1967: SerializableTransactionRunner loses records on retry
Hi,

I have got a batch insert occuring on a table 'test' with the use of `@SqlBatch` annotation. The batch size is 512 and there are about 1000 records to insert to the table. The configuration for JDBI transaction handler is [SerializableTransactionRunner](https://github.com/jdbi/jdbi/blob/0420cc462f98565062bd577cf5d3a4d92a621060/core/src/main/java/org/jdbi/v3/core/transaction/SerializableTransactionRunner.java).

When a failure occurs from the database during commit especially around [here](https://github.com/jdbi/jdbi/blob/0420cc462f98565062bd577cf5d3a4d92a621060/core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java#L210)
```
 public <R, X extends Exception> R inTransaction(Handle handle,
                                                        HandleCallback<R, X> callback) throws X {
            if (isInTransaction(handle)) {
                throw new IllegalStateException("Already in transaction");
            }
            didTxnRollback = false;
            final R returnValue;
            try {
                handle.begin();
                returnValue = callback.withHandle(handle);
                if (!didTxnRollback) {
                    **// ISSUE: Database throws exception here
                    handle.commit();**
                }
            } catch (Throwable e) {
                try {
                    handle.rollback();
                    e.addSuppressed(rollback);
                }
                throw e;
            } finally {
                didTxnRollback = false;
            }

            return returnValue;
        }

```

, the transaction runner tries to re-try the batch around [here](https://github.com/jdbi/jdbi/blob/0420cc462f98565062bd577cf5d3a4d92a621060/core/src/main/java/org/jdbi/v3/core/transaction/SerializableTransactionRunner.java#L53-L79)
```
while (true) {
	try {
               **// Second attempt has 0 bindings and commit transaction completes in 0 ms
		R result = getDelegate().inTransaction(handle, callback);**
		config.onSuccess.accept(new ArrayList<>(failures));
		return result;
	} catch (Exception last) {
		X x = (X) last;

		// throw immediately if the exception is unexpected
		if (!isSqlState(config.serializationFailureSqlState, x)) {
			throw last;
		}

		failures.addLast(x);
		config.onFailure.accept(new ArrayList<>(failures));

		// no more attempts left? Throw ALL the exceptions! \o/
		attempts -= 1;
		if (attempts <= 0) {
			X toThrow = failures.removeLast();
			while (!failures.isEmpty()) {
				toThrow.addSuppressed(failures.removeLast());
			}
			throw toThrow;
		}
	}
}
```

The issue is the retry-attempt will always be successful as the bind is cleared by execute. The definition for [execute](https://github.com/jdbi/jdbi/blob/0420cc462f98565062bd577cf5d3a4d92a621060/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java#L68-L71) is:

`Function<PreparedBatch, ResultIterator<?>> modCounts = PreparedBatch::executeAndGetModCount;
batchIntermediate = method.getReturnType().equals(boolean[].class) ? mapToBoolean(modCounts) : modCounts;`

This particular execution [clears the bind](https://github.com/jdbi/jdbi/blob/127fda14d9875518670a4b568fdd84688a2a95ea/core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java#L236) as a result, the subsequent retry doesn't have any bind and is marked successful. 

An example scenario per my batch insert for 1000 records with batch size of 512 is as follows:

1. Batch#1 with 512 records fails from the database and records are not committed.
2. SerializableTransactionRunner retries and is successful with empty binds.
3. Batch#2 with 488 records are successful.

Queried the table and Batch#1 consisting of the 512 records are missing.

Edit:
The issue fix should be around at [SqlBatchTransactionHandler](https://github.com/jdbi/jdbi/blob/0420cc462f98565062bd577cf5d3a4d92a621060/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java#L311-L319):

```
private ResultIterator<?> executeBatch(final Handle handle, final PreparedBatch batch) {
	if (!handle.isInTransaction() && sqlBatch.transactional()) {
		// it is safe to use same prepared batch as the inTransaction passes in the same
		// Handle instance.
		// Actual:
                // return handle.inTransaction(c -> batchIntermediate.apply(batch));
                // Avoid clearing the original batch with binds by cloning
		return handle.inTransaction(c -> batchIntermediate.apply(deepCopy(batch)));


	} else {
		return batchIntermediate.apply(batch);
	}
}
```

sudharsannr:
Fixes #1967 . When SQLBatchHandler is used in conjuction with SerializableTransactionRunner (retry on failure), subsequent retry has the binding for PreparedBatch cleared. Therefore, make a copy of the binding arguments before executing and this ensures the binding's are not cleared.

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java
@@ -13,6 +13,19 @@
  */
 package org.jdbi.v3.sqlobject.statement.internal;
 
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.lang.reflect.Type;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.extension.HandleSupplier;
@@ -31,20 +44,6 @@ import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
 import org.jdbi.v3.sqlobject.statement.SqlBatch;
 import org.jdbi.v3.sqlobject.statement.UseRowMapper;
 import org.jdbi.v3.sqlobject.statement.UseRowReducer;
-
-import java.lang.annotation.Annotation;
-import java.lang.reflect.Method;
-import java.lang.reflect.Parameter;
-import java.lang.reflect.Type;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.List;
-import java.util.NoSuchElementException;
-import java.util.function.Function;
-import java.util.function.Supplier;
-import java.util.stream.IntStream;
-import java.util.stream.Stream;
 
 public class SqlBatchHandler extends CustomizingStatementHandler<PreparedBatch> {
     private final SqlBatch sqlBatch;
@@ -198,8 +197,7 @@ public class SqlBatchHandler extends CustomizingStatementHandler<PreparedBatch> 
                 // execute a single chunk and buffer
                 List<Object[]> currArgs = new ArrayList<>();
                 for (int i = 0; i < chunkSize && batchArgs.hasNext(); i++) {
-                    Object[] batchArg = batchArgs.next();
-                    currArgs.add(Arrays.copyOf(batchArg, batchArg.length));
+                    currArgs.add(batchArgs.next());
                 }
                 Supplier<PreparedBatch> preparedBatchSupplier = () -> createPreparedBatch(handle, sql, currArgs);
                 return executeBatch(handle, preparedBatchSupplier);
@@ -316,7 +314,6 @@ public class SqlBatchHandler extends CustomizingStatementHandler<PreparedBatch> 
                     + " did you mean @SqlQuery?", null, null);
         }
 
-        final Object[] sharedArg = new Object[args.length];
         return new Iterator<Object[]>() {
             @Override
             public boolean hasNext() {
@@ -330,6 +327,7 @@ public class SqlBatchHandler extends CustomizingStatementHandler<PreparedBatch> 
 
             @Override
             public Object[] next() {
+                final Object[] sharedArg = new Object[args.length];
                 for (int i = 0; i < extras.size(); i++) {
                     sharedArg[i] = extras.get(i).next();
                 }
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java
@@ -17,11 +17,9 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 import java.lang.reflect.Type;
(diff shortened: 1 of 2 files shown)


sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java
```diff
@@ -17,11 +17,9 @@
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 import java.lang.reflect.Type;
-import java.util.ArrayList;
-import java.util.Iterator;
-import java.util.List;
-import java.util.NoSuchElementException;
+import java.util.*;
```
stevenschlansker, line 20: Our style rules don't allow `*` imports

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java
```diff
@@ -194,12 +192,22 @@ public Object invoke(Object target, Object[] args, HandleSupplier h) {
 
             private ResultIterator<?> loadChunk() {
                 // execute a single chunk and buffer
-                PreparedBatch batch = handle.prepareBatch(sql);
+                List<Object[]> currArgs = new ArrayList<>();
                 for (int i = 0; i < chunkSize && batchArgs.hasNext(); i++) {
-                    applyCustomizers(batch, batchArgs.next());
+                    Object[] batchArg = batchArgs.next();
+                    currArgs.add(Arrays.copyOf(batchArg, batchArg.length));
```
stevenschlansker, line 198: This is to avoid keeping references to the `sharedArg` at line 307, right? If so, there is no longer a point to sharing that array - instead of copying a shared array repeatedly we can just allocate a new one for each iteration.

stevenschlansker:
Thank you so much for contributing this @sudharsannr ! I've enabled checks and left a couple of minor comments to fix. We should be able to integrate and release this shortly after all checks and comments pass.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java
sudharsannr: Got it. Working on it and also checking failures

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java
sudharsannr: No, this is to ensure that the `batchArgs` are copied before submitting to `PreparedBatch` to ensure the args aren't cleared after execute. This is the fix where `SerializableTransactionRunner` (line 55) - on a second retry will preserve these args.

sudharsannr: Thanks for this. Initializing sharedArg everytime we do this.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java
```diff
@@ -330,6 +327,7 @@ public boolean hasNext() {
 
             @Override
             public Object[] next() {
+                final Object[] sharedArg = new Object[args.length];
```
stevenschlansker, line 330: The argument array is no longer shared, so how about just `args` or `argArray`

sudharsannr: Updated

stevenschlansker (approved):

PR #1997, opened 2022-01-30
inline plugin experiment

hgschmie:
This is a highly experimental build that inlines the antlr4-runtime into the jdbi3-core plugin. 

The build passes all the tests but might need a bit more baking before we could mainline this. If you want to try it out, check out this branch, build locally and then build your service / application etc. using the -SNAPSHOT jars. Please report any errors etc. here. 

See https://github.com/basepom/inline-maven-plugin for more information about the inline plugin.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -44,10 +44,6 @@
     </properties>
 
     <dependencies>
-        <dependency>
-            <groupId>org.antlr</groupId>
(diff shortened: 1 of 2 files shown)


stevenschlansker:
Looks reasonable enough. I'll try to run some tests on this when I get a chance.

hgschmie:
superseded by #2006

Issue #1961, opened 2021-12-03
`Slf4JSqlLogger.logAfterExecution()` should handle a null `ParsedSql`


bseib:
In the [Slf4JSqlLogger.logAfterExecution()](https://github.com/jdbi/jdbi/blob/bde09ea617f228ff9a8218c0f12fcc1fc216e9fe/core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java#L43) function:

```java
    @Override
    public void logAfterExecution(StatementContext context) {
        if (log.isDebugEnabled()) {
            log.debug("Executed in {} '{}' with parameters '{}'",
                    format(Duration.between(context.getExecutionMoment(), context.getCompletionMoment())),
                    context.getParsedSql().getSql(),
                    context.getBinding());
        }
    }
```

The function `context.getParsedSql()` can return null and cause a NPE. This showed up in some of our app's unit tests as soon as we added SQL logging, i.e. `jdbi.setSqlLogger(Slf4JSqlLogger(logger))`.

I have not created a completely isolated test case yet, but the code that created this situation looks approximately like this (kotlin):

```kotlin
    private fun getInitializationSql(): String {
        return """
            CREATE TABLE public.foo_table(
              foo_id BIGINT NOT NULL,
              foo_name TEXT NOT NULL
            ) 
            WITH (oids = false);
            
            CREATE INDEX foo_name_idx ON public.foo_table
              USING btree (foo_name COLLATE pg_catalog."default");

            INSERT INTO foo_table(
                foo_id,
                foo_name
            ) VALUES (
                42,
                'Person Name'
            );
        """.trimMargin()
    }

    private fun initDatabase(jdbi: Jdbi) {
        jdbi.open().use { handle ->
            handle.createScript(getInitializationSql()).execute()
        }
    }

    // and jdbi has been setup something like:
    jdbi = Jdbi.create(dataSource)
    // install jdbi plugins and register column mappers
    jdbi.setSqlLogger(Slf4JSqlLogger(logger))

```

Perhaps the name of this issue is not correct, if the parsed sql string should never be null. It is not clear to me when the `ParsedSql` value could be null. The above example does contain multiple SQL statements in a single string, so perhaps that is a contributing factor to not being able to create a `ParsedSql` object.  ??

----

[...]

PR #2000, opened 2022-02-17
Fix NPE for Slf4JSqlLogger#logAfterExecution

tmichel:
The NPE issue (https://github.com/jdbi/jdbi/issues/1961) during logging has already come up but the fix (https://github.com/jdbi/jdbi/commit/00172225de8f2feb165c5832d382620391d0217e) only addressed the NPE for `logException` and not for `logAfterExecution`.

I extracted getting the SQL string from the `StatementContext` into a helper method. I opted for not using an Optional and rather use a conditional.

There were no tests for `Slf4jSqlLogger` so I added some that reproduced the issue. I opted to not write a comprehensive test suite to keep the pull request more focused.


stevenschlansker (approved):
Thank you for fixing this!

tmichel:
I fixed the unused import error. What else do you need to get this merged?

stevenschlansker:
Thanks @tmichel , released as 3.27.2

Issue #1798, opened 2020-12-15
antlr has hard dependency on runtime, but version might be changed by integrators


stevenschlansker:
Unfortunately the ANTLr-generated code has a version check embedded, see #1793
Maybe we need to shade in the runtime, or perhaps there's another fix...

hgschmie:
We may want to bump this upward. shadowing the antlr runtime would fix a lot of pain for users

PR #2006, opened 2022-02-24
Remove antlr4-runtime dependency

hgschmie:
Inline the dependency into the core jar. This removes the long
standing problem that jdbi creates version conflicts when used in
projects that also use antlr4 and that use a different runtime
version. Unfortunately antlr insists on a strong relationship between
the generated classes for the grammar and the runtime.

Inlining this dependency makes allows JDBI users to choose their antlr
version arbitrarily as JDBI will use its private version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -85,7 +85,7 @@
         <basepom.test.reuse-vm>true</basepom.test.reuse-vm>
         <basepom.test.timeout>240</basepom.test.timeout>
 
-        <dep.antlr.version>4.9.2</dep.antlr.version>
+        <dep.antlr.version>4.9.3</dep.antlr.version>
         <dep.caffeine.version>3.0.3</dep.caffeine.version>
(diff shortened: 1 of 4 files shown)


stevenschlansker:
Fixes #1798

stevenschlansker (approved):

pom.xml
```diff
@@ -792,6 +793,16 @@
                         </pomElements>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.basepom.maven</groupId>
+                    <artifactId>inline-maven-plugin</artifactId>
```
electrum, line 798: Cool! Looks like the repository is not available? https://github.com/basepom/inline-maven-plugin

hgschmie: oops. It is public now. I kept it private for a while because the code changed a lot.

PR #2015, opened 2022-03-14
Fix markup

maffe:
Because the backtick is inside a word, the rendered output was not correct (see http://jdbi.org/#_immutables).


stevenschlansker (approved):
Thanks, good catch!

PR #2017, opened 2022-03-30
Allow FieldMapper to use inaccessible default constructors

paladin235:
Currently, the FieldMapper requires the default constructor be public in order to instantiate a new instance prior to setting the fields. The FieldMapper already allows private fields to be set via reflection, so it makes sense that it should work with private constructors as well. 

Making a constructor private provides the benefit of preventing its use outside of reflective mapping.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/FieldMapper.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/FieldMapper.java
@@ -13,6 +13,7 @@
  */
 package org.jdbi.v3.core.mapper.reflect;
 
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -201,9 +202,12 @@ public class FieldMapper<T> implements RowMapper<T> {
         return format("%s.%s", type.getSimpleName(), field.getName());
     }
 
+    @SuppressWarnings("PMD.AvoidAccessibilityAlteration")
     private T construct() {
         try {
-            return type.getDeclaredConstructor().newInstance();
+            Constructor<T> ctor = type.getDeclaredConstructor();
+            ctor.setAccessible(true);
+            return ctor.newInstance();
         } catch (ReflectiveOperationException e) {
             throw new IllegalArgumentException(format("A type, %s, was mapped which was not instantiable", type.getName()), e);
         }
--- a/core/src/test/java/org/jdbi/v3/core/SamplePrivateBean.java
+++ b/core/src/test/java/org/jdbi/v3/core/SamplePrivateBean.java
@@ -23,9 +23,7 @@ public class SamplePrivateBean {
     BigDecimal privateBigDecimalField;
     ValueType valueTypeField;
 
-    private SamplePrivateBean() {
-
-    }
+    private SamplePrivateBean() {}
 
     public Long getLongField() {
         return longField;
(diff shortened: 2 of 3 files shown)


hgschmie:
I am pretty opposed to that change because it will become difficult in later java versions. java 9 starts to warn, java 17 deprecates the whole AccessibleObject.

stevenschlansker:
So, this stuff is pretty confusing, but my understanding is that the `isAccessible` method is deprecated but that `setAccessible` isn't actually deprecated. What is deprecated is using it _across modules_ to a module that doesn't explicitly open deep reflection to Jdbi.

So users that move to modulepath would then find that they cannot see private fields / constructors - but classpath would continue to work much the same as before.

I also think that it makes sense to align the behavior - either FieldMapper should work with `private` or it should not. If I were building this class today, I would forbid any private access at all. But it already works with fields, so I think that's a fairly compelling reason to also allow it with constructors.

@hgschmie Do you have a reference on where the `setAccessible` method is intended to be deprecated? I wasn't able to find one.

uschindler:
My perosn opinion: It should be aligned. But FieldMapper should have a config setting to disable access to private APIs at all. It should only call setAccessible when enabled. 
As of JDBI 3.x the default may be enabled, but later versions it should for sure be disabled. This cuased major pain with security critical application where (now deprecated) SecurityManager is used to prevent access to private APIs also in classpath applications.

hgschmie:
I glanced at the docs and it seems only the c'tor is deprecated. The changes to accessibility rules in Java 9+ break private field access for anything anyway (unless you open your module to jdbi). 

I feel we should look to the future and not support private fields and c'tors (it requires runtime setup across modules and will be a source of confusion and error reports. At the very least we need to throw exceptions that clearly state what the problem is (and how to resolve it, which requires handling InaccessibleObjectException) (the current code throws `new IllegalArgumentException(format("A type, %s, was mapped which was not instantiable", type.getName()), e)` which is probably not helpful ("... I upgraded from Java 8 to Java 11 and now JDBI is broken with that error...") 

I feel this is a place where we need to rework the access logic for both c'tor and field to differ for Java 8 and Java 9+ (multi-version code). The current change simply calls "setAccessible". 

On a personal note, I am not a big fan of the test. We have similar code where we actually set up a database table and a real result set mapper (e.g. the bean mapper) and then exercise the real code paths and not drive it out of mocks. I would very much prefer this over the "mock everything somehow" style. If that is needed to test this specific piece of code, it needs to be refactored to be more test friendly.

paladin235:
@hgschmie The tests in this change are a copy of your existing FieldMapper tests, but with a modification to map to a class with a private constructor.

uschindler:
The problem is if JDBI is a module in your app. JDBI as caller can't make a class in your module accessible.

paladin235:
> The problem is if JDBI is a module in your app. JDBI as caller can't make a class in your module accessible.

Yes, I forgot to add the important part that FieldMapper could be opened for extension. I edited my above comment and added that just before your comment. This allows users to create a FieldMapper in their own module that sets accessibility as desired. 

Alternatively, you could provide some interface for users to provide a callback which can set accessibility. This also avoids opening the module to Jdbi.

stevenschlansker:
We'd already talked about making it configurable. Instead of calling `setAccessible`, we could provide an extension point, possibly a `Consumer<AccessibleObject>`. The default could continue to call `setAccessible` as today, with an eye toward changing that default in the future. Other options would include doing nothing (the future default) or providing your own class that calls `setAccessible` from within your own module, instead of from the scope of Jdbi.

How does that sound?

stevenschlansker:
@paladin235 , I have rebased your PR and updated it to have a configurable strategy. See #2193

PR #2019, opened 2022-04-06
Add toString method to SqlStatement that outputs sql and bindings

sman-81:
The addition of this method will greatly help in troubleshooting and error-handling of queries, updates, prepared batches and other subclasses of `SqlStatement`.


stevenschlansker (approved):
Thanks for this improvement!

PR #2022, opened 2022-05-01
Build updates

hgschmie:
- basepom 44 to allow full JDK 17 compatibility
- workarounds for MDEP-791/MDEP-804
- kotlin 1.6.21, detekt 1.20.0, ktlint 1.13.1
- remove formatting plugin for detekt, it is now optional (same as ktlint)


stevenschlansker:
Nice

PR #2034, opened 2022-05-23
DefineStatementLexer: improve backslash escape in quotes handling

### LINKED ISSUE #1906: DefinedAttributeTemplateEngine parsing bug
Passing the following sql query
```
String sql = "SELECT '\\\\' = '\\\\'";
Handle handle = jdbi.open());
       handle.createQuery(sql)

```

In `SqlStatement.java` https://github.com/jdbi/jdbi/blob/34a81e7cdd7ced4156aebf9708d084f68a9b6e04/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java#L1806

`renderedSql` 
- will then have string value of `SELECT '\\' = '\'`  (single backslash on the right side of the equals sign)
-  instead of `SELECT '\\' = '\\'` (double backslash on the right side of the equals sign)

stevenschlansker:
Fixes #1906

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  - Fix DefinedAttributeTemplateEngine lexer bug swallowing single-quoted text with escapes (#1906)
   - ANTLr 4.10.1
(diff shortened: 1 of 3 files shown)


hgschmie (approved):
LGTM.

Issue #2021, opened 2022-04-28
JDBI incorrectly splits statements for Oracle SQL script


galzetta:
Hello, I want to use JDBI to run a plain SQL script for an Oracle DB (i.e. JDBI should just forward the SQL to the server and do no interpretation whatsoever ideally). Unfortunately it seems impossible to achieve this due to how JDBI interprets the SQL.

In particulare `createScript` seems to "split" the SQL into "batches" and execute them separately... but it fails to split the statements correctly. In particular my SQL file contains a TRIGGER definition containing multiple lines of code but JDBI considers the first `;` inside the trigger to be the end of the statement resulting in invalid SQL.

The Oracle script is similar to this:

```sql
CREATE TABLE EXAMPLE
(
    ID            NUMBER PRIMARY KEY,
    USERNAME      VARCHAR2(128) NOT NULL,
    CREATED_AT    TIMESTAMP WITH TIME ZONE NOT NULL
);

CREATE SEQUENCE EXAMPLE_ID_SEQUENCE
    START WITH 1
    INCREMENT BY 1;


CREATE OR REPLACE TRIGGER EXAMPLE_TRIGGER
    BEFORE INSERT
    ON EXAMPLE
    FOR EACH ROW
BEGIN
    SELECT sys_context('USERENV', 'SESSION_USER') INTO :new.USERNAME FROM DUAL; -- JDBI STOPS HERE!
    IF :new.USERNAME = 'something'  THEN
        raise_application_error(-20000, 'Some error');
    END IF;
    SELECT EXAMPLE_ID_SEQUENCE.nextval INTO :new.ID FROM DUAL;
    SELECT CURRENT_TIMESTAMP INTO :new.CREATED_AT FROM DUAL;
    SELECT CURRENT_TIMESTAMP INTO :new.UPDATED_AT FROM DUAL;
END;
```

When executing this using `createScript` what happens is that an `EXAMPLE_TRIGGER` is created but its source code ends after the first line of the body where I wrote `JDBI STOPS HERE` in the codeblock above. It then gives an error because it finds some SQL which is invalid at top level.

Is this a JDBI bug? Shouldn't JDBI understand where the TRIGGER statement ends? If not how can I run the SQL script without JDBI messing around with splitting it in batches?

galzetta:
@stevenschlansker Thank you for the fast reply.

[...]

PR #2051, opened 2022-06-04
Fix#2021 Enhance SqlScriptLexer/SqlScriptParser to correctly deal with BEGIN…END blocks in sql scripts

sman-81:
This PR fixes issue #2021 (JDBI incorrectly splits statements for Oracle SQL script) by enhancing classes `SqlScriptLexer` and `SqlScriptParser` to correctly deal with `BEGIN…END` blocks in sql scripts.

It also adds a JUnit test (`testOracleScriptWithBeginEndBlock`) based on the failing script received from the original bug reporter (amended for better coverage).


sman-81:
Hi there, could one of the maintainers kindly take a look at this PR and approve it?

core/src/main/java/org/jdbi/v3/core/internal/SqlScriptParser.java
```diff
@@ -37,26 +47,37 @@ public String parse(CharStream charStream) {
...
-                case SqlScriptLexer.COMMENT:
-                case SqlScriptLexer.MULTI_LINE_COMMENT:
+                case COMMENT:
+                case MULTI_LINE_COMMENT:
                     break;
-                case SqlScriptLexer.NEWLINES:
+                case NEWLINES:
                     if (sb.length() > 0) {
-                        sb.append(' ');
+                        sb.append(System.lineSeparator()); // normalize newline sequences
```
stevenschlansker, line 75: Hm. This is a bit odd - if you run a Jdbi client program on Windows, suddenly your Postgres server on Linux will see Windows line endings. Different text will be sent to the server depending on the client OS which I find a bit surprising.

Is it harmful to keep the existing `' '`, or we could use a fixed `'\n'`?

core/src/test/java/org/jdbi/v3/core/statement/TestScript.java
```diff
@@ -98,9 +101,44 @@ public void testScriptAsSetOfSeparateStatements() {
...
+
+            assertThat(statements).hasSize(3);
+
+            String lastStmt = statements.get(2);
+            assertThat(lastStmt).startsWith("CREATE OR REPLACE TRIGGER EXAMPLE_TRIGGER");
+            assertThat(lastStmt).endsWith("END;");
+            assertThat(lastStmt).hasLineCount(15);
+            assertThat(lastStmt).has(new Condition<>(s -> 7 == s.chars().filter(ch -> ch == ';').count(), "count semicolons"));
+
+        } catch (SQLException ex) {
```
stevenschlansker, line 136: You could just make the test `throws SQLException` which will automatically fail the test.

core/src/test/resources/script/oracle-with-begin-end-blocks.sql
```diff
@@ -0,0 +1,40 @@
...
+
+CREATE SEQUENCE EXAMPLE_ID_SEQUENCE
+    START WITH 1
+    INCREMENT BY 1;
+
+CREATE OR REPLACE TRIGGER EXAMPLE_TRIGGER
+    BEFORE INSERT
+    ON EXAMPLE
+    FOR EACH ROW
+BEGIN
```
stevenschlansker, line 30: Would it be worth testing that `BEGIN` and `END` are ok to include in comments?
like `-- a comment that includes BEGIN`

stevenschlansker:
Sorry for the delay @sman-81 , I was on vacation 🚢 
I think the test failure is a false positive. I left some comments for consideration.

core/src/test/java/org/jdbi/v3/core/statement/TestScript.java
sman-81: Ok, will do

core/src/main/java/org/jdbi/v3/core/internal/SqlScriptParser.java
sman-81: Database servers are newline-agnostic. They must be, dealing with clients of all sorts and os.
I didn't like the super-long lines the original code creates. But sure, I can roll this change back.

core/src/test/resources/script/oracle-with-begin-end-blocks.sql
sman-81: I've amended the test but please be aware I am anything but an Antlr expert ;)

sman-81:
> Sorry for the delay @sman-81 , I was on vacation

Thanks for the feedback @stevenschlansker and welcome back from a well-deserved vacation ;)
you know PR authors can be eager beavers 😂
I have incorporated your recommendations and force-pushed the update.

stevenschlansker (approved):

Issue #2050, opened 2022-06-02
Build requires Docker but doesn't support Podman


stevenschlansker:
The pg-embedded component now requires Docker.
I have Podman on my machine, but it isn't picked up:

```
    [ERROR] Errors: 
    [ERROR]   JdbiRulePostgresTest.isAlive » IllegalState Could not find a valid Docker environment. Please see logs and check configuration
```

The runner should consider Podman as a viable alternate to Docker, at least.

stevenschlansker:
I tried linking `docker-machine` to run `podman machine "$@"`, but that doesn't work, because `podman machine ls` doesn't accept a `-q` option the runner uses.

PR #2052, opened 2022-06-04
Do not require docker to build the code

hgschmie:
The OTJ Postgres Plugin has switched to testcontainers which in turn
requires a local docker environment to build the code base.

Add a check for docker availability for all OTJ related tests.


hgschmie:
@stevenschlansker Please review. This should fix #2050 by turning off docker.

However, testcontainers *should* support podman, can you check https://stackoverflow.com/questions/71549856/testcontainers-with-podman-in-java-tests to see whether it is possible for you to activate the docker socket. I looked at the testcontainer code and it should support it. 

Plz comment here.

hgschmie:
abandoned in favor of #2054

PR #2056, opened 2022-06-11
Enhancements to ResultIterable

sman-81:
- Add method `ResultIterable#filter` to apply a filter (predicate) to a result iterable without use of streams.
  Similarities with `ResultIterable#map` extracted into class `ResultIteratorDelegate`
- Add enhanced `ResultIterable#forEachWithCount`, an enhanced forEach method that returns the iteration/record count
- Add unit tests


sman-81:
Hi @stevenschlansker,

thanks for your feedback!

We avoid switching to `Stream` for 'intermediate' operations and stay within the `ResultIterable` type to perform custom, value-added 'terminal' operations such as `forEachWithCount`.
Frequently working with resultsets and Jdbi one will do `forEach` to iterate over the resultset (doing this does not use Stream API).
Along the same lines, the `map`, `findOne`, `first`, `findFirst` methods, which all have existed for a long time, do not use Stream API. My new `filter` method works along those lines.

> In particular, we avoid doing filter in Java code, since the predicate cannot be lifted up to the database.

What do you mean?
`filter` could be done in most if not all dbms using WHERE.
Similarly `map` could be done inside db using built-in functions.

Yet, having both in JDBI serves a very good purpose, as it allows you to keep a db simple and shift more complex processing to Java.

sman-81:
Hi @stevenschlansker,

Hope this finds you well :)
I am rather excited about the positive impact of this PR.
Could you kindly get back to me on my above comments?

Cheers, Markus

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
```diff
@@ -87,39 +90,48 @@ static <T> ResultIterable<T> of(ResultIterator<T> iterator) {
...
+     * It is often useful (e.g. for logging) to know the record count while processing result sets.
+     * <pre>
+     * {@code
+         int cnt = h.createQuery("select * from something").mapTo(String.class)
+                    .forEachWithCount(System.out::println);
+         System.out.println(cnt + " records selected");
+       }
+     *  </pre>
+     *
+     * @param action action top apply (required)
```
stevenschlansker, line 121: `top` -> `to`

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
```diff
@@ -303,4 +315,103 @@ default <U> U reduce(U identity, BiFunction<U, T, U> accumulator) {
                 });
         }
     }
+
+    /**
+     * Filters the {@link ResultIterable} by applying the specified predicate.<br>
+     * This method has the look and feel of {@link Stream#filter(Predicate)} without making use of streams.
+     *
+     * @param predicate a non-interfering, non-null predicate to apply to each element
```
stevenschlansker, line 323: I like that you are referring to Stream documentation to help guide this, but I think the term `non-interfering` might not be quite accurate here. Since the data source is from the database, it explicitly handles concurrency (for example, Postgres would continue to provide the same snapshot data from the start of the transaction regardless of later interference).

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
```diff
@@ -303,4 +315,103 @@ default <U> U reduce(U identity, BiFunction<U, T, U> accumulator) {
...
+
+            /**
+             * Forwards the resultset to the next result that passes the filter (i.e. {@link Predicate} tests {@code true}).<br>
+             * @return true if another such result exists, false otherwise
+             */
+            private boolean findNext() {
+                synchronized (this) {
+                    next = null;
+                    while (getDelegate().hasNext()) {
+                        T n = getDelegate().next();
```
stevenschlansker, line 368: It looks like calling:

```
filtered = iter.filter(...);
filtered.hasNext();
filtered.hasNext();
```
would incorrectly consume an element of the filtered results. I think there should be tests to ensure correct behavior of `hasNext` even when called repeatedly.

Maybe `hasNext` needs to check if `next` already is non-null.

core/src/test/java/org/jdbi/v3/core/statement/TestQueries.java
```diff
@@ -468,4 +472,58 @@ public void testBindPrimitiveWithExplicitNull() {
...
+
+        ResultIterator<String> iter = ri2.iterator();
+        while (iter.hasNext()) {
+            results.add(iter.next());
+        }
+
+        assertThatThrownBy(iter::next).isInstanceOf(NoSuchElementException.class)
+                .hasMessage("No more filtered results");
+
+        assertThat(results).startsWith("john")
```
stevenschlansker, line 525: `assertThat(results).containsExactly("john")` is a more concise way of saying the same thing, I think.

stevenschlansker:
And sorry for the delay @sman-81 , I have been on vacation 🌴

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
sman-81: good points, Javadoc updated to clarify

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
```diff
@@ -303,4 +315,103 @@ default <U> U reduce(U identity, BiFunction<U, T, U> accumulator) {
...
+
+    /**
+     * Filters the {@link ResultIterable} by applying the specified predicate.<br>
+     * This method has the look and feel of {@link Stream#filter(Predicate)} without making use of streams.
+     *
+     * @param predicate a non-interfering, non-null predicate to apply to each element
+     *                  to determine whether it should be included in the result
+     * @return the new result iterable
+     *
+     * @since 3.30.1
```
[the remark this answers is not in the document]
sman-81: I've amended the version number. But would it better to remove it altogether?

core/src/test/java/org/jdbi/v3/core/statement/TestQueries.java
sman-81: very true 🤭

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
sman-81: well spotted

core/src/test/java/org/jdbi/v3/core/statement/TestQueries.java
```diff
@@ -468,4 +472,58 @@ public void testBindPrimitiveWithExplicitNull() {
...
+
+        ResultIterable<String> ri1 = h.createQuery("select name from something").mapTo(String.class);
+
+        assertThatThrownBy(() -> ri1.filter(null)).isInstanceOf(NullPointerException.class)
+                .hasMessage("Filter required");
+
+        ResultIterable<String> ri2 = ri1
+                .filter(Objects::nonNull)
+                .filter(s -> s.startsWith("j"))
+                .filter(s -> s.contains("h"));
```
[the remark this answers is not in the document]
sman-81: Unit test expanded

sman-81:
Your feedback incorporated and pr updated.

core/src/main/java/org/jdbi/v3/core/result/ResultIterable.java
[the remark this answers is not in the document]
stevenschlansker: I don't mind it, there is some value to knowing what version of Jdbi a feature dates to, but we don't typically include it.

stevenschlansker (approved):

PR #2057, opened 2022-06-15
Prepared statements: Support bound parameters of type CharSequence

sman-81:
This small pull request adds support for bound parameters of type `CharSequence` in prepared statements.
Such parameters are to be treated same as `String`.


sman-81:
good morning @stevenschlansker 
I'll look into adding a test case to this PR.

sman-81:
Hi @stevenschlansker,
I found a more versatile solution for `CharSequence` support in parameters and added tests.
`EssentialsArgumentFactory` only works on the explicit type. A (low-priority) factory that handles parameters that implement `CharSequence` and that are not handled by a more specific or user-defined factory is what we would want.
I have squashed and force pushed the update.
Please take a look - thanks.

core/src/main/java/org/jdbi/v3/core/argument/CharSequenceArgumentFactory.java
```diff
@@ -0,0 +1,49 @@
...
+public class CharSequenceArgumentFactory extends AbstractArgumentFactory<CharSequence> {
+
+    public CharSequenceArgumentFactory() {
+        super(Types.VARCHAR);
+    }
+
+    @Override
+    protected Argument build(CharSequence value, ConfigRegistry config) {
+        return (position, statement, ctx) -> {
+            StatementBinder<String> binder = PreparedStatement::setString;
```
stevenschlansker, line 44: I'm not sure why the `binder` here is helpful. Wouldn't it be simpler to just call `statement.setString(position, Objects.toString(value, null));`?

sman-81: You're right! Fixed.

PR #2068, opened 2022-07-18
Fix issue 2065

### LINKED ISSUE #2065: JDBI 3 - Handle.close() throws Failed to clear transaction status on close error
We use jdbi api in our drop wizard based application which internally uses tomcat connection pool for db operations.
Below code snippet is used to delete user from our oracle database.

```
jdbi.useHandle(
 handle ->
 handle.createUpdate("DROP USER XXXXX CASCADE")
 .execute());
```

The above code throws an error -

```
Caused by: org.jdbi.v3.core.CloseException: Failed to clear transaction status on close
at org.jdbi.v3.core.Handle.close(Handle.java:183)
at org.jdbi.v3.core.Jdbi.withHandle(Jdbi.java:358)
at org.jdbi.v3.core.Jdbi.useHandle(Jdbi.java:373)
at com.test.db.DBSchemaProviderImpl.deleteDBSchema(DBSchemaProviderImpl.java:57)
... 7 common frames omitted
Caused by: org.jdbi.v3.core.transaction.TransactionException: Failed to test for transaction status
at org.jdbi.v3.core.transaction.LocalTransactionHandler$BoundLocalTransactionHandler.isInTransaction(LocalTransactionHandler.java:194)
at org.jdbi.v3.core.Handle.isInTransaction(Handle.java:346)
at org.jdbi.v3.core.Handle.close(Handle.java:154)
... 10 common frames omitted
Caused by: java.sql.SQLException: Connection has already been closed.
at org.apache.tomcat.jdbc.pool.ProxyConnection.invoke(ProxyConnection.java:121)
at org.apache.tomcat.jdbc.pool.JdbcInterceptor.invoke(JdbcInterceptor.java:109)
at org.apache.tomcat.jdbc.pool.interceptor.AbstractCreateStatementInterceptor.invoke(AbstractCreateStatementInterceptor.java:79)
at org.apache.tomcat.jdbc.pool.JdbcInterceptor.invoke(JdbcInterceptor.java:109)
at org.apache.tomcat.jdbc.pool.DisposableConnectionFacade.invoke(DisposableConnectionFacade.java:81)
at com.sun.proxy.$Proxy110.getAutoCommit(Unknown Source)
at org.jdbi.v3.core.transaction.LocalTransactionHandler$BoundLocalTransactionHandler.isInTransaction(LocalTransactionHandler.java:192)
... 12 common frames omitted
```

Please suggest a way of working with this code. Thanks

hgschmie:
Check connection status in handle.close()

Ensure that the handle will tolerate a closed connection when close()
is called. If the database connection gets closed between the last
operation finishing successfully and the handle being closed, checking
the autocommit status will cause an exception to be thrown that is
user-visible but not actionable. Check the connection state before
executing any operation on the database connection.

Fixes #2065

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -148,14 +148,14 @@ public class Handle implements Closeable, Configurable<Handle> {
             return;
         }
 
-        boolean connectionIsLive = false;
+        boolean connectionIsLive;
 
         try {
             connectionIsLive = !connection.isClosed();
         } catch (SQLException e) {
             // if the connection state can not be determined, assume that the
-            // connection is closed
-            suppressed.add(e);
+            // connection is closed and ignore the exception
+            connectionIsLive = false;
         }
 
         boolean wasInTransaction = false;
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -148,8 +148,19 @@ public class Handle implements Closeable, Configurable<Handle> {
             return;
         }
 
+        boolean connectionIsLive = false;
+
+        try {
+            connectionIsLive = !connection.isClosed();
+        } catch (SQLException e) {
+            // if the connection state can not be determined, assume that the
+            // connection is closed
+            suppressed.add(e);
+        }
+
         boolean wasInTransaction = false;
-        if (forceEndTransactions && localConfig.get().get(Handles.class).isForceEndTransactions()) {
+
+        if (connectionIsLive && forceEndTransactions && localConfig.get().get(Handles.class).isForceEndTransactions()) {
             try {
                 wasInTransaction = isInTransaction();
             } catch (Exception e) {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -148,8 +148,19 @@ public void close() {
...
         }
 
+        boolean connectionIsLive = false;
+
+        try {
+            connectionIsLive = !connection.isClosed();
+        } catch (SQLException e) {
+            // if the connection state can not be determined, assume that the
+            // connection is closed
+            suppressed.add(e);
```
stevenschlansker, line 158: If we add a thrown exception to `suppressed` here, won't it get thrown later on after we call `closer.close`? Then it would be thrown anyway when it seemed like the intent is that `close()` shouldn't throw in that case.

hgschmie: that is a good point. I guess we should just drop the exception then. It will be a "connection closed" or "IO exception" anyway.

PR #2070, opened 2022-07-22
Replace lambda class with inner class to support GraalVM native image

Eng-Fouad:


Eng-Fouad:
Fix #2071

Eng-Fouad:
Closes #2071

hgschmie:
Hi @Eng-Fouad, 

Thank you for trying out JDBI on GraalVM and submitting patches to the project! We really appreciate it. 

I have to admit that I am a bit conflicted with this PR. I am very excited about GraalVM and having JDBI run on it. On the other hand, lambdas (and in this case method handles) are such a fundamental part of Java since Java 8 that I feel replacing or doing workarounds for the sake of supporting GraalVM is questionable. I wish there were a good timeline for GraalVM when they support this (the stack overflow post seems to imply it) so that we could say to hold off until a version of GraalVM that supports lambdas has been released. 

I feel this warrants more discussion before we apply or reject this PR. Others should chime in here as well.

Eng-Fouad:
Thank you @hgschmie for looking into this PR. Regarding GraalVM timeline to support lambda classes, I couldn't find any related tasks in their [project page](https://github.com/oracle/graal/projects/2). I opened a new issue for that: oracle/graal#4753 

> lambdas (and in this case method handles) are such a fundamental part of Java since Java 8 that I feel replacing or doing workarounds for the sake of supporting GraalVM is questionable

Lambda can be used normally in GraalVM native image. It's just the reflection part where GraalVM does not support it yet. I believe the proposed solution doesn't really change anything in the code other than giving stable name to the class that will be reflectively accessed by Caffeine library.

Eng-Fouad:
My bad, I found the related task: oracle/graal#2990

hgschmie:
it seems that they keep moving this from milestone to milestone. I agree that having a stable name would solve those problems. 

@stevenschlansker @qualidafial  opinions?

stevenschlansker:
I agree that Graal needs to fully support Java without needing to make changes to all the various ecosystem libraries.
However, if this is the only case in Jdbi where it breaks, and accepting this patch allows it to work, I think we could make a compromise here and accept the patch.

Eng-Fouad:
Before accepting this PR, let me double check and test it with Quarkus native image. Thanks.

Eng-Fouad:
I had to explicitly add the method `loadAll()` to the new class to make it work in native image:

    public final class SqlParserCacheLoader implements CacheLoader<String, ParsedSql> {
        @Override
        public ParsedSql load(String key) {
            return internalParse(key);
        }

        @Override
        public Map<String, ParsedSql> loadAll(Iterable<? extends String> keys) throws Exception {
            return CacheLoader.super.loadAll(keys);
        }
    }

Eng-Fouad:
It turns out that no modifications are needed in JDBI. The solution is to include the default method `CacheLoader.loadAll` into `reflection-config.json`:

    {"name":"com.github.benmanes.caffeine.cache.CacheLoader","methods":[{"name":"loadAll","parameterTypes":["java.lang.Iterable"]}]}

Sorry for the mistake. I am closing this PR.

stevenschlansker:
I am glad it ends up working for you with no changes! Let us know how your Graal-Native + Jdbi experience is, I think others are excited to use it as well

PR #2072, opened 2022-07-22
Add postgis plugin with support for JTS data types

bchapuis:
This PR adds support for postgis and map geospatial data types to JTS data types.

As this is my first contribution to JDBI, I would love to get feedbacks:
- Does the plugin fits in the postgres module or should I create a new postgis module?
- Should the factory and mapper class be moved at the level of the package?
- How to test postgis without testcontainers as the setup is slighly different from postgres?
- Anything else that raises eyebrows.

Thanks a lot for maintaining JDBI.

### DIFF
--- a/internal/build/pom.xml
+++ b/internal/build/pom.xml
@@ -58,6 +58,7 @@
         <dep.jdbi-policy.version>3.31.1-SNAPSHOT</dep.jdbi-policy.version>
         <dep.jetbrainsAnnotations.version>21.0.1</dep.jetbrainsAnnotations.version>
         <dep.joda-time.version>2.9.9</dep.joda-time.version>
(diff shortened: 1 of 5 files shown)


hgschmie:
Hi @bchapuis  !

Thank you for contributing to JDBI! Your code is generally sound but you did a lot of work that was not necessary. :-) 

I cleaned up the code in #2074 (and also made the test work and pass with our normal test framework). Please review and let me know if that works for your use case.

bchapuis:
Thanks a lot for your feedback, it simplifies things a lot! Do you mind if we merge this PR? I cherry picked your change, added you as a co-author and force pushed. I'd prefer to appear in the commit history than in the git comment. ;)

hgschmie:
merged as #2074

Issue #2078, opened 2022-07-26
PreparedBatch with first bindNull() and then bind() afterwards fails


doppelrittberger:
Hi,
I think my issue is related to https://github.com/jdbi/jdbi/issues/1901
but I have a different call order which I cannot change. In my case I first use bindNull() on PreparedBatch and add another binding with bind() after it. You can simply check that this is failing if you change TestPreparedBatch:273 and move this line up:
```java
    @Test
    public void testBindNull() {
        Handle handle = h2Extension.getSharedHandle();
        handle.execute("CREATE TABLE record (b bool)");
        assertThat(handle.prepareBatch("INSERT INTO record (b) VALUES (:bVal)")
            .bindNull("bVal", Types.BOOLEAN).add()
            .bind("bVal", false).add()
            .bindByType("bVal", null, boolean.class).add()
            .execute()).containsExactly(1, 1, 1);
    }
```
This will cause the test to fail.
If there is anything I can help you with please let me know.

Best regards
Markus Ritter

hgschmie:
This is related to #1968 and #1991

hgschmie:
Closing, still tracked in https://github.com/jdbi/jdbi/issues/2390.

Issue #1968, opened 2021-12-14
UnableToCreateStatementException when using @SqlBatch with binding of a nullable enum values


levyshachar2:
Hi everyone.
We have an entity class that has multiple fields, some of the fields are Nullable enums (we use Kotlin)
So for example in our data class, we have `val fooType: FooType?`
Which is the following class
`enum class FooType() {
  OTHER,
  SOMETHING_ELSE,
}`
 
When using @SqlBatch to update multiple rows we are getting the following exception:
`org.jdbi.v3.core.statement.UnableToCreateStatementException: No argument factory registered for 'OTHER' of qualified type java.lang.Object`

this is the code we are doing for the bind
val fooType = entity.fooType
`q.bind("fooType", fooType)`

Several things I noticed:

1. This only happens when using the @SqlBatch annotation - when using @SqlUpdate in a loop for the same values we don't encounter this issue
2. This only happens if the first value of the batch values has a null value in fooType.
So for example, If I have two rows in the DB that I want to update, 
                 PK, foo_type
First row: id1, null
Second row: id2, OTHER

This will fail, however, if the first value will be OTHER it will work.

3. If I'll try in the binding `q.bind("fooType", fooType.name)` this will work

Some technical information:
We are using Kotlin, JDBI3 version 3.20.0 (part of dropwizard), mySQL 

let me know if you need anymore information

hgschmie:
This sounds a bit like #1944 related. Looks similar (can not find a factory for the enum). 

can you 
- try 3.24.0 or later
- make sure that you install the KotlinPlugin in Jdbi

ideal would be to have an isolated test case that shows the problem, if you could whip up some test code, that would be great.

hgschmie:
Actually, it is not related to that bug. I have another test that fails similarly:

```java
@Test
public void testNonUniformBatchInserts() throws Exception {
    int batchCount = 50;

     try (Handle h = h2Extension.openHandle()) {
        PreparedBatch batch = h.prepareBatch("INSERT INTO something (id, name) VALUES(:id, :name)");
        for (int i = 1; i <= batchCount; i++) {
            Map<String, Object> data;
            if (i % 2 == 0) {
                data = ImmutableMap.of("id", i, "name", "User:" + i);
            } else {
                data = ImmutableMap.of("id", i, "name", i);
            }

            batch.bindMap(data).add();
        }

        int[] counts = batch.execute();

        assertEquals(batchCount, counts.length);
[...]

PR #2079, opened 2022-07-26
Possible fix for https://github.com/jdbi/jdbi/issues/2078

doppelrittberger:
Hi, 
I created a possible fix for https://github.com/jdbi/jdbi/issues/2078 and added more test cases.
Feel free to reach out to me if there is something missing.
Best regards
Markus Ritter

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/argument/DelegatingArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/DelegatingArgumentFactory.java
@@ -35,11 +35,12 @@ abstract class DelegatingArgumentFactory implements ArgumentFactory.Preparable {
 
     @Override
     public Optional<Argument> build(Type expectedType, Object value, ConfigRegistry config) {
+        Class<?> expectedClass;
         if (expectedType == NullArgument.class && value != null) {
-            expectedType = value.getClass();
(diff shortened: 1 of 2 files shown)


hgschmie:
just making sure that this is not slipping through the cracks: This is the same issue as #1968 and there are tests in #1991 for this. The problem is unfortunately not trivial to fix because JDBI allows batches to have non-uniform rows which wreaks havoc with cached argument binders. But not caching the argument binders will kill all the performance gains that the caching gives us.

This fix, while it address the "null, then non-null" use cases for some columns, is not a general fix.

stevenschlansker:
Yes, I think this is currently the most important problem in Jdbi to fix. Unfortunately I haven't seen a solution I love yet.
I have considered whether as a stop-gap we might include a configuration option to select between "fast and strict" (no flexibility on argument type) and "slow but forgiving" (the classic behavior) so users may select the choice appropriate for their use.

Of course, finding the general solution would be much preferred. I am on vacation for two weeks though (Hello from Switzerland!) and will consider it more when I return :)

doppelrittberger:
Hi
Thanks for your effort. At least the fix I made works for our usecases 😁 Please let me know if I can support you. 
Best regards

hgschmie:
Closing, still tracked in #2390.

PR #2104, opened 2022-09-12
use root locale for uppercase/lowercase

hgschmie:


stevenschlansker:
Should we make the Locale configurable? I remember that we have at least one user who relies on French locale working (we had a regression before).

hgschmie:
locale for things like column name lower/uppercasing? This is interesting. Right now, we have a mix of "Locale.ROOT" and not using a locale. The only place that I can see where this is deliberate is the CaseStrategy and I am not sure how consistent we are using this.

stevenschlansker (approved):
Yeah, being consistent with ROOT is definitely better than inconsistent. I am not sure configurability is desirable, I just thought I'd mention it.

PR #2121, opened 2022-09-30
Some improvements to the Sql wrapper

hgschmie:
- #of(CharSequence ...) can never be null, drop the null check
- #of(Collection<CharSequence>) should be of(Iterable<? extends CharSequence>)

These changes are backwards incompatible; this is ok as Sql is still marked as @Beta.


stevenschlansker (approved):

core/src/main/java/org/jdbi/v3/core/Sql.java
```diff
@@ -61,29 +65,27 @@ private Sql(String sql) {
...
+    public static Sql of(Iterable<? extends CharSequence> tokens) {
+        return tokens == null ? EMPTY_SQL : new Sql(format(tokens));
     }
 
     /**
-     * Formats an sql statement from multiple tokens.<br>
-     * Leading and trailing whitespace is removed from each token and empty tokens ignored.<br>
-     * The tokens are joined using a single blank character to create the sql string.<br>
-     * Finally, any trailing semicolons are removed from the resulting sql.
+     * Formats an sql statement from multiple tokens.<br> Leading and trailing whitespace is removed from each token and empty tokens ignored.<br> The tokens
```
sman-81, line 76: I personally find the original formatting easier to read.
Was the Javadoc perhaps auto-formatted?

core/src/test/java/org/jdbi/v3/core/SqlTest.java
```diff
@@ -29,13 +30,20 @@ public class SqlTest {
...
+        assertEquals("", Sql.of((CharSequence) null).toString());
         assertEquals("", Sql.of((Collection<CharSequence>) null).toString());
         assertEquals("", Sql.of(new CharSequence[0]).toString());
         assertEquals("", Sql.of(new ArrayList<>()).toString());
         assertEquals("", Sql.of().toString());
     }
 
+    @Test
+    void testMultipleStrings() {
+        assertEquals("A B C", Sql.of(ImmutableList.of("A", "B", "C")).toString());
```
sman-81, line 42: I would recommend the JDK's `List.of` which is fully immutable (including set operations). This would prevent the usage of Guava in this test class.

sman-81 (changes requested):
@hgschmie reviewed as requested ;) Two small comments. Please take a look

core/src/main/java/org/jdbi/v3/core/Sql.java
hgschmie: Yes; IntelliJ. I can revert that.

core/src/test/java/org/jdbi/v3/core/SqlTest.java
hgschmie: Correct. List.of requires JDK9. ;-)

PR #2123, opened 2022-10-01
Remove Closeable from BaseStatement

hgschmie:
Interestingly enough, there are only very few changes required to make BaseStatement not implement Closable. And the places are few and all internal.

As we have never actually published that the various SqlStatement things are closable (and basically no one ever should have to deal with them directly anyway), it should be possible to make this change without causing too much scare amongst JDBI users. Or not?

Discussion wanted!

@hjohn would that change help your evaluation? I can follow your arguments and I am not actually sure why BaseStatement extends Closable. I don't think it is really needed. :-)

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
@@ -122,7 +122,7 @@ public class PreparedBatch extends SqlStatement<PreparedBatch> implements Result
         try {
             return internalBatchExecute().updateCounts;
         } finally {
-            getContext().close();
+            PreparedBatch.nullSafeCleanUp(this);
         }
     }
 
@@ -206,11 +206,7 @@ public class PreparedBatch extends SqlStatement<PreparedBatch> implements Result
                 return batch.stmt;
             }, getContext());
         } catch (SQLException e) {
-            try {
-                close();
-            } catch (Exception e1) {
-                e.addSuppressed(e1);
-            }
+            cleanUpForException(e);
             throw new UnableToProduceResultException("Exception producing batch result", e, getContext());
         }
     }
--- a/core/src/main/java/org/jdbi/v3/core/statement/BaseStatement.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/BaseStatement.java
@@ -17,6 +17,7 @@ import java.io.Closeable;
 import java.sql.SQLException;
 import java.util.Collection;
 
+import org.jdbi.v3.core.CloseException;
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.config.Configurable;
@@ -51,6 +52,22 @@ abstract class BaseStatement<This> implements Closeable, Configurable<This> {
         return ctx;
     }
 
+    public static void nullSafeCleanUp(BaseStatement<?> statement) {
+        if (statement != null) {
+            statement.getContext().close();
+        }
+    }
+
+    protected final void cleanUpForException(SQLException e) {
+        try {
+            nullSafeCleanUp(this);
+        } catch (CloseException ce) {
+            e.addSuppressed(ce.getCause());
+        } catch (Exception e1) {
+            e.addSuppressed(e1);
+        }
+    }
+
     /**
      * Registers the given {@link Cleanable} to be executed when this statement is closed.
      *
@@ -82,7 +99,7 @@ abstract class BaseStatement<This> implements Closeable, Configurable<This> {
 
     @Override
     public void close() {
-        getContext().close();
+        nullSafeCleanUp(this);
     }
 
     @FunctionalInterface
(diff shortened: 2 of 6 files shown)


hjohn:
The change indeed looks far simpler than I could have hoped, perhaps `Closeable` was only used because it was a convenient interface with a `close` method. As for my evaluation (if that really matters), I would definitely reconsider as I think JDBI is still exactly what I've been looking for.

core/src/main/java/org/jdbi/v3/core/statement/Cleanable.java
```diff
@@ -27,4 +27,23 @@
 public interface Cleanable extends AutoCloseable {
     @Override
     void close() throws SQLException;
+
+    /*
+     * TODO - This is *NOT* the final location for these methods. But I needed something that is public and somewhat related to cleaning up.
```
hjohn, line 32: You could consider making a class `Cleanables` with the helper methods.

hjohn, line 32: You could consider adding a class `Cleanables` with these helpers.

core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
```diff
@@ -204,11 +204,7 @@ private <R> R execute(ResultProducer<R> producer, Consumer<ExecutedBatch> batchC
                 return batch.stmt;
             }, getContext());
         } catch (SQLException e) {
-            try {
-                close();
-            } catch (Exception e1) {
-                e.addSuppressed(e1);
-            }
+            Cleanable.cleanUpForException(e, this);
```
hjohn, line 207: I noticed this works slightly different.  Originally it added the exception thrown by `close` as surpressed, the new code (in `cleanUpForException` adds the cause of the exception (it may not have any!).

core/src/main/java/org/jdbi/v3/core/statement/Cleanable.java
```diff
@@ -27,4 +27,23 @@
...
+            statement.getContext().close();
+        }
+    }
+
+    static void cleanUpForException(SQLException e, BaseStatement<?> statement) {
+
+        try {
+            cleanUp(statement);
+        } catch (Exception e1) {
+            e.addSuppressed(e1.getCause());
```
hjohn, line 46: This may not have a cause, in which case it will throw a NPE as adding a `null` surpressed exception is not allowed.

hgschmie: good catch.

hgschmie: I was noodling around with catching just the CloseException (which comes out of the StatementContext close) but then changed it back to the original "catch Exception" but did not change the getCause back. I now treat these exceptions differently (CloseException gets unwound, everything else is added "as is").

hgschmie:
> The change indeed looks far simpler than I could have hoped, perhaps `Closeable` was only used because it was a convenient interface with a `close` method. As for my evaluation (if that really matters), I would definitely reconsider as I think JDBI is still exactly what I've been looking for.

I think it is unfortunate that there are some checkers that now consider anything that implements "Closeable" as sacrosanct and requiring try-with-resources. Actually, this is what AutoCloseable is about. I believe we used that as a convenient marker interface for a close() method. Unwinding this is not hard, the question for us is more whether we can do this (and break backwards compatibility with something that most likely no one is actually using). Until someone shows up with a use-case.

hjohn:
> > The change indeed looks far simpler than I could have hoped, perhaps `Closeable` was only used because it was a convenient interface with a `close` method. As for my evaluation (if that really matters), I would definitely reconsider as I think JDBI is still exactly what I've been looking for.
> 
> I think it is unfortunate that there are some checkers that now consider anything that implements "Closeable" as sacrosanct and requiring try-with-resources. Actually, this is what AutoCloseable is about. I believe we used that as a convenient marker interface for a close() method. Unwinding this is not hard, the question for us is more whether we can do this (and break backwards compatibility with something that most likely no one is actually using). Until someone shows up with a use-case.

Well, it is hard to blame the checkers. `Closeable` (introduces in Java 1.5) was retrofitted in Java 1.7 to extend `AutoCloseable`, so in effect it is now an `AutoCloseable`.  The only difference between the two now is that one throws an `IOException` while the other throws the generic exception.

And you're right, it is not a strictly backwards compatible change, if someone is currently using try with resources on the `Update` returned (it wouldn't compile anymore as `Update` isn't `AutoCloseable` anymore). It doesn't make sense to do it, but it was possible.

     try (Update u = jdbi.createUpdate("<sql>")) {  // <-- compile error here
          u.execute();
     }

hgschmie:
agreed but before 1.7, "Closeable" implied "has a close() method". Checkers started to aggressively change that semantic to "needs to be closed".

hgschmie:
I have changed the scope of the commit for now to just clean up the code but leave the interface in place. I am planning to write some docs and then we can revisit whether we want to remove the interface or not. I am very much leaning towards removing it once we have documented what we expect users to do.

PR #2126, opened 2022-10-04
Factor lombok code into its own module

hgschmie:
It seems that using lombok crashes the compiler in some IDEs. Factor the lombok code into a separate module so that core still compiles in such IDEs.

The lombok module is tests-only, so it does not ship any artifacts.


hgschmie:
not super-excited but if it makes Eclipse happy... Or maybe use IntelliJ? :-)

stevenschlansker (approved):
I did end up getting Eclipse to work OK with Lombok. It does feel like it's special enough that it could be a separate module. Up to you whether you want to merge this or not.

PR #2129, opened 2022-10-04
Replace `@Unmappable` with `@JdbiProperty` and support skipping binding arguments

stevenschlansker:
This allows you to skip binding, same as you can skip mapping, on reflectively discovered properties.
In our use case, we have a helper method that returns a non-bindable type that we just want Jdbi to ignore.

I changed the meaning from the negative (unmappable) to the positive, otherwise you end up with double negatives (unmappable = false)

A `@JdbiIgnore` would be nice to have, but is technically redundant. WDYT?


core/src/main/java/org/jdbi/v3/core/annotation/JdbiProperty.java
```diff
@@ -21,16 +21,22 @@
...
-     * Returns true if the property is unmappable. It will be skipped when read from the database into a pojo.
+     * Returns true if the property is mapped in a result. It will be read from the database into a pojo.
      *
-     * @return True if the property is unmappable. It will be skipped when read from the database into a pojo.
+     * @return true if the property is unmappable
      */
-    boolean value() default true;
+    boolean map() default true;
+    /**
+     * Returns true if the property is bound as an argument. Pojo property will be bound as an argument.
```
hgschmie, line 37: it also applies to fields (for kotlin and the field mapper). Maybe have a more neutral explanation?

core/src/main/java/org/jdbi/v3/core/annotation/internal/JdbiAnnotations.java
```diff
@@ -0,0 +1,52 @@
...
+package org.jdbi.v3.core.annotation.internal;
+
+import java.lang.reflect.AnnotatedElement;
+import java.util.Optional;
+
+import org.jdbi.v3.core.annotation.JdbiProperty;
+import org.jdbi.v3.core.mapper.reflect.internal.PojoProperties.PojoProperty;
+
+public final class JdbiAnnotations {
+    private JdbiAnnotations() {}
```
hgschmie, line 23: throws a utility class exception.

core/src/main/java/org/jdbi/v3/core/annotation/internal/JdbiAnnotations.java
```diff
@@ -0,0 +1,52 @@
...
+
+    private static boolean isBound(Optional<JdbiProperty> anno) {
+        return anno.map(JdbiProperty::bind).orElse(true);
+    }
+
+    private static boolean isMapped(Optional<JdbiProperty> anno) {
+        return anno.map(JdbiProperty::map).orElse(true);
+    }
+
+    private static Optional<JdbiProperty> jdbiProperty(AnnotatedElement elem) {
```
hgschmie, line 49: not really a fan of "anno" and "elem". maybe use `jdbiProperty` and `annotatedElement` instead?

core/src/main/java/org/jdbi/v3/core/argument/internal/PojoPropertyArguments.java
```diff
@@ -45,18 +49,26 @@ protected PojoPropertyArguments(String prefix, Object obj, PojoProperties<?> pro
         super(prefix, obj);
         this.properties = properties;
         this.config = config;
+        names = new HashSet<>();
```
hgschmie, line 52: can we do this as an expression? names = `properties.getProperties().stream().filter.collect(...)` ?

docs/src/adoc/index.adoc
```diff
@@ -1211,7 +1211,7 @@ Reflection mappers support the following annotations. Unless otherwise noted, al
 * `@PropagateNull` on a given attribute propagates a null value for a specific attribute up. So if the attribute or column is null, the whole bean will be discarded and a null value will be used instead of the bean itself. This annotation can also be used on the bean
   class with a column name that must be present in the result set. When used on an attribute, no value must be set.
 * `@Nullable` for `ConstructorMapper` arguments. If no column is mapped onto a specific constructor argument, don't fail but use `null` as value. Any annotation called `Nullable` is accepted.
-* `@Unmappable` for `BeanMapper` bean properties and `FieldMapper` fields. Any attribute with this annotation will be ignored.
+* `@JdbiProperty(map=false, bind=false)` for `BeanMapper` bean properties and `FieldMapper` fields. Any attribute with this annotation will be ignored for binding and mapping respectively.
```
hgschmie, line 1214: this is a very terse explanation and it is explained in context of "Bean mapping".  This should focus exclusively on this (not talk about `bind=true|false`). 

Unless someone knows where to look, no one is going to find this if they look for ways to suppress binding. It would be good to have some explanation at the end of section 4.3.4 (http://jdbi.org/#_supported_argument_types) that talks more about using this annotation to control binding (`@JdbiProperty(bind=false)`)

kotlin/src/main/kotlin/org/jdbi/v3/core/kotlin/KotlinMapper.kt
```diff
@@ -134,7 +134,7 @@ class KotlinMapper(val kClass: KClass<*>, private val prefix: String = "") : Row
         }
 
         val memberPropertyMappers = memberProperties.filter {
-            !(it.javaField?.getAnnotation(Unmappable::class.java)?.value ?: false)
```
hgschmie, line 137: can't we use the helper here as well?    `JdbiAnnotations.isBound(it.javaField?)` should work, should it not? We can skip this for now if Eclipse does not like kotlin.

hgschmie (commented):
Bunch of smaller nits to pick. I would like to see better docs. People are struggling with the documentation and adding really only the minimal amount of information does not help. 

Unless of course we put it at the bottom of a locked filing cabinet stuck in a disused lavatory with a sign on the door saying ‘Beware of the Leopard'. ;-)

PR #2134, opened 2022-10-08
Unit Test Refactoring

sman-81:
- Substitute all JUnit5 assertions by AssertJ (assertj-core) i.e. `assertEquals`/`assertNull`/`assertTrue` become `assertThat`
- Bump JUnit version from 5.8.2 to 5.9.1
- ~~Remove scarcely used dependency on `org.assertj:assertj-guava` by rewriting asserts using `assertj-core`~~
- ~~Introduce test base class `JdbiBaseTest` which extends `JdbiAssertions` which extends AssertJ Assertions~~
- ~~Bump SLF4J version from 1.7.32 to 1.7.36~~

Note: jdbi3-testing depends on jdbi3-core, therefore jdbi3-core should not currently use testing (avoid circular dependency). This is something we might want to refactor going forward.


hgschmie (changes requested):
I do not want to delve into the "extends AbstractTestBase" hell that is inevitable once some "test base class" is introduced. I think we have about zero tests that extend a base class unless that inheritance is a fundamental concept that is tested.

That is not what inheritance is intended for. If we factor out code, we do so in helper classes and use those helpers in the tests.

testing/src/main/java/org/jdbi/v3/testing/JdbiAssertions.java
```diff
@@ -0,0 +1,21 @@
...
+ * limitations under the License.
+ */
+package org.jdbi.v3.testing;
+
+import org.assertj.core.api.Assertions;
+
+/**
+ * Extension entry-point for custom JDBI assertions.
+ */
+public class JdbiAssertions extends Assertions {}
```
hgschmie, line 21: this makes no sense to me. Why would this code extend an assertions class?

hgschmie (changes requested):

hgschmie:
There is some value in this PR (updating some versions and dependencies). Replacing all the different tests is a labor of love, having done the Junit4 -> Junit5 transition I really appreciate that work. Some are improving the tests (e.g. `assertThat(someOptional.get()).isTrue()` to `assertThat(someOptional).isTrue()` ), for some others I don't really see the value (`assertEquals("", something)` -> `assertThat(something).isBlank()`) but I don't mind the change. 

Splitting these things into a number of smaller PRs would be good.

testing/src/main/java/org/jdbi/v3/testing/JdbiAssertions.java
sman-81: Extending `Assertions` gives access to all AssertJ `assertThat` methods without the need for static imports.

sman-81:
> I do not want to delve into the "extends AbstractTestBase" hell that is inevitable once some "test base class" is introduced. I think we have about zero tests that extend a base class unless that inheritance is a fundamental concept that is tested.
> 
> That is not what inheritance is intended for. If we factor out code, we do so in helper classes and use those helpers in the tests.

One idea behind a test base class is for it to contain common testing behaviour such as the test begin / test end logging. How would this be achieved this without inheritance?
Also, by extending Assertions these become part of the tests' namespace and are available without static imports. The test extends AssertJ Assertions not JUnit5 Assertions. All of the tests have been updated to use only AssertJ, while before the mix of JUnit / AssertJ was more or less random.

sman-81:
First of all thanks for your feedback @hgschmie.

> There is some value in this PR (updating some versions and dependencies). Replacing all the different tests is a labor of love, having done the Junit4 -> Junit5 transition I really appreciate that work. Some are improving the tests (e.g. `assertThat(someOptional.get()).isTrue()` to `assertThat(someOptional).isTrue()` ), for some others I don't really see the value (`assertEquals("", something)` -> `assertThat(something).isBlank()`) but I don't mind the change.

The changes make tests somewhat cleaner and easier to read IMO by using only AssertJ and fluent style where possible. I should note that this is a purely technical update without any effect on test coverage or functionality.
It fixes all remaining 'SonarCloud Quality Gate' code smells that I was aware of. This brings the project to zero code smells. It also solves the oddity in `TestScript.testOracleScriptWithBeginEndBlock` where Java 9 API `String.chars()` had slipped in.

I read from your comments that you don't like the AbstractTestBase and extends Assertions thing much which I understand and am fine with. I'll go ahead and remove these parts from the PR. Then shall we take another look?

testing/src/main/java/org/jdbi/v3/testing/JdbiAssertions.java
hgschmie: Why would static imports be bad? Inheritance is not intended for "make some methods available without import". These are public methods and now I can write "TestFoo.assertThat" in some other test class. 

JDBI uses inheritance where it makes sense. Most of the time, aggregation and encapsulation work better.

hgschmie:
> First of all thanks for your feedback @hgschmie.
> 
> > There is some value in this PR (updating some versions and dependencies). Replacing all the different tests is a labor of love, having done the Junit4 -> Junit5 transition I really appreciate that work. Some are improving the tests (e.g. `assertThat(someOptional.get()).isTrue()` to `assertThat(someOptional).isTrue()` ), for some others I don't really see the value (`assertEquals("", something)` -> `assertThat(something).isBlank()`) but I don't mind the change.
> 
> The changes make tests somewhat cleaner and easier to read IMO by using only AssertJ and fluent style where possible. I should note that this is a purely technical update without any effect on test coverage or functionality. It fixes all remaining 'SonarCloud Quality Gate' code smells that I was aware of. This brings the project to zero code smells. It also solves the oddity in `TestScript.testOracleScriptWithBeginEndBlock` where Java 9 API `String.chars()` had slipped in.
> 
> I read from your comments that you don't like the AbstractTestBase and extends Assertions thing much which I understand and am fine with. I'll go ahead and remove these parts from the PR. Then shall we take another look?

Yes, definitely. Thank you for doing this work.

sman-81:
Pull request updated and rebased.

sman-81:
Hi Henning (@hgschmie)
please take another look at this PR. I just rebased it to master. It can be merged without conflicts.
Have a great week!

guava/src/test/java/org/jdbi/v3/guava/TestGuavaCollectors.java
```diff
@@ -99,21 +98,21 @@ public void immutableSortedSetWithComparator() {
...
-                .collect(toList()));
+                .collect(Collectors.toList()));
     }
 
     @Test
     public void optionalPresent() {
         Optional<Integer> shouldBePresent = h2Extension.getSharedHandle().createQuery("select intValue from something where intValue = 1")
             .collectInto(new GenericType<Optional<Integer>>() {});
-        assertThat(shouldBePresent).contains(1);
+        assertThat(shouldBePresent.get()).isOne();
```
hgschmie, line 108: this change is necessary because now, the guava assert no longer is present. We use this library specifically because it tests the guava types. Why is that removal useful? It is part of the test code, so it does not increase the footprint of what we ship but make our tests less specific

testing/pom.xml
```diff
@@ -55,6 +55,11 @@
             <optional>true</optional>
         </dependency>
 
+        <dependency>
```
hgschmie, line 58: I am not a fan of this. We are now forcing all users of the testing module to also use assertj. While I am comfortable with using this in our tests, this is expanding our footprint as a library to the users.

hgschmie: can this be test scope?

hgschmie (changes requested):
Most of the changes are good.

- I don't like the crusade against assertj-guava. The tests that use guava do test guava functionality and this needs to be preserved. There is no need to remove a testing dependency, we do not ship it
- We should not force the users of -testing to also pull in assertj. This will cause problems .

guava/src/test/java/org/jdbi/v3/guava/TestGuavaCollectors.java
```diff
@@ -170,9 +169,20 @@ private <M extends Multimap<Long, String>> void testMultimapCollector(Class<? ex
                 .findFor(genericType.getType())
                 .orElseThrow(() -> new IllegalStateException("Missing collector for " + genericType));
 
-        M map = Stream.of(entry(1L, "foo"), entry(2L, "bar")).collect(collector);
```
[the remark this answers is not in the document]
sman-81: > I don't like the crusade against assertj-guava

it's no crusade :) There are so few uses of `assertj-guava` in the project, that personally find little value in depending on it.
But it's your call. I will restore assertj-guava in the PR.

>  don't understand this rewrite.

The non assertj-guava version of the test code. I'll leave roll it back

guava/src/test/java/org/jdbi/v3/guava/MultimapEntryMapperTest.java
```diff
@@ -25,7 +25,7 @@
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
-import static org.assertj.guava.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThat;
```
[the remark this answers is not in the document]
sman-81: same thing

Both are `org.assertj.core.api.AbstractAssert.isEqualTo(Object)`

guava/src/test/java/org/jdbi/v3/guava/TestGuavaCollectors.java
sman-81: changed it back

testing/pom.xml
sman-81: removed

sman-81:
Hi @hgschmie, all your requested changes implemented, commited and rebased!

testing/pom.xml
```diff
@@ -90,6 +90,7 @@
             <artifactId>flyway-core</artifactId>
             <optional>true</optional>
         </dependency>
+
```
hgschmie, line 93: :-)

hgschmie (approved):

hgschmie:
Thank you for doing all this work. Tests are super-important but are often considered a chore.

PR #2137, opened 2022-10-11
Add Maven defaultGoal to support invocation of builds with only 'mvn'

sman-81:


stevenschlansker (approved):
Wow, I didn't even know this existed. I do think it's useful to be explicit about what you want Maven to do, but I don't see a way this is harmful, so OK.

sman-81:
> [..] I do think it's useful to be explicit about what you want Maven to do

I agree. The explicitness comes through the very well crafted POM of this project. The defaultGoal is for convenience, but also carries the message 'this is the build command to invoke'.

hgschmie:
I am actually hoping to steer people towards "make install" as the default goal. At some point, we want to switch from `./mvnw clean install` to `mvnd clean install` to use the maven daemon.

sman-81:
> I am actually hoping to steer people towards "make install" as the default goal.

I've never heard of a 'make' goal :)

`mvnw` and `mvnd` should honor the defaultGoal just the same as it is defined in the POM.

hgschmie:
no, not a "make" goal. There is a Makefile in the root of the tree and you can use it to execute most build goals.

sman-81:
> no, not a "make" goal. There is a Makefile in the root of the tree and you can use it to execute most build goals.

Will check that out. Thanks!

PR #2138, opened 2022-10-11
Add Maven profile 'fast'

sman-81:
This PR adds Maven profile 'fast' to the internal build parent POM.
The profile is intended to build the project while skipping all time-consuming steps such as test execution, pmd, checkstyle, javadoc, etc. It is of course not active by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@
 #
 SHELL = /bin/sh
 .SUFFIXES:
-.PHONY: help clean install install-nodocker docs tests tests-nodocker publish-docs deploy release
+.PHONY: help clean install install-nodocker install-fast docs tests tests-nodocker publish-docs deploy release
 
 # replace JDBI_MAVEN_OPTS with implicit MAVEN_OPTS, once 3.9.x or later has been released
 MAVEN = ./mvnw ${JDBI_MAVEN_OPTS}
@@ -30,6 +30,9 @@ install:
 
 install-nodocker: JDBI_MAVEN_OPTS += -Dno-docker
 install-nodocker: install
+
+install-fast: JDBI_MAVEN_OPTS += -Pfast
+install-fast: install
 
 docs: JDBI_MAVEN_OPTS += -Dbasepom.check.skip-all=true -Dbasepom.javadoc.skip=false -DskipTests=true
 docs: install
@@ -57,6 +60,7 @@ help:
 	@echo " * clean            - clean local build tree"
 	@echo " * install          - builds and installs the current version in the local repository"
 	@echo " * install-nodocker - same as 'install', but skip all tests that require a local docker installation"
+	@echo " * install-fast     - same as 'install', but skip test execution and code analysis (Checkstyle/PMD/Spotbugs)
 	@echo " * docs             - build up-to-date documentation in docs/target/generated-docs/"
 	@echo " * tests            - run all unit and integration tests"
 	@echo " * tests-nodocker   - same as 'tests', but skip all tests that require a local docker installation"
--- a/internal/build/pom.xml
+++ b/internal/build/pom.xml
@@ -1057,5 +1057,15 @@
                 </plugins>
             </build>
         </profile>
+
+        <profile>
+            <!-- Profile to skip time-consuming steps. -->
+            <id>fast</id>
+            <properties>
+                <basepom.check.skip-all>true</basepom.check.skip-all>
+                <skipTests>true</skipTests>
+            </properties>
+        </profile>
+
     </profiles>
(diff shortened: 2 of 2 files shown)


internal/build/pom.xml
```diff
@@ -1054,5 +1057,21 @@
                 </plugins>
             </build>
         </profile>
+
+        <profile>
+            <!-- Profile to skip time-consuming steps. -->
+            <id>fast</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
```
hgschmie, line 1065: this is not necessary. And `activeByDefault` is somewhat funky because marking a profiles as "active as default true" actually disables all other profiles.

internal/build/pom.xml
```diff
@@ -1054,5 +1057,21 @@
...
+            </activation>
+            <properties>
+                <basepom.check.skip-all>true</basepom.check.skip-all>
+                <!-- skip _compiling_ the tests -->
+                <maven.test.skip>true</maven.test.skip>
```
hgschmie, line 1070: this skips the test compilation; this is a habit we should not get into.

hgschmie (changes requested):
we should not skip test compilation.

internal/build/pom.xml
sman-81: True that the default is false
`<xs:element minOccurs="0" name="activeByDefault" type="xs:boolean" default="false"></xs:element>`

> [..] `activeByDefault` is somewhat funky because marking a profiles as "active as default true" actually disables all other profiles

The documentation describes it differently: If set to true, this profile will be active unless another profile in this pom is activated using the command line -P option or by one of that profile's activators.

internal/build/pom.xml
sman-81: I'll update the PR. There isn't much of it left however 😂

internal/build/pom.xml
hgschmie: as I said, it is funky. :-)

internal/build/pom.xml
hgschmie: you could add another target to the makefile that allows "make install-fast"

testing/src/main/java/org/jdbi/v3/testing/junit5/JdbiPostgresExtension.java
```diff
@@ -63,8 +63,7 @@ public class JdbiPostgresExtension extends JdbiExtension {
 
     private volatile DatabaseInfo info;
 
-    static final JdbiExtension instance(EmbeddedPgExtension pg) {
```
hgschmie, line 66: this seems unrelated?

internal/build/pom.xml
```diff
@@ -1054,5 +1057,14 @@
                 </plugins>
             </build>
         </profile>
+
+        <profile>
+            <!-- Profile to skip time-consuming steps. -->
+            <id>fast</id>
+            <properties>
+                <basepom.check.skip-all>true</basepom.check.skip-all>
```
hgschmie, line 1065: sorry, I was not clear in my review. the `<skipTests>true</skipTests` can stay just fine. What I don't want us is to not compile the classes. Not running the tests is ok.

testing/src/main/java/org/jdbi/v3/testing/junit5/JdbiPostgresExtension.java
hgschmie: or maybe needs a rebase?

sman-81: Not part my PR. Not sure how Github crunches the data to produce this?!
I will rebase the pull request.

internal/build/pom.xml
sman-81: > you could add another target to the makefile that allows "make install-fast"

good point!
done

internal/build/pom.xml
sman-81: glad to read that

done

Makefile
```diff
@@ -31,6 +31,8 @@ install:
 install-nodocker: JDBI_MAVEN_OPTS += -Dno-docker
 install-nodocker: install
 
+install-fast: ${MAVEN} clean install -Pfast
```
hgschmie, line 34: this is not how make works. :-)  This should probably be
```
install-fast: JDBI_MAVEN_OPTS += -Pfast
install-fast: install
```

hgschmie (changes requested):
the change to the Makefile, then it is good to go.

PR #2145, opened 2022-10-18
BaseStatement: why do we need a helper method for close()?

stevenschlansker:
This is a followup to #2123 - I don't understand why a public static helper method is preferable to simply using the existing close method.

Even if we choose to ditch Closeable, if there is any cleanup to do, the close method seems like the natural place.
I do not intend to start an edit war, but why is the current state preferable to the simpler place we were at before?


hgschmie (approved):

hgschmie:
I am fine with this. I am not actually convinced that we want to remove Closeable. There is precedent in the JDK (Stream) that things that are closeable do not *need* to be closed.

Issue #2119, opened 2022-09-30
Evaluating JDBI. Why use `Closeable` on objects that are returned to the user that are not supposed to be closed?


hjohn:
So, after just 30 minutes of evaluating JDBI, I ran into this issue that I think will deter me from using it.

I want to do the following:

    jdbi.useTransaction(handle -> {
        handle.createUpdate("INSERT INTO xyz (data) VALUES (:data)")
            .bind("data", data)
            .execute();
    });

Now, `createUpdate` returns a `Closeable`.  My question is, why would it do this when there is no need to create the actual closeable resource until one of `execute` or `one` is called?  You are basically giving me a simple "binder" type object that allows me to bind things to a simple `String`, but it seems this object that is returned is already an "expensive" resource (perhaps there is even an actual prepared statement created already) that will require closing.  Why not defer creating the prepared statement until it is actually required?

Now the user is getting a warning on `createUpdate` that this resource needs closing, and IMHO these warnings should not be ignored without good reason.  Note that in this scenario, the user never needs to close anything, so why return something that is `Closeable` in the first place?

I would expect the implementation of `createUpdate` to be something like:
    
      class Handle {
           UpdateBinder createUpdate(String sql) {
                return new Binder(this, sql);  // this = Handle     
           }
      }

      class UpdateBinder {  // not Closeable, why would it be?
           UpdateBinder (Handle handle, String sql) {
                ...
           }

           ... bind methods ...

           int execute() {
                // - use handle to create prepared statement -- use try with resources on that
                // - bind everything
                // - execute
                // - close
           }
      }

No `Closeable` is ever returned to the user.

stevenschlansker:
@hjohn , the reason that our statement types are closeable is that they own resources that must be returned to the database system after use. For example, an `Update` holds on to (at least) a `Statement`. It is true that in the case of `one()`, we are able to prove that the user does not need to close the statement, since `one()` finishes the entire process.

However, in more advanced cases, you might be returning a type like `Iterable<T>` or `Stream<T>` which holds on to an open database cursor. In those cases, you are responsible to close these resources in case the iterator isn't exhausted, otherwise you leak the cursor until the connection is closed entirely.

It would be great if we could statically differentiate between terminal operations that always clean themselves up, and non-terminal operations where you are responsible to finish cleaning up. Unfortunately I think that would be a large change and I'm not sure what the API would look like.

So, there is a good reason, even if it's annoying in some cases.
Hope that helps.

[...]

PR #2146, opened 2022-10-18
Add ResultIterable.withIterator and useIterator

stevenschlansker:
Inspired by #2119


hgschmie (changes requested):
The corresponding stream methods (`useStream` and `withStream`) use explicit callback interfaces (`StreamCallback` and `StreamConsumer`) that allow throwing exceptions from the called code. This feels uneven if the iterator code does not.

hgschmie (approved):
Ship it.

image

PR #2149, opened 2022-10-19
hsqldb 2.7.1

stevenschlansker:


hgschmie:
there is no 2.7.1 on central. It is supposed to come out today but I don't see  it here: https://central.sonatype.dev/artifact/org.hsqldb/hsqldb/2.7.0

If this is driven out of a dependabot alert, can we link those here?  (https://github.com/jdbi/jdbi/security/dependabot/16)

stevenschlansker:
Ah, I did not realize you could link them. I don't seem to be able to comment or otherwise modify the Dependabot alert in any way other than dismissing.

Yes, that's why I tried to upgrade - but still no artifact to upgrade to...

hgschmie (approved):

PR #2157, opened 2022-10-25
Enhance TestInetAddress* unit tests

### LINKED ISSUE #1998: Unable to use bound InetAddress with PostgreSQL
I am trying to insert an Inet4Address to a PostgreSQL table containing a column of type "inet". According to the documentation this is a supported data type for bound parameters (see https://jdbi.org/#_supported_argument_types and https://github.com/jdbi/jdbi/pull/446 ). However, when I try to insert an Inet4Address created with

`InetAddress inetAddress = Inet4Address.getByName("8.8.8.8");`

I get an error message

> org.jdbi.v3.core.statement.UnableToExecuteStatementException: org.postgresql.util.PSQLException: ERROR: column "hostipv4" is of type inet but expression is of type character varying
>   Hint: You will need to rewrite or cast the expression.

Test case:

```
@Test
    public void inetTest(){
        try (Handle handle = jdbi.open()) {
            String createTableSql = """
                    CREATE  TABLE "public".inet_test (
                    	hostipv4             inet
                     );
                    """;
            handle.execute(createTableSql);

            String hostAddress = "8.8.8.8";

            InetAddress inetAddress = Inet4Address.getByName(hostAddress);

            handle.createUpdate("insert into inet_test(hostipv4) values (?)")
                    .bind(0, inetAddress)
                    .execute();

        }catch(Exception e){
            e.printStackTrace();
        }
    }
```

The exception is produced at line "handle.createUpdate". When I use [p6spy](https://github.com/p6spy/p6spy) to observe what SQL is created and sent to the PostgreSQL driver and execute the created SQL statements manually in the database, it works. The error seems to be in the validation.

sman-81:
Fixes #1998

see https://www.postgresql.org/docs/current/datatype-net-types.html

While JDBI had an `InetArgumentFactory` an `InetColumnMapper` was strangely missing.
Enhanced the unit test from insert-only to insert/select and update/select and testing with null.

### DIFF
--- a/core/src/test/java/org/jdbi/v3/core/argument/TestInetAddressH2.java
+++ b/core/src/test/java/org/jdbi/v3/core/argument/TestInetAddressH2.java
@@ -13,10 +13,13 @@
  */
 package org.jdbi.v3.core.argument;
 
+import java.net.Inet4Address;
 import java.net.InetAddress;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.Sql;
 import org.jdbi.v3.core.junit5.DatabaseExtension;
 import org.jdbi.v3.core.junit5.H2DatabaseExtension;
 import org.junit.jupiter.api.Test;
@@ -53,7 +56,44 @@ public class TestInetAddressH2 {
         });
     }
 
+    @Test
+    public void testInetAddress2() throws Exception {
+        try (Handle h = dbExtension.getJdbi().open()) {
+            h.execute("CREATE TABLE inet_test (ipaddress " + getInetType() + ")");
+
+            InetAddress inetAddrIn = InetAddress.getByName("8.8.8.8");
+            assertThat(inetAddrIn).isInstanceOf(Inet4Address.class);
+
+            // insert IP address into column of type INET
+            int update1Count = h.createUpdate(Sql.of("INSERT INTO inet_test (ipaddress) VALUES (:address)"))
+                .bind("address", inetAddrIn)
+                .execute();
+            assertThat(update1Count).isOne();
+
+            // read back the inserted IP address
+            InetAddress inetAddr1Out = h.createQuery(Sql.of("SELECT ipaddress FROM inet_test"))
+                .mapTo(InetAddress.class)
+                .one();
+
+            // test for equality
+            assertThat(inetAddrIn).isEqualTo(inetAddr1Out);
+
+            // update the record to a null IP address
+            int update2Count = h.createUpdate(Sql.of("UPDATE inet_test SET ipaddress = :address"))
+                .bind("address", (InetAddress) null)
+                .execute();
+            assertThat(update2Count).isOne();
+
+            // read back the record ensuring proper null handling
+            InetAddress inetAddr2Out = h.createQuery(Sql.of("SELECT ipaddress FROM inet_test"))
+                .mapTo(InetAddress.class)
+                .one();
+            assertThat(inetAddr2Out).isNull();
+        }
+    }
+
     protected String getInetType() {
         return "VARCHAR";
     }
+
 }


postgres/src/main/java/org/jdbi/v3/postgres/InetColumnMapper.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+class InetColumnMapper implements ColumnMapper<InetAddress> {
+    @Override
+    public InetAddress map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
+        String val = Optional.ofNullable(r.getObject(columnNumber))
+            .map(PGobject.class::cast)
+            .map(PGobject::getValue).orElse(null);
+
+        try {
+            return val == null ? null : InetAddress.getByName(val);
```
hgschmie, line 37: this makes me super-nervous. I am not sure that we want to do actual DNS resolution on the fly when mapping data from the database. It also only returns a single value, there is a method (getAllByName) that returns an array of names (which makes sense as these names can map to multiple addresses).

But then again, we already do that in the InternetMapperFactory, so I guess the point is moot.

postgres/src/main/java/org/jdbi/v3/postgres/InetColumnMapper.java
```diff
@@ -0,0 +1,43 @@
...
+        } catch (UnknownHostException ex) {
+            throw new IllegalArgumentException(String.format(
+                "could not get %s from %s (%s)", InetAddress.class.getSimpleName(), val, ex));
```
hgschmie, line 40: use the (message, throwable) constructor and not format the exception as part of the message. This will break the exception chain

also, in the InternetMapperFactory, we throw a MappingException, so this should match

postgres/src/main/java/org/jdbi/v3/postgres/InetColumnMapper.java
```diff
@@ -0,0 +1,43 @@
...
+import org.jdbi.v3.core.statement.StatementContext;
+import org.postgresql.util.PGobject;
+
+/**
+ * A column mapper which maps Postgres's {@code inet} type to Java's {@link InetAddress}.
+ */
+class InetColumnMapper implements ColumnMapper<InetAddress> {
+    @Override
+    public InetAddress map(ResultSet r, int columnNumber, StatementContext ctx) throws SQLException {
+        String val = Optional.ofNullable(r.getObject(columnNumber))
```
hgschmie, line 32: this is the only difference to the code in the InternetMapperFactory in core. Why is this better/needed compared to just doing getString() ?

postgres/src/test/java/org/jdbi/v3/postgres/TestInetAddressPg.java
```diff
@@ -13,19 +13,79 @@
...
+            ")");
+
+        try (Handle h = dbExtension.getJdbi().open()) {
+            h.execute(createTableSql);
+
+            InetAddress inetAddrIn = InetAddress.getByName("8.8.8.8");
+            assertThat(inetAddrIn).isInstanceOf(Inet4Address.class);
+
+            // insert IP address into column of type INET
+            int update1Count = h.createUpdate(Sql.of("INSERT INTO", tblName, "(", colName, ") VALUES (?)"))
```
hgschmie, line 64: I am not convinced that this is more readable than just "INSERT INTO inet_test (ipaddress) VALUES (:address)".  Our tests tend to focus on readability more than type safety or constant out-factoring. personally I am very fine with writing the string "inet_test" three times, because if I get it wrong once, the test will fail. this test was not intuitive or easy for me to read.

postgres/src/test/java/org/jdbi/v3/postgres/TestInetAddressPg.java
```diff
@@ -13,19 +13,79 @@
...
+            // read back the inserted IP address
+            InetAddress inetAddr1Out = h.createQuery(Sql.of("SELECT", colName, "FROM", tblName))
+                .mapTo(InetAddress.class)
+                .one();
+
+            // test for equality
+            assertThat(inetAddrIn).isEqualTo(inetAddr1Out);
+
+            // update the record to a null IP address
+            int update2Count = h.createUpdate(Sql.of("UPDATE", tblName, "SET", colName, "= ?"))
```
hgschmie, line 78: I think I never used positional bindings, so I guess that works. It is somewhat brittle, the main selling point of Jdbi is using named bindings. :-)

hgschmie (changes requested):

hgschmie:
Is this even needed? There is a factory in core that provides both Column and SingleColumnRow mappers for InetAddress

postgres/src/main/java/org/jdbi/v3/postgres/InetColumnMapper.java
sman-81: `InternetMapperFactory` does already what is needed. I'll use that (must make it public though).

postgres/src/main/java/org/jdbi/v3/postgres/InetColumnMapper.java
sman-81: `InetColumnMapper` ist gone

I find quite a few of mappers that throw `IllegalArgumentException`. Those should be reviewed then and possibly replaced by `MappingException`. I'll take a note and may provide a separate pull request.

postgres/src/test/java/org/jdbi/v3/postgres/TestInetAddressPg.java
sman-81: Very good feedback. You've convinced me. I can get carried away with _"clean code"_ ...

postgres/src/test/java/org/jdbi/v3/postgres/TestInetAddressPg.java
sman-81: That slipped in from the original bug report. Will update to named bindings :)

postgres/src/test/java/org/jdbi/v3/postgres/TestInetAddressPg.java
```diff
@@ -13,19 +13,72 @@
...
+
+    {
+        dbExtension = PgDatabaseExtension.instance(pg).withPlugin(new PostgresPlugin());
+    }
 
     @RegisterExtension
     public static EmbeddedPgExtension pg = MultiDatabaseBuilder.instanceWithDefaults().build();
 
+    @RegisterExtension
+    public JdbiExtension pgExtension = JdbiExtension.postgres(pg).withPlugin(new PostgresPlugin());
```
hgschmie, line 41: this should not be needed. this test extends a test in the core module, so it uses the core specific (non-testing) test classes. dbExtension should work just fine, no need to bring up another Jdbi.

postgres/src/main/java/org/jdbi/v3/postgres/PostgresPlugin.java
```diff
@@ -146,6 +147,7 @@ public void customizeJdbi(Jdbi jdbi) {
         jdbi.registerArrayType(new ByteaArrayType());
 
         jdbi.registerColumnMapper(new JavaTimeMapperFactory());
+        jdbi.registerColumnMapper(new InternetMapperFactory());
```
hgschmie, line 150: this is registered with the core mappers in the BuiltinMapperFactory, so this should not be needed.

core/src/main/java/org/jdbi/v3/core/mapper/InternetMapperFactory.java
```diff
@@ -67,7 +67,7 @@ private static InetAddress getInetAddress(ResultSet r, int i, StatementContext c
         try {
             return hostname == null ? null : InetAddress.getByName(hostname);
         } catch (UnknownHostException e) {
-            throw new MappingException("Could not map InetAddress", e);
+            throw new MappingException("Could not map hostname " + hostname + " to InetAddress", e);
```
hgschmie, line 70: this is not needed, as the UnknownHostException is constructed as `new UnknownHostException(hostname)` :-)

postgres/src/test/java/org/jdbi/v3/postgres/TestInetAddressPg.java
```diff
@@ -13,19 +13,72 @@
...
+
+            // insert IP address into column of type INET
+            int update1Count = h.createUpdate(Sql.of("INSERT INTO inet_test (ipaddress) VALUES (:address)"))
+                .bind("address", inetAddrIn)
+                .execute();
+            assertThat(update1Count).isOne();
+
+            // read back the inserted IP address
+            InetAddress inetAddr1Out = h.createQuery(Sql.of("SELECT ipaddress FROM inet_test"))
+                .mapTo(InetAddress.class)
```
hgschmie, line 64: this is the big difference between this unit test and the bug report. You are doing this right. I am pretty sure that your test will work even without the mapper registered.

core/src/main/java/org/jdbi/v3/core/mapper/InternetMapperFactory.java
```diff
@@ -37,10 +37,10 @@
  *     <li>{@link URI}</li>
  * </ul>
  */
-class InternetMapperFactory implements ColumnMapperFactory {
+public class InternetMapperFactory implements ColumnMapperFactory {
```
hgschmie, line 40: neither of these public should be needed. The actual mapper can be retrieved with `ctx.getConfig(ColumnMappers.class).forType(InetAddress.class)` if needed and the actual type is an implementation detail that can easily be hidden. No user should ever have the need to actually instantiate the mapper directly outside a test (and those can be package private).

hgschmie (changes requested):
I am still not convinced that this is needed at all. The current code should "just work", even the bug report #1998 works, once the data is mapped onto the right data type.

What is the problem that this PR is addressing?

postgres/src/test/java/org/jdbi/v3/postgres/TestInetAddressPg.java
```diff
@@ -13,19 +13,72 @@
...
+    @RegisterExtension
+    public JdbiExtension pgExtension = JdbiExtension.postgres(pg).withPlugin(new PostgresPlugin());
+
     @Override
     protected String getInetType() {
         return "INET";
     }
+
+    @Test
+    public void testPgInetAddress() throws Exception {
```
hgschmie, line 49: you can actually paste this test "as is' into TestInetAddressH2 and it will pass both with H2 and Pg.

core/src/test/java/org/jdbi/v3/core/argument/TestInetAddressH2.java
```diff
@@ -29,6 +32,10 @@ public class TestInetAddressH2 {
     @RegisterExtension
     public DatabaseExtension<?> dbExtension = H2DatabaseExtension.instance();
 
+    protected String getInetType() {
```
hgschmie, line 35: we tend to order the code "test methods" -> public / protected / package private / private methods from top to bottom. So this method should stay at the bottom of the class.

core/src/main/java/org/jdbi/v3/core/mapper/InternetMapperFactory.java
sman-81: true :)

postgres/src/test/java/org/jdbi/v3/postgres/TestInetAddressPg.java
sman-81: done

postgres/src/test/java/org/jdbi/v3/postgres/TestInetAddressPg.java
sman-81: makes sense now

core/src/test/java/org/jdbi/v3/core/argument/TestInetAddressH2.java
sman-81: I wished this could be enforced by Checkstyle. There is no core Checkstyle check to do this afaik (only 3rd party).

sman-81:
PR title renamed from `Configure column mapper for Postgres's inet type to Java InetAddress`.

postgres/src/test/java/org/jdbi/v3/postgres/TestInetAddressPg.java
```diff
@@ -19,7 +19,11 @@
 import org.jdbi.v3.core.junit5.PgDatabaseExtension;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
-public class TestInetAddressPg extends TestInetAddressH2 { { dbExtension = PgDatabaseExtension.instance(pg).withPlugin(new PostgresPlugin()); }
```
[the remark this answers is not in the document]
sman-81: sorry

PR #2158, opened 2022-10-26
Cleanup StatementBuilder usage

hgschmie:
- Consistent calling of close for statements created by the builder
- Explicit close methods for PreparedStatement and regular Statement

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
@@ -1766,10 +1766,10 @@ public abstract class SqlStatement<This extends SqlStatement<This>> extends Base
         ParsedSql parsedSql = parseSql();
 
         try {
-            stmt = createStatement(ctx, parsedSql);
+            stmt = createStatement(parsedSql.getSql());
             // The statement builder might (or might not) clean up the statement when called. E.g. the
             // caching statement builder relies on the statement *not* being closed.
-            getContext().addCleanable(() -> getHandle().getStatementBuilder().close(getHandle().getConnection(), this.sql, stmt));
+            getContext().addCleanable(() -> cleanupStatement(stmt));
             getConfig(SqlStatements.class).customize(stmt);
         } catch (SQLException e) {
             throw new UnableToCreateStatementException(e, ctx);
@@ -1794,8 +1794,12 @@ public abstract class SqlStatement<This extends SqlStatement<This>> extends Base
         return stmt;
     }
 
-    PreparedStatement createStatement(final StatementContext ctx, ParsedSql parsedSql) throws SQLException {
-        return getHandle().getStatementBuilder().create(getHandle().getConnection(), parsedSql.getSql(), ctx);
+    PreparedStatement createStatement(final String parsedSql) throws SQLException {
+        return getHandle().getStatementBuilder().create(getHandle().getConnection(), parsedSql, getContext());
+    }
+
+    void cleanupStatement(final PreparedStatement statement) throws SQLException {
+        getHandle().getStatementBuilder().close(getHandle().getConnection(), this.sql, statement);
     }
 
     ParsedSql parseSql() {
--- a/core/src/main/java/org/jdbi/v3/core/statement/StatementBuilder.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/StatementBuilder.java
@@ -57,7 +57,7 @@ public interface StatementBuilder {
      *
      * @param conn the JDBC Connection the statement is being created for
      * @param sql the translated SQL which should be prepared
-     * @param ctx Statement context associated with the SqlStatement this is building for
+     * @param ctx Statement context associated with the SqlStatement
      *
      * @return a CallableStatement for the given arguments
      *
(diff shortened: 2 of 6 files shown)


core/src/main/java/org/jdbi/v3/core/statement/StatementBuilder.java
```diff
@@ -57,29 +57,56 @@ public interface StatementBuilder {
...
      */
-    void close(Connection conn, String sql, Statement stmt) throws SQLException;
+    default void close(Statement stmt, Connection conn, StatementContext ctx) throws SQLException {
+        close(conn, "", stmt);
+    }
 
     /**
      * Called when the handle this StatementBuilder is attached to is closed.
      *
      * @param conn the connection to close.
```
stevenschlansker, line 109: This doc (not part of your immediate change) is slightly misleading, since you are not supposed to close the connection here.

core/src/main/java/org/jdbi/v3/core/statement/StatementBuilder.java
```diff
@@ -57,29 +57,56 @@ public interface StatementBuilder {
...
+     * Called to close an individual statement created from this builder.
      *
-     * @throws SQLException if anything goes wrong closing the statement
+     * @param stmt the prepared statement.
+     * @param conn the connection to close.
+     * @param ctx Statement context associated with the SqlStatement.
+     * @throws SQLException if anything goes wrong closing the statement.
      */
-    void close(Connection conn, String sql, Statement stmt) throws SQLException;
+    default void close(Statement stmt, Connection conn, StatementContext ctx) throws SQLException {
```
stevenschlansker, line 102: It is a little odd to have two method overloads that accept different sub-types, since method dispatch is decided by the static type. So for example
```
Statement stmt = createPreparedStatement();
close(stmt, conn, ctx);
```
will invoke the *Statement* overload instead of the *PreparedStatement* overload, which probably isn't what you wanted.

If it is important to do cleanup based on the dynamic type of the statement, maybe instead of two methods, we should just have one that accepts `Statement` and then checks `instanceof PreparedStatement` for special logic.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -1794,8 +1794,12 @@ PreparedStatement internalExecute() {
         return stmt;
     }
 
-    PreparedStatement createStatement(final StatementContext ctx, ParsedSql parsedSql) throws SQLException {
-        return getHandle().getStatementBuilder().create(getHandle().getConnection(), parsedSql.getSql(), ctx);
+    PreparedStatement createStatement(final String parsedSql) throws SQLException {
```
stevenschlansker, line 1797: Can this be pushed up to `BaseStatement` so it can be shared with `Batch`?

core/src/main/java/org/jdbi/v3/core/statement/StatementBuilder.java
hgschmie: this was mostly based on the fact that there is a method that actually takes the sql and there is now a code path that wants to hand back a statement that never saw any SQL (from the Batch). I can simplify that.

core/src/main/java/org/jdbi/v3/core/statement/StatementBuilder.java
hgschmie: ok, not sure I understand that comment. :-)

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
hgschmie: yes-ish but not a fan. BaseStatement then slowly becomes a collection of utility methods. We could push it to a helper.

hgschmie: also means we would need to split the prepare/cleanup pair as the cleanup method needs the sql in the statement (that the base statement does not have).

Why do we need the sql on the close BTW?

hgschmie: > Can this be pushed up to `BaseStatement` so it can be shared with `Batch`?

actually, batch does not have sql. So the create method there does not take a sql string.

hgschmie: to sum this up: We *could* push all of this into the handle. Not sure that this is a good place, but it would become shorthand for getHandle().getStatementBuilder()

I think we should rethink the whole statementbuilder thing. it is actually handle specific (because that is where the connection lives) and the various sqlstatement pieces should just go there and say "stamp me a new Statement/PrepareStatement" from this statementcontext. And this should be very doable because we actually assign the parsed sql to the statement context (in SqlStatement#parseSql), so passing it in is actually not needed.

It is a pretty mess. I think I will not touch it on this PR but just clean up the uses and then we may consider a backwards incompatible changes of an API that most likely no one uses anyway. :-)

core/src/main/java/org/jdbi/v3/core/statement/StatementBuilder.java
hgschmie: I think I understood it later. I changed the wording of the javadoc.

PR #2161, opened 2022-10-28
Replace ConnectionCloser with Cleanable

hgschmie:
Remove a single-use interface that works just as well with a lambda.


stevenschlansker (commented):
LGTM

PR #2165, opened 2022-11-03
Update contributing document and IDE support

hgschmie:
- move IDE files into a more obvious place
- add intellij code style
- reword the CONTRIBUTING document to be a bit more welcoming


hgschmie:
@stevenschlansker @sman-81

CONTRIBUTING.md
```diff
@@ -23,7 +23,7 @@ If you must make some internal code `public` to access it from other packages, p
 
 ### Forward
 
-Completely new API should, in most cases, be marked with `@Beta`. This lets users know not to rely too much on your changes yet, as the public release might reveal that more work needs to be done.
+Completely new APIs should be marked with `@Alpha` or `@Beta`. This lets users know not to rely too much on your changes yet, as the public release might reveal that more work needs to be done.
```
sman-81, line 26: We could add hyperlinks to Javadoc of the Alpha / Beta annotations. The Javadoc distinguishes them nicely.

CONTRIBUTING.md
```diff
@@ -41,90 +41,95 @@ Unit tests are nice for atomic components, but since jdbi is a complex ecosystem
 
 Since our tests essentially describe and verify jdbi's behavior, changing their specifics where it isn't inherently necessary is considered a red flag.
 
+The use of mocks and mocking framework is generally discouraged. There are a number of existing tests that use Mockito and they are a pain to maintain.
```
sman-81, line 44: It's not obvious why tests with mocking are a pain to maintain?

CONTRIBUTING.md
```diff
@@ -41,90 +41,95 @@ Unit tests are nice for atomic components, but since jdbi is a complex ecosystem
...
+
 ## Pull requests
 
-Expect your pull request to be scrutinized in even the tiniest details (down to grammar in javadoc), and to need to address many remarks even for small changes. We strive for a healthy balance between subjective perfection and practical considerations, but we are firmly against doing a quick and sloppy job that will require a lot of follow-up work later.
+We strive for a healthy balance between subjective perfection and practical considerations, but we are firmly against doing a quick and sloppy job that will require a lot of follow-up work later.
```
sman-81, line 48: We could add a word or two about meaningful PR titles and commit messages.

CONTRIBUTING.md
```diff
@@ -41,90 +41,95 @@ Unit tests are nice for atomic components, but since jdbi is a complex ecosystem
...
 The project uses a set of code style and formatting rules. These are enforced by Checkstyle and PMD as part of the build cycle.
-We ask that you configure the project in your development environment to support these rules as closely as possible.
 
-The JDBI Eclipse formatting preferences are available for download [here](internal/policy/src/main/resources/ide/jdbi-eclipse-formatter.xml).
+We do not review or merge PRs that do not pass our pre-merge checks.
 
 ### IntelliJ IDEA
 
-Import the file with the 'Adapter for Eclipse Code Formatter'. Additional info is available [here](https://plugins.jetbrains.com/plugin/6546-adapter-for-eclipse-code-formatter).
+* [IntelliJ IDEA formatting rules](https://github.com/jdbi/jdbi/blob/master/ide-support/intellij/jdbi.xml).
```
sman-81, line 64: Rename the IDEA settings file to `jdbi-intellij.xml` for alignment with names of the Eclipse settings files?
You could then do away with the `intellij` / `eclipse` sub directories (but it's a matter of taste).

CONTRIBUTING.md
```diff
@@ -41,90 +41,95 @@ Unit tests are nice for atomic components, but since jdbi is a complex ecosystem
...
 static *
 ```
 
-A blank line is required between each group.
-Imports in a group must be ordered alphabetically.
+A blank line is required between each group. Imports in a group must be ordered alphabetically.
 
 Wildcard (aka star) imports e.g. `import org.apache.*;`, including static imports, are not allowed.
 
 Javadoc may not cause an import statement i.e. use FQCN in Javadoc unless the import statement is already caused by code.
```
sman-81, line 113: in hindsight, not everyone might understand _FQCN_. I would suggest to change into _fully qualified class name (FQCN)_

sman-81 (approved):
nice update

CONTRIBUTING.md
hgschmie: good idea.

CONTRIBUTING.md
hgschmie: the collide e.g. with the leak checker. I am not against using mocks if they are done right (and most of the time, you end up using stubs or dummies, not mocks). What is abhorred are the pseudo-"convenient", reflection-driven, most of the time "record and regurgitate" frameworks.  

There are the inevitable links:
* https://martinfowler.com/bliki/TestDouble.html
* https://blog.cleancoder.com/uncle-bob/2014/05/14/TheLittleMocker.html

After every larger rewrite, some tests break even if functionality did not change. And most of the time, these are the few mockito tests that we have because they do some naive "record this, then this, then this, then check that these methods were called n times" checking and when rewriting code, these patterns change. So you end up testing the mock invocation, not the functionality. And that is a waste of time.

CONTRIBUTING.md
hgschmie: Personally, I don't care much about ceremony here. If someone sends a PR with "fix" as the commit message, we will call it out. But I want people to focus on code and functionality, not ceremony. We are a pretty informal bunch anyway. ;-)

CONTRIBUTING.md
hgschmie: the file name is used by the IDE.

CONTRIBUTING.md
hgschmie: +1

PR #2166, opened 2022-11-03
Remove pom versions that are equal to parent pom versions (avoiding warnings)

sman-81:


hgschmie (changes requested):
This needs to be rebased before it can be reviewed. There are additional commits on the PR that should not be here.

sman-81:
> This needs to be rebased before it can be reviewed. There are additional commits on the PR that should not be here.

Thanks for the heads up. Now it's fine :)

PR #2173, opened 2022-11-05
Add leak checkers

hgschmie:
Those are the leak checkers for the core tests and the testing test support module. They are activated by default and track whether all Handles are closed correctly and all cleanable objects get released.


hgschmie:
incorporated the feedback from the wip PR.

PR #2178, opened 2022-11-07
Replace handle thread locals

hgschmie:
Remove the convoluted thread local code in the handle, replace with a simple stack structure in the handle suppliers. This also gets rid of the undocumented and needlessly complicated Invocations class.


core/src/main/java/org/jdbi/v3/core/extension/ExtensionContext.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.jdbi.v3.core.extension;
+
+import org.jdbi.v3.core.config.ConfigRegistry;
+
+/**
+ * Accepted by a handle when executing extension code. This temporarily reconfigures the handle to execute
+ * extension code. Reconfigurations are e.g. annotations from the sql objects that define mappers or bindings.
+ */
+public final class ExtensionContext {
```
hgschmie, line 22: I would have loved to just roll this into ExtensionMethod but it would change the constructor in a non-backwards compat way.

core/src/main/java/org/jdbi/v3/core/LazyHandleSupplier.java
```diff
@@ -13,36 +13,40 @@
...
 import org.jdbi.v3.core.extension.ExtensionMethod;
 import org.jdbi.v3.core.extension.HandleSupplier;
+import org.jdbi.v3.core.internal.MemoizingSupplier;
 import org.jdbi.v3.core.internal.OnDemandHandleSupplier;
 
-import static org.jdbi.v3.core.internal.Invocations.invokeWith;
-
 class LazyHandleSupplier implements HandleSupplier, AutoCloseable, OnDemandHandleSupplier {
 
     private final Object[] lock = new Object[0];
```
stevenschlansker, line 30: Since this is always used as a ThreadLocal, this lock can go away, right?

core/src/main/java/org/jdbi/v3/core/LazyHandleSupplier.java
```diff
@@ -13,36 +13,40 @@
...
-
     private final Jdbi db;
-    private final ThreadLocal<ConfigRegistry> localConfig;
-
-    @SuppressWarnings("ThreadLocalUsage")
-    private final ThreadLocal<ExtensionMethod> localExtensionMethod = new ThreadLocal<>();
-
-    private volatile Handle handle;
-    private volatile boolean closed = false;
+    private final AtomicBoolean closed = new AtomicBoolean();
```
stevenschlansker, line 32: We don't seem to actually use any of the added features `AtomicBoolean` brings over the `volatile` we had before.

core/src/main/java/org/jdbi/v3/core/LazyHandleSupplier.java
```diff
@@ -52,45 +56,56 @@ public Jdbi getJdbi() {
...
     }
 
     @Override
     public void close() {
+        if (closed.get()) {
+            throw new IllegalStateException("Handle is closed");
+        }
+
         synchronized (lock) {
-            closed = true;
```
stevenschlansker, line 86: I don't see a corresponding `closed.set(true)` to replace the old `closed = true`

core/src/main/java/org/jdbi/v3/core/extension/ExtensionContext.java
stevenschlansker: If you add equivalent getters to this class, could we then deprecate `ExtensionMethod` and eventually retire it?

stevenschlansker (approved):

core/src/main/java/org/jdbi/v3/core/extension/ExtensionContext.java
hgschmie: there is a place where the config and the extension method show up on different call paths (the config comes from a lambda creation point and the extension method in the lambda execution), so I think for now we should keep them separate. We can't make it an internal class though because it is part of the extension API (HandleSupplier).

PR #2181, opened 2022-11-10
fix cleanAfter() method, add changelog

hgschmie:


hgschmie:
the `hiddenField` checkstyle thing is somewhat annoying...

PR #2190, opened 2022-11-16
Use inspectable code for the memoizer

hgschmie:
With the existing code, it is impossible to inspect a memoizer instance for the value that it contains which makes debugging really hard.

Replace the current code with the memoizer from Guava.


stevenschlansker (approved):
I think `transient` is not necessary

PR #2193, opened 2022-11-16
FieldMapper: make constructor accessible, make accessibility strategy configurable

stevenschlansker:
Replaces #2017
cc @paladin235


core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
```diff
@@ -115,6 +118,35 @@ public ReflectionMappers setCaseChange(UnaryOperator<String> caseChange) {
         return this;
     }
 
+    /**
+     * Set the strategy Jdbi uses for Java accessibility rules.
+     * The legacy default is to call {@code setAccessible(true)} in certain cases when we try to use a Constructor, Method, or Field.
+     * In the future, this default will be changed to a no-op, to better interact with the Java module system.
+     */
+    @Beta
+    public ReflectionMappers setAccessibleObjectStrategy(Consumer<AccessibleObject> makeAccessible) {
```
hgschmie, line 127: this seems an incredible convoluted way to say "call setAccessible(true)" on any accessible object. Are there real use cases where people will register actual code that looks at the passed in Accessible object (which is a constructor or a method, so not even a class) and decide "I make this accessible but not that other thing")? 

This seems overkill. A simple boolean flag might suffice. or a `Function<Class<?>, Boolean>` that decides on a per-class base whether to make its members accessible or not.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
```diff
@@ -30,6 +32,7 @@ public class ReflectionMappers implements JdbiConfig<ReflectionMappers> {
     private List<ColumnNameMatcher> columnNameMatchers;
     private boolean strictMatching;
     private UnaryOperator<String> caseChange;
+    private Consumer<AccessibleObject> makeAccessible = accessibleObject -> accessibleObject.setAccessible(true);
```
hgschmie, line 35: "accessibleStrategy"

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
```diff
@@ -115,6 +118,35 @@ public ReflectionMappers setCaseChange(UnaryOperator<String> caseChange) {
...
     }
 
+    /**
+     * Set the strategy Jdbi uses for Java accessibility rules.
+     * The legacy default is to call {@code setAccessible(true)} in certain cases when we try to use a Constructor, Method, or Field.
+     * In the future, this default will be changed to a no-op, to better interact with the Java module system.
+     */
+    @Beta
+    public ReflectionMappers setAccessibleObjectStrategy(Consumer<AccessibleObject> makeAccessible) {
+        this.makeAccessible = makeAccessible;
```
hgschmie, line 128: this.accessibleStrategy = accessibleStrategy. Heh, why does that not complain about "shadowing a class field" ?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
```diff
@@ -115,6 +118,35 @@ public ReflectionMappers setCaseChange(UnaryOperator<String> caseChange) {
...
+    public ReflectionMappers disableAccessibleObjectStrategy() {
+        this.makeAccessible = accessibleObject -> {};
+        return this;
+    }
+
+    /**
+     * Use the accessibility strategy to potentially make a reflective operation accessible.
+     */
+    @Beta
+    public <T extends AccessibleObject> T makeAccessible(T accessibleObject) {
```
hgschmie, line 145: I always felt that the config classes should be config. Now they actually contain logic.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
hgschmie: this introduces a bug. If someone *really* bothers to register a strategy, then after the first copy operation (which calls `new ReflectionMappers(oldReflectionMappers)`, the strategy will be reset to the default because the field was not added to the copy constructor.

core/src/test/java/org/jdbi/v3/core/mapper/reflect/FieldMapperPrivateMockTest.java
```diff
@@ -0,0 +1,211 @@
...
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
```
hgschmie, line 41: There are straightforward ways to use the db framework and "withSomething" or "users()" from the testing initializers to get known data into the database. We can not talk about "please spin up a database for your test" and then add tests that use mocks when it is neither needed nor useful.

hgschmie (changes requested):

core/src/test/java/org/jdbi/v3/core/mapper/reflect/FieldMapperPrivateMockTest.java
```diff
@@ -0,0 +1,211 @@
...
+        when(resultSet.getLong(1)).thenReturn(aLongVal);
+        when(resultSet.wasNull()).thenReturn(false);
+
+        SamplePrivateBean sampleBean = mapper.map(resultSet, ctx);
+
+        assertThat(sampleBean.getLongField()).isEqualTo(aLongVal);
+    }
+
+    @Test
+    public void shouldHandleEmptyResult() throws Exception {
```
hgschmie, line 83: ironically, this test makes no sense in real life. This sets up the ResultSet to contain zero columns, which, from what I can see, is not possible in actual SQL. - https://stackoverflow.com/a/2969590/4401166

Another reason why mock tests are bad.

hgschmie: and before suggesting using a join mapper where no column maps onto the FieldMapper: There is an explicit check to throw an exception if any column exists in the result set and none maps onto the fieldmapper. This may actually be a bug / limitation.

hgschmie (approved):

PR #2207, opened 2022-11-23
Performance fix for StatementContext#close

hgschmie:
Skip most of the execution if there are no cleanables.


stevenschlansker:
Is it ever actually possible to benefit from this improvement? It seems that any statement context that ever executed a query or produced any result would have at least one cleanable.

PR #2209, opened 2022-11-23
Review the resource management for ResultSetScanner and friends

hgschmie:
Review the terminal operations on ResultBearing, fix a few small problems and a resource leak with the stream() method.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/result/ResultBearing.java
+++ b/core/src/main/java/org/jdbi/v3/core/result/ResultBearing.java
@@ -224,19 +224,18 @@ public interface ResultBearing {
      * @see RowReducer
      */
     default <C, R> Stream<R> reduceRows(RowReducer<C, R> rowReducer) {
-        return scanResultSet((resultSetSupplier, ctx) -> {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/jdbi/v3/core/result/ResultBearing.java
```diff
@@ -224,19 +224,18 @@ default <T> ResultIterable<T> map(RowViewMapper<T> mapper) {
...
+            try (StatementContext ctx = context) {
+                ResultSet resultSet = resultSetSupplier.get();
                 RowView rowView = new RowViewImpl(resultSet, ctx);
 
                 C container = rowReducer.container();
                 while (resultSet.next()) {
                     rowReducer.accumulate(container, rowView);
                 }
-                return rowReducer.stream(container);
+                return rowReducer.stream(container).onClose(ctx::close);
```
stevenschlansker, line 236: Isn't this `onClose(ctx::close)` redundant with the `try (ctx) { ... }` resource management? The context will be closed as if in a finally block here.

PR #2223, opened 2022-11-30
Start a new, java 17+ module

hgschmie:
add some tests for records

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -91,4 +91,16 @@
             </dependency>
         </dependencies>
     </dependencyManagement>
+
+    <profiles>
+        <profile>
+            <id>java17</id>
+            <activation>
+                <jdk>[17,)</jdk>
+            </activation>
+            <modules>
+                <module>java17</module>
+            </modules>
+        </profile>
+    </profiles>
 </project>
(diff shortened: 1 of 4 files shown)


hgschmie:
this is intended to go in after 3.35.0 release

pom.xml
```diff
@@ -90,4 +90,16 @@
...
     </dependencyManagement>
+
+    <profiles>
+        <profile>
+            <id>jdk17_plus</id>
+            <activation>
+                <jdk>[17,)</jdk>
+            </activation>
+            <modules>
+                <module>modern</module>
```
stevenschlansker, line 101: I am not sure how many fancy new compiler features Jdbi will want to adopt, but it might be good to name the profiles after the JDK version they target. We might want a jdk17 module, and later want to add a jdk20 module, but the name `modern` can only be used once.

hgschmie: ok, makes sense. Right now I am using this for JDK 17+ tests (such as records).

stevenschlansker: Oh, if it will only be tests, it probably doesn't matter.

hgschmie: right now there are only tests. It may eventually evolve into actual java 17 code. I renamed it "java17" and made it internal for now.

PR #2233, opened 2022-12-13
Add support for jdbc databases in testcontainers

hgschmie:
Let's face it: The Jdbi developers are PostgreSQL fanboys. Our tests run on PostgreSQL, we have three flavors of PostgreSQL testing and in general, if we have to grab a non-in-memory database, we use ... PostgreSQL. The only exception so far is ... Oracle of all things.

This PR adds full support to all databases that are provided by the testcontainers.org project and offers similar convenience (running multiple test methods on the same database instance) as the Postgres based test framework offers.

It supports (and runs test for!) MySQL, MariaDB, TiDB, PostgreSQL (with sub-flavors such as PostGIS and TimescaleDB), CockroachDB, YugabyteDB, ClickHouse, Oracle XE, and TrinoDB out of the box and any other testcontainer supported database can easily be configured and used.

### DIFF
--- a/testcontainers/src/test/java/org/jdbi/v3/testing/junit5/tc/TrinoJdbiTestContainersExtensionTest.java
+++ b/testcontainers/src/test/java/org/jdbi/v3/testing/junit5/tc/TrinoJdbiTestContainersExtensionTest.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.testing.junit5.tc;
+
+
+import org.junit.jupiter.api.BeforeEach;
+import org.testcontainers.containers.JdbcDatabaseContainer;
+import org.testcontainers.containers.TrinoContainer;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+@Testcontainers
+class TrinoJdbiTestContainersExtensionTest extends AbstractJdbiTestcontainersExtensionTest {
+
+    @Container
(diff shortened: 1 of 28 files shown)


testcontainers/src/test/java/org/jdbi/v3/testing/junit5/tc/TrinoJdbiTestContainersExtensionTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    JdbcDatabaseContainer<?> getDbContainer() {
+        return dbContainer;
+    }
+
+    @BeforeEach
+    public void beforeEach() throws Exception {
+        // we still love you, trino. But you are a mess.
+        Thread.sleep(5_000);
```
stevenschlansker, line 37: Yuck.

stevenschlansker (approved):
I'm still not in love with using containers for everything, but it does let us test against other databases. I do not see many examples of actual tests that test other databases' features here, though.

What is the effect on build time? I would be somewhat worried that having so many database engines slows down all future PRs, or introduces random failures that we then have to diagnose.  "sleep 5" is not a reliable way to wait for a resource to become available.

testcontainers/src/test/java/org/jdbi/v3/testing/junit5/tc/TrinoJdbiTestContainersExtensionTest.java
hgschmie: effect on local build times is zero. All the tests are turned off locally unless you pass `-Dbasepom.it.skip=false`. Running `make test` does that, but `mvn clean install` does not.

I don't want to spend many minutes locally to run tests either.

hgschmie:
Primary goal for this PR is to widen the tent to others using local database testing. It opens a path to more modules like jdbi3-postgres or jdbi3-sqlite by allowing to run tests against the actual databases; something we could not do before.

testcontainers/src/test/java/org/jdbi/v3/testing/junit5/tc/TrinoJdbiTestContainersExtensionTest.java
hgschmie: the "sleep 5" is a bug workaround, when the trino container comes up, it waits for a query to succeed, but it actually calls the container "ready" when trino accepts the query and I had the test repeatedly fail with "no nodes available" because the discovery logic inside trino had not fully converged yet. The sleep ensures that the trino "cluster" is actually ready. This is a clash between the testcontainer setup and the trino engine.

hgschmie: see https://github.com/testcontainers/testcontainers-java/issues/6310

PR #2235, opened 2022-12-17
Some build tweaks

hgschmie:
- use -fn (fail never) instead of -fae (builds all the way through)
- use -Pfast
- add missing invoker:install to make tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
     - name: build distribution
       env:
         JAVA_HOME: ${{ steps.build_jdk.outputs.path }}
-        JDBI_MAVEN_OPTS: "-DskipTests -DskipITs -Dbasepom.check.skip-all=true -Dbasepom.check.skip-enforcer=false -B -fae"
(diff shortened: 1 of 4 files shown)


stevenschlansker:
Why would we want to "fail never"? It is important for the build to fail if there is an error.

hgschmie:
it will still not build any modules that are in error or their descendants. but it will build all the modules that are not affected by a specific error so it will give us a wider footprint of things built. 

There is a longer debate here: https://stackoverflow.com/questions/4174696/making-maven-run-all-tests-even-when-some-fail

Main difference is that if module B depends on module A, and module A has failing tests, then module B is not built or tested with -fae but with -fn. There are some pros and cons to this.

stevenschlansker:
One of the comments:

> I just tested the option --fail-never. Maven will not fail the build even if there are compile errors. If I use this option on Jenkins, the build looks successful even if it has lots of compile errors. 

This is exactly what I am concerned about. This sounds like a bad thing.

hgschmie:
I removed all the `-fn` and kept the other changes

PR #2237, opened 2022-12-20
Make PojoTypes thread-safe

### LINKED ISSUE #2236: Potential race-condition in PojoTypes
**JDBI Version:** 3.30.0

[PojoTypes](https://github.com/jdbi/jdbi/blob/master/core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoTypes.java) contains an internal map of factories which is copied in a number of situations, such as `Jdbi.open()`, or in my case when an on demand extension is used. This map is also modified by calls to `register`.

Today, for the first time that I've noticed, calling a `@SqlBatch` annotated method on an on-demand extension threw a mysterious `ConcurrentModificationException`, presumably because a factory was being registered at the same time that `PojoTypes.factory` was being copied (stacktrace below)

The good news is that subsequent calls appear to work just fine.

I haven't yet tried to repro this because I have a feeling it will be very tricky, and I wanted some validation from people more familiar with the guts of JDBI before embarking on that road.

Alternatively, since this error implies `PojoTypes` is not thread-safe, I would be happy to write up a PR making it thread-safe without repro'ing if we like the risk-to-reward ratio.

---

```
java.util.ConcurrentModificationException: null
at java.base/java.util.HashMap$HashIterator.nextNode(Unknown Source)
at java.base/java.util.HashMap$EntryIterator.next(Unknown Source)
at java.base/java.util.HashMap$EntryIterator.next(Unknown Source)
at java.base/java.util.HashMap.putMapEntries(Unknown Source)
at java.base/java.util.HashMap.putAll(Unknown Source)
at org.jdbi.v3.core.mapper.reflect.internal.PojoTypes.<init>(PojoTypes.java:32)
at org.jdbi.v3.core.mapper.reflect.internal.PojoTypes.createCopy(PojoTypes.java:52)
at org.jdbi.v3.core.mapper.reflect.internal.PojoTypes.createCopy(PojoTypes.java:25)
at org.jdbi.v3.core.config.ConfigRegistry.lambda$new$0(ConfigRegistry.java:58)
at java.base/java.util.concurrent.ConcurrentHashMap.forEach(Unknown Source)
at org.jdbi.v3.core.config.ConfigRegistry.<init>(ConfigRegistry.java:57)
at org.jdbi.v3.core.config.ConfigRegistry.createCopy(ConfigRegistry.java:121)
at org.jdbi.v3.sqlobject.SqlObjectFactory.attach(SqlObjectFactory.java:97)
at org.jdbi.v3.core.extension.Extensions.lambda$findFor$0(Extensions.java:75)
at java.base/java.util.Optional.map(Unknown Source)
at org.jdbi.v3.core.extension.Extensions.findFor(Extensions.java:75)
at org.jdbi.v3.core.Jdbi.callWithExtension(Jdbi.java:488)
at org.jdbi.v3.core.Jdbi.withExtension(Jdbi.java:478)
at org.jdbi.v3.core.internal.OnDemandExtensions.lambda$createProxy$3(OnDemandExtensions.java:82)
at jdk.proxy3/jdk.proxy3.$Proxy172.insert(Unknown Source)

... My Code ....
jdbi.onDemand(Dao.class).insert(...)
```

If I'm reading the following trace right, it implies that [`other.factories`](https://github.com/jdbi/jdbi/blob/de7883067b9c2dcc06ebe5f4e91377e6521557b5/core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoTypes.java#L32) is being called concurrently - presumably because of a call to `pojoTypes.register`. I _believe_ the bulk of my calls come from [`JdbiImmutables.registerImmutable`](https://github.com/jdbi/jdbi/blob/de7883067b9c2dcc06ebe5f4e91377e6521557b5/core/src/main/java/org/jdbi/v3/core/mapper/immutables/JdbiImmutables.java#L54) and my framework _ought_ to be making these calls during startup only, but I saw this exception regardless.

npetryk:
Fixes #2236

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoTypes.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoTypes.java
@@ -29,7 +29,9 @@ public class PojoTypes implements JdbiConfig<PojoTypes> {
     public PojoTypes() {}
 
     private PojoTypes(PojoTypes other) {
-        factories.putAll(other.factories);
+        synchronized (other.factories) {
+            factories.putAll(other.factories);
+        }
     }
 
     @Override
@@ -38,13 +40,17 @@ public class PojoTypes implements JdbiConfig<PojoTypes> {
     }
 
     public PojoTypes register(Class<?> key, PojoPropertiesFactory factory) {
-        factories.put(key, factory);
+        synchronized (factories) {
+            factories.put(key, factory);
+        }
         return this;
     }
 
     public Optional<PojoProperties<?>> findFor(Type type) {
-        return Optional.ofNullable(factories.get(GenericTypes.getErasedType(type)))
+        synchronized (factories) {
+            return Optional.ofNullable(factories.get(GenericTypes.getErasedType(type)))
                 .map(ppf -> ppf.create(type, registry));
+        }
     }
 
     @Override
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoTypes.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoTypes.java
@@ -14,7 +14,7 @@
 package org.jdbi.v3.core.mapper.reflect.internal;
 
 import java.lang.reflect.Type;
-import java.util.HashMap;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.Map;
 import java.util.Optional;
 
@@ -23,15 +23,13 @@ import org.jdbi.v3.core.config.JdbiConfig;
 import org.jdbi.v3.core.generic.GenericTypes;
 
 public class PojoTypes implements JdbiConfig<PojoTypes> {
-    private final Map<Class<?>, PojoPropertiesFactory> factories = new HashMap<>();
+    private final Map<Class<?>, PojoPropertiesFactory> factories = new ConcurrentHashMap<>();
     private ConfigRegistry registry;
 
     public PojoTypes() {}
 
     private PojoTypes(PojoTypes other) {
-        synchronized (other.factories) {
-            factories.putAll(other.factories);
-        }
+        factories.putAll(other.factories);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoTypes.java
```diff
@@ -29,7 +29,9 @@ public class PojoTypes implements JdbiConfig<PojoTypes> {
     public PojoTypes() {}
 
     private PojoTypes(PojoTypes other) {
-        factories.putAll(other.factories);
+        synchronized (other.factories) {
```
hgschmie, line 32: e.g. PostgresTypes uses a ConcurrentHashMap, which is (to me) the preferred approach over adding synchronized blocks.

hgschmie (approved):

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/PojoTypes.java
```diff
@@ -14,7 +14,7 @@
 package org.jdbi.v3.core.mapper.reflect.internal;
 
 import java.lang.reflect.Type;
-import java.util.HashMap;
+import java.util.concurrent.ConcurrentHashMap;
```
hgschmie, line 17: order is wrong, thus fails the build. please run the build locally (`make install` in the root folder).

hgschmie (changes requested):
fails static code checkers

stevenschlansker (approved):
I don't think we can reliable add tests for race conditions. This looks fine to me.

hgschmie:
yes, it is. I would have approved the changes but the build is failing.

npetryk:
Thanks for y'all's patience. Got `make install` working locally

(Incidentally, I had to run `mvn license:format` to get around a `Some files do not have the expected license header` error 🤷 )

npetryk:
We good to go @hgschmie ?

Issue #2238, opened 2022-12-20
Review other JdbiConfig classes for thread safety


hgschmie:
In the spirit of #2236, review

- ReflectionMappers
- SqlStatements (template cache)

sman-81:
The `SqlStatements.templateCache` is of type `com.github.benmanes.caffeine.cache.BoundedLocalCache` which according to its Javadoc is "an in-memory cache implementation that supports full concurrency of retrievals, a high expected concurrency for updates, and multiple ways to bound the cache." This default cache should do.
However there is public setter `SqlStatements.setTemplateCache(Caffeine<Object, Object>)` through which a non-threadsafe cache could be set. This setter could either be removed (still `@Beta`) or documented in such way that the parameter must be a thread-safe cache instance.

sman-81:
In `ReflectionMappers` the member `columnNameMatchers` is not currently thread-safe.
The list reference as well as list contents change when the public setter `setColumnNameMatchers` is invoked.
The getter `getColumnNameMatchers` is public, and callers likely need to iterate the list (as done in `ConstructorMapper` and `FieldMapper`) thus risking `ConcurrentModificationException`.
Is the public setter `setColumnNameMatchers` needed? Unfortunately it is part of the public API.

I'd make `columnNameMatchers` final and call `clear` and `replaceAll` in the setter (with synchronization on the list). The public getter should return a copy (which may be modified by the caller. Javadoc would need to be updated.

hgschmie:
The setTemplateCache takes a `Caffeine` object which is final. It is not possible to set a different implementation than the output value of `Caffeine.build()` unless you manually crack open the class and change the implementation. At that point, we take no guarantees. So `SqlStatements` is threadsafe (we would have had tons of bug reports if it weren't as this is a very hot path in the SqlStatement class).

hgschmie:
In ReflectionMappers
- the c'tor creates an internal Arrays.ArrayList which does not support add and remove and the iterator does not support remove
- the copy c'tor creates a copy of the passed in list, so you can not force a race by keeping a reference of the list passed in
- setColumnNameMappers creates a copy of the passed in list, see above
- getColumnNameMappers wraps the internal list into Collections.unmodifiableList which does not support any manipulation of the underlying list (add, remove, iterator.remove)

I don't see how that member can be used in a thread unsafe way. I would like to see a failing test case before making changes.

hgschmie:
[...]

PR #2239, opened 2022-12-21
Review JdbiConfig classes ReflectionMappers and SqlStatements for thread safety (#2238)

sman-81:


core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
```diff
@@ -44,9 +42,9 @@ public class ReflectionMappers implements JdbiConfig<ReflectionMappers> {
      * snake_case matching for names.
      */
     public ReflectionMappers() {
-        columnNameMatchers = Arrays.asList(
```
hgschmie, line 47: I am not sure these change is needed at all. The list is never modified and the only place where it is copied is in the c'tors which are by definition single-threaded.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
```diff
@@ -60,21 +58,25 @@ private ReflectionMappers(ReflectionMappers that) {
...
 
     /**
-     * Returns the registered column name mappers.
+     * Returns the registered column name mappers in a new list
+     * that may be iterated and modified by the caller.
      *
      * @return the registered column name mappers.
      */
     public List<ColumnNameMatcher> getColumnNameMatchers() {
-        return Collections.unmodifiableList(columnNameMatchers);
```
hgschmie, line 68: the unmodifiable list is IMHO the better change because it makes it impossible for the caller to modify the returned list.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
```diff
@@ -60,21 +58,25 @@ private ReflectionMappers(ReflectionMappers that) {
...
     /**
-     * Replace all column name matchers with the given list.
+     * Replaces all column name matchers with the given list.
      * @param columnNameMatchers the column name matchers to use
      * @return this
      */
     public ReflectionMappers setColumnNameMatchers(List<ColumnNameMatcher> columnNameMatchers) {
-        this.columnNameMatchers = new ArrayList<>(columnNameMatchers);
+        synchronized (this.columnNameMatchers) {
+            this.columnNameMatchers.clear();
```
hgschmie, line 77: this changes the semantics from "set the new value, discard the old" to the actual possible race of "clear the object, then refill it". That is why the synchronized block is needed. Which in turn will cause some of the checkers to whine ("partial synchronized access" or so).

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -158,7 +158,8 @@ public SqlStatements setTemplateEngine(TemplateEngine templateEngine) {
     }
 
     /**
-     * Sets the Caffeine cache used to avoid repeatedly parsing SQL statements.
+     * Sets the Caffeine cache used to avoid repeatedly parsing SQL statements.<br>
```
hgschmie, line 161: yeah, but ... is it? :-)

hgschmie (changes requested):
Not sure any of those changes are actually needed. The code may be gnarly but I think it is threadsafe.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
sman-81: I left two comments on the original ticket #2238 to explain.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
sman-81: The two constructors created two different `ArrayList`s with different behaviour.
`java.util.ArrayList != java.util.Arrays.ArrayList`
I prefer if such subtle differences are avoided.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
sman-81: Where is there "partial synchronized access"?
The member is written only here and in the constructor. The outside world only gets copies.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
sman-81: I added another sentence and for nicer formatting the `<br>` in between.
The concurrency behaviour of SqlStatements depends on the type of cache which can be set from outside therefore thread-safety is up to the user.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
hgschmie: what is the different behavior? Where does it manifest to the end user. We give zero guarantees besides "you can get a list from this class". I would prefer us using `List.of` as those are explicitly called unmodifiable. Alas, that is Java 9...

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
hgschmie: They can not. For concurrent modification exception, it would need to be possible to modify the internal list object. You can not do that.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
hgschmie: Yes, and creating the copy needs to be synchronized, otherwise you access it here in a synchronized block and when you create the copy, you do not.  "partial synchronized access"

hgschmie:
I think these changes are not necessary and don't change anything. Unless there is a test that proves that there is really a way to force a ConcurrentModificationException (and that is entirely possible; I am not that smart after all), I think this is just change for the sake of making that one field final which actually introduces problems. I like the atomic swap that is there today and I think we should keep it.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
sman-81: Much prefer using one way of creating an array list instead of two: avoid one import and improved readability
I would also prefer `List.of` ...
https://www.baeldung.com/java-arrays-aslist-vs-new-arraylist

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ReflectionMappers.java
sman-81: You're right.

PR #2241, opened 2023-01-02
Rename JdbiCache to ConfigCache

hgschmie:
The generic named class "JdbiCache" is actually a very special cache class that is used to cache specific configuration information and configuration metadata. Rename the interface and its related config class to "ConfigCache" and "ConfigCaches".

This is user visible as the "JdbiCaches" element is technically part of the public API. However it is marked as @Beta so we could rename it.

"JdbiCache" is never used in a public API except on the methods of the "JdbiCaches" method.


hgschmie:
I want to rename this to allow creating a more generic "Cache" thing in Jdbi to get rid of the caffeine dependency. @stevenschlansker  @sman-81  please weigh in.

PR #2249, opened 2023-01-10
Allow null in Optional or UnableToCreateStatementException cannnot be thrown

sman-81:


core/src/main/java/org/jdbi/v3/core/argument/AbstractArgumentFactory.java
```diff
@@ -95,12 +95,16 @@ public final Optional<Argument> build(Type type, Object value, ConfigRegistry co
...
+        if (!arg.isPresent()) {
+            throw new UnableToCreateStatementException("Prepared argument " + value + " of type " + type + " failed to build");
+        }
+        return arg;
     }
 
     @SuppressWarnings("unchecked")
     private Optional<Argument> innerBuild(Object value, ConfigRegistry config) {
-        return Optional.of(value == null
+        return Optional.ofNullable(value == null
```
hgschmie, line 107: wouldn't

```java
...
    Argument arg = (value == null ? new NullArgument(sqlType) : build ((T) value, config));
    if (arg == null) {
        throw new UnableToCreateStatementException("Prepared argument " + value + " of type " + type + " failed to build");
    }
   return Optional.of(arg);
}
```

and leave the other part unchanged be simpler? The innerBuild is called in multiple places and the check above only covers one place? Worst case, you need to pass the type argument into innerBuild to make the error message work.

hgschmie (changes requested):
I think this change only covers half of the innerBuild call sites and it changes the behavior (now it can return an optional that is empty, something that was not possible before).

Tests would be nice.

sman-81:
@hgschmie thanks for your comment.

`innerBuild` is called twice. Previously `innerBuild` returned an `Optional` that could never be empty (`Optional.of`). Therefore the first call site `innerBuild(value, config).orElseThrow(..)` can never throw `UnableToCreateStatementException`. This is a bug.

I let `innerBuild` return an `Optional.ofNullable`, so it can be empty and the exception can fire as intended by the original author of the code.
This change requires that `build(Type, Object, ConfigRegistry)` is updated to not change its behaviour. I chose an UnableToCreateStatementException over NullPointerException to resemble the behaviour of `prepare`.

A quick word why I found this: I find `Optional.of` to be misnomer and frequent source of bugs. An Optional should allow null really. But it only does so by calling the non-intuitive factory `Optional.ofNullable`.

hgschmie:
That whole thing is gnarly. 

- if we are cool with `build((T) value, config)` returning null, then the change to use `Optional.ofNullable` is fine, but then we should lose the `if (!arg.isPresent()) { throw new UnableToCreateStatementException(...` error. Because, we are fine with build returning null and therefore the `build(Type type, Object value, ConfigRegistry config)` method returning Optional.empty is ok as well.
- if we are *not* cool with `build((T) value, config)` returning null, *then* we should error out right at the place where that method returns. which is inside `innerBuild`. 

The `Optional.of(value -> innerBuild().orElseThrow())` is IMHO a red herring and actually an artifact of someone wanting to write code in a functional style and then getting lost in the intricacies of unwinding the various Optionals. Well, Java is not Haskell after all. 

a) we should not be cool with `build((T) value, config)` returning null. That means, we do not have an argument to bind the value here. So that is an error. So we should error out in the inner build.
b) erroring out in the innerBuild means that the method will *always* return an Argument or throw an exception. So we don't need an optional to return because there is always a value.
c) if innerBuild always returns a value, there is no need anymore for `orElseThrow`

I would argue that this should look like this:

```diff
diff --git a/core/src/main/java/org/jdbi/v3/core/argument/AbstractArgumentFactory.java b/core/src/main/java/org/jdbi/v3/core/argument/AbstractArgumentFactory.java
index 0acf2ad52..d4f69e8b4 100644
--- a/core/src/main/java/org/jdbi/v3/core/argument/AbstractArgumentFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/AbstractArgumentFactory.java
@@ -80,8 +80,7 @@ public abstract class AbstractArgumentFactory<T> implements ArgumentFactory.Prep
     @Override
     public Optional<Function<Object, Argument>> prepare(Type type, ConfigRegistry config) {
         return isInstance.test(type, null)
-                ? Optional.of(value -> innerBuild(value, config)
-                        .orElseThrow(() -> new UnableToCreateStatementException("Prepared argument " + value + " of type " + type + " failed to bind")))
+                ? Optional.of(value -> innerBuild(type, value, config))
                 : Optional.empty();
     }

@@ -92,17 +91,21 @@ public abstract class AbstractArgumentFactory<T> implements ArgumentFactory.Prep

     @Override
     public final Optional<Argument> build(Type type, Object value, ConfigRegistry config) {
-        if (!isInstance.test(type, value)) {
-            return Optional.empty();
-        }
-        return innerBuild(value, config);
+        return isInstance.test(type, value)
+                ? Optional.of(innerBuild(type, value, config))
+                : Optional.empty();
     }

     @SuppressWarnings("unchecked")
-    private Optional<Argument> innerBuild(Object value, ConfigRegistry config) {
-        return Optional.of(value == null
-                ? new NullArgument(sqlType)
-                : build((T) value, config));
+    private Argument innerBuild(Type type, Object value, ConfigRegistry config) {
+        if (value == null) {
+            return new NullArgument(sqlType);
+        }
+        Argument argument = build ((T) value, config);
+        if (argument == null) {
+            throw new UnableToCreateStatementException("Prepared argument " + value + " of type " + type + " failed to build");
+        }
+        return argument;
     }

     /**
```
  
and now the `prepare` and `build` methods align very nicely, which they are suppose to do (because one is just a curried version of the other even though we got the argument order wrong).

My rant for this post is   `build` calls `innerBuild` calls `build`. Seriously people what were you smoking? Can I have some?

hgschmie:
We should also update the documentation of the `protected abstract Argument build(T value, ConfigRegistry config);` method that it must not return null. We currently have 24 implementations (15 in code, 9 in tests) and none returns null. So this is possible for outside implementations (I guess that is how you stumbled upon this) but should not be allowed.

sman-81:
Hi @hgschmie I'm not even close to your level of insight, so it would be cool if you take over this issue (commit to this PR directly or close it and fix it in master, it's really up to you). Thanks!

hgschmie:
not sure about "the level of insight". :-) It is basically asking "what is that piece of code supposed to do" and then be willing to look at code and ask "why does it do that?". Once you lose the Optional from innerBuild, the structure becomes visible. 

Happy to push that change for review. Out of curiosity: How did you find this problem? Do you have a custom argument factory that returns null on the implemented build method?

sman-81:
> Out of curiosity: How did you find this problem? Do you have a custom argument factory that returns null on the implemented build method?

I reviewed usages of `Optional.of`. I find `Optional.of` to be a misnomer and frequent source of bugs. An Optional should allow null really. It only does so by calling the non-intuitive factory `Optional.ofNullable`.

sman-81:
Superseded by pr #2252

PR #2252, opened 2023-01-12
Clarify the logic for AbstractArgumentFactory.

hgschmie:
Document that the abstract build method must return an argument instance. Align the implementation of build() and prepare().

Addresses #2249


sman-81 (approved):
Looks good to me. It's cleaner and more understandable than before.

PR #2256, opened 2023-01-22
caffeine cache plugin using caffeine cache

hgschmie:
Using caffeine for the internal caches. Installing the caffeine module replaces the internal caches with a caffeine implementation.


cache/caffeine-cache/src/main/java/org/jdbi/v3/cache/caffeine/CaffeineCache.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final Cache<K, V> cache;
+
+    CaffeineCache(Caffeine<Object, Object> caffeine) {
+        this.cache = caffeine.build();
+    }
+
+    @Override
+    public V get(K key) {
+        throw new UnsupportedOperationException();
```
ben-manes, line 38: Shouldn’t this be calling cache.getIfPresent?

PR #2258, opened 2023-01-25
Add stream and iterator test suite

hgschmie:


hgschmie:
70ish test cases covering all combinations of streams, iterators, attach, onDemand and extensions.

PR #2260, opened 2023-01-26
Add null safety to PostGISCodec

tprebs:
If a column contained a null geometry the registered Geometry ColumnMapper (from PostGISCodec) would throw a NPE while trying to convert the null value to a byte array. 

If a null value is now processed by the ColumnMapper, a null value is returned


sman-81 (approved):

stevenschlansker (approved):
Thank you for the improvement!

Issue #2264, opened 2023-01-31
Update Caffeine Dependency Version


alkamo:
Caffeine had a [weird issue in version 3.0.3 that caused a problem with Gradle tar tasks](https://github.com/ben-manes/caffeine/issues/716). That was resolved in version 3.1.1, but JDBI is still using 3.0.3. JDBI should be updated to resolve this transient issue.

PR #2265, opened 2023-01-31
Updated Caffeine dependency version to eliminate Gradle conflict

alkamo:
https://github.com/jdbi/jdbi/issues/2264
* Updated Caffeine dependency version to eliminate Gradle conflict

### DIFF
--- a/internal/build/pom.xml
+++ b/internal/build/pom.xml
@@ -69,7 +69,7 @@
         <dep.antlr.version>4.10.1</dep.antlr.version>
         <dep.assertj-core.version>3.23.1</dep.assertj-core.version>
         <dep.assertj-guava.version>3.4.0</dep.assertj-guava.version>
-        <dep.caffeine.version>3.0.3</dep.caffeine.version>
+        <dep.caffeine.version>3.1.2</dep.caffeine.version>
         <dep.commons-text.version>1.10.0</dep.commons-text.version>
         <dep.derby.version>10.15.2.0</dep.derby.version>
         <dep.detekt.version>1.21.0-RC2</dep.detekt.version>


stevenschlansker:
Thanks!

stevenschlansker (approved):

hgschmie:
Hi @alkamo,

Thank you for opening an issue with the Jdbi project. 

Are you sure that this is needed? Wouldn't it be enough to version-manage the dependency in your project? We are trying to keep the lowest version possible as baseline and have a number of tests to ensure that we run with later versions as well.

Upgrading this version will ripple through all projects that use Jdbi and may cause conflicts down the line.

As this is a problem with gradle and the gradle build process, I would assume that adding a version requirement to your build file of caffeine 3.1.x should suffice. I am not sure that there is a need to change this version globally.

hgschmie (changes requested):
Let's make sure that a global change is justified before comitting this.

stevenschlansker:
There are correctness bugs that are fixed in the later releases. Also, caffeine is used by jdbi itself, so there is some incentive to keep it up to date. If a user uses the caffeine cache but doesn't otherwise use caffeine in their project, they should get the latest version without having to manage it themselves preferably.

alkamo:
> If a user uses the caffeine cache but doesn't otherwise use caffeine in their project, they should get the latest version without having to manage it themselves preferably.

Now I'm wondering if something else is going on here because I'm **not** using caffeine cache. I'm seeing caffeine pulled in as a dependency of _jdbi3-core_. Below is the dependency graph that Gradle reports.

Looking at the JDBI code, I can't see any reason why this would happen.
```

+--- org.jdbi:jdbi3-core:3.36.0
|    +--- org.slf4j:slf4j-api:1.7.36
|    +--- io.leangen.geantyref:geantyref:1.3.13
|    \--- com.github.ben-manes.caffeine:caffeine:3.0.3
|         +--- org.checkerframework:checker-qual:3.27.0
|         +--- com.google.errorprone:error_prone_annotations:2.16
|         +--- com.fasterxml.jackson:jackson-bom:2.14.1 (c)
|         \--- org.apache.httpcomponents:httpclient:4.5.13 (c)
```

alkamo:
I think I might have found the reason: the published POM for jdbi3-core (at mvnrepository.com) differs from the one found at  /core/pom.xml and does include caffeine. I'm not familiar enough with Maven builds to know why this is or what to do about it, but it seems like that is unintentional?

hgschmie:
the release pom for 3.36.0 at mvnrepo points at central, so it is the same as the central pom.
the 3.36.0 tagged release pom is the same as the one on the central repo.

The current head no longer contains caffeine in core, because the next (not yet released) version refactors the code to no longer require it. But that is independent of the 3.36.0 release and very intentional. The v3.36.0 tag contains the dependency.

The gradle dependency tree looks incorrect. core depends on caffeine in 3.36.0, but caffeine 3.0.3 only depends on checker framework and error prone (see https://repo1.maven.org/maven2/com/github/ben-manes/caffeine/caffeine/3.0.3/caffeine-3.0.3.pom) but the tree shows jackson-bom (!?!) of all things and a http client. 

I think it is most likely that the gradle module file (https://repo1.maven.org/maven2/com/github/ben-manes/caffeine/caffeine/3.0.3/caffeine-3.0.3.module) is interfering with this.

hgschmie:
can you try to simply version manage the dependency version to 3.1.2? I don't really understand how gradle resolves dependencies (and https://docs.gradle.org/current/userguide/dependency_capability_conflict.html#sub:selecting-between-candidates was not helpful), so this would be the equivalent of adding

```xml
<dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>com.github.ben-manes.caffeine</groupId>
      <artifactId>caffeine</artifactId>
      <version>3.1.2</version>
    </dependency>
  </dependencies>
</dependencyManagement>
```

to your gradle build. This should solve your problem.

hgschmie:
it might be something like

```groovy
dependencies {
    constraints {
        implementation('com.github.ben-manes.caffeine:caffeine:3.1.2') {
            because 'build problems'
        }
}
```

or so. Don't take my word for it.

alkamo:
> The gradle dependency tree looks incorrect. core depends on caffeine in 3.36.0, but caffeine 3.0.3 only depends on checker framework and error prone

You caught me. I had edited the dependency tree to show 3.0.3 to show the dependency from jdbi-3-core, not realizing that the caffeine dependencies had changed. The real one was from where I forced the upgrade to 3.1.2 and looked like this:

```
+--- org.jdbi:jdbi3-core:3.36.0
|    +--- org.slf4j:slf4j-api:1.7.36
|    +--- io.leangen.geantyref:geantyref:1.3.13
|    \--- com.github.ben-manes.caffeine:caffeine:3.0.3 -> 3.1.2
|         +--- org.checkerframework:checker-qual:3.27.0
|         +--- com.google.errorprone:error_prone_annotations:2.16
|         +--- com.fasterxml.jackson:jackson-bom:2.14.1 (c)
|         \--- org.apache.httpcomponents:httpclient:4.5.13 (c)
```

So, yes managing the dependency manually works. Since it looks like the next version of JDBI will remove that dependency altogether I will likely just wait to update until that version is released.

PR #2266, opened 2023-02-02
SqlObject Factory rewrite

hgschmie:
- factor all the Generator specific code into GeneratorSqlObjectFactory, a new ExtensionFactory that deals specifically with the generated classes.
- strip all code that is not specific to the SqlObject Proxy extensions out of SqlObjectFactory
- register both factories in SqlObjectModule, so no user visible changes
- remove the specific code for generated classes (initData and lookupMethod) out of the SqlObjectInitData, this is now in the generator factory. Leave deprecated stubs in place for old, non-compiled classes. Those stubs will go away
- move the threadlocal that holds the init data into the generator factory
- move all init code for SqlObjectInitData into the class itself
- move duplicate superTypes code also into JdbiClassUtils

Even if the work stops here, this is a big improvement to make the SqlObject code actually understandable. The code is not that complicated, it was just split in very complicated ways and hard to follow. The PR splits it into three major pieces:

- everything that deals with the Proxy SqlObjects
- everything that deals with the generated SqlObjects
- everything that initializes the SqlObjectInitData object and its cache

When this is done, it should be possible to address the whole `invokeInContext` stuff on the HandleSupplier, the plan is to move this out of the handle supplier and into a wrapper around the method invocations itself, which would then allow things like annotations on default methods. Still a work in progress.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
@@ -16,10 +16,14 @@ package org.jdbi.v3.core.internal;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static java.lang.String.format;
 
 public final class JdbiClassUtils {
+
     private JdbiClassUtils() {
         throw new UtilityClassException();
     }
@@ -37,6 +41,11 @@ public final class JdbiClassUtils {
         try {
             return klass.getMethod(methodName, parameterTypes);
         } catch (ReflectiveOperationException | SecurityException e) {
+            try {
+                return klass.getDeclaredMethod(methodName, parameterTypes);
+            } catch (ReflectiveOperationException | SecurityException e2) {
+                e.addSuppressed(e2);
+            }
             throw new IllegalStateException(format("can't find %s#%s%s", klass.getName(), methodName, Arrays.asList(parameterTypes)), e);
         }
     }
@@ -45,7 +54,11 @@ public final class JdbiClassUtils {
         try {
             return Optional.of(klass.getMethod(methodName, parameterTypes));
         } catch (ReflectiveOperationException | SecurityException ignored) {
-            return Optional.empty();
+            try {
+                return Optional.of(klass.getDeclaredMethod(methodName, parameterTypes));
+            } catch (ReflectiveOperationException | SecurityException ignored2) {
+                return Optional.empty();
+            }
         }
     }
 
@@ -55,5 +68,17 @@ public final class JdbiClassUtils {
         } catch (ReflectiveOperationException | SecurityException e) {
             throw new IllegalStateException(format("Unable to instantiate class %s", factoryClass), e);
         }
+    }
+
+    public static Stream<Class<?>> superTypes(Class<?> type) {
+        Class<?>[] interfaces = type.getInterfaces();
+        // collect into a set to deduplicate the classes found.
+        // this can happen if e.g. a extends b and both implement c.
+        Set<Class<?>> result = Stream.concat(
+                Arrays.stream(interfaces).flatMap(JdbiClassUtils::superTypes),
+                Arrays.stream(interfaces))
+                .collect(Collectors.toSet());
+
+        return result.stream();
     }
 }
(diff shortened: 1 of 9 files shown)


hgschmie:
@stevenschlansker, as a major user of the generator, this would really benefit from some real world testing. I don't think this PR broke the generator and the generated code and all the tests pass, but the test suite is pretty light when it comes to the generator.

core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
```diff
@@ -56,4 +67,11 @@ public static <T> T createInstance(Class<T> factoryClass) {
             throw new IllegalStateException(format("Unable to instantiate class %s", factoryClass), e);
         }
     }
+
+    public static Stream<Class<?>> superTypes(Class<?> type) {
+        Class<?>[] interfaces = type.getInterfaces();
```
stevenschlansker, line 72: Should this helper method emit duplicate super types? Like if A implements B,C and B implements C, you would emit C twice. (I realize this is not a new problem, but if we provide a helper, maybe we should specify it one way or the other.)

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
```diff
@@ -87,182 +64,29 @@ private boolean looksLikeSqlObject(Class<?> extensionType) {
...
+        final SqlObjectInitData sqlObjectInitData = SqlObjectInitData.lookup(extensionType, handleSupplier.getConfig());
+        final ConfigRegistry instanceConfig = sqlObjectInitData.configureInstance(handleSupplier.getConfig().createCopy());
+
         instanceConfig.get(Extensions.class).onCreateProxy();
 
         Map<Method, Supplier<InContextInvoker>> handlers = new HashMap<>();
         final Object proxy = Proxy.newProxyInstance(
                 extensionType.getClassLoader(),
-                new Class[] {extensionType},
+                new Class[]{extensionType},
```
stevenschlansker, line 85: the style I use would usually keep the space between `[] {`

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
```diff
@@ -121,6 +102,139 @@ public Supplier<InContextInvoker> lazyInvoker(Object target, Method method, Hand
...
+
+        // tostring can not be constant, different for every object
+        Handler toStringHandler = (target, args, handleSupplier) ->
+                "Jdbi sqlobject proxy for " + sqlObjectType.getName() + "@" + Integer.toHexString(target.hashCode());
+
+        addMethodHandler(handlerMap, toStringHandler, Object.class, "toString");
+        addMethodHandler(handlerMap, EQUALS_HANDLER, Object.class, "equals", Object.class);
+        addMethodHandler(handlerMap, HASHCODE_HANDLER, Object.class, "hashCode");
+
+        // this should probably have a check whether the sqlObjectType actually implements SqlObject
```
stevenschlansker, line 167: I don't think it is harmful to have extra handler methods here that you can't actually call.

stevenschlansker (approved):

sqlobject/src/main/java/org/jdbi/v3/sqlobject/internal/SqlObjectInitData.java
hgschmie: that comment has always been there.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
hgschmie: weird.

core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
hgschmie: I added a deduplication step.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/SqlObjectFactory.java
hgschmie: The answer seems to be "IntelliJ has a new checkbox and it was off" :-P

stevenschlansker:
The test suite is light, but does have pretty good coverage of a lot of different cases. Will let you know if I see any problems.

PR #2268, opened 2023-02-02
Factor out common code in HandleSupplier

hgschmie:
Starting to reduce the code sprawl around HandleSupplier. Remove the convoluted back and forth between the invokeInContext methods that no user can actually get right and the only implementations are in Jdbi anyway.


hgschmie:
This is the first in a number of smaller PRs that rework the convoluted extension / handlesupplier and ultimately sqlobject logic. I try to split it into smaller pieces that are digestible. Feedback wanted!

PR #2271, opened 2023-02-03
Refactor method lookup methods

hgschmie:
There is a random helper in Handlers, move that to a core class and use it across the codebase.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
@@ -13,7 +13,13 @@
  */
 package org.jdbi.v3.core.internal;
 
-public class JdbiClassUtils {
+import java.lang.reflect.Method;
+import java.util.Arrays;
+import java.util.Optional;
+
+import static java.lang.String.format;
+
+public final class JdbiClassUtils {
     private JdbiClassUtils() {
         throw new UtilityClassException();
     }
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
```diff
@@ -22,8 +28,32 @@ public static boolean isPresent(String klass) {
         try {
             Class.forName(klass);
             return true;
-        } catch (ReflectiveOperationException ignored) {
+        } catch (ReflectiveOperationException | SecurityException ignored) {
```
stevenschlansker, line 31: How can you get a `SecurityException` here? From my reading of the `Class.forName` javadoc, this does not seem possible.

core/src/main/java/org/jdbi/v3/core/internal/OnDemandExtensions.java
```diff
@@ -37,11 +37,11 @@ public class OnDemandExtensions implements JdbiConfig<OnDemandExtensions> {
...
-            EQUALS_METHOD = Object.class.getMethod("equals", Object.class);
-            HASHCODE_METHOD = Object.class.getMethod("hashCode");
-            TOSTRING_METHOD = Object.class.getMethod("toString");
-        } catch (NoSuchMethodException wat) {
-            throw new IllegalStateException("OnDemandExtensions initialization failed", wat);
+            EQUALS_METHOD = JdbiClassUtils.methodLookup(Object.class, "equals", Object.class);
+            HASHCODE_METHOD = JdbiClassUtils.methodLookup(Object.class, "hashCode");
+            TOSTRING_METHOD = JdbiClassUtils.methodLookup(Object.class, "toString");
+        } catch (Exception e) {
+            throw new ExceptionInInitializerError(e);
```
stevenschlansker, line 44: Is this catch necessary at all now? `methodLookup` declares no checked exceptions

core/src/test/java/org/jdbi/v3/core/extension/TestExtensionContext.java
```diff
@@ -59,8 +60,8 @@ public interface TestExtension {
         Handle getHandle();
 
         default void checkContextInDefaultMethod() throws Exception {
-            Method m = TestExtension.class.getMethod("checkContextInDefaultMethod");
-            // there is not extension method set in default methods. These methods
+            Method m = JdbiClassUtils.methodLookup(TestExtension.class, "checkContextInDefaultMethod");
```
stevenschlansker, line 63: This change seems to add another layer of abstraction to this code without really making a specific improvement.

stevenschlansker (approved):

core/src/main/java/org/jdbi/v3/core/internal/OnDemandExtensions.java
hgschmie: catch Exception also catches the runtime exception and so this throws ExceptionInInitializerError (which is the right thing to do in an initializer)

core/src/test/java/org/jdbi/v3/core/extension/TestExtensionContext.java
hgschmie: mechanic search-and-replace. :-)

core/src/main/java/org/jdbi/v3/core/internal/OnDemandExtensions.java
stevenschlansker: That wrapping is already done by the compiler if you just leave the catch out, pretty sure.

PR #2278, opened 2023-02-09
Rewrite the SqlObject code generator

hgschmie:
- make execution more robust, use init() method
- don't fail on IOException, makes it possible to debug generated code
- resolve static method fields at build time, no need for the thread local anymore
- omit object creation if arrays are empty

### DIFF
--- a/generator/src/main/java/org/jdbi/v3/generator/GenerateSqlObjectProcessor.java
+++ b/generator/src/main/java/org/jdbi/v3/generator/GenerateSqlObjectProcessor.java
@@ -16,7 +16,7 @@ package org.jdbi.v3.generator;
 import java.io.IOException;
 import java.io.Writer;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
+import java.lang.reflect.Type;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


hgschmie:
this is pretty much new code that writes files similar to the old sql object source files. It is best reviewed side-by-side, not unified (it is pretty unreadable unified). 

The code will no longer support implementations built with generators pre-3.38.0; this can be restored (it is the code in GeneratorSqlObjectFactory and SqlObjectInitData that would need to stay) if asked for. I would prefer a clean break but happy to do that in 3.39.0

the new classes are much smaller and still pass all the tests. It would be good to give it some real world testing, not just the few simple tests here.

hgschmie:
Main reason why I want a clean break is that I plan to change the InContextInvoker so that it can be pulled "up" to get the "invokeInContext" out of the handle supplier. That will *definitely* break backwards compatibility to classes generated with the old generator. I can hold off on that until past 3.38 but right now, that is  a hindrance. :-)

hgschmie:
@stevenschlansker @sman-81 planning to move forward with this. Review would still be appreciated, but this is open for a week now.

stevenschlansker (approved):

PR #2280, opened 2023-02-12
executor wrapper to use jdbi in async services

leblonk:
This comes from having seen jdbi used in async services. The common pattern is to use jdbi like this:

```java
CompletionStage<String> doSelect() {
  return CompletableFuture.supplyAsync(() -> {
      return jdbi.withHandle(h ->
        h.createQuery(....
    }, executor);
}
```

This turns out to be very error prone (especially with multiple contributors), because each call to the database requires this exact boilerplate to surround the call. Easy to forget to pass in the executor, exceptions have to be handled. etc. Some of these errors don't show in tests either, only in prod, because the database call is executed on the wrong threadpool, which only shows up under load.

Having a class like this JdbiExecutor to pass around, instead of Jdbi, eliminates most possibilities of using things wrong, which in turn eliminates risk. Same code now looks like:

```java
CompletionStage<String> doSelect() {
  return jdbiExecutor.withHandle(h ->
    h.createQuery(....
  );
}
```

Would like to add this as a standard feature of jdbi.

In the end, it's just a very thin wrapper on top of the Jdbi class, but it eliminates a class of errors in Jdbi usage in async services

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
+++ b/core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
@@ -1,0 +1,93 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.async;
+
+import java.util.concurrent.CompletionStage;
+
+import org.jdbi.v3.core.HandleCallback;
+import org.jdbi.v3.core.HandleConsumer;
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.extension.ExtensionCallback;
+import org.jdbi.v3.core.extension.ExtensionConsumer;
+import org.jdbi.v3.core.transaction.TransactionIsolationLevel;
+
+public abstract class AbstractJdbiExecutor implements JdbiExecutor {
+
+    /**
+     * Single method through which all other methods converge. If you want to override JdbiExecutorImpl, you only need to override this one method.
+     *
+     * @param handler the handler that takes a Jdbi instance and returns a value
+     * @param <R>     type returned by the callback
+     * @param <X>     exception type thrown by the callback, if any.
+     * @return a completion stage that will complete when the handler returns or throws an exception
+     */
+    protected abstract <R, X extends Exception> CompletionStage<R> doExecute(Handler<R, X> handler);
+
+    @Override
+    public <R, X extends Exception> CompletionStage<R> withHandle(final HandleCallback<R, X> callback) {
+        return doExecute(jdbi -> jdbi.withHandle(callback));
+    }
+
+    @Override
+    public <R, X extends Exception> CompletionStage<R> inTransaction(final HandleCallback<R, X> callback) {
+        return doExecute(jdbi -> jdbi.inTransaction(callback));
+    }
+
+    @Override
+    public <R, X extends Exception> CompletionStage<R> inTransaction(final TransactionIsolationLevel level, final HandleCallback<R, X> callback) {
+        return doExecute(jdbi -> jdbi.inTransaction(level, callback));
+    }
+
+    @Override
+    public <X extends Exception> CompletionStage<Void> useHandle(final HandleConsumer<X> consumer) {
+        return doExecute(jdbi -> {
+            jdbi.useHandle(consumer);
+            return null;
+        });
+    }
+
+    @Override
+    public <X extends Exception> CompletionStage<Void> useTransaction(final HandleConsumer<X> callback) {
+        return doExecute(jdbi -> {
+            jdbi.useTransaction(callback);
+            return null;
+        });
+    }
+
+    @Override
+    public <X extends Exception> CompletionStage<Void> useTransaction(final TransactionIsolationLevel level, final HandleConsumer<X> callback) {
+        return doExecute(jdbi -> {
+            jdbi.useTransaction(level, callback);
+            return null;
+        });
+    }
+
+    @Override
+    public <R, E, X extends Exception> CompletionStage<R> withExtension(final Class<E> extensionType, final ExtensionCallback<R, E, X> callback) {
+        return doExecute(jdbi -> jdbi.withExtension(extensionType, callback));
+    }
+
+    @Override
+    public <E, X extends Exception> CompletionStage<Void> useExtension(final Class<E> extensionType, final ExtensionConsumer<E, X> callback) {
+        return doExecute(jdbi -> {
+            jdbi.useExtension(extensionType, callback);
+            return null;
+        });
+    }
+
+    protected interface Handler<R, X extends Exception> {
+
+        R apply(Jdbi jdbi) throws X;
+    }
+}
--- a/core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
+++ b/core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
@@ -20,74 +20,72 @@ import org.jdbi.v3.core.HandleConsumer;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.core.extension.ExtensionCallback;
 import org.jdbi.v3.core.extension.ExtensionConsumer;
+import org.jdbi.v3.core.internal.exceptions.CheckedConsumer;
+import org.jdbi.v3.core.internal.exceptions.CheckedFunction;
 import org.jdbi.v3.core.transaction.TransactionIsolationLevel;
 
 public abstract class AbstractJdbiExecutor implements JdbiExecutor {
 
     /**
-     * Single method through which all other methods converge. If you want to override AbstractJdbiExecutor, you only need to override this one method.
+     * Single method through which all other with* methods converge. Since useExecute also calls this, any implementation of AbstractJdbiExecutor only needs to
+     * implement this method
      *
-     * @param handler the handler that takes a Jdbi instance and returns a value
-     * @param <R>     type returned by the callback
-     * @param <X>     exception type thrown by the callback, if any.
+     * @param callback the callback that takes a Jdbi instance and returns a value
+     * @param <T>      type returned by the callback
+     * @return a completion stage that will complete when the handler returns a value or throws an exception
+     */
+    protected abstract <T> CompletionStage<T> withExecute(CheckedFunction<Jdbi, T> callback);
+
+    /**
+     * Single method through which all other use* methods converge. This method calls {@link #withExecute(CheckedFunction)}
+     *
+     * @param callback the callback that takes a Jdbi instance
      * @return a completion stage that will complete when the handler returns or throws an exception
      */
-    protected abstract <R, X extends Exception> CompletionStage<R> doExecute(Handler<R, X> handler);
+    protected CompletionStage<Void> useExecute(CheckedConsumer<Jdbi> callback) {
+        return withExecute(jdbi -> {
+            callback.accept(jdbi);
+            return null;
+        });
+    }
 
     @Override
     public <R, X extends Exception> CompletionStage<R> withHandle(final HandleCallback<R, X> callback) {
-        return doExecute(jdbi -> jdbi.withHandle(callback));
+        return withExecute(jdbi -> jdbi.withHandle(callback));
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


hgschmie:
Hi @leblonk! Thank you for opening a PR with the JDBI team. 

I looked through the tests and you run those with a single threaded executor. I know there are some thread locals deep within Jdbi, so it would be good to try this change using an executor where the different stages are executed by multiple threads to see whether the state is preserved correctly.

The Handle is ultimately not a multi-thread safe object; it is reasonably safe if only one thread at a time uses it.

stevenschlansker:
It looks like each Handle is acquired, operated on, and then closed from within the same executor-executed action. So, as long as the user does not further pass the Handle off to a different thread, I don't think it should be a problem here.

core/src/main/java/org/jdbi/v3/core/async/JdbiExecutorImpl.java
```diff
@@ -0,0 +1,127 @@
...
+     * connection pool. The worker queue in the executor should probably be bounded, unless
+     * the caller(s) already has a bound for the number of outstanding requests. Making the
+     * queue bounded will mean you are blocking the calling thread when the queue fills up.
+     * Whether or not that is a
+     * </p>
+     *
+     * @param jdbi the initialized Jdbi class
+     * @param executor an executor to use for all database calls
+     */
+    public JdbiExecutorImpl(Jdbi jdbi, Executor executor) {
```
stevenschlansker, line 48: Rather than have a `Impl` class be public, it would be nice to have the `Impl` class be package-private and provide a static factory method like `JdbiExecutor.create(...)`

leblonk: The nice thing about the Impl is that it can be overridden by only implementing the one method and chain the JdbiExecutor (so you can for example have a JdbiExecutorMetrics). One alternative is to make all the methods in the JdbiExecutor interface defaults, and have the defaults implementation all redirect to the one doExecute method. Same end result, but then you can make the Impl package private without losing the capability to make an easy delegate. At the cost of exposing a bit more details in the interface.

Would look like:
```java
public interface JdbiExecutor {
    /**
     * Single delegate handler to which all other methods delegate
     * ...
     */
    <R, X extends Exception> CompletionStage<R> doExecute(final Handler<R, X> handler);

    /**
     * A convenience function which manages the lifecycle of a handle and yields it to a callback
     * for use by clients.
     * ...
     */
    default <R, X extends Exception> CompletionStage<R> withHandle(HandleCallback<R, X> callback) {
        return doExecute(jdbi -> jdbi.withHandle(callback));
    }
```

leblonk:
@hgschmie although it's unlikely, as @stevenschlansker pointed out, I added a test nonetheless, just in case.

core/src/main/java/org/jdbi/v3/core/async/JdbiExecutorImpl.java
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * Construct a {@link JdbiExecutor}.
+     *
+     * <p>
+     * The executor to pass in needs to be sized to the use case. A rule of thumb is to
+     * have the max number of workers be equal to the max number of connections in the
+     * connection pool. The worker queue in the executor should probably be bounded, unless
+     * the caller(s) already has a bound for the number of outstanding requests. Making the
+     * queue bounded will mean you are blocking the calling thread when the queue fills up.
+     * Whether or not that is a
```
stevenschlansker, line 42: Incomplete sentence

core/src/main/java/org/jdbi/v3/core/async/JdbiExecutorImpl.java
stevenschlansker: That is a fair point. Unless Henning has a strong opinion, I guess I will leave it up to you which style to prefer.

core/src/main/java/org/jdbi/v3/core/async/JdbiExecutor.java
```diff
@@ -0,0 +1,178 @@
...
+import org.jdbi.v3.core.transaction.TransactionIsolationLevel;
+
+public interface JdbiExecutor {
+
+    /**
+     * A convenience function which manages the lifecycle of a handle and yields it to a callback
+     * for use by clients.
+     *
+     * <p>
+     * The callback will be executed in a separate thread
```
stevenschlansker, line 32: Strictly speaking, this is not necessarily true. For example Guava has a "direct" executor which runs on the current thread, and there is the `ThreadPoolExecutor.CallerRunsPolicy` which does the same.

leblonk: Fair point. Changed it to say "will be executed in a thread supplied by the executor". That should cover these cases

core/src/main/java/org/jdbi/v3/core/async/JdbiExecutorImpl.java
leblonk: Ah, yes. Amended (and moved this comment block to the JdbiExecutor.create method)

core/src/main/java/org/jdbi/v3/core/async/JdbiExecutorImpl.java
leblonk: I've changed it to use a public abstract class AbstractJdbiExecutor that does the redirect to the single method and then a JdbiExecutor.create which instantiates a package private JdbiExecutorImpl which only implements that single method.

This way, it maintains the clean interface, but still allows easy chaining of the implementation.

core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+import org.jdbi.v3.core.HandleConsumer;
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.extension.ExtensionCallback;
+import org.jdbi.v3.core.extension.ExtensionConsumer;
+import org.jdbi.v3.core.transaction.TransactionIsolationLevel;
+
+public abstract class AbstractJdbiExecutor implements JdbiExecutor {
+
+    /**
+     * Single method through which all other methods converge. If you want to override JdbiExecutorImpl, you only need to override this one method.
```
stevenschlansker, line 28: JdbiExecutorImpl -> AbstractJdbiExecutor

stevenschlansker (approved):

core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
leblonk: Fixed

core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public <R, E, X extends Exception> CompletionStage<R> withExtension(final Class<E> extensionType, final ExtensionCallback<R, E, X> callback) {
+        return doExecute(jdbi -> jdbi.withExtension(extensionType, callback));
+    }
+
+    @Override
+    public <E, X extends Exception> CompletionStage<Void> useExtension(final Class<E> extensionType, final ExtensionConsumer<E, X> callback) {
+        return doExecute(jdbi -> {
```
hgschmie, line 83: for most of our code, we use two distinct interfaces (one would be the `JdbiCallback` below, the other would be

```
public interface JdbiConsumer<X extends Exception> {
    void useJdbi(Jdbi jdbi) throws X;
}
```

The second one should be used here, so this would fold into 

```
return doExecute(jdbi -> jdbi.useExtension(extensionType, callback));
```

core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+        });
+    }
+
+    @Override
+    public <R, E, X extends Exception> CompletionStage<R> withExtension(final Class<E> extensionType, final ExtensionCallback<R, E, X> callback) {
+        return doExecute(jdbi -> jdbi.withExtension(extensionType, callback));
+    }
+
+    @Override
+    public <E, X extends Exception> CompletionStage<Void> useExtension(final Class<E> extensionType, final ExtensionConsumer<E, X> callback) {
```
hgschmie, line 82: rename `callback` to `consumer`. The object is a consumer, not a callback.

core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+    public <X extends Exception> CompletionStage<Void> useTransaction(final HandleConsumer<X> callback) {
+        return doExecute(jdbi -> {
+            jdbi.useTransaction(callback);
+            return null;
+        });
+    }
+
+    @Override
+    public <X extends Exception> CompletionStage<Void> useTransaction(final TransactionIsolationLevel level, final HandleConsumer<X> callback) {
+        return doExecute(jdbi -> {
```
hgschmie, line 70: same as below.

core/src/main/java/org/jdbi/v3/core/async/JdbiExecutorImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Make sure to run the handler in a thread supplied by the executor
+     *
+     * @param handler the handler that takes a Jdbi instance and returns a value
+     * @param <R>     type returned by the callback
+     * @param <X>     exception type thrown by the callback, if any.
+     * @return a completion stage that will complete when the handler returns or throws an exception
+     */
+    @Override
+    protected <R, X extends Exception> CompletionStage<R> doExecute(final Handler<R, X> handler) {
```
hgschmie, line 48: similar to above, there should be one "use" and one "with" method, same naming pattern we use everywhere else.

hgschmie (changes requested):

hgschmie:
Generally looks good; some minor nits:

- Please mark everything as `@Alpha` or `@Beta`. I let you decide what is appropriate. But I want the option to change this without being locked into public API compatibility on the very first attempt
- We have too many "handle", "handler" or similar named classes and interfaces already. When we introduce this, it needs to use the same patterns and naming conventions as the rest of the code ("with" and "use")
- make the two callback interfaces public and not nested interfaces.

hgschmie:
an alternate route that @stevenschlansker and I briefly discussed was to abandon the custom handler completely and just use a `org.jdbi.v3.core.internal.exceptions.CheckedFunction`. That would work for me as well.

Again, thank you for contributing this. I think this will be useful and we should merge that PR.

core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public <E, X extends Exception> CompletionStage<Void> useExtension(final Class<E> extensionType, final ExtensionConsumer<E, X> callback) {
+        return doExecute(jdbi -> {
+            jdbi.useExtension(extensionType, callback);
+            return null;
+        });
+    }
+
+    protected interface Handler<R, X extends Exception> {
```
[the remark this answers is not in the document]
stevenschlansker: I thought the proposal was to re-use an existing type, like CheckedFunction.

If we are to introduce a new type, maybe we should wait for a second use case before introducing new public api for it.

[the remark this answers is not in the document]
leblonk: Changing it to use CheckedFunction. With that there isn't a need to introduce yet another class

core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
leblonk: I can certainly change that. It was based on the naming convention in Jdbi though, where these are all called callback?

core/src/main/java/org/jdbi/v3/core/async/JdbiExecutorImpl.java
leblonk: Changed it in AbstractJdbiExecutor to have a withExecute and useExecute method. But added an implementation to useExecute, so this impl still only has to implement one method. Can change it to 2 overrides needed if you think that will be more consistent

core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
leblonk: Changed to use withExecute and useExecute

leblonk:
Added the beta label (not alpha, since I have used similar for several years successfully)

core/src/main/java/org/jdbi/v3/core/async/AbstractJdbiExecutor.java
hgschmie: ok

hgschmie (approved):

hgschmie:
@leblonk Thank you, this looks really good now. I will be merging it once the CI passes. 

It would be good to get some docs for this as well so that users can take advantage of it.

PR #2284, opened 2023-02-26
JdbiExecutor documentation

leblonk:
Initial docs on JdbiExecutor.

I'm not sure if this is the best location for this (4.4 under 4. Core API Concepts), so please let me know if there's a better option. Also, should there be more expansion on why this is needed in async applications (I'm kind of presuming the reader of this chapter is familiar with async)

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -783,6 +783,60 @@ List<String> catalogNames = h.queryMetadata(DatabaseMetaData::getCatalogs)
 ----
 
 
+
+
+=== Usage in asynchronous applications
+
+Calling Jdbc and by extension Jdbi is inherently a blocking operation.
+In asynchronous applications
+(where results are returned as a link:https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html[CompletionStage^]
+or a link:https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html[CompletableFuture^])
+it is very important never to block on threads that the caller doesn't control. For this, there is the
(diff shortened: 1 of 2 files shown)


stevenschlansker (approved):
I am fine with it living in section 4. Other options would be section 20 (Advanced Topics) or it could even be its own short top level topic near the end.

Regardless of where it ends up, thank you for adding this excellent documentation 😸

leblonk:
Let's stick with section 4 then

docs/src/test/java/jdbi/doc/AsyncTest.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    private Jdbi jdbi;
+    private JdbiExecutor jdbiExecutor;
+    private Handle handle;
+
+    @BeforeEach
+    public void setUp() {
+        // H2 in-memory database
+        Jdbi jdbi = Jdbi.create("jdbc:h2:mem:test").installPlugin(new SqlObjectPlugin());
```
hgschmie, line 50: plz use the jdbi testing framework (see https://jdbi.org/#_junit_5)

docs/src/adoc/index.adoc
```diff
@@ -783,6 +783,60 @@ List<String> catalogNames = h.queryMetadata(DatabaseMetaData::getCatalogs)
 ----
 
 
+
+
+=== Usage in asynchronous applications
+
+Calling Jdbc and by extension Jdbi is inherently a blocking operation.
+In asynchronous applications
+(where results are returned as a link:https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html[CompletionStage^]
```
hgschmie, line 792: we have a macro for that: `link:{jdkdocs}/javax/sql/DataSource.html[DataSource^]`

PR #2287, opened 2023-03-02
Use jdkdocs and test extension

leblonk:
Addresses PR comments from previously merged PR.
- Use {jdkdocs} instead of full path
- Use h2Extension instead of jdbc string


hgschmie:
Thank you! I really appreciate you addressing the comments.

Issue #2293, opened 2023-03-13
Use attachToHandleForCleanup() everywhere


jodastephen:
The method `attachToHandleForCleanup()` has been added recently, and looks to be a good solution to always ensuring things are closed. I could be wrong, but it seems like it might be desirable to set `attachToHandleForCleanup()` always, however I can see no easy way to do this. ie. I can see no easy way at the framework-level, obviously I could update all the use sites, but adding try-with-resources is a fair bit more verbose. I don't believe `StatementCustomizer` can do the job here.

Some possible approaches I came up with:

* could a new method be added to `Handle` - `attachStatementsToHandleForCleanup()` that is used to auto-set the flag whenever a `BaseStatement` subclass is created - feels quite nice, and can be configured in `JdbiPlugin`
* could there be a callback to `JdbiPlugin` whenever `Handle` creates a `Query`/`Update`/`Call` etc. - not sure what else the callbacks would be used for
* could there be a built-in subclass of `Handle` that calls `attachToHandleForCleanup()` after creating the statement - doesn't feel like JDBI design style
* could JDBI add methods like `withQuery(String, QueryCallback)` to `Handle` - feels like overkill

Background info:
* We use JDBI as a Java API, no annotations/Spring etc.
* We have our own JDBI plugin that is always called, so ideally we'd like to set `attachToHandleForCleanup()` there
* We use Java 11 at present

Thanks for JDBI!

hgschmie:
Hi @jodastephen ,

Thank you for opening an issue with the Jdbi project! 

This is an interesting proposal. We could add a flag to `SqlStatements` to set this as a default.

hgschmie:
As a caveat: Adding every statement to a handle for cleanup means that resources may be held for a while (until the handle is actually closed). There is a chance that these resources "pile up", especially when many statements are executed.

hgschmie:
adding the flag as proposed should do the trick. If you have a factory that creates the Jdbi objects, it should just call `jdbi.getConfig(SqlStatements.class).setAttachAllStatementsForCleanup(true)`. Alternatively you can use a simple plugin:

```java
public class AttachHandlesPlugin extends JdbiPlugin.Singleton {
    @Override
    public void customizeJdbi(Jdbi jdbi) {
        jdbi.getConfig(SqlStatements.class).setAttachAllStatementsForCleanup(true);
    }
}
```

and then call `jdbi.installPlugin(new AttachHandlesPlugin());`

[...]

PR #2294, opened 2023-03-14
Adds `setAttachAllStatementsForCleanup` flag

hgschmie:
Allows all statements to be attached to their handle by default and cleaned up when the handle closes.

Addresses #2293.

@jodastephen would that work for your use case?


hgschmie:
@stevenschlansker any opinion?

stevenschlansker:
Seems reasonable enough.

PR #2304, opened 2023-03-22
StringTemplate engine: handle st4 errors, allow configuring missing attribute as error

### LINKED ISSUE #1556: Attribute isn't defined error
Hi everyone!
I'm doing a query with StringTemplate4 like this:
```kotlin
    @SqlUpdate("UPDATE users " +
            "<if(first_name)>first_name = :first_name,<endif>" +
            "<if(last_name)>last_name = :last_name,<endif>" +
            "updated_at = :now WHERE user_id=:id")
    @RegisterConstructorMapper(User::class)
    @DefineNamedBindings @UseStringTemplateEngine @Timestamped @GetGeneratedKeys
    fun update(@Bind("id") id: Int, @BindMap map: Map<String, Any>): User
```

All runs flawless, but if I don't send "first_name" key, I get this message in console:
`attribute first_name isn't defined`

Of course, I researched the reason and we need set a new [STErrorListener](https://github.com/antlr/stringtemplate4/blob/master/doc/listeners.md) to delete the message, inside this `STGroup`:
https://github.com/jdbi/jdbi/blob/f3173087f9fbfd3a9c722fe9218fafbea506decd/stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateEngine.java#L27

Here's any plan to include this configuration in JDBI?
Thanks!

stevenschlansker:
Fixes #1556

### DIFF
--- a/internal/build/pom.xml
+++ b/internal/build/pom.xml
@@ -450,7 +450,7 @@
             <dependency>
                 <groupId>org.antlr</groupId>
                 <artifactId>ST4</artifactId>
-                <version>4.3.1</version>
+                <version>4.3.4</version>
             </dependency>
 
             <dependency>
--- a/stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateEngine.java
+++ b/stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateEngine.java
@@ -15,19 +15,63 @@ package org.jdbi.v3.stringtemplate4;
 
 import org.jdbi.v3.core.statement.StatementContext;
 import org.jdbi.v3.core.statement.TemplateEngine;
+import org.jdbi.v3.core.statement.UnableToCreateStatementException;
+import org.jdbi.v3.core.statement.UnableToExecuteStatementException;
 import org.stringtemplate.v4.ST;
+import org.stringtemplate.v4.STErrorListener;
 import org.stringtemplate.v4.STGroup;
+import org.stringtemplate.v4.misc.STMessage;
 
 /**
  * Rewrites a StringTemplate template, using the attributes on the {@link StatementContext} as template parameters.
+ * For configuration, see {@link StringTemplates}.
  */
 public class StringTemplateEngine implements TemplateEngine {
     @Override
     public String render(String sql, StatementContext ctx) {
-        ST template = new ST(new STGroup(), sql);
+        STGroup group = new STGroup();
+        group.setListener(new ErrorListener(ctx));
+        ST template = new ST(group, sql);
 
         ctx.getAttributes().forEach(template::add);
 
         return template.render();
+    }
+
+    static class ErrorListener implements STErrorListener {
+        private final StatementContext ctx;
+
+        ErrorListener(StatementContext ctx) {
+            this.ctx = ctx;
+        }
+
+        @Override
+        public void compileTimeError(STMessage msg) {
+            throw new UnableToCreateStatementException("Compiling StringTemplate failed: " + msg, msg.cause, ctx);
+        }
+
+        @Override
+        public void runTimeError(STMessage msg) {
+            switch (msg.error) {
+                case NO_SUCH_PROPERTY:
+                    break;
+                case NO_SUCH_ATTRIBUTE:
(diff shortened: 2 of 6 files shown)


stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateEngine.java
```diff
@@ -15,19 +15,62 @@
...
+    static class ErrorListener implements STErrorListener {
+        private final StatementContext ctx;
+
+        ErrorListener(StatementContext ctx) {
+            this.ctx = ctx;
+        }
+
+        @Override
+        public void compileTimeError(STMessage msg) {
+            throw new UnableToCreateStatementException("Compiling StringTemplate failed: " + msg, msg.cause);
```
hgschmie, line 49: why not pass in statement context?

stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplateEngine.java
```diff
@@ -15,19 +15,62 @@
...
+        public void runTimeError(STMessage msg) {
+            switch (msg.error) {
+                case NO_SUCH_PROPERTY:
+                    break;
+                case NO_SUCH_ATTRIBUTE:
+                    if (!ctx.getConfig(StringTemplates.class).failOnMissingAttribute()) {
+                        break;
+                    }
+                default:
+                    throw new UnableToExecuteStatementException("Executing StringTemplate failed: " + msg, msg.cause);
```
hgschmie, line 62: why not pass in statement context?

core/src/main/java/org/jdbi/v3/core/statement/UnableToExecuteStatementException.java
```diff
@@ -31,6 +31,10 @@ public UnableToExecuteStatementException(String message, StatementContext ctx) {
         super(message, ctx);
     }
 
+    public UnableToExecuteStatementException(String message, Throwable throwable) {
```
hgschmie, line 34: this is not neeeded if you pass in the statement context below.

stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplates.java
```diff
@@ -0,0 +1,40 @@
...
+    private boolean failOnMissingAttribute = false;
+
+    public StringTemplates() {}
+
+    StringTemplates(StringTemplates other) {
+        this.failOnMissingAttribute = other.failOnMissingAttribute;
+    }
+
+    public boolean failOnMissingAttribute() {
+        return failOnMissingAttribute;
```
hgschmie, line 28: I cringe a bit at the "getter has the same name as field" pattern. `isFailOnMissingAttribute`?

hgschmie (approved):
Code itself is fine. Users will benefit from

- javadocs for `StringTemplates` configuration class with docs for the new config option.
- couple sentences in the `string template` section of the jdbi docs.

internal/build/pom.xml
```diff
@@ -450,7 +450,7 @@
             <dependency>
                 <groupId>org.antlr</groupId>
                 <artifactId>ST4</artifactId>
-                <version>4.3.1</version>
+                <version>4.3.3</version>
```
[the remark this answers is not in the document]
stevenschlansker: Well, the problem with the lowest version declaration is that in real use cases that means every user is exposed to bugs that have long since been fixed. The vast majority of users will not manage the dependency and would prefer us to have a reasonably up to date version.

I think it is better to treat downgrading as the exception and having the latest as the default - if someone really needs to keep an old version, they are always free to manage it down themselves.

Regarding 4.3.4, because their releases page is out of date :(
https://github.com/antlr/stringtemplate4/releases

stringtemplate4/src/main/java/org/jdbi/v3/stringtemplate4/StringTemplates.java
stevenschlansker: Ah, I cringe at unnecessarily modifying property names for their getter which makes all property discovery much harder, since you have to guess what prefix is maybe or maybe not there :/
I will change it.

Issue #2262, opened 2023-01-30
Feature: Provide collect methods instead of GenericType


gpsfl:
Right now the most simple way to convert a Query to a typed `Collection` or `Map` is the following:

```java
query.collectInto(new GenericType<Map<K, V>>() {});
query.collectInto(new GenericType<List<V>>() {});
```

This however has two disadvantages:
- It creates an anonymous inner class for each query
- It doesn't allow for the usage of Java's `Collectors` api

This is why I suggest adding the following two methods to handle the most common cases. From my understanding they could be implemented using the already existing logic from above:

```java
query.collectToMap(K.class, V.class, Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue))
query.collectTo(V.class, Collectors.toList())
```

We're using JDBI in our production environment and I think this would save quite a lot of code lines / output classes while remaining as readable. If you agree with this change I can submit a pull request for it.

hgschmie:
Hi @gpsfl, 

thanks for opening an issue with the Jdbi project.

Use  `query.map(V.class).list()` for `query.collectInto(new GenericType<List<V>() {});`, which is equivalent to the proposed `query.collectTo(V.class, Collectors.toList()`

I feel that the list collection is as succinct as a new `collectTo` method, which is basically implemented as `mapTo(type).collect(collector)`

* use `query.map(V.class).collect(... collector ...)` for any generic collector

For maps, you can use 
```java
Map<K, V> result = query.reduceRows(new LinkedHashMap<K, V>(), (map, rowView) -> {
    map.put(rowView.getRow(K.class), rowView.getRow(V.class));
    return map;
});
``` 

```java

Map<K, V> map = query.mapTo(new GenericType<Map.Entry<K, V>>() {})
        .collect(Collectors.toMap(Map.Entry::getKey, Map::Entry::getValue));
```

If you are concerned about class counts / readability, use a constant:

```java
private static final CUSTOM_MAP_TYPE = new GenericType<Map.Entry<KeyType, ValueType>>() {};
[...]
Map<KeyType, ValueType> map = query.mapTo(CUSTOM_MAP_TYPE).collect(toMap(Map.Entry::getKey, Map::Entry::getValue));
```

which has a similar amount of characters / type as

```java
Map<KeyType, ValueType> map = query.collectToMap(KeyType.class, ValueType.class, toMap(Map.Entry::getKey, Map.Entry::getValue));
```

[...]

PR #2320, opened 2023-04-13
Provide collect methods on ResultIterable

hgschmie:
This is the first piece to address #2262

- ResultIterable#set()
- ResultIterable#collectToMap(keyFunction, valueFunction)

- add test suite
- some code reshuffling


hgschmie:
these are the pieces of #2303 that we agree upon.

PR #2323, opened 2023-04-19
FreemarkerEngine: use a singleton instance to improve cacheability

stevenschlansker:


core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -389,12 +389,11 @@ private static final class StatementCacheKey {
             this.template = template;
         }
 
-        @SuppressWarnings("PMD.UnusedPrivateMethod") // PMD gets that one wrong...
```
stevenschlansker, line 392: the JVM does not understand class nesting, so while the Java language allows accessing a private method of a nested class, it just ends up generating a synthetic non-private accessor anyway

hgschmie:
Lgtm

PR #2328, opened 2023-04-21
Support Consumer<Iterable<T>> 

hgschmie:
We support Consumer<Stream<T>> and Consumer<Iterator<T>> already

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
@@ -345,13 +345,17 @@ abstract class ResultReturner {
                             .orElseThrow(() -> new IllegalStateException("Couldn't find Iterator type on " + elementType))));
                 }
                 throw new IllegalArgumentException(format("Consumer argument for %s can not use a subtype of Iterator (found %s)!", method, type));
-                // special case: Stream<Iterator<T>>
+                // special case: Consumer<Stream<T>>
             } else if (GenericTypes.isSuperType(Stream.class, type)) {
                 if (GenericTypes.getErasedType(type) == Stream.class) {
                     return new ConsumeStreamResultReturner(consumerIndex, elementType.mapType(t -> GenericTypes.findGenericParameter(t, Stream.class)
                             .orElseThrow(() -> new IllegalStateException("Couldn't find Stream type on " + elementType))));
                 }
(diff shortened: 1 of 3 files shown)


stevenschlansker:
Iterable, unlike Iterator or Stream, expects to be able to produce results more than once. So this oversight is at least somewhat intentional, since we cannot really fulfill Iterable's contract.

hgschmie:
The PR implementation supports calling Iterable#iterator() multiple times just fine. It does cache the result in memory though (in a linked list). This could be implemented as a single-shot iterator, see https://stackoverflow.com/questions/15781739/is-there-any-official-contract-for-the-iterable-interface-with-respect-to-multip for some discussion.

There seems to be no official "must be able to call this multiple times" documentation, just implicit. And it can be mitigated with documentation (see https://docs.oracle.com/javase/7/docs/api/java/nio/file/DirectoryStream.html)

hgschmie:
single shot iterable PoC:

```java
               consumer.accept(new Iterable() {

                    @Override
                    public Iterator iterator() {
                        return stream.iterator();
                    }

                    @Override
                    public void forEach(Consumer action) {
                        stream.iterator().forEachRemaining(action);
                    }

                    @Override
                    public Spliterator spliterator() {
                        return stream.spliterator();
                    }
                });
```

hgschmie:
this would behave exactly like the DirectoryStream thing. Biggest advantage is that the memory use for the iterable would be the same as for iterator and stream. Otherwise, it will cache everything in memory (which may surprise people).

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/ResultReturner.java
```diff
@@ -430,4 +436,21 @@ protected void accept(Stream<?> stream, @SuppressWarnings("rawtypes") Consumer c
...
+
+        ConsumeIterableResultReturner(int consumerIndex, QualifiedType<?> elementType) {
+            super(consumerIndex, elementType);
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        protected void accept(Stream<?> stream, @SuppressWarnings("rawtypes") Consumer consumer) {
+            try {
+                consumer.accept(stream.collect(Collectors.toCollection(LinkedList::new)));
```
stevenschlansker, line 450: This looks like it eagerly consumes all results. We should probably stream the results out through the returned iterator, if possible?

stevenschlansker (approved):

PR #2329, opened 2023-04-24
Performance: CopyOnWriteArray(List,Set) have fast-path shallow copy constructors

stevenschlansker:


stevenschlansker:
No, I don't see it documented anywhere, but both JDK implementations (at least as of 20) do a runtime check for their own type and fast-path it.

Isn't your suggestion to implement copy-on-write ourselves almost the same as just using the JDK implementation, relying on the fast-copy constructor? I am not seeing how it is different.

hgschmie (approved):
LGTM. We should document somewhere that implementors of configuration classes should use CopyOnWriteMap/CopyOnWriteSet for that property.

PR #2330, opened 2023-04-24
Performance: introduce CopyOnWriteHashMap

stevenschlansker:
to make cloning mapper and argument registries faster

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/internal/CopyOnWriteHashMap.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/CopyOnWriteHashMap.java
@@ -1,0 +1,262 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.internal;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.UnaryOperator;
+
+public final class CopyOnWriteHashMap<K, V> implements ConcurrentMap<K, V> {
+    @SuppressWarnings("rawtypes")
+    private static final AtomicReferenceFieldUpdater<CopyOnWriteHashMap, Map> U
+            = AtomicReferenceFieldUpdater.newUpdater(CopyOnWriteHashMap.class, Map.class, "m");
+    @SuppressWarnings("unused")
+    private volatile Map<K, V> m;
+
+    public CopyOnWriteHashMap() {
+        clear();
+    }
+
+    public CopyOnWriteHashMap(Map<K, V> that) {
+        if (that.getClass() == CopyOnWriteHashMap.class) {
+            set(((CopyOnWriteHashMap<K, V>) that).get());
+        } else {
+            set(new HashMap<>(that));
+        }
+    }
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    private AtomicReferenceFieldUpdater<CopyOnWriteHashMap<K, V>, Map<K, V>> u() {
+        return (AtomicReferenceFieldUpdater) U;
+    }
+
+    private Map<K, V> get() {
+        return u().get(this);
+    }
+
+    private void set(Map<K, V> m) {
+        u().set(this, m);
+    }
+
+    private Map<K, V> getAndUpdate(UnaryOperator<Map<K, V>> updater) {
+        return u().getAndUpdate(this, updater);
+    }
+
+    private Map<K, V> updateAndGet(UnaryOperator<Map<K, V>> updater) {
+        return u().updateAndGet(this, updater);
+    }
+
+    @Override
+    public int size() {
+        return get().size();
+    }
+
+    @Override
+    public boolean isEmpty() {
+        return get().isEmpty();
+    }
+
+    @Override
+    public boolean containsKey(Object key) {
+        return get().containsKey(key);
+    }
+
+    @Override
+    public boolean containsValue(Object value) {
+        return get().containsValue(value);
+    }
+
+    @Override
+    public V get(Object key) {
+        return get().get(key);
+    }
+
+    @Override
+    public V put(K key, V value) {
+        return getAndUpdate(old -> {
+            Map<K, V> updated = new HashMap<>(old);
+            updated.put(key, value);
+            return updated;
+        }).get(key);
+    }
+
+    @Override
+    public V putIfAbsent(K key, V value) {
+        return getAndUpdate(old -> {
+            if (old.get(key) != null) {
+                return old;
+            }
+            Map<K, V> updated = new HashMap<>(old);
+            updated.putIfAbsent(key, value);
+            return updated;
+        }).get(key);
+    }
+
+    @Override
+    public void putAll(Map<? extends K, ? extends V> that) {
+        updateAndGet(old -> {
+            Map<K, V> updated = new HashMap<>(old);
+            updated.putAll(that);
+            return updated;
+        });
+    }
+
+    @Override
+    public V remove(Object key) {
+        return getAndUpdate(old -> {
+            if (!old.containsKey(key)) {
+                return old;
+            }
+            Map<K, V> updated = new HashMap<>(old);
+            updated.remove(key);
+            return updated;
+        }).get(key);
+    }
+
+    @Override
+    public boolean remove(Object key, Object value) {
+        return Objects.equals(getAndUpdate(old -> {
+            if (!Objects.equals(old.get(key), value)) {
+                return old;
+            }
+            Map<K, V> updated = new HashMap<>(old);
+            updated.remove(key, value);
+            return updated;
+        }).get(key), value);
+    }
+
+    @Override
+    public V replace(K key, V value) {
+        return getAndUpdate(old -> {
+            if (!old.containsKey(key)) {
+                return old;
+            }
+            Map<K, V> updated = new HashMap<>(old);
+            updated.replace(key, value);
+            return updated;
+        }).get(key);
+    }
+
+    @Override
+    public boolean replace(K key, V oldValue, V newValue) {
+        return Objects.equals(getAndUpdate(old -> {
+            if (!Objects.equals(old.get(key), oldValue)) {
+                return old;
+            }
+            Map<K, V> updated = new HashMap<>(old);
+            updated.put(key, newValue);
+            return updated;
+        }).get(key), oldValue);
+    }
+
+    @Override
+    public V compute(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction) {
+        return updateAndGet(old -> {
+            Map<K, V> updated = new HashMap<>(old);
+            updated.compute(key, remappingFunction);
+            return updated;
+        }).get(key);
+    }
+
+    @Override
+    public V computeIfAbsent(K key, Function<? super K, ? extends V> mappingFunction) {
+        return updateAndGet(old -> {
+            if (old.containsKey(key)) {
+                return old;
+            }
+            Map<K, V> updated = new HashMap<>(old);
+            updated.computeIfAbsent(key, mappingFunction);
+            return updated;
+        }).get(key);
+    }
+
+    @Override
+    public V computeIfPresent(K key, BiFunction<? super K, ? super V, ? extends V> remappingFunction) {
+        return updateAndGet(old -> {
+            if (!old.containsKey(key)) {
+                return old;
+            }
+            Map<K, V> updated = new HashMap<>(old);
+            updated.computeIfPresent(key, remappingFunction);
+            return updated;
+        }).get(key);
+    }
+
+    @Override
+    public V merge(K key, V value, BiFunction<? super V, ? super V, ? extends V> remappingFunction) {
+        return updateAndGet(old -> {
+            Map<K, V> updated = new HashMap<>(old);
+            updated.merge(key, value, remappingFunction);
+            return updated;
+        }).get(key);
+    }
+
+    @Override
+    public void clear() {
+        set(new HashMap<>());
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return get().equals(obj);
+    }
+
+    @Override
+    public int hashCode() {
+        return get().hashCode();
+    }
+
+    @Override
+    public V getOrDefault(Object key, V defaultValue) {
+        return get().getOrDefault(key, defaultValue);
+    }
+
+    @Override
+    public void forEach(BiConsumer<? super K, ? super V> action) {
+        get().forEach(action);
+    }
+
+    @Override
+    public void replaceAll(BiFunction<? super K, ? super V, ? extends V> function) {
+        updateAndGet(old -> {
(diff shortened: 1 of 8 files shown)


hgschmie (changes requested):
What worries me with this change is that the configuration is explicitly multithread safe and this class is most likely not. 

I would like to see a few tests that test multithreading.

core/src/main/java/org/jdbi/v3/core/internal/CopyOnWriteHashMap.java
```diff
@@ -0,0 +1,325 @@
...
+    }
+
+    @Override
+    public V putIfAbsent(K key, V value) {
+        return m.getAndUpdate(old -> {
+            if (old.get(key) != null) {
+                return old;
+            }
+            Map<K, V> updated = new HashMap<>(old);
+            updated.putIfAbsent(key, value);
```
jarlah, line 87: return value is ignored. How do you know if the map was updated ? You can't verify this with compile time checks, so here we return a copy on write even if the old map hasn't even changed.

stevenschlansker:
Generally, it is not possible to show something is thread safe through unit tests. I am happy to explain or test any specific concerns you have with the code, but "test it to show it is multithread-safe" is not really a solvable problem, at least not with the effort I can commit here.

> More worried about performance. But why aren't we using ConcurrentHashMap ? Then we could have avoided all this AtomicReference stuff ? EDIT: Oh we did use ConcurrentHashMap before, and now we suddenly stopped using it in the the new class. I think we should continue using ConcurrentHashMap inside the new wrapper class. The implementation would be a lot simpler and easier to read and understand.

Yes, the point of this change is that right now we are spending the majority of our time (as measured in production in my work app) copying ConcurrentHashMap around. The semantics of Jdbi configuration is that each layer inherits the configuration from its creator (Jdbi -> Handle -> Statement) but further changes are private and do not back-propagate.
Previously we had to eagerly copy each ConcurrentHashMap in case someone further down might change the configuration.
Now, we use a CopyOnWriteMap (which I wish was in JDK but there is no built-in impl) so that when we clone configurations we do not actually have to copy all the maps and nodes until a re-configuration happens (which is almost never).

Does that help? I don't think moving back to ConcurrentHashMap would let me reach my goal of not having to eagerly copy maps on config-clone, but if I missed anything about your suggestion please let me know.

jarlah:
> I don't think moving back to ConcurrentHashMap would let me reach my goal of not having to eagerly copy maps on config-clone, but if I missed anything about your suggestion please let me know.

I just think there is a simpler way. I might not have understood the reasoning behind the AtomicReference and stuff from the get go, but I do now. However, I'm trying to suggest things that make the code less "noisy". For ex this implementation of CopyOnWriteHashMap is much more simpler and more readable https://github.com/djasonpenney/jersey-1.x/blob/2442996200a96989beb76c34e449aeb40a49ced4/jersey-client/src/main/java/com/sun/jersey/client/impl/CopyOnWriteHashMap.java ... instead of always copying, you make a copy only when the instance is cloned, or in this case when you construct a copyonwrite from another copyonwrite, or when you construct a copyonwrite with an existing map (since you want to leave the original map untouched). Is this class from jersey not what we want ? 

Not trying to make noise or anything. I just recently starting to poke into jdbi again after some years away from it, and if my comments is garbage please ignore them :D

stevenschlansker:
Thanks for pointing me at the class from Jersey. I wasn't aware of this one, but I do not think it is correctly multi-threaded. For example, I think there is a race in e.g. the `put` method. Imagine this interleaving: T1 calls put(...), which calls copy(), which does the `compareAndSet`, and is then preempted. T2 comes along, and calls put(...), which calls copy(), which thinks the map has _already been copied_ (due to T1's incomplete work). It then calls put on a _shared_ map. Then, T1 wakes back up, and completes the copy method. In this case T2 would modify a shared, non-thread-safe map that other users expect to be immutable.

This is part of why my implementation runs all modifications through an AtomicReference update cycle. The bug sneaks in because of unsynchronized mutation of _multiple_ fields in concert (`requiresCopyOnWrite` and `core`). By running all updates through a single volatile variable, there's no possibility to slice the work in half - it either completes totally or does not.

Let me know if you think my analysis isn't correct.

hgschmie:
The copyonwritemap from jersey is great, except that it is GPL (and now that we looked at it, we are actually "tainted"). But it is a straightforward implementation of the map semantics along the lines of CoWSet and CoWList

hgschmie:
@stevenschlansker  - https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/utils/CopyOnWriteMap.java

that has a compatible license, maybe just copy that?

hgschmie:
I would be comfortable just using the map from kafka.

stevenschlansker:
I think the Kafka CopyOnWriteMap is also incorrectly implemented. It implements `ConcurrentMap` which implements the default e.g. `Map.computeIfAbsent`. These default implementations are explicitly marked as not thread safe, and required to re-implement if you implement `ConcurrentMap`, but Kafka makes no attempt to actually provide a correct implementation.

Here is a specific failure I see possible with the Kafka impl:
T1 calls computeIfAbsent("a", "1")
-> T1 calls get("a") and observes `null`
T2 calls put("a", "2"), and stores.
-> T1 calls put("a", "1") and overwrites T2's value

Now, the map contains {"a": "1"} which is not consistent with any serial execution. The only valid serial-execution result is {"a":"2"} - either T1 completes first, writing 1, which is then replaced by 2. Or, T2 completes first, putting 2, then T1 is a no-op.

Rather disappointing for a project that I would have thought would do better...

stevenschlansker:
I filed https://issues.apache.org/jira/browse/KAFKA-14942 to at least let the Kafka team know about the potential shortcomings in their map implementation.

stevenschlansker:
If the `AtomicReference` use really is a dealbreaker, it looks like this could be inlined using `AtomicReferenceFieldUpdater`.

stevenschlansker:
I updated the code to avoid the unnecessary `AtomicReference` wrapping. The code is a bit uglier now, but it avoids an extra indirection.

stevenschlansker:
I was at least somewhat wrong about the Kafka implementation. It actually uses default methods from computeIfAbsent, which do use a retry-loop.
Regardless, the implementation here should perform better, as it does not need any locking at all.

jarlah:
I'm very unsure about this. And I feel like a turd bringing it up. But can you show me an example of what this PR is going to solve ? "to make cloning mapper and argument registries faster" is not enough for me to understand, since I literally just "jumped" into the conversation.

So, basically, we have several options to go with:

1. AtomicReference with nesting (😷)
2. AtomicReferenceFieldUpdater, with volatile "unused" map field (😷)
3. volatile map with synchronisation on write operations (Kafka version) (🤧)
4. volatile core map with volatile view map (jersey version, mix of both safe and optimised read and copy on write) ✅ 

Using ideas is not a problem with GPL. but copying code is. Plus this javaee/jersey-1.x is both archived and insanely old (8 years) and copyright notice isn't updated (now it's 2010-2015).  I would argue it's possible to use (reimplement) the jerseys version, but it's not good enough for us (I presume). I would make sure it always copy on write instead of using the atomicboolean, cutting down a bit on the amount of code and the amount of cognitive overhead. We are basically only "tainted" by old and archived GPL code, if we avoid looking at any possible newer version of this file wherever it went after the archival.

What do you think?

PS. I checked out the code to investigate this now, so I have actually looked at and played with the code.

jarlah:
Im toying with the idea of using a combination of several of the ideas, just give me some time and I can come up with an idea for it, maybe. And why do we make all these extra classes when we could just wrap the map in unmodifiableMap? it does make sure the map is unmodifiable.

jarlah:
```java
package org.jdbi.v3.core.internal;

import java.util.*;
import java.util.concurrent.atomic.AtomicReference;
import java.util.function.Consumer;

public class CopyOnWriteHashMapV2<K,V> implements Map<K,V> {
    private volatile AtomicReference<Map<K,V>> core;

    private volatile Map<K,V> view;

    public CopyOnWriteHashMapV2() {
        core.set(new HashMap<>());
        view = null;
    }

    private CopyOnWriteHashMapV2(Map<K,V> that) {
        core.set(new HashMap<>(that));
        view = null;
    }

    private CopyOnWriteHashMapV2(CopyOnWriteHashMapV2<K,V> that) {
        core.set(that.core.get());
        view = null;
    }

    @Override
    public int size() {
        return core.get().size();
    }

    @Override
    public boolean isEmpty() {
        return core.get().isEmpty();
    }

    @Override
    public boolean containsKey(Object key) {
        return core.get().containsKey(key);
    }

    @Override
    public boolean containsValue(Object value) {
        return core.get().containsValue(value);
    }

    @Override
    public V get(Object key) {
        return core.get().get(key);
    }

    @Override
    public V put(K key, V value) {
        V oldValue = get(key);
        copyAndUpdate(map -> map.put(key, value));
        return oldValue;
    }

    @Override
    public V remove(Object key) {
        if (!containsKey(key)) {
            return null;
        }
        V oldValue = get(key);
        copyAndUpdate(map -> map.remove(key));
        // we return the old value, not the value we inserted
        return oldValue;
    }

    @Override
    public void putAll(Map<? extends K, ? extends V> t) {
        copyAndUpdate(map -> map.putAll(t));
    }

    @Override
    public void clear() {
        core.set(new HashMap<>());
        view = null;
    }

    @Override
    public Set<K> keySet() {
        return getView().keySet();
    }

    @Override
    public Collection<V> values() {
        return getView().values();
    }

    @Override
    public Set<Entry<K,V>> entrySet() {
        return getView().entrySet();
    }

    @Override
    public String toString() {
        return core.get().toString();
    }

    private Map<K, V> getView() {
        Map<K, V> result = view;
        if (result == null) {
            result = Collections.unmodifiableMap(core.get());
            view = result;
        }
        return result;
    }

    private void copyAndUpdate(Consumer<Map<K, V>> updater) {
        core.getAndUpdate(old -> {
            Map<K, V> copy = new HashMap<>(old);
            updater.accept(copy);
            return copy;
        });
        view = null;
    }
}
```

jarlah:
the suggestion above is terser and more read optimised. And it will always copy on write. And it will implement put and remove correctly.

hgschmie (approved):

hgschmie:
Personally, I am not convinced that we should taking on implementing a hashmap (I would have just taken the implementation from kafka and be done) and I would prefer a simpler structure with immutable data structures (and then documentation what others should do), but I am ok with having this and shipping it to improve performance.

jarlah (commented):
Im ok to to ship this, _but_, I don't understand the need for SetEntryEntrySet, SetEntryIterator and SetEntrySpliterator. Which is basically ripped out of the java collections api just for this new class. Instead, make an immutable copy of the map and just call the respective functions on on the immutable copy.

I suggest this:

```java
private Map<K, V> getView() {
  return Collections.unmodifiableMap(m); // <- direct read access to the map
}

public Set<K> keySet() {
  return getView().keySet();
}

public Collection<V> values() {
  return getView().values();
}

public Set<Entry<K, V>> entrySet() {
  return getView().entrySet();
}
```

Because, afaik, there is no reason to call get() on the updater. If the updater updates the map atomically the updated value will be represented across the threads because of its volatile keyword.

Unless there is a really good reason to involve these extra classes, I would suggest this. 

If not feel free to ignore.

stevenschlansker:
> I'm very unsure about this. And I feel like a turd bringing it up. But can you show me an example of what this PR is going to solve ? "to make cloning mapper and argument registries faster" is not enough for me to understand, since I literally just "jumped" into the conversation.

No problem @jarlah, it is always good to make sure we are solving a real problem. The specific case that is being solved here is right now every time you open a statement, we make a deep copy of the jdbi config tree. This showed up in our production performance profiling as the single hottest object allocation site, as well as a very hot code path, to do this deep copying. A specific bad case is the `PojoTypes` class, which remembers a user-provided set of classes, keeps its data in a CHM today. In order to make a "child" `PojoTypes` which has the same initial condition but is independent going forward, we must copy the CHM element by element, which is O(n). In our case, N is starting to get into dozens, and when you run many requests per second it shows up on top of our profiler.

I appreciate that you like the Jersey approach, but I think it is inherently much more dangerous than other approaches, because it uses three unsynchronized fields together to maintain state. Since `core`, `view`, and `requiresCopyOnWrite` are all mutated without external synchronization, we must assume that any mutation in the code is _non-atomic_. Specifically,
```
    private void copy() {
        if (requiresCopyOnWrite.compareAndSet(true, false)) {
            core = new HashMap<K, V>(core);
            view = null;
        }
    }
```
looks quite problematic. Since this is not synchronized, we must assume that any other thread might observe a state where e.g. `requiresCopyOnWrite` has already been cleared, but the actual map copy has not happened yet.
In order for this to be safe, either all writes must hold an exclusive lock, or we must prove to ourselves that any partial completion of this method is just as safe as full completion. That's where the potential race comes in around e.g. `computeIfAbsent`. Here's another place I think there is a problem, that remains even if you rewrite to get rid of `requiresCopyOnWrite`:
```
    private Map<K, V> getView() {
        Map<K, V> result = view; // volatile read
        if (result == null) {
            result = Collections.unmodifiableMap(core);
            view = result; // volatile write
        }
```
Here, it is possible for the core and view writes to ping pong. Imagine `getView` starts and reads view as null, so starts to set `view`. It reads `core`, and wraps it in an `unmodifiableMap`. Now, a competing write comes in and as such calls `copy`. The copy is made, the change is made at time T2, and view is cleared to `null`. Now, the original `getView` resumes, and stores the _old map from T1_ into `view`, leading to stale reads until the next update operation comes by.

Basically, I think if you want to do this lock-free, you should aim to have all writes go through a single volatile variable. Any coordinated mutation is hard to prove correct.

I think your updated code is vulnerable to the same race: both `getView` and `copyAndUpdate` mutate multiple volatile variables without external synchronization. So if two threads ping-pong between `getView` and `copyAndUpdate`, you can end up storing a stale `core` into `view`.

> even if the returned value from said functions might be "out of date", its more correct than to return the current value.
This is not true, unfortunately, and my coworker actually just ran into this! Some of the map operations specify specifically returning the _old_ value. I think you actually cover this in your code, so I am not 100% clear on what you meant here.

stevenschlansker:
> Personally, I am not convinced that we should taking on implementing a hashmap 

@hgschmie , totally agree, but unfortunately this is not something that comes with the standard libraries. The Kafka implementation is okay, but given how much work I have been doing to eliminate locking in our hot paths, it seemed unfortunate to go right back to having a core datastructure wrapped in a big lock. The updater approach is really not that bad once you understand the primitives used (there are really only two, "volatile read of current state", and "update state and return (old|new) state".

I have been toying with the idea of trying to contribute this to the JDK, but I am not sure how fruitful that process would be.

stevenschlansker:
> Im ok to to ship this, but, I don't understand the need for SetEntryEntrySet, SetEntryIterator and SetEntrySpliterator. Which is basically ripped out of the java collections api just for this new class. Instead, make an immutable copy of the map and just call the respective functions on on the immutable copy.

@jarlah , fair point, I can just return `Collections.unmodifiableMap(get()).entrySet()` and the extra classes go away. I made this change. (it'll push after tests complete)

jarlah (approved):
With no extra classes in the COWHMP implementation, I approve :) An unused private map field will just have to do for now :)

jarlah:
@stevenschlansker I just looked at the original AtomicReference implementation, and I must sadly admit that I would be better using AtomicReference. I'll admit that the code might have been drowned by the extra complexity from the extra classes 🙈 You decide. But if you don't use any other classes, and use AtomicReference, the stuff is golden. But we can also go with the current approach of course. Im just worried its basically same as going for an AtomicReference 🙈

stevenschlansker:
I agree it would look nicer, but it comes at the cost of an extra indirection. Since I already did the work to inline it, let's go with this. If there's problems, we can revisit again, maybe taking Henning's idea to have truly immutable copies at the expense of ditching the jdk collections implementations.

stevenschlansker:
In the future, with value types, the AtomicReference might get inlined. But that's not in the present, yet.

sman-81:
hi, I am not in favor of adding complicated fundamental data structures such as this to Jdbi.
Is the anticipated gain in performance worth the cost of
- extra complexity?
- increased barrier to entry for new or less experienced contributors?
- additional maintenance burden?

Is there a general merit or is this merely satisfying a narrow use case?
Has before / after been profiled?

stevenschlansker:
Yes, I am fully convinced some change is necessary here. This might not be the change we eventually settle on, but our production application was allocating gigabytes per second of garbage due simply to copying these maps over and over again. We have run our production app with these changes for a few days now, and the GC is totally idle now instead of being on fire under heavy load.

PR #2341, opened 2023-05-02
JdbiUtil: use concurrent data structure to store static state

### LINKED ISSUE #2336: jdbi3-spring plugin  JdbiUtil.TRANSACTIONAL_HANDLES should be made from a concurrent hash map
This set should be either synchronized or a set made from a concurrent hash map.
In a multi-thread scenario, the element in TRANSACTIONAL_HANDLES may be a Hashmap.TreeNode, but it is a Hashmap.Node actually when you try to remove it from the set because of the 'untreeify'.

stevenschlansker:
Fixes #2336

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,4 +1,5 @@
 # Unreleased
+  - spring5 JdbiUtil: fix thread safety #2341
 
 # 3.38.1
   - Dramatic performance improvement around copying configuration objects when creating statements
--- a/spring5/src/main/java/org/jdbi/v3/spring5/JdbiUtil.java
+++ b/spring5/src/main/java/org/jdbi/v3/spring5/JdbiUtil.java
@@ -13,20 +13,21 @@
  */
 package org.jdbi.v3.spring5;
 
-import java.util.HashSet;
+import java.util.Collections;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.core.internal.UtilityClassException;
-import org.springframework.transaction.support.TransactionSynchronizationAdapter;
+import org.springframework.transaction.support.TransactionSynchronization;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 
 /**
  * Utility for working with Jdbi and Spring transaction bound resources
  */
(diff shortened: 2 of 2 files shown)


spring5/src/main/java/org/jdbi/v3/spring5/JdbiUtil.java
```diff
@@ -13,20 +13,21 @@
...
-import org.springframework.transaction.support.TransactionSynchronizationAdapter;
+import org.springframework.transaction.support.TransactionSynchronization;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
 
 /**
  * Utility for working with Jdbi and Spring transaction bound resources
  */
 public class JdbiUtil {
-    private static final Set<Handle> TRANSACTIONAL_HANDLES = new HashSet<>();
+    private static final Set<Handle> TRANSACTIONAL_HANDLES = Collections.newSetFromMap(new ConcurrentHashMap<>());
```
hgschmie, line 30: I am cool with this, but we did use CopyOnWriteSet in other places where concurrency was required.

RELEASE_NOTES.md
```diff
@@ -1,4 +1,5 @@
 # Unreleased
+  - spring5 JdbiUtil: fix thread safety
```
hgschmie, line 2: maybe add a reference to the issue that this addresses

hgschmie (approved):

stevenschlansker:
Right, unless I am missing something, this is quite different. The copy on write map is useful with mostly immutable config where the most common operation is copying the map. In this case we never copy, we are just adding and removing elements to a single static collection, which is a better fit for a normal concurrent map. On May 3, 2023, at 6:40 PM, Henning Schmiedehausen ***@***.***> wrote:﻿
@hgschmie commented on this pull request.



In spring5/src/main/java/org/jdbi/v3/spring5/JdbiUtil.java:
>  import org.springframework.transaction.support.TransactionSynchronizationManager;
 
 /**
  * Utility for working with Jdbi and Spring transaction bound resources
  */
 public class JdbiUtil {
-    private static final Set<Handle> TRANSACTIONAL_HANDLES = new HashSet<>();
+    private static final Set<Handle> TRANSACTIONAL_HANDLES = Collections.newSetFromMap(new ConcurrentHashMap<>());

I am cool with this, but we did use CopyOnWriteSet in other places where concurrency was required.

PR #2344, opened 2023-05-04
Handle.inTransaction: improve exception when isolation restore fails

### LINKED ISSUE #2343: `Handle` `inTransaction` throws in finally block
In [the finally block](https://github.com/jdbi/jdbi/blob/c399c2b273aee3ee7864e9137f86747a87f83375/core/src/main/java/org/jdbi/v3/core/Handle.java#L769) of the Handle inTransaction, the isolation level is reset to the initial value.
If the transaction handler throws an error and resetting the isolation error throws an error as well, the error thrown by the transaction handler is ignored so we lose a lot of context about why the transaction actually failed.
Adding a catch block and storing the initial exception would allow adding suppressed exceptions in the finally block and hence we can display both exceptions to the user.

stevenschlansker:
Fixes #2343

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,4 +1,5 @@
 # Unreleased
+  - Handle.inTransaction: improve exception thrown when restoring transaction isolation #2343
 
 # 3.38.2
   - spring5 JdbiUtil: fix thread safety #2341
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -761,12 +761,8 @@ public class Handle implements Closeable, Configurable<Handle> {
             return callback.withHandle(this);
         }
 
-        TransactionIsolationLevel currentLevel = getTransactionIsolationLevel();
-        try {
-            setTransactionIsolationLevel(level);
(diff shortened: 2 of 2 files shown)


hgschmie (commented):
This is a perfectly fine way to do this, the old way was fine IMHO as well. From the AutoCloseable doc: `An object that may hold resources (such as file or socket handles) until it is closed`. So this is not intended to be a general purpose "let's replace try - finally with try-with-resources" pattern. It introduces new objects that need to be gc'ed and I don't fail to see the real value in this change.

 I don't care too strongly about one way or the other here, but we should not go on a quest to replace all finally blocks with closeables.

hgschmie:
The more I think about this PR, the less I like the change. The old code read pretty straightforward "Save old value, set new value, execute something, restore old value".

The new code reads "set level in a block, do something" (you need to know that the thing that set the level will restore the old level when the block is exited).

hgschmie (approved):

stevenschlansker:
Do you have a suggestion for a better name?

hgschmie:
No. :-) `TemporaryTransactionLevelChangeFactoryProviderSupplier` sounds too enterprisey to me. :-)  

Let's ship what we have.

Issue #2342, opened 2023-05-04
When updating to version 3.38.1, an error message "No row mapper registered for map key class java.lang.String" was prompted.


buzzxu:
Hello, when updating to version 3.38.1, an exception occurred: org.jdbi.v3.core.mapper.NoSuchMapperException: No row mapper registered for map key class java.lang.String. This issue did not occur in version 3.37.1.

buzzxu:
Sorry, I just saw your reply。
I registered many Arguments and ColumnMappers, and I'm not sure if they are the cause of the problem. Without changing the original configuration, version 3.37.1 works fine, but upgrading directly to version 3.38.1 results in an exception.
```java
private void customizeJdbi(Jdbi jdbi){
        jdbi.registerColumnMapper(new GenericType<List<IdName>>() {
        }, (r, columnNumber, ctx) -> {
            String val = r.getString(columnNumber);
            return Strings.isNullOrEmpty(val) ? Collections.emptyList() : Jackson.json2List(val, IdName.class);
        });
        jdbi.registerColumnMapper(new GenericType<List<Map<String, Object>>>() {
        }, (r, columnNumber, ctx) -> {
            String val = r.getString(columnNumber);
            return Strings.isNullOrEmpty(val) ? null : Jackson.json2Object(val, Jackson.buildCollectionType(List.class, Map.class));
        });
        jdbi.registerColumnMapper(new GenericType<Map<String, Object>>() {
        }, (r, columnNumber, ctx) -> {
            String val = r.getString(columnNumber);
            return Strings.isNullOrEmpty(val) ? Collections.emptyMap() : Jackson.json2Map(val);
        });
        jdbi.registerColumnMapper(new GenericType<List<String>>() {
        }, (r, columnNumber, ctx) -> {
            String val = r.getString(columnNumber);
            return Strings.isNullOrEmpty(val) ? Collections.emptyList() : Jackson.isJSON(val) ? Jackson.json2Object(val, new TypeReference<List<String>>() {
            }) : Splitter.on(",").splitToList(val);
        });
        jdbi.registerColumnMapper(new GenericType<List<Integer>>() {
        }, (r, columnNumber, ctx) -> {
            String val = r.getString(columnNumber);
            return Strings.isNullOrEmpty(val) ? Collections.emptyList() :  Jackson.isJSON(val) ? Jackson.json2Object(val, new TypeReference<List<Integer>>() {
[...]

PR #2351, opened 2023-05-11
Ignore unknown mappers when warming DAO classes

hgschmie:
Restore the pre-3.38.0 behavior where errors during warming were only thrown when a method was called for the first time. The 3.38.0 changes made warming up much more aggressive which resulted in errors being reported for extension methods that were not in use (but present on the extension class).

Add some clarification to the error messages as well. 

This addresses #2342


stevenschlansker (approved):
Do we want to make this configurable? Often, fail-fast is considered a feature. But this is fine as is.

hgschmie:
We could; it would just be another knob that needs documenting. I will add a task for it.

PR #2357, opened 2023-05-15
Implement `@RegisterCollector`

TheRealRyGuy:
Adds an annotation call to `JdbiCollectors#registerCollector(Type, Collector)`, allowing for QoL complex collectors as an alternative to making a `CollectorFactory` for use in interface DAOs. 
This could probably use some tests though

### DIFF
--- a/sqlobject/src/main/java/org/jdbi/v3/sqlobject/config/internal/RegisterCollectorImpl.java
+++ b/sqlobject/src/main/java/org/jdbi/v3/sqlobject/config/internal/RegisterCollectorImpl.java
@@ -7,16 +7,26 @@ import org.jdbi.v3.sqlobject.config.RegisterCollector;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
 
 public class RegisterCollectorImpl extends SimpleExtensionConfigurer {
     @Override
     public void configure(ConfigRegistry config, Annotation annotation, Class<?> extensionType) {
         RegisterCollector registerCollector = (RegisterCollector) annotation;
-        JdbiCollectors mappers = config.get(JdbiCollectors.class);
+        JdbiCollectors collectors = config.get(JdbiCollectors.class);
+
         try {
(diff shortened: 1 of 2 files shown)


hgschmie:
Hi @TheRealRyGuy,

Thank you for opening a PR with the Jdbi project. This seems a useful addition to our code base. As you pointed out, it would be good to have some tests (and also add some documentation). I will see if I can address this later this week.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/config/internal/RegisterCollectorImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
+public class RegisterCollectorImpl extends SimpleExtensionConfigurer {
+    @Override
+    public void configure(ConfigRegistry config, Annotation annotation, Class<?> extensionType) {
+        RegisterCollector registerCollector = (RegisterCollector) annotation;
+        JdbiCollectors collectors = config.get(JdbiCollectors.class);
+
+        try {
+            Type type = null; //resultant type
+            for(Type t : registerCollector.value().getGenericInterfaces()) {
```
stevenschlansker, line 20: This looks like it could be something along the lines of `GenericTypes.findGenericParameter(registerCollector.value(), Collector.class).orElseThrow(...)`

stevenschlansker (commented):
Thank you for this - I agree, we should have a test case.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/config/internal/RegisterCollectorImpl.java
TheRealRyGuy: This calls `Class#getGenericSuperclass` instead of `Class#getGenericInterfaces()`, not applicable as far as I know

TheRealRyGuy:
Would appreciate some assistance in writing a test - not 100% sure what generics to pass. Noticed `CollectorFactory` also has wildcard generics on its Collectors and no test so that didn't help me

PR #2377, opened 2023-06-09
Implement @RegisterCollector with tests and short doc

stevenschlansker:
I took the PR and hopefully got it over the finish line with a test and some simple docs.
Bonus fix: try to use the Collector's declared element type, so that you may collect into things that aren't `Iterable`s.

Fixes #2357

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -4636,7 +4636,14 @@ the `User.id` and `User.name` fields. Likewise for `r_id` and `r_name` into the
 The link:{jdbidocs}/sqlobject/config/RegisterConstructorMapper.html[@RegisterConstructorMapper^] annotation may be repeated multiple times on
 the same type or method to register multiple constructor mappers.
 
+==== @RegisterCollectorFactory
 
+Convenience annotation to register a `CollectorFactory` for this SqlObject method. See <<Collectors>> for more details.
+
+==== @RegisterCollectorFactory
+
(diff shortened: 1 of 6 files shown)


stevenschlansker:
cc @TheRealRyGuy

TheRealRyGuy:
thank you so much, have had no time lately 
my only thought that I was hopefully gonna get around to - would it be better to use a CollectorSupplier kind of class in this case? Allows for shorthand Collector.of and Collectors usage

docs/src/adoc/index.adoc
```diff
@@ -4636,7 +4636,14 @@ the `User.id` and `User.name` fields. Likewise for `r_id` and `r_name` into the
 The link:{jdbidocs}/sqlobject/config/RegisterConstructorMapper.html[@RegisterConstructorMapper^] annotation may be repeated multiple times on
 the same type or method to register multiple constructor mappers.
 
+==== @RegisterCollectorFactory
```
hgschmie, line 4639: RegisterCollector ?

hgschmie (changes requested):

docs/src/adoc/index.adoc
hgschmie: I think this would be a better fit after `@UseRowReducer`. The docs are so bare bones that, without an example, this will probably not find any users.

hgschmie:
I will merge this and then fix the docs.

Issue #1813, opened 2021-02-24
Stored Procedures without declared out paramters


lavinj87:
Hello

I've been review this PR https://github.com/jdbi/jdbi/pull/1592 and in particular the unit test. In that unit test I have stored procedure that declares two output parameters is correctly executed and parsed, so far so good.

I was wondering, how may you support invocation of stored procedures without output parameters defined as in (sql-server)

```
CREATE PROCEDURE [Foo].[GetAll()] AS
BEGIN TRY
  SELECT BAR, BAZ FROM FROO
END TRY
BEGIN CATCH
 -- error management
END CATCH
```

The use case is to create a new service that makes use of dropwizard and jdbi but the database access is already defined as stored procedures that can't change.

Thanks for your time.

hgschmie:
I am not sure that Jdbi can support this. Ultimately, this will depend on the jdbc driver for sql-server. Jdbi registers outparameters with the driver, if that would ignore the parameters if they are undeclared, then you can just add some dummy out parameters like this:

```java
OutParameters ret = h.createCall("CALL TO_DEGREES(?, ?)")
            .registerOutParameter(0, Types.REF_CURSOR)
            .registerOutParameter(1, Types.REF_CURSOR)
```

and then retrieve the result sets using

```java
List<MyStuff> myStuff = ret.getRowSet(0).map(MyStuff.class).list();
List<MyOtherStuff> myOtherStuff = ret.getRowSet(1).map(MyOtherStuff.class).list();
```

This pretty much all depends on whether the sql server JDBC driver supports that. 

A test would be helpful, a full test case using https://java.testcontainers.org/modules/databases/mssqlserver/ would be fantastic.

PR #2394, opened 2023-06-15
Fix null return values for outparameters

hgschmie:
- converted the procedure tests in TestCallable to postgres
- enabled the disabled tests (pg has outparameters)
- investigated #1813 (and added a test)

Turns out that do not correctly return null values for outparameters
(we always return the default value if an out parameter is null). Fix
that by calling #wasNull() on the statement. Add a test.

### DIFF
--- a/cache/caffeine-cache/src/it/test-caffeine-cache/pom.xml
+++ b/cache/caffeine-cache/src/it/test-caffeine-cache/pom.xml
@@ -147,6 +147,7 @@
                                 <exclude>**/ConstructorMapperPgTest.*</exclude>
                                 <exclude>**/TestArgumentBinder.*</exclude>
                                 <exclude>**/TestQueriesPG.*</exclude>
+                                <exclude>**/TestCallable.*</exclude>
                             </excludes>
                         </configuration>
                     </plugin>
--- a/core/src/main/java/org/jdbi/v3/core/statement/Call.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/Call.java
@@ -123,11 +123,16 @@ public class Call extends SqlStatement<Call> {
             internalExecute();
             OutParameters out = new OutParameters(getContext());
             for (OutParamArgument param : params) {
-                Object obj = param.map((CallableStatement) stmt);
+                final Object obj = param.map((CallableStatement) stmt);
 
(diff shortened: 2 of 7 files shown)


hgschmie:
@stevenschlansker this warrants some cursory glance, it is obviously correct but it changes behavior slightly.

cache/caffeine-cache/src/it/test-caffeine-cache/pom.xml
```diff
@@ -147,6 +147,7 @@
                                 <exclude>**/ConstructorMapperPgTest.*</exclude>
                                 <exclude>**/TestArgumentBinder.*</exclude>
                                 <exclude>**/TestQueriesPG.*</exclude>
```
stevenschlansker, line 149: Is it possible to hold this configuration somehow by annotating the code? It sucks to edit three pom files just to add a new test case.

core/src/main/java/org/jdbi/v3/core/statement/Call.java
```diff
@@ -198,5 +203,13 @@ public Object map(CallableStatement stmt) {
                 throw new UnableToExecuteStatementException("Could not get OUT parameter from statement", e, getContext());
             }
         }
+
+        private boolean isNull(CallableStatement stmt) {
```
stevenschlansker, line 207: Should this test go into the `OutParamArgument.map` logic instead? Right now it hides all the logic about fetching the value from the statement, and now it is split across two places. Imagine testing `OutParamArgument.map` in isolation; you would expect it to return `null` if the fetched value `wasNull`.

stevenschlansker (approved):

cache/caffeine-cache/src/it/test-caffeine-cache/pom.xml
hgschmie: will all be gone once JDK8 is gone

core/src/main/java/org/jdbi/v3/core/statement/Call.java
hgschmie: i tried a few shapes, this seemed least painful. Otherwise, I need to assign to an object, do null check at the end and return the object or null.

Issue #2395, opened 2023-06-15
Error encountered while converting json string to List during JSON deserialization in JDBI 3


satviksach:
Context:
I'm facing an issue with JSON deserialization in JDBI 3 while using the jdbi3-json module with the Jackson2 plugin. My goal is to map a JSON column to a List<ColumnMetadata> field, but I'm encountering an error when trying to convert the JSON string to a List object. To provide further context, here is the schema of the table involved:

Table Schema:

    Table Name: TableConfig
    Columns:
        organizationId (int)
        tableName (string)
        columnMetadata (json)

Code Snippet - TableConfig Class:

```
public class TableConfig {
    private int organizationId;
    private String tableName;

    @Json
    private List<ColumnMetadata> columnMetadata;

    // Constructors, getters, and other methods
}

```
Code Snippet - ColumnMetadata Class:
```

public class ColumnMetadata {
    private String fieldName;
    private boolean visible;
    private Integer order;

    // Constructors, getters, and other methods
}

```
Code Snippet - DAO Get Call:

java[EDIT]
```
@SqlQuery("""
            SELECT *
            FROM OrganizationTableConfig
            WHERE OrganizationID = :organizationId AND TableName = :tableName
            """)
@RegisterBeanMapper(TableConfig.class)
    TableConfig getTableConfig(
            @Bind("organizationId") int organizationId,
            @Bind("tableName") String tableName
    );

```
Sample Error Message:
Upon attempting the JSON deserialization, I receive the following error message: "Cannot cast java.lang.String to java.util.List". This error prevents the successful conversion of the JSON string into the List<ColumnMetadata> field.

Expected Outcome:
I'm seeking assistance in resolving this issue and correctly performing the JSON deserialization for the columnMetadata field in the given TableConfig class while utilizing the jdbi3-json module. Any guidance, suggestions, or alternative approaches would be greatly appreciated.

stevenschlansker:
Hi @satviksach , it looks like in your DAO:

```
@SqlQuery("SELECT * FROM TableConfig")
List getAllTableConfigs();
}
```

This uses the raw type `List`, so Jdbi does not know what type you want it to return. Try using a parameterized type like `List<TableConfig>`.

natebranchapp:
Better question is: Does JDBI support converting a `JSON` SQL column type that contains an JSON array (vs. an object) to a java `List<MyJson> listOfJsonObjects` Or only JSON objects?

natebranchapp:
[...]

PR #2396, opened 2023-06-16
Add a unit test for issue 2395

hgschmie:
Ensure that this works just fine. :-)

### DIFF
--- a/json/src/test/java/org/jdbi/v3/json/AbstractIssue2395Test.java
+++ b/json/src/test/java/org/jdbi/v3/json/AbstractIssue2395Test.java
@@ -1,0 +1,320 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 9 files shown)


json/src/test/java/org/jdbi/v3/json/AbstractIssue2395Test.java
```diff
@@ -0,0 +1,161 @@
...
+        assertThat(dao.retrieve(2)).isEqualTo(data2);
+    }
+
+    public interface Dao {
+
+        @SqlUpdate("insert into test2395 (id, name, data) values (:id, :name, :data)")
+        void write(@BindBean Wrapper bean);
+
+        @SqlQuery("select * from test2395")
+        List<Wrapper> readAll();
```
stevenschlansker, line 71: If this issue is testing #2395 , I think this is slightly different. The OP asked about mapping `@Json List<T>` from a `[T, T]` json response, which is not the same as mapping `List<@Json T>` from a result set with many rows of `T`, which is what this test checks

stevenschlansker (approved):

PR #2403, opened 2023-06-23
RowMapper, ColumnMapper.findFor: reduce Stream usage in hot path

stevenschlansker:
This showed up as an allocation hotspot in our app since finding mappers is one of the most common Jdbi ops

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
@@ -26,7 +26,6 @@ import org.jdbi.v3.core.enums.internal.EnumMapperFactory;
 import org.jdbi.v3.core.generic.GenericType;
 import org.jdbi.v3.core.interceptor.JdbiInterceptionChainHolder;
 import org.jdbi.v3.core.internal.CopyOnWriteHashMap;
-import org.jdbi.v3.core.internal.JdbiOptionals;
 import org.jdbi.v3.core.qualifier.QualifiedType;
 import org.jdbi.v3.meta.Alpha;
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
```diff
@@ -213,15 +212,18 @@ public <T> Optional<ColumnMapper<T>> findFor(QualifiedType<T> type) {
...
         }
 
-        Optional<ColumnMapper<T>> mapper = (Optional) factories.stream()
-                .flatMap(factory -> JdbiOptionals.stream(factory.build(type, registry)))
-                .findFirst();
-
-        mapper.ifPresent(m -> m.init(registry));
-
-        cache.put(type, mapper);
+        for (QualifiedColumnMapperFactory factory : factories) {
```
hgschmie, line 215: this seems to be an excellent opportunity to fix some technical debt from the past (this will go into main branch):

```java
    @SuppressWarnings({ "unchecked" })
     public <T> Optional<ColumnMapper<T>> findFor(QualifiedType<T> type) {
        return Optional.ofNullable((ColumnMapper<T>) cache.computeIfAbsent(type, t -> {
            for (QualifiedColumnMapperFactory factory : factories) {
                ColumnMapper<?> mapper = factory.build(t, registry).orElse(null);
                if (mapper != null) {
                    mapper.init(registry);
                    return mapper;
                }
            }
            return null;
        }));
```

and change the cache to no longer hold the optionals but only the actual mappers (which makes the cache smaller because it no longer needs to cache negative hits).

hgschmie: ok, after some testing, removing the negative cache is a bad idea. One ends up with an infinite loop when resolving e.g. a map type. So keep the optional. :-) It still makes the code nicer.

hgschmie:
what surprises me still is that this becomes a hotspot. the actual operations should be expensive only once and then served from the cache. Are you by any chance using `@RegisterRowMapper` or `@RegisterColumnMapper` annotations? *Any* newly registered mapper trashes the cache and in that case, you would re-register the mapper all the time for each operations. You would be much better off registering them once (ideally with the Jdbi) and then leverage the cache. 

I wonder if the right fix would be to make cache invalidation smarter (registering a mapper only invalidates the cache entry for that type).

stevenschlansker:
I agree, this indicates something is clearing the cache unexpectedly. I'll try to trace that down as a followup. We don't dynamically register any types.

It is hard to make the cache invalidation smarter since the factories are allowed to inspect a `Type` and decide if they will map it or not. There is no way to enumerate all types that a factory might map. In fact, doing so might not be possible - what about a mapper that matches nearly any type, like a json mapper.

core/src/main/java/org/jdbi/v3/core/mapper/ColumnMappers.java
stevenschlansker: I think your infinite looping might be re-discovering this `computeIfAbsent` behavior, rather than anything to do with the negative cache. It might still be possible to remove the negative caching, although I am not sure it hurts or helps much, since missing mappers should be rare (usually throws an exception)

stevenschlansker:
I will go ahead with this change, since it should help cases where you cannot re-use the mapper registrations.
I am still following up on why mapper re-use doesn't work in the first place and will have test cases soon.

PR #2408, opened 2023-06-27
Remove JSR 305 annotations

hgschmie:
The findbugs jsr305 annotations jar uses annotations in the
javax.annotation namespace, which causes problems with JPMS.

Remove the annotations from the code base. The Jdbi code itself
transitions to error prone and spotbugs.

For code that uses Jdbi (and all tests), a new
annotation (org.jdbi.v3.core.mapper.reflect.Nullable) is introduced as
all candidates either do not have Runtime retention (spotbugs,
jetbrains), have problems with JPMS (jsr305), are very
specialized (spring) or obscure (checker framework).


stevenschlansker (changes requested):
Let's use the `jakarta.annotation-api` that provides `Nullable` and `Nonnull` rather than rolling our own, unless there's a reason not to.

hgschmie:
replaced custom annotation with jakarta.

PR #2414, opened 2023-06-29
move method reflection from ExtensionFactoryDelegate to ExtensionMetadata

### LINKED ISSUE #2413: ExtensionFactoryDelegate uses reflection checks for every time you attach a dao
Right now, the `ExtensionFactoryDelegate.attach` method will use reflection to look up methods `toString`, `equals`, and `hashCode` every time you attach the dao.

This logic should probably be pushed up to only happen the first time we see a type, since its methods can never change.

stevenschlansker:
fixes #2413

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
+++ b/core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
@@ -35,7 +35,7 @@ import static java.lang.String.format;
 @Alpha
 public interface ExtensionHandler {
 
-    /** Implementation for the {@link Object#equals(Object)} method. Each object using this handler is only equal to ifself. */
+    /** Implementation for the {@link Object#equals(Object)} method. Each object using this handler is only equal to itself. */
     ExtensionHandler EQUALS_HANDLER = (handleSupplier, target, args) -> target == args[0];
 
     /** Implementation for the {@link Object#hashCode()} method. */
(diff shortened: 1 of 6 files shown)


hgschmie:
Turns out that there is actually a lot of stupidity in the original code. An interface can not override any method from Object (because ... rules). So all of the tests can go away, because we only create the metadata for interfaces.

core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
```diff
@@ -44,6 +45,15 @@ public interface ExtensionHandler {
     /** Handler that only returns null independent of any input parameters. */
     ExtensionHandler NULL_HANDLER = (handleSupplier, target, args) -> null;
 
+    /** Constant for {@link Object#equals(Object)}. */
+    Method EQUALS_METHOD = JdbiClassUtils.methodLookup(Object.class, "equals", Object.class);
```
stevenschlansker, line 49: I originally had these in `JdbiClassUtils` in order to have them not be part of the public API. Do we really want to ship this as part of public API?

core/src/main/java/org/jdbi/v3/core/extension/ExtensionMetadata.java
```diff
@@ -109,6 +112,14 @@ public Set<Method> getExtensionMethods() {
         return methodHandlers.keySet();
     }
 
+    /**
+     * Returns a reference to a method that overrides {@link Object#finalize()} if it exists.
+     * @return An {@link Optional} containing a {@link Method} if a finalizer exists.
+     */
+    public Optional<Method> getFinalizer() {
```
stevenschlansker, line 119: I originally had this as package-private so it is not part of the public API.
Do we really think this should be a fully supported part of the public API?

PR #2416, opened 2023-06-29
Improve config customizer performance

hgschmie:
Allow config customizers that do not need the config object itself to
create internal state at metadata creation time and cache that state as
part of the extension type metadata. This moves the creation of a
class (and potentially other expensive operations) away from the
invocation of extension methods.


stevenschlansker (approved):

hgschmie:
any chance you can test this before merging it? I want to make sure it actually has benefits.

PR #2418, opened 2023-06-29
Make caches for row and column mappers smarter

hgschmie:
If a mapper factory can report a type that it supports (the inferred
factories and the static `...of()` mappers do as well), only invalidate
a single cache entry, not the whole cache.

This preserves elements that have been registered on the handle or the
jdbi ifself when using with sqlobject @RegisterColumnMapper and
@RegisterRowMapper annotations

Self-rolled factories can implement the `getType()` method on the
RowMapper and ColumnMapper interfaces to benefit from this as well.


hgschmie:
Not really effective, shelving this for now

PR #2420, opened 2023-06-29
Json plugin: do json type resolution during prepare stage

stevenschlansker:
for arguments and column mappers

This allows type lookup and object reader / writer to be reused

### DIFF
--- a/json/src/main/java/org/jdbi/v3/json/JsonMapper.java
+++ b/json/src/main/java/org/jdbi/v3/json/JsonMapper.java
@@ -26,6 +26,19 @@ import org.jdbi.v3.core.config.ConfigRegistry;
  * jdbi3-jackson2 and jdbi3-gson2 are readily available for this.
  */
 public interface JsonMapper {
-    String toJson(Type type, Object value, ConfigRegistry config);
-    Object fromJson(Type type, String json, ConfigRegistry config);
+    @Deprecated // forRemoval=true
+    default String toJson(Type type, Object value, ConfigRegistry config) {
+        return forType(type, config).toJson(value, config);
+    }
+    @Deprecated // forRemoval=true
+    default Object fromJson(Type type, String json, ConfigRegistry config) {
+        return forType(type, config).fromJson(json, config);
+    }
+
+    TypedJsonMapper forType(Type type, ConfigRegistry config);
+
+    interface TypedJsonMapper {
+        String toJson(Object value, ConfigRegistry config);
+        Object fromJson(String json, ConfigRegistry config);
+    }
 }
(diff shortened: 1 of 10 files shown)


json/src/main/java/org/jdbi/v3/json/JsonMapper.java
```diff
@@ -26,6 +26,19 @@
...
+        return forType(type, config).toJson(value, config);
+    }
+    @Deprecated // forRemoval=true
+    default Object fromJson(Type type, String json, ConfigRegistry config) {
+        return forType(type, config).fromJson(json, config);
+    }
+
+    TypedMapper forType(Type type, ConfigRegistry config);
+
+    interface TypedMapper {
```
hgschmie, line 40: JsonTypedMapper ? "TypedMapper" is too generic for my taste. 

This is basically a 

```java
public interface SerDe<T> {
    String serialize(T value, ConfigRegistry config);
    T deserialize(String value, ConfigRegistry config);
}
```

This may be useful in other contexts as well.

json/src/test/java/org/jdbi/v3/json/JsonPluginTest.java
```diff
@@ -13,44 +13,52 @@
...
 import org.jdbi.v3.testing.junit5.JdbiExtension;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.api.extension.RegisterExtension;
-import org.mockito.Mock;
-import org.mockito.junit.jupiter.MockitoExtension;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.ArgumentMatchers.any;
```
hgschmie, line 27: I applaud this. :-)

hgschmie (approved):

json/src/main/java/org/jdbi/v3/json/JsonMapper.java
stevenschlansker: Sure, it is really a nested type, so the name is `JsonMapper.TypedMapper` which already feels a bit redundant and `JsonMapper.JsonTypedMapper` feels even more redundant. But I do not care too much.

I am happy for the interface to be re-used but I do not see another place for it to be used right now.

json/src/test/java/org/jdbi/v3/json/JsonPluginTest.java
stevenschlansker: I rather dislike that mocks make your tests depend on exactly what methods are called in what order, rather than the actual behavior of the implementation...

PR #2429, opened 2023-07-10
ConfigRegistry, BeanPropertiesFactory: restore behavior of reusing method lookups

### LINKED ISSUE #2427: Followup #2425 - restore eager Constructor and Method lookup
#2425 introduces some nice centralization, but causes a regression where instead of looking up constructor or methodhandle objects once during initialization, we now look them up again every time we invoke.

Let's please restore this to eagerly look up the constructor or method the first time we encounter a type, and then re-use that each time we actually need to invoke. cc @hgschmie

stevenschlansker:
Fixes #2427

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
+++ b/core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
@@ -13,6 +13,9 @@
  */
 package org.jdbi.v3.core.config;
 
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
@@ -21,7 +24,7 @@ import java.util.function.Function;
 import org.jdbi.v3.core.argument.Arguments;
 import org.jdbi.v3.core.collector.JdbiCollectors;
 import org.jdbi.v3.core.config.internal.ConfigCaches;
-import org.jdbi.v3.core.internal.JdbiClassUtils;
+import org.jdbi.v3.core.internal.exceptions.Sneaky;
 import org.jdbi.v3.core.mapper.ColumnMappers;
 import org.jdbi.v3.core.mapper.Mappers;
 import org.jdbi.v3.core.mapper.RowMappers;
@@ -33,9 +36,6 @@ import org.jdbi.v3.core.statement.SqlStatements;
  * @see Configurable
  */
 public final class ConfigRegistry {
-
-    private static final Class<?>[] JDBI_CONFIG_TYPES = {ConfigRegistry.class};
-
     private final Map<Class<? extends JdbiConfig<?>>, JdbiConfig<?>> configs = new ConcurrentHashMap<>(32);
     private final Map<Class<? extends JdbiConfig<?>>, Function<ConfigRegistry, JdbiConfig<?>>> configFactories;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
```diff
@@ -81,12 +81,45 @@ public <C extends JdbiConfig<C>> C get(Class<C> configClass) {
     }
 
     private Function<ConfigRegistry, JdbiConfig<?>> configFactory(Class<? extends JdbiConfig<?>> configClass) {
-        return configFactories.computeIfAbsent(configClass, klass ->
-                registry -> {
-                    var config = JdbiClassUtils.findConstructorAndCreateInstance(klass, JDBI_CONFIG_TYPES, registry);
```
stevenschlansker, line 86: This unfortunately does not fit well into the new helper paradigm, so I mostly reverted it to how it was, since we take care to use different argument lists with `invokeExact`.

core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
```diff
@@ -187,12 +184,10 @@ private static <T> T checkedCreateInstance(Class<T> type, Class<?>[] parameters,
      * @param parameters Parameters for the constructor.
      * @return An {@link Optional} wrapping the instantiated type or {@link Optional#empty()} if no matching constructor was found.
      */
-    @Alpha
```
stevenschlansker, line 190: `internal` package is never public API, so I am not sure `@Alpha` means anything here, since it is always subject to change.

hgschmie (changes requested):

core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
stevenschlansker: @hgschmie , yes, we can benchmark this. I am fine to have benchmarks before changing sensitive code, but please hold yourself to that standard too - #2425 was merged with no benchmarking and no review, and introduces a significant performance regression. Please consider #2432 which demonstrates that `MH.invokeExact` is by far the fastest implementation.

I object to your characterization of this as unreadable. What parts could be written better to be more readable? Or, maybe you need to spend some time familiarizing yourself with `MethodHandle`? It is not the prettiest code but it does what it needs to do, and is relatively straightforward.

MH is the new way of doing things and will not be going away. I think at this point it should be the default in any performance-sensitive code over "classic" reflection, unless there is a specific reason not to use it. Also, always prefer `invokeExact` when possible.

I like the idea of improving the `JdbiClassUtils` to return a `MethodHandle` that is stored for later re-use. However, it is a little tricky because the returned MH would have different method signatures depending on whether a `ConfigRegistry` is expected as an argument or not. It is entirely possible to fix this as well, but at some point the helper method is not generic anymore and ends up looking very much like just moving this code to `JdbiClassUtils`, which doesn't really solve any problem. I played with this for a bit but did not make good progress on deriving a helper method that is both fast and actually re-usable. So for now, this just brings us back to the baseline we had before.

> In its current shape, this seems to be more an exercise in proving that this can be done somehow using MethodHandles at any cost. I want to understand first what the value is before reverting the change. The loss of readability is IMHO a problem.

Right, because using `MH.invokeExact` carries significant performance gains. So, I spent significant effort converting the old code to be much faster. It is a little frustrating to leave for one week to find all that work thrown in the trash. I am on board with the goal to improve readability, but there very much is a reason that this code was the way it was.

> Why can't the findConstructorAndCreateInstance method not be rewritten to iterate over the list of parameters using method handles in the same way it does right now for finding constructors?

This is the tricky bit, because then the returned MH will accept an unknown number of parameters. We'd need to return some composite result that has both the MH and which signature it should be called with, otherwise you run into `WrongMethodTypeException`.

stevenschlansker: I guess what we could do is take the MethodHandle we want to return and compose with `MethodHandles.dropArguments` to avoid the caller needing to know what kind of call we actually make. I will briefly look into this, although this is very much not what I was planning to spend this morning on.

stevenschlansker: I played around with this a bit and I think it would work, however there is actually only one place that we need both 1) MethodHandle performance and 2) accept multiple type signatures, and that is here in `ConfigRegistry`. So, I am thinking that we should defer further extracting a general helper until we have a second use site that actually needs it.

core/src/main/java/org/jdbi/v3/core/config/ConfigRegistry.java
```diff
@@ -81,12 +81,45 @@ public <C extends JdbiConfig<C>> C get(Class<C> configClass) {
...
-                    config.setRegistry(registry);
-                    return config;
-                });
+        return configFactories.computeIfAbsent(configClass, klass -> {
+            final Exception notFound;
+            try {
+                MethodHandle mh = MethodHandles.lookup().findConstructor(klass,
+                        MethodType.methodType(void.class, ConfigRegistry.class))
+                    .asType(MethodType.methodType(JdbiConfig.class, ConfigRegistry.class));
+                return registry -> {
```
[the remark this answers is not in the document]
stevenschlansker: Unreadable is very much a style opinion. I moved the cache computation function to a helper method, so there are no nested lambdas now.

The sneaky throws were intentionally unrolled - previously, we used `Unchecked`, but this introduces more stack frames. I thought it was preferable to handle it locally rather than nesting more lambdas. Do you like the `Unchecked.function` style better?

Do you have further concrete suggestions for how it could be more readable?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/BeanPropertiesFactory.java
```diff
@@ -203,10 +205,22 @@ static class PropertiesHolder<T> {
                 } catch (IntrospectionException e) {
                     throw new IllegalArgumentException("Failed to inspect bean " + clazz, e);
                 }
+                Supplier<T> myConstructor;
```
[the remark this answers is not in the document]
stevenschlansker: I think you mis-understand what MethodHandles are supposed to achieve. They are not supposed to be "reflection but easier". In fact, it is somewhat more complicated to think about, but in exchange for understanding a bit more about how the JVM implements it under the hood, it is much easier for the JVM to optimize it.

JEP-416 is a great change but is more about the code maintainability of the JVM itself. It misses a significant part of the performance puzzle, which is that "classic" reflection did not have enough static information available to the JVM to allow optimizations. In order to utilize these new optimizations, you must use the MH interface, not classic reflection.

MethodHandles are "a bit more complicated than reflection, but easier for the JVM to optimize, and better modelling how it actually works under the hood".

PR #2431, opened 2023-07-11
Cache the discovered c'tors for types

hgschmie:
This ensures that each c'tor is discovered only once and then
retrieved from a cache.

Alternative approach to #2427

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
@@ -16,12 +16,15 @@ package org.jdbi.v3.core.internal;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Arrays;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jdbi/v3/core/internal/JdbiClassUtils.java
```diff
@@ -192,6 +195,8 @@ public static <T> T findConstructorAndCreateInstance(Class<T> type, Class<?>[] t
         return findConstructorAndCreateInstance(type, types, DEFAULT_EXCEPTION_HANDLER, parameters);
     }
 
+    private static final Map<Class<?>, CheckedFunction<Object[], Object>> CTOR_CACHE = new ConcurrentHashMap<>();
```
stevenschlansker, line 198: This will cause Jdbi to retain a reference to any class it observes forever, preventing class unloading.
Probably most users don't care but it could be considered a memory leak.

stevenschlansker:
I created a benchmark in #2432 which demonstrates that while caching the discovered constructors helps by about a factor of 2x, it is still 10x slower than the code that was there before:

```
ConstructorBenchmark.constructorMhInvokeExact        avgt    5   2.021 ± 0.179  ns/op
ConstructorBenchmark.jcuFindInstantiate              avgt    5  32.425 ± 1.115  ns/op
ConstructorBenchmark.jcuFindInstantiateWithCache     avgt    5  18.645 ± 0.207  ns/op
```

hgschmie:
Superseded by #2441

PR #2433, opened 2023-07-11
JdbiClassUtils: add a MethodHandle based findConstructor

### LINKED ISSUE #2427: Followup #2425 - restore eager Constructor and Method lookup

stevenschlansker:
Fixes #2427

Replaces #2429


hgschmie:
Superseded by #2441

PR #2436, opened 2023-07-13
MapToFactory: reuse a single ResultReturner

stevenschlansker:
rather than re-creating it each time

Inspecting methods is relatively expensive and they don't change.


hgschmie (approved):
I assume that you checked that "result returners should be stateless" is true for all returners that we ship.

stevenschlansker:
Yes, I did.

PR #2439, opened 2023-07-13
Generator: improve performance of creating instances

stevenschlansker:
Right now, every time you call attach or onDemand we repeatedly lookup the class and methods we use to construct them. Now, hold a cache of the lookups so we can re-use them.


hgschmie:
this should benefit from #2441 as well.

hgschmie (approved):

PR #2440, opened 2023-07-13
OnDemandExtensions: flatten stack a bit by inlining Unchecked

stevenschlansker:
We call Unchecked three times, and this ends up adding a number of stack frames in the debugger or profiler any time you are looking at an ondemand sqlobject

This fights back just a little bit against ever-deepening stack traces, and makes the logic easier to read too without the distracting Unchecked signatures in the middle


hgschmie (approved):
shipit.

PR #2443, opened 2023-07-18
update jackson version b/c CVE

hgschmie:


stevenschlansker (approved):
This is a pretty old date to be updating to - in 2022? Is the CVE old and we just didn't notice until now?

hgschmie:
dependabot did not flag it. I only saw it in the sonatype vuln report. We might want to upgrade past 2.12 (which is no longer the stable branch) at some point.

PR #2451, opened 2023-08-03
Handle: unconditionally close Connection

### LINKED ISSUE #2446: [Bug] Jdbi3-core - connection object can be stuck in a connection pool forever if connection was closed.
Hi,
it seems that `jdbi3-core` might have a code path that can make broken connection objects stuck in the connection pool forever in "ALLOCATED" state. This can lead to max out the connection pool capacity. This is what most likely have happen in our system.

I have created a reproducer for this issue with detailed instructions - [here](https://github.com/kristoffSC/Jdbi3ObjectStuckInPool)

In Handle's constructor there is a check, that verifies if used connection for this handle is live. The Cleanable is added to the created Handle instance ONLY if connection is Live.

```
addCleanable(() -> {
            // shut down statement builder
            if (checkConnectionIsLive()) {
                statementBuilder.close(connection);
            }
        });
```
If the connection was closed or had any connection issue (client or server side) after it was borrowed from the pool and before this check, then no Cleanable will be registered making this connection object never be returned to the pool.

**###### UPDATE ######**
After deeper analysis it seems that problem is more complex.

The generalization is that if there will be an exception thrown anywhere in `org.jdbi.v3.core.Jdbi::open()` method, after the connection was acquired from pool, then the connection object will not be returned to the pool.

I would say that `org.jdbi.v3.core.Jdbi::open()` should have exception handling logic which will guaranty that connection object will be properly returned to the pool in case of exception. Currently it is not the case.

Specific use cases:
1. The connection is closed from the client side, as described in this bug report.
In this case it turns out that exception will be thrown by this [line](https://github.com/jdbi/jdbi/blob/299268ba7b00a2845e05f7942fb416147d603f18/core/src/main/java/org/jdbi/v3/core/Handle.java#L115C17-L115C17) in Handle's constructor.
`this.transactionHandler = transactionHandler.specialize(this);` so even before registering any Cleanable objects 

In this case, since exception was thrown from Handle constructor, there is no Handle instance to close/clean at the end.

2. The second use case is where the connection is closed from server side, for example Data Base restart.
In our system we are using Jdbi's PostgresPlugin. In  `org.jdbi.v3.core.Jdbi::open()` method, after Handle object is created there is this call:
```
for (JdbiPlugin p : plugins) {
        h = p.customizeHandle(h);
}
```
This will throw an exception for PostgresPlugin.
In result the `open()` method will not return any Handle reference so no Handle will be closed/clean up hence connection object will not be returned to the pool. The stack trace for this case - 
[JdbiTrace.txt](https://github.com/jdbi/jdbi/files/12276766/JdbiTrace.txt)

To reproduce this, please pull my repository (I added PostgresPlugin) and simply restart Postgresql container on a break point just after where connection object is acquired from the pool.

The PoC patch for this could be something like this:
```
public Handle open() {
    Connection conn = null;
    Handle h = null;
    try {
      conn = Objects.requireNonNull(connectionFactory.openConnection(),
          () -> "Connection factory " + connectionFactory + " returned a null connection");
      final long stop = System.nanoTime();

      for (JdbiPlugin p : plugins) {
        conn = p.customizeConnection(conn);
      }

      StatementBuilder cache = statementBuilderFactory.get().createStatementBuilder(conn);

      h = Handle.createHandle(this,
          connectionFactory.getCleanableFor(conn), // don't use conn::close, the cleanup must be done by the connection factory!
          transactionhandler.get(),
          cache,
          conn);

      for (JdbiPlugin p : plugins) {
        h = p.customizeHandle(h);
      }
      LOG.trace("Jdbi [{}] obtain handle [{}] in {}ms", this, h, MILLISECONDS.convert(stop - start, NANOSECONDS));
      return h;
    } catch (Exception e) {
     // NEW EXCEPTION HANDLING TO GUARANTEE THAT CONNECTION OBJECT IS WILL BE RETURNED TO THE POOL.
      if (h != null) {
        // close Handle if we have one in case of Exception, connection will be clean up by this.
        h.close();
      } else if (conn != null) {
        try {
         // clean up a connection if we have one since exception must have been thrown before Handle instance was created
          this.connectionFactory.closeConnection(conn);
        } catch (SQLException ex) {
          throw new ConnectionException(e);
        }
      } if (e instanceof SQLException) { // not sure about this part, whatever Exception type we would like to have here
        throw new ConnectionException(e);
      }
    }
  }
```

stevenschlansker:
Right now, we check if a connection is closed before closing it.

However, this isn't good enough: even if the server closed the connection, we might have borrowed this from a pool, which needs to know we've returned it.

Connection.close is specified to no-op if called more than once, so it's safe to just do it regardless.

Fixes #2446

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Jdbi.java
+++ b/core/src/main/java/org/jdbi/v3/core/Jdbi.java
@@ -33,6 +33,7 @@ import org.jdbi.v3.core.extension.Extensions;
 import org.jdbi.v3.core.extension.HandleSupplier;
 import org.jdbi.v3.core.extension.NoSuchExtensionException;
 import org.jdbi.v3.core.internal.OnDemandExtensions;
+import org.jdbi.v3.core.internal.exceptions.Sneaky;
 import org.jdbi.v3.core.internal.exceptions.Unchecked;
 import org.jdbi.v3.core.spi.JdbiPlugin;
 import org.jdbi.v3.core.statement.DefaultStatementBuilder;
@@ -337,11 +338,13 @@ public class Jdbi implements Configurable<Jdbi> {
      * @see #useHandle(HandleConsumer)
      * @see #withHandle(HandleCallback)
      */
+    @SuppressWarnings({ "PMD.CloseResource", "PMD.AvoidInstanceofChecksInCatchClause" })
     public Handle open() {
+        Connection conn = null;
+        Handle h = null;
         try {
             final long start = System.nanoTime();
-            @SuppressWarnings("PMD.CloseResource")
-            Connection conn = Objects.requireNonNull(connectionFactory.openConnection(),
+            conn = Objects.requireNonNull(connectionFactory.openConnection(),
                     () -> "Connection factory " + connectionFactory + " returned a null connection");
             final long stop = System.nanoTime();
 
@@ -351,7 +354,7 @@ public class Jdbi implements Configurable<Jdbi> {
 
             StatementBuilder cache = statementBuilderFactory.get().createStatementBuilder(conn);
 
-            Handle h = Handle.createHandle(this,
+            h = Handle.createHandle(this,
                 connectionFactory.getCleanableFor(conn), // don't use conn::close, the cleanup must be done by the connection factory!
                 transactionhandler.get(),
                 cache,
@@ -362,8 +365,21 @@ public class Jdbi implements Configurable<Jdbi> {
             }
             LOG.trace("Jdbi [{}] obtain handle [{}] in {}ms", this, h, MILLISECONDS.convert(stop - start, NANOSECONDS));
             return h;
-        } catch (SQLException e) {
-            throw new ConnectionException(e);
+        } catch (Exception e) {
+            try {
+                if (h != null) {
+                    h.close();
+                } else if (conn != null) {
+                    connectionFactory.closeConnection(conn);
+                }
+            } catch (Exception e1) {
+                e.addSuppressed(e1);
+            }
+            if (e instanceof SQLException) {
+                throw new ConnectionException(e);
+            } else {
+                throw Sneaky.throwAnyway(e);
+            }
         }
     }
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+# Unreleased
+  - Handle: fix potential Connection leak if connection is closed early
+
 # 3.40.0
 
   - add extension point to decorate withHandle and friends (#2448)
--- a/core/src/main/java/org/jdbi/v3/core/ConnectionFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/ConnectionFactory.java
@@ -45,6 +45,7 @@ public interface ConnectionFactory {
 
     /**
      * Returns a {@link Cleanable} that will close the connection and release all necessary resources.
+     * Like {@link Connection#close()}, the Cleanable should no-op if called more than once.
      *
      * @param conn A {@link Connection} object.
      * @return A {@link Cleanable} instance. Calling the {@link Cleanable#close()} method will close the connection and release all resources.
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -115,12 +115,8 @@ public class Handle implements Closeable, Configurable<Handle> {
         this.transactionHandler = transactionHandler.specialize(this);
         this.forceEndTransactions = !transactionHandler.isInTransaction(this);
 
-        addCleanable(() -> {
-            // shut down statement builder
-            if (checkConnectionIsLive()) {
-                statementBuilder.close(connection);
-            }
-        });
+        addCleanable(() ->
+            statementBuilder.close(connection));
     }
 
     /**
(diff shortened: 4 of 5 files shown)


hgschmie:
This might need a bit more thinking,. It is one of these "damned if I do, damned if I don't" situation because of the plethora of drivers. The check exists because there are a number of drivers that throw a "ConnectionAlreadyClosedException" if calling Connection#close on a closed connection (yes, it's against the spec, but but ...).  Making the close unconditional might pop up more errors in more popular paths.

The isAlive check was ported over in the 3.35.0 rewrite from the previous code. 

I can take a closer look at the change over the weekend when I am no longer in an airport departure lounge, please hold off merging till then. 

The bug report worries me. It implies that there are connection pools that report a connection "dead" (IAW closed) while still holding onto the actual database driver connection. This would be a pool bug.

stevenschlansker:
No urgency to merge this.

Throwing a `ConnectionAlreadyClosedException` is explicitly violating the spec. Can you be more specific about which driver(s) do this?

I think it is understandable that if the pool leases a connection, it expects it to be returned, even if the database server asynchronously-closes the connection. Maybe there is a pool bug but I am not as sure as you are.

kristoffSC:
Hi,
Thanks for looking at this, I'm an author of that `[bug report](https://github.com/jdbi/jdbi/issues/2446)`.

kristoffSC:
Hi,
Thanks for looking at this, I'm an author of that [bug report](https://github.com/jdbi/jdbi/issues/2446).

Regarding:
> It implies that there are connection pools that report a connection "dead" (IAW closed) while still holding onto the actual database driver connection. This would be a pool bug.

vs
> I think it is understandable that if the pool leases a connection, it expects it to be returned

I think that the latter is more accurate for my bug report.
The bottom line is that jdbi will not return borrowed object to the pool. It's up to the pool what should be done with such object.

In my case, the connection was never returned to the pool. The pool can have "health" checks while creating, borrowing or returning object to the pool but in this case this would not work since connection was closed after it was acquired from the pool and before `Cleanable` was registered on the Handle object which happens in jdbi.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -363,6 +365,15 @@ public Handle open() {
             LOG.trace("Jdbi [{}] obtain handle [{}] in {}ms", this, h, MILLISECONDS.convert(stop - start, NANOSECONDS));
             return h;
         } catch (SQLException e) {
```
kristoffSC, line 367: This will not catch exception thrown from Jdbi's PostgresPlugin. This one throws `LargeObjectException`

stevenschlansker: Good point, updated.

stevenschlansker: @kristoffSC please take another look and see if it looks good now :)

hgschmie:
> If drivers violating the spec is a problem, we can catch and ignore the thrown exception. Checking the connection `isValid` isn't a proper way to avoid this anyway, since it leaves open a race where the connection is closed behind your back between the valid check and doing the close.

Yeah, but by doing so, we now penalize all the code that behaves correctly (reporting the connection as dead when it is dead). Catching the exception is expensive.

This seems to be a case where we should just have a config switch. If the underlying pool/thing requires an unconditional close, you can set a flag and we force that.

hgschmie:
> Hi, Thanks for looking at this, I'm an author of that [bug report](https://github.com/jdbi/jdbi/issues/2446).
> 
> Regarding:
> 
> > It implies that there are connection pools that report a connection "dead" (IAW closed) while still holding onto the actual database driver connection. This would be a pool bug.
> 
> vs
> 
> > I think it is understandable that if the pool leases a connection, it expects it to be returned
> 
> I think that the latter is more accurate for my bug report. The bottom line is that jdbi will not return borrowed object to the pool. It's up to the pool what should be done with such object.
> 
> In my case, the connection was never returned to the pool. The pool can have "health" checks while creating, borrowing or returning object to the pool but in this case this would not work since connection was closed after it was acquired from the pool and before `Cleanable` was registered on the Handle object which happens in jdbi.

Thank you for the explanation but that is then a pool bug. We call "checkConnectionLive()" which needs to return `true` so that we close the connection. This code calls `Connection#isClosed()` which is explicitly defined as `A connection is closed if the method close has been called on it or if certain fatal errors have occurred. This method is guaranteed to return true only when it is called after the method Connection.close has been called.`. So the pool connection returned "I am closed, no need to do anything", which then was not correct. This is not a Jdbi bug, but a pool bug.

Is there a specific pool library that shows that behavior? (please don't say "c3p0").

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -362,7 +363,16 @@ public Handle open() {
             }
             LOG.trace("Jdbi [{}] obtain handle [{}] in {}ms", this, h, MILLISECONDS.convert(stop - start, NANOSECONDS));
             return h;
-        } catch (SQLException e) {
+        } catch (Exception e) {
```
hgschmie, line 366: this makes me super-uncomfortable. There is a single bug report about a problem with connection closing and this is a change at one of the most fundamental pieces of Jdbi, affecting the most fundamental path in the whole code base which is probably executed a billion times a day all over the java ecosystem without any problems.

Why touch this and if touching it, there needs to be a bunch of tests that *prove* that this low level code (this replaces one line of straightforward catch and rethrow with a bunch of manual resource manipulations) actually is an improvement. I don't believe it is.

hgschmie: I spend a lot of time and wrote a lot of tests in the 3.35.0 cycle to prove that the new code not only behaves better but does not leak resources anymore. I don't think introducing this manual piece-by-piece resource releasing and bypassing the mechanism that is used everywhere else in post-3.35.0 code is the right thing to do.

hgschmie (changes requested):

hgschmie:
I actually found the real bug. It is embarrassing.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
kristoffSC: Hi @hgschmie 
I totally understand your concerns. 

> There is a single bug report about a problem with connection closing

This is true, it is a single bug report and it was originally about connection closing. However yesterday I have updated the bug report because it turns out that the problem that I found is more generic and case when db connection was closed can be one of the "implementation" of this issue.

The pool has basically two sets of methods, 
one for acquiring the object and second one for returning/destroying the object when it is no longer needed. Both methods are called by pool client which in this case is JDBI. It is the pool client who knows when to return the object back to the pool, so its up to the caller (JDBI i this case) to make sure that object is always return back to the pool right? At lest to make sure that proper API is called. Forgive me but I can not say that this is a pool bug, but maybe I'm missing something.

The `org.jdbi.v3.core.Jdbi::open()` gets object from the pool. If there will be an exception thrown anywhere in `org.jdbi.v3.core.Jdbi::open()` method, after the connection was acquired from pool, the connection object will not be returned back to the pool.

I would say that `org.jdbi.v3.core.Jdbi::open()` should have exception handling logic which will guaranty that connection object will be properly returned to the pool in case of exception. Currently it is not the case.

**The specific use cases for connection close:**

1. The connection is closed from the client side, as described in this [bug report](https://github.com/jdbi/jdbi/issues/2446).
In this case it turns out that exception will be thrown by this [line](https://github.com/jdbi/jdbi/blob/299268ba7b00a2845e05f7942fb416147d603f18/core/src/main/java/org/jdbi/v3/core/Handle.java#L115C17-L115C17) in Handle's constructor.
this.transactionHandler = transactionHandler.specialize(this); so even before registering any Cleanable objects
In this case, since exception was thrown from Handle constructor, there is no Handle instance to close/clean at the end.

2. The second use case is where the connection is closed from server side, for example Data Base restart.
In our system we are using Jdbi's PostgresPlugin. In org.jdbi.v3.core.Jdbi::open() method, after Handle object is created there is this call:
```
for (JdbiPlugin p : plugins) {
        h = p.customizeHandle(h);
}
```
This will throw an exception for PostgresPlugin.
In result the open() method will not return any Handle reference so no Handle will be closed/clean up hence connection object will not be returned to the pool. The stack trace for this case -
[JdbiTrace.txt](https://github.com/jdbi/jdbi/files/12276766/JdbiTrace.txt)

To reproduce this, please pull my repository (I added PostgresPlugin) and simply restart Postgresql container on a break point just after where connection object is acquired from the pool.

However this problem will happen in case of any exception thrown in Jdbi::open() method, connection issue is only one of them.

kristoffSC:
Hi @hgschmie 
> Thank you for the explanation but that is then a pool bug. We call "checkConnectionLive()" which needs to return `true` so that we close the connection. This code calls `Connection#isClosed()` which is explicitly defined as `A connection is closed if the method close has been called on it or if certain fatal errors have occurred. This method is guaranteed to return true only when it is called after the method Connection.close has been called.`. So the pool connection returned "I am closed, no need to do anything", which then was not correct. This is not a Jdbi bug, but a pool bug.

Not sure if I can agree with that this is pool bug. Originally bug report was about connection closing, but after further investigation I believe that there is an issue with exception handling in `org.jdbi.v3.core.Jdbi::open()`. I believe that its up to the client code (JDBI code) to make sure or at least try to return borrowed object to the pool in any case. Please see my comment here for details: https://github.com/jdbi/jdbi/pull/2451#discussion_r1286715734

> Is there a specific pool library that shows that behavior? (please don't say "c3p0").
I'm using `org.apache.commons.dbcp2.PoolingDataSource` wrapping `org.apache.commons.pool2.impl.GenericObjectPool`
My JDBI setup looks like this:

```
BasicDataSource dataSource = new BasicDataSource();
    dataSource.setUrl(container.getJdbcUrl());
    dataSource.setPassword("password");
    dataSource.setUsername("user");
    dataSource.setMinIdle(2);
    dataSource.setMaxIdle(4);
    dataSource.setMaxTotal(6);

    this.jdbi = Jdbi.create(dataSource)
        .installPlugin(new SqlObjectPlugin())
        .installPlugin(new PostgresPlugin());
```

stevenschlansker:
> Yeah, but by doing so, we now penalize all the code that behaves correctly (reporting the connection as dead when it is dead). Catching the exception is expensive.

I think this is the other way around: the current code penalizes all code that behaves correctly, because calling `isClosed` does a lot of work (in Pg, it will require a server roundtrip), and we can always safely close without that roundtrip. The Pg code is careful that calling close() a second time is very cheap. If there are other concrete databases that behave differently, we can talk about that, but we need a specific example else we should go with the spec which says "don't use isClosed since it is not a reliable test; close is safe to call multiple times". They specifically call out not to rely on it.

Throwing an exception on duplicate close is forbidden by spec, so it will not happen, so it has no cost. Unless there's a specific case of a driver that incorrectly does this.

> This seems to be a case where we should just have a config switch. If the underlying pool/thing requires an unconditional close, you can set a flag and we force that.

If we have a specific database where this is needed, we can add it. I don't think we should pre-emptively add configuration in contradiction to the JDBC spec without a specific reason. That just adds complexity.

I agree with Kristoff that we have a potential leak in `Jdbi.open`, if e.g. a plugin `customizeConnection` throws, we never return the Handle nor close the Connection.

stevenschlansker:
The changes you pushed over mine undoes the cleanup of resources.

I wrote two tests: one to show the leak when customization fails, and one to show the leak when isClosed throws. Both tests fail in master today, and pass with this fix.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
stevenschlansker: I added test cases to show this behavior in a unit test that we can commit to the repo.

stevenschlansker:
closed in favor of #2458

PR #2456, opened 2023-08-07
Remove findbugs annotations

stevenschlansker:
As discussed in #2408, we should prefer the `jakarta` annotations, but that's not what actually got committed here. So clean that up

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Sql.java
+++ b/core/src/main/java/org/jdbi/v3/core/Sql.java
@@ -18,7 +18,8 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
-import edu.umd.cs.findbugs.annotations.NonNull;
+import jakarta.annotation.Nonnull;
+
 import org.jdbi.v3.meta.Beta;
 
 /**
@@ -103,7 +104,7 @@ public final class Sql implements CharSequence {
         return str.charAt(index);
     }
 
-    @NonNull
+    @Nonnull
     @Override
     public CharSequence subSequence(int start, int end) {
         return str.subSequence(start, end);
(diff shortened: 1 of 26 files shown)


core/src/main/java/org/jdbi/v3/core/Sql.java
```diff
@@ -18,7 +18,8 @@
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
-import edu.umd.cs.findbugs.annotations.NonNull;
```
hgschmie, line 21: spotbugs-annotations

hgschmie (changes requested):
The code in the PR replaces annotations from the spotbugs-annotations jar, not the findbugs jsr305 jar. That was eliminated in #2408. 

While it good to do so, the PR also needs to remove the jar references.

I am ok with the PR but it needs to get updated to remove the spotbugs-annotation references.

It may also make sense to add a section to the build pom akin to the basepom:

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-dependency-plugin</artifactId>
    <configuration>
        <ignoredDependencies combine.children="append">
            <ignoredDependency>jakarta.annotation:jakarta.annotation-api</ignoredDependency>
    </configuration>
</plugin>
```

hgschmie (approved):

hgschmie:
actually never mind, it seems I have unlearned on how to read diffs. :-) All good. The exclusion block would still be good.

core/src/main/java/org/jdbi/v3/core/interceptor/JdbiInterceptionChainHolder.java
```diff
@@ -34,15 +34,6 @@
 @Alpha
 public final class JdbiInterceptionChainHolder<S, T> {
 
-    @SuppressWarnings("UnnecessaryLambda") // constant for readablity
```
hgschmie, line 37: I would have preferred that change in a separate PR. It kind of gets lost between the annotation changes.

stevenschlansker:
Why should we add a dependency ignore to the checker plugin? It passes the plugin checks fine right now.
I'll merge this and we can always tweak the build in a followup.

PR #2458, opened 2023-08-08
Ensure that the connection is always closed

### LINKED ISSUE #2446: [Bug] Jdbi3-core - connection object can be stuck in a connection pool forever if connection was closed.

hgschmie:
Even if the init code in Jdbi or Handle fails, ensure that we don't leak connections.

Fixes #2446

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+# Unreleased
+  - Handle: fix potential Connection leak if connection is closed early
+
 # 3.40.0
 
   - add extension point to decorate withHandle and friends (#2448)
--- a/core/src/main/java/org/jdbi/v3/core/ConnectionFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/ConnectionFactory.java
@@ -45,6 +45,7 @@ public interface ConnectionFactory {
 
     /**
      * Returns a {@link Cleanable} that will close the connection and release all necessary resources.
+     * Like {@link Connection#close()}, the Cleanable should no-op if called more than once.
      *
      * @param conn A {@link Connection} object.
      * @return A {@link Cleanable} instance. Calling the {@link Cleanable#close()} method will close the connection and release all resources.
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -115,12 +115,8 @@ public class Handle implements Closeable, Configurable<Handle> {
         this.transactionHandler = transactionHandler.specialize(this);
         this.forceEndTransactions = !transactionHandler.isInTransaction(this);
 
-        addCleanable(() -> {
-            // shut down statement builder
-            if (checkConnectionIsLive()) {
-                statementBuilder.close(connection);
-            }
-        });
+        addCleanable(() ->
+            statementBuilder.close(connection));
     }
 
     /**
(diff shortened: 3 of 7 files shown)


stevenschlansker (approved):

stevenschlansker:
@kristoffSC, do you want to verify this proposed change fixes the connection leak for you?

kristoffSC:
> @kristoffSC, do you want to verify this proposed change fixes the connection leak for you?

@hgschmie @stevenschlansker 
I confirm that this PR fixes https://github.com/jdbi/jdbi/issues/2446

Is there any estimate date for new release?
Thanks.

PR #2460, opened 2023-08-09
Revise the HandleCallback decorator code

hgschmie:
- Rename "Handler" to "HandleCallbackDecorator". "Handler" is a way too generic name and there is potential for confusion with "Handle" which is a main concept in Jdbi while this change introduces a minor feature.
- Mark the added code with `@Alpha` annotation, as the API should not have been set in stone.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/HandleCallbackDecorator.java
+++ b/core/src/main/java/org/jdbi/v3/core/HandleCallbackDecorator.java
@@ -13,8 +13,15 @@
  */
 package org.jdbi.v3.core;
 
-public interface Handler {
-    Handler STANDARD_HANDLER = new Handler() {
+import org.jdbi.v3.meta.Alpha;
+
+/**
+ * Decorates the {@link HandleCallback} instance for {@link Jdbi#useHandle(HandleConsumer)}, {@link Jdbi#withHandle(HandleCallback)},
+ * {@link Jdbi#inTransaction(HandleCallback)} and {@link Jdbi#useTransaction(HandleConsumer)}.
+ */
+@Alpha
+public interface HandleCallbackDecorator {
+    HandleCallbackDecorator STANDARD_HANDLE_CALLBACK_DECORATOR = new HandleCallbackDecorator() {
         @Override
         public <R, X extends Exception> HandleCallback<R, X> decorate(HandleCallback<R, X> callback) {
             return callback;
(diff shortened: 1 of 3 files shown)


hgschmie:
TBH, I am pretty unhappy that #2448 was merged right before the 3.40.0 release and then immediately released. This PR marks the code as `@Alpha` (it should have been alpha to begin with IMHO).

It is not clear to me what that code is intended to accomplish. 

The way I see it, the code can do one of two things:

One use case is to select an alternate HandleCallback instead of the one supplied. That is similar to one test where within the decorator, the code tests and does not return the HandleCallback passed in but an alternate version. This is IMHO much simpler accomplished by doing

```java
var handleCallback = someCondition()
    ? handleCallback1
    : handleCallback2;

var result = jdbi.withHandle(handleCallback);
```

The other use case is that the decorator is used to modify the handle:

```java
    var decorator = new HandleCallbackDecorator() {
        @Override
        public <R, X extends Exception> HandleCallback<R, X> decorate(HandleCallback<R, X> callback) {
            return handle -> {
                handle.setReadOnly(true); // this is the handle modification code
                return callback.withHandle(handle);
            };
        }
    };

    jdbi.setHandleCallbackDecorator(decorator);

    jdbi.withHandle(...); // handle here will always be read-only
```

This seems to be a very complicated way to do this. The same effect can be accomplished today by doing

```java
    var plugin = new JdbiPlugin() {
        @Override
        public Handle customizeHandle(Handle handle) throws SQLException {
            handle.setReadOnly(true);
            return handle;
        }
    };

    jdbi.installPlugin(plugin);
```

which is as global as the registered HandleCallbackDecorator (all handles will pass through the `customizeHandle` method) and accomplished the same effect.

The feature is undocumented (no examples, no adding to index.adoc). This is destined to be another of the myriad of undocumented features that were added at some time or another for a very specific use case without documentation and in a year or so, no one remembers why.

@Randgalt could you elaborate what you are trying to do? I read the description about `Specialized behavior can be added for an app's entire JDBI usage without project devs needing to do anything special. This is very useful when introducing protocols such as custom retry behavior, custom connection management,[...]` and I would like to understand what specific thing can not be done with a Jdbi plugin and needs this specific decoration of the handle callback? 

Also, it would be good if you could contribute some documentation and examples beyond the test.

Randgalt:
We are working with a database, CRDB, that has some unfortunate edge cases. The TL;DR is that our connection pool can get poisoned with invalid connections that cannot be discovered without making a query with the connection. Inside of a transaction this is not a problem as we have a custom JDBI `TransactionHandler` that allows us to notice this event and substitute a good connection for the bad one. In non-transactions, however, we have no opportunity to do this. 

> This is IMHO much simpler accomplished by doing

This requires changes at every call site. This change I submitted makes it possible to hook every instance of withHandle/useHandle just as we do with the `TransactionHandler`. Going through our code and wrapping every call to `useHandle` is a non-starter. 

I hope this helps.

hgschmie:
Thanks for the example, this is useful. It would be good to contribute some documentation (e.g for chapter 14 or chapter 19) about the callback so that this feature is properly documented.

stevenschlansker (approved):

Randgalt:
https://github.com/jdbi/jdbi/pull/2463

PR #2470, opened 2023-08-16
Jdbi JPMS (java module) support

hgschmie:
*THIS IS NOT FOR THE FAINT OF HEART*

Add full JPMS module support using the moditect jpms tooling.

Adds an integration test suite that integration-tests all generated jars using their
own unit tests.

This is at best lightly tested and probably requires a lot of additional works

- Main jars are modularized using the moditect jpms tooling
- Test jars are non-modularized

Please test. good luck. ;-)

See also #2453 / @SentryMan


SentryMan:
the only thing I can really say against it is that you define a proper module-info file and have moditect pick it up(much easier to read than XML). [see this](https://github.com/javalin/javalin/pull/1923/files)

PR #2473, opened 2023-08-18
pom fixes

hgschmie:
Some small fixes that came out of the JPMS work

- don't add `basepom.it.skip` to maven modules that have integration
  tests, add it globally.
- have the core artifact declare a dep on checker-qual (which is
  ignored by the dep plugin so it never showed as a bug and is an
  optional provided dep. This was found by the JPMS work).
- move version for gson-java8 to the build pom
- change the way we add JPMS automatic named modules, don't use
  the .jdbi-parent files any longer; only check and enforce for
  jar artifacts. Activate the automatic module name by checking
  for module-info.java file.

### DIFF
--- a/internal/build/pom.xml
+++ b/internal/build/pom.xml
@@ -750,6 +750,7 @@
                         <configLocation>policy/checkstyle.xml</configLocation>
                         <suppressionsLocation>policy/checkstyle-suppress.xml</suppressionsLocation>
                         <sourceDirectories>${project.build.sourceDirectory},${project.build.testSourceDirectory}</sourceDirectories>
+                        <excludes>**\/module-info.java</excludes>
                     </configuration>
                     <dependencies>
                         <dependency>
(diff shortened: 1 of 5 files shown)


internal/build/pom.xml
```diff
@@ -868,18 +863,59 @@
...
-            <id>in-parent</id>
+            <id>jdbi.jpms-auto-module</id>
             <activation>
+                <property>
+                    <name>packaging</name>
+                    <value>jar</value>
+                </property>
                 <file>
-                    <exists>.jdbi-parent</exists>
+                    <missing>src/main/java/module-info.java</missing>
```
stevenschlansker, line 874: Why do we want conditional auto-module activation? I would have thought it would be an auto-module in main branch, and then as part of the module PR work we change it to not use auto-modules at all.

hgschmie: because right now, we don't have any explicit modules, so we ship everything with auto modules. This is refactoring the existing code (removing the .jdbi-parent touch files).

The explicit JPMS code is still a long way out. This allows us to slowly fade it in (jdbi maven modules with a module-info will use that, all the others will use automatic module names)

stevenschlansker (approved):

PR #2481, opened 2023-09-01
Support '.' and '-' as characters in named parameters

### LINKED ISSUE #2471: Can't bind list named parameter if the alias contains a dot
Hi and thanks for JDBI,
I've noticed that bindList doesn't work correctly if the name of a list named parameters contains a dot, e.g. <like.this>, it seems that when building the query the parameter list is not being replaced in the query in this case. Normal named parameters with a dot work without any issues.

For example, this snippet works:

        Jdbi jdbi = Jdbi.create("jdbc:h2:mem:test;Mode=Oracle;DB_CLOSE_DELAY=-1"); // (H2 in-memory database)
        jdbi.setSqlLogger(new Slf4JSqlLogger());

        List<String> names = jdbi.withHandle(handle -> {
            handle.execute("CREATE TABLE \"user\" (id INTEGER PRIMARY KEY, name_id VARCHAR)");

            handle.createUpdate("INSERT INTO \"user\" (id, name_id) VALUES (?, ?)")
                    .bind(0, 1) // 0-based parameter indexes
                    .bind(1, "Bob")
                    .execute();

            handle.createUpdate("INSERT INTO \"user\" (id, name_id) VALUES (:id, :name)")
                    .bind("id", 2)
                    .bind("name", "Clarice")
                    .execute();

            List<String> res = handle.createQuery("SELECT name_id FROM \"user\" WHERE ((id >:dot.id) AND name_id IN (<names>) AND id >:dot.id)")
                    .bindList("names",List.of("1","2","Bob","Clarice"))
                    .bind("dot.id",-1)
                    .mapTo(String.class)
                    .list();

But when I rename `names` to `nam.es`:

            List<String> res = handle.createQuery("SELECT name_id FROM \"user\" WHERE ((id >:dot.id) AND name_id IN (<nam.es>) AND id >:dot.id)")
                    .bindList("nam.es",List.of("1","2","Bob","Clarice"))
                    .bind("dot.id",-1)
                    .mapTo(String.class)
                    .list();

The example above fails with an exception (generic H2 exception):

      org.jdbi.v3.core.statement.UnableToCreateStatementException: org.h2.jdbc.JdbcSQLSyntaxErrorException: Syntax error in SQL statement "SELECT name_id FROM ""user"" WHERE ((id >?) AND name_id IN ([*]<nam.es>) AND id >?)"; expected "INTERSECTS (, NOT, EXISTS, UNIQUE, INTERSECTS"; SQL statement:
      SELECT name_id FROM "user" WHERE ((id >?) AND name_id IN (<nam.es>) AND id >?) [42001-220] [statement:"SELECT name_id FROM "user" WHERE ((id >:dot.id) AND name_id IN (<nam.es>) AND id >:dot.id)", arguments:{positional:{}, named:{dot.id:-1,__nam.es_1:2,__nam.es_2:Bob,__nam.es_3:Clarice,__nam.es_0:1}, finder:[]}]
      
	      at org.jdbi.v3.core.statement.SqlStatement.internalExecute(SqlStatement.java:1789)
	      at org.jdbi.v3.core.result.ResultProducers.lambda$createResultBearing$3(ResultProducers.java:94)
	      at org.jdbi.v3.core.result.internal.ResultSetResultIterable.iterator(ResultSetResultIterable.java:51)
	      at org.jdbi.v3.core.result.ResultIterable.stream(ResultIterable.java:266)
	      at org.jdbi.v3.core.result.ResultIterable.collect(ResultIterable.java:340)

I would expect to see something like `AND name_id IN (__nam.es_0,__nam.es_1, __nam.es_2,__nam.es_3)` but it seems that the replacement is not taking place.
Luckily, just not using dots for those aliases works as a workaround.

hgschmie:
Fixes #2471

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  - Support `.` and `-` as part of named parameter names. (#2471)
   - Fix incorrect attempt to rollback txn when exception is thrown after commit (#2478)
 
 # 3.41.0
--- a/core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g4
+++ b/core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/ColonStatementLexer.g4
@@ -22,7 +22,7 @@ fragment COLON: {_input.LA(2) != ':'}? ':';
 fragment DOUBLE_COLON: {_input.LA(2) == ':'}? '::';
 fragment QUESTION: {_input.LA(2) != '?'}? '?';
 fragment DOUBLE_QUESTION: {_input.LA(2) == '?'}? '??';
-fragment NAME: JAVA_LETTER | [0-9] | '.' | '?.';
+fragment NAME: JAVA_LETTER | [0-9] | '.' | '?.' | '-';
 
 /* Lovingly lifted from https://github.com/antlr/grammars-v4/blob/master/java/JavaLexer.g4 */
 fragment JAVA_LETTER : [a-zA-Z$_] | ~[\u0000-\u007F\uD800-\uDBFF] | [\uD800-\uDBFF] [\uDC00-\uDFFF];
--- a/core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/DefineStatementLexer.g4
+++ b/core/src/main/antlr4/org/jdbi/v3/core/internal/lexer/DefineStatementLexer.g4
@@ -19,7 +19,7 @@ fragment ESCAPE_QUOTE: ESCAPE QUOTE ;
 fragment DOUBLE_QUOTE: '"' ;
 fragment LT: '<' ;
 fragment GT: '>' ;
-fragment NAME: JAVA_LETTER | [0-9];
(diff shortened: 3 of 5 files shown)


stevenschlansker (approved):

stevenschlansker:
Thank you for reporting this regression. @hgschmie , do we need to revert this change?

stevenschlansker:
Or, maybe it would be sufficient to allow `-` only when it is followed by more alphabetic characters, so that `->` and `-3` are unambiguously operators not names.

hgschmie:
@gokristian https://github.com/jdbi/jdbi/pull/2499 should fix this. I plan to release a 3.41.3 around this regression.

hgschmie:
I changed the grammar to match the old version and not accept `-` at the end. To be able to differentiate between "next character is alphanumeric or not", this requires lookahead which can probably be done by the lexer but I don't know exactly how. :-)

hgschmie:
@gokristian again, thank you for reporting the regression. Please try with 3.41.3, this fixes the problem.

gokristian:
@hgschmie No problem, thank you for fixing it. 3.41.3 works great with param json field accessor!

Issue #2415, opened 2023-06-29
jdbi should have a MySQL artifact


stevenschlansker:
We could support `set` mysql type, and it would be helpful to have a test rule that spins up a mysql database to write unit tests against. Then we should write tests that date types work reasonably.

Ref: #1911 #1806 #1614

stevenschlansker:
Also would help: #2402

hgschmie:
We have that: 

```java
@Testcontainers
class MySQLTest {

    @Container
    static JdbcDatabaseContainer<?> dbContainer = new MySQLContainer<>("mysql");

    @RegisterExtension
    JdbiExtension extension = JdbiTestcontainersExtension.instance(dbContainer);

...
}
```

This should work for most of our test suite. Yes, using testcontainers is annoying but not worse than using the OtjPg library.

stevenschlansker:
We'd want to wrap that up in an extension for reuse like `JdbiH2Extension`. Also, would probably want to include the driver.

hgschmie:
This would be similar to the [OtjExtension](https://github.com/jdbi/jdbi/blob/master/testing/src/main/java/org/jdbi/v3/testing/junit5/JdbiOtjPostgresExtension.java). Not sure that this warrants another module (it would be a module similar to testing), we can just add this to testing. 

The glue code that exists in the otj postgres extension still would need to be written, I would consider that "outside jdbi scope", but it also should not be too hard, it might boil down to code similar to [EmbeddedPostgres](https://github.com/opentable/otj-pg-embedded/blob/master/src/main/java/com/opentable/db/postgres/embedded/EmbeddedPostgres.java), only for MySQL.

hgschmie:
The main reason I put the testcontainers code together is to avoid having lots of special case "for database x" modules. The testcontainer thing is pretty straightforward. For H2, Postgres and Hsql, we already had a large testsuite so those were "grandfathered" in. 

I am not objecting to having a `JdbiMysqlExtension`, in fact I would welcome it. Someone has to write it, though. :'-)

Issue #2402, opened 2023-06-21
@SqlQuery containing "LIKE" returns no results


marmor7:
Hello,

I have a query that uses MySql "LIKE" that runs fine directly in MySQL returning expected results, 
however when run within Java using @SqlQuery it returns no results.

Here's the query I need to run:
```
SELECT contact_id
FROM contacts
WHERE etag LIKE CONCAT('%',?)  LIMIT 1
```

And here's the 4 different options I've tried, all producing 0 results:
```
@SqlQuery("SELECT contact_id " +
    " FROM contacts " +
    " WHERE etag LIKE CONCAT('%',:etagSuffix) " +
    " LIMIT 1")

@SqlQuery("SELECT contact_id " +
    " FROM contacts " +
    " WHERE etag LIKE '%':etagSuffix " +
    " LIMIT 1")

@SqlQuery("SELECT contact_id " +
    " FROM contacts " +
    " WHERE etag LIKE '%' || :etagSuffix " +
    " LIMIT 1")

// in this option I pass ("%" + etagSuffix) instead of just etagSuffix
@SqlQuery("SELECT contact_id " +
    " FROM contacts " +
    " WHERE etag LIKE :etagPattern " +
    " LIMIT 1")
```

I've tried it on an older jdbi 3.8.2 version, and also on the latest 3.39.1.

Would be glad to understand how the above query can be run or if this is a bug in jdbi
Thanks.

stevenschlansker:
Hi @marmor7 , I am not sure what is wrong here. Can you show the method signature for the methods your `@SqlQuery` annotations are attached to, just for completeness? I think we might have to add a `jdbi3-mysql` artifact as a home for things like mysql unit tests...

marmor7:
Sure, here are the 4 methods I tried:

```
@SqlQuery("SELECT contact_id " +
    " FROM contacts " +
    " WHERE etag LIKE :etagPattern " +
    " LIMIT 1")
public List<MySqlContact> getContactByEtagSuffix1(@Bind("etagPattern") String etagPattern);

@SqlQuery("SELECT contact_id " +
    " FROM contacts " +
    " WHERE etag LIKE CONCAT('%',:etagSuffix) " +
    " LIMIT 1")
public List<MySqlContact> getContactByEtagSuffix2(@Bind("etagSuffix") String etagSuffix);

@SqlQuery("SELECT contact_id " +
    " FROM contacts " +
[...]

PR #2488, opened 2023-09-18
Start a mysql module

hgschmie:
Currently does not create an artifact, just runs tests.

Tests are disabled by default (because they use testcontainers which is
slow).

Run tests with `mvn -Dbasepom.test.skip=false -pl :jdbi3-mysql clean test`


hgschmie:
related to #2402 (tests run are related to the queries in question) and #2415

stevenschlansker (approved):

PR #2492, opened 2023-09-20
Fix transactions for non-autocommitting handles

### LINKED ISSUE #2491: Handle.begin() must be explicitly called in transaction
On the latest version (`3.41.1`), if `autocommit` is disabled, you need to explicitly call `handle.begin()` before any insert/update/delete to make any changes. If not, it silently exits without making any changes.

Not sure if it was an intended change or if the previous behaviour was wrong, but it wasn't clear from the release notes or documentation that this would be needed.

Example:
```java
// ... jdbi - postgres connection with autocommit set to false
int result = jdbi.inTransaction(handle -> {
      int inserted = handle.execute("INSERT INTO example_table (example_column) VALUES ('example_value')");
      handle.commit();
      return inserted;
    });
```

Running the above code on `3.41.0`, will insert into DB.
Running the above code on `3.41.1`, will not insert into DB.

To fix the above on `3.41.1` you would need to add `handle.begin()`:
```java
// ... jdbi - postgres connection with autocommit set to false
int result = jdbi.inTransaction(handle -> {
      handle.begin();
      int inserted = handle.execute("INSERT INTO example_table (example_column) VALUES ('example_value')");
      handle.commit();
      return inserted;
    });
```

From what I found, this might be caused by the new logic around `handlerState` property added in the `BoundLocalTransactionHandler` implementation (#2479).

hgschmie:
If a handle is by default using autocommit == false, then transactions
could be commited (or rolled back) without having to call
Handle#begin() first. This broke in 3.41.1.

Restore the old behavior. Fixes #2491

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,7 @@
 # Unreleased
 
   - Deprecate the `otjPostgres` support in jdbi-testing. This will be undeprecated if they ship a version that provides an automatic module name for JPMS, otherwise it will be removed when Jdbi ships with full JPMS support.
+  - Restore pre-3.41.0 behavior for handles using auto-commit == false where transactions don't need `Handle#begin()` before `Handle#commit()` (#2491, thanks @grigorem)
 
 
 # 3.41.1
--- a/core/src/main/java/org/jdbi/v3/core/Handle.java
+++ b/core/src/main/java/org/jdbi/v3/core/Handle.java
@@ -114,7 +114,7 @@ public class Handle implements Closeable, Configurable<Handle> {
 
         // both of these methods are bad because they leak a reference to this handle before the c'tor finished.
         this.transactionHandler = transactionHandler.specialize(this);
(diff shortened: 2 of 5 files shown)


stevenschlansker (approved):
Is this expected to work? It seems a little weird to hand off an in-flight transaction to Jdbi. But I suppose breaking user code sucks too

hgschmie:
well, not sure how to answer that. But up until 3.41.1, you could turn autocommit off on a connection, pile a bunch of statements onto it and then call `Handle#commit()`. You probably *shouldn't* but it still worked. 

What the change does is basically looking at the commit state of the connection when the transaction handler is created and if it has autocommit already turned off, it moves the state machine "one state further". So the handler is in the same state if you opened it with a connection that has auto-commit turned off as with a regular (autocommit on) + `Handle#begin()`. 

I still think it is a cornercase that the user stumbled upon and if it hadn't worked in the past, I would not make that change. I also think it is benign for "regular" use one can call begin multiple times.

PR #2496, opened 2023-09-28
Add reference to `@DatabaseValue` to the documentation

hpoettker:
The PR adds references to the annotation `@DatabaseValue` from #1558 in the reference documentation. The annotation is a really neat feature but currently not mentioned in the reference documentation.


hgschmie:
Hi @hpoettker!

Thank you for your contribution to Jdbi!

PR #2507, opened 2023-10-05
Silence Java 21 warnings

hgschmie:
- warning about annotation processing without `-proc:full`
- warning about dynamic byte code agent loading (fix is only temporary,
  JDK 23 will most likely permanently disable dynamic byte code loading

### DIFF
--- a/internal/build/pom.xml
+++ b/internal/build/pom.xml
@@ -875,6 +875,31 @@
 
     <profiles>
         <profile>
+            <id>java21</id>
+            <activation>
+                <jdk>[21,)</jdk>
+            </activation>
+            <properties>
+                <!-- see https://github.com/mockito/mockito/issues/3037 -->
+                <basepom.it.arguments>-XX:+EnableDynamicAgentLoading</basepom.it.arguments>
+                <basepom.test.arguments>-XX:+EnableDynamicAgentLoading</basepom.test.arguments>
+            </properties>
+            <build>
+                <pluginManagement>
+                    <plugins>
+                        <plugin>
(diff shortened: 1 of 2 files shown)


internal/build/pom.xml
```diff
@@ -874,6 +874,31 @@
...
+            </properties>
+            <build>
+                <pluginManagement>
+                    <plugins>
+                        <plugin>
+                            <groupId>org.apache.maven.plugins</groupId>
+                            <artifactId>maven-compiler-plugin</artifactId>
+                            <configuration>
+                                <!-- see https://inside.java/2023/07/29/quality-heads-up/ -->
+                                <proc>full</proc>
```
stevenschlansker, line 895: should this configuration apply to all jdks, even earlier ones? It's not harmful to specify this and then we have less per-jdk configuration.

hgschmie: pre-JDK 21 does not support "full" as a valid value 🤦 ...

```
❯ java -version
openjdk version "17.0.8.1" 2023-08-24
OpenJDK Runtime Environment Temurin-17.0.8.1+1 (build 17.0.8.1+1)
OpenJDK 64-Bit Server VM Temurin-17.0.8.1+1 (build 17.0.8.1+1, mixed mode)
❯ javac -proc:full
error: invalid flag: -proc:full
Usage: javac <options> <source files>
use --help for a list of possible options

❯ java -version
openjdk version "21" 2023-09-19
OpenJDK Runtime Environment Homebrew (build 21)
OpenJDK 64-Bit Server VM Homebrew (build 21, mixed mode, sharing)
❯ javac -proc:full
error: no source files or class names
```

PR #2523, opened 2023-10-27
Refactor HandleSupplier management

hgschmie:
Pull the thread local into a pluggable object. This will open the path
to more sophisticated ways of managing handles across different execution
paths than just "by thread". This is most interesting right now for Kotlin
coroutines but also future structured concurrency will benefit from this.

This is a core piece of Jdbi, so basically every test checks whether this
change breaks anything. So there are no explicit tests.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Jdbi.java
+++ b/core/src/main/java/org/jdbi/v3/core/Jdbi.java
@@ -32,6 +32,7 @@ import org.jdbi.v3.core.extension.ExtensionFactory;
 import org.jdbi.v3.core.extension.Extensions;
 import org.jdbi.v3.core.extension.HandleSupplier;
 import org.jdbi.v3.core.extension.NoSuchExtensionException;
+import org.jdbi.v3.core.internal.HandleScope;
 import org.jdbi.v3.core.internal.OnDemandExtensions;
 import org.jdbi.v3.core.internal.exceptions.Unchecked;
 import org.jdbi.v3.core.spi.JdbiPlugin;
@@ -64,11 +65,9 @@ public class Jdbi implements Configurable<Jdbi> {
     private final AtomicReference<TransactionHandler> transactionhandler = new AtomicReference<>(LocalTransactionHandler.binding());
     private final AtomicReference<StatementBuilderFactory> statementBuilderFactory = new AtomicReference<>(DefaultStatementBuilder.FACTORY);
     private final AtomicReference<HandleCallbackDecorator> handleCallbackDecorator = new AtomicReference<>(HandleCallbackDecorator.STANDARD_HANDLE_CALLBACK_DECORATOR);
+    private HandleScope handleScope = HandleScope.threadLocal();
 
     private final CopyOnWriteArrayList<JdbiPlugin> plugins = new CopyOnWriteArrayList<>();
-
-    @SuppressWarnings("ThreadLocalUsage")
-    private final ThreadLocal<HandleSupplier> threadHandleSupplier = new ThreadLocal<>();
 
     private Jdbi(ConnectionFactory connectionFactory) {
         Objects.requireNonNull(connectionFactory, "null connectionFactory");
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -391,11 +413,9 @@ public Handle open() {
      * @throws X any exception thrown by the callback
      */
     public <R, X extends Exception> R withHandle(HandleCallback<R, X> callback) throws X {
+        final HandleCallback<R, X> decoratedCallback = handleCallbackDecorator.get().decorate(callback);
 
-        HandleSupplier handleSupplier = threadHandleSupplier.get();
-
-        HandleCallback<R, X> decoratedCallback = handleCallbackDecorator.get().decorate(callback);
-
+        final HandleSupplier handleSupplier = getHandleScope().get();
```
hgschmie, line 418: I keep going back and forth around "should this be `HandleScope handleScope = getHandleScope()` and then use the handleScope object everywhere.

This seems to be a case of second-guessing the JVM; my hunch is that it will inline the shallow getters anyway (they are on a very hot path) so it should not actually matter..

stevenschlansker: I am not thinking about the nano-optimization necessarily, although that's always a consideration. I am thinking along the lines of, for the resource to be managed correctly, the acquire / release calls must be paired. So it is most correct to ensure we always pair the calls to the same instance specifically, so you can store it in a local variable.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -332,6 +330,29 @@ public HandleCallbackDecorator getHandleCallbackDecorator() {
...
 
+    /**
+     * Returns the internal {@link HandleScope} object. The Jdbi instance uses this to provide handles in a given scope.
+     * The default scope is <i>per-thread</i>, so every thread manages its own handle.
+     * <br>
+     * <b>This is an internal method and not part of the public API!</b>
+     * @return A {@link HandleScope} object
+     * @param <T> type parameter
+     */
+    @Alpha
```
stevenschlansker, line 341: Non-public API is never supported or stable, so `@Alpha` doesn't make too much sense here IMO

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -332,6 +330,29 @@ public HandleCallbackDecorator getHandleCallbackDecorator() {
...
+    public final <T extends HandleScope> T getHandleScope() {
```
stevenschlansker, line 342: It would be really nice to find a way to hide this either as a non-public method, or in a class in the `.internal` package

core/src/main/java/org/jdbi/v3/core/internal/ThreadLocalHandleScope.java
```diff
@@ -0,0 +1,44 @@
...
+    protected final ThreadLocal<HandleSupplier> threadLocal = new ThreadLocal<>();
+
+    @Override
+    public HandleSupplier get() {
+        return threadLocal.get();
+    }
+
+    @Override
+    public void set(HandleSupplier handleSupplier) {
+        if (handleSupplier != null) {
```
stevenschlansker, line 33: These `null` semantics are subtly different from ThreadLocal. Should we just forbid calling `set(null)` at all? since we never intend to do that, it seems that it would be a bug.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
stevenschlansker, line 342: Do we actually want this `<T>` return type? Why not just return the interface? For callers to actually work if we substitute the implementation, they cannot rely on a specific subtype anyway.

stevenschlansker (commented):
This seems fine, except that it would be very nice to find a way to hide the `getHandleScope` api as not public.

core/src/main/java/org/jdbi/v3/core/internal/ThreadLocalHandleScope.java
hgschmie: good call.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
hgschmie: I removed that. It was for some convenience but in kotlin (as in modern java), it is not necessary.

hgschmie:
> This seems fine, except that it would be very nice to find a way to hide the `getHandleScope` api as not public.

I can not see a way to do that in a way that will work with the Kotlin extensions. it is an embedded object in the Jdbi instance and it needs to be exposed to the kotlin coroutine machinery somehow.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
hgschmie: removed it. Also put handlescope into the internal package, so it is not part of the public API footprint

Issue #2522, opened 2023-10-27
Jdbi should be able to work with Kotlin Coroutines


hgschmie:
See https://github.com/anderssv/jdbi/blob/handle-coroutines/kotlin/src/test/kotlin/org/jdbi/v3/core/kotlin/coroutines/CoRoutineContextTest.kt for an example 

From a mailing list post: (@anderssv)

We're using JDBI with the Kotlin additions, but as we switched to async for several methods we realized that jdbi.inTransaction {...} etc. uses a ThreadLocal to keep the transaction. Since Kotlin co-routines jump threads, this of course won't work. 

I've been looking and can not really find much. Is there something that I can use that will handle transactions, just with CoroutineContext instead of ThreadLocals?

hgschmie:
So I finally (!) found time to dig into this. I was using the test class above.

This can actually get to work. Even much simpler than what you have proposed in the PR. What is needed is

- split out the ThreadLocal in Jdbi into a holder object. I was planning to do that anyway because it is needed for Java Scoped Concurrency. In a way, kotlin coroutines are basically scoped concurrency in Kotlin land.
- Make that holder class configurable.
- Allow some kotlin code to add the threadlocal to the coroutine context using the `.asContextElement()` method.

Then use this as a context element. There is coordination required between kotlin coroutines and jdbi (I don't think that a "magic" class like ThreadLocal exists in Kotlin coroutine land).

I should have a PoC PR in a bit.

hgschmie:
Fixed by #2524

PR #2524, opened 2023-10-28
Kotlin coroutine support

hgschmie:
Support Kotlin coroutines

Allows using Jdbi objects (handles and extension objects) with kotlin coroutines. Objects can be shared between coroutines (even across threads) as long as the caller can guarantee that not multiple coroutines use them at the same time.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/Jdbi.java
+++ b/core/src/main/java/org/jdbi/v3/core/Jdbi.java
@@ -32,7 +32,6 @@ import org.jdbi.v3.core.extension.ExtensionFactory;
 import org.jdbi.v3.core.extension.Extensions;
 import org.jdbi.v3.core.extension.HandleSupplier;
 import org.jdbi.v3.core.extension.NoSuchExtensionException;
-import org.jdbi.v3.core.internal.HandleScope;
 import org.jdbi.v3.core.internal.OnDemandExtensions;
 import org.jdbi.v3.core.internal.exceptions.Unchecked;
 import org.jdbi.v3.core.spi.JdbiPlugin;
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4,6 +4,7 @@
 - Move from Kotlin 1.5 to 1.6 as 1.5 is deprecated and will be removed.
 - Add support for Function arguments, similar to Consumer arguments, to SQL objects (#2326)
 - Fix kotlin deprecation warnings (#2511) - Thanks @lwach-allegro for contributing a PR
+- Support Kotlin coroutine scope. (#2524) - Suggested by @anderssv on the mailing list
 
 # 3.41.3
 - Fix regression introduced by #2481 where `-` at the end of named parameters get swallowed. (#2499, thanks @gokristian for reporting).
--- a/core/src/main/java/org/jdbi/v3/core/HandleScope.java
+++ b/core/src/main/java/org/jdbi/v3/core/HandleScope.java
@@ -11,10 +11,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jdbi.v3.core.internal;
+package org.jdbi.v3.core;
 
-import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.core.extension.HandleSupplier;
+import org.jdbi.v3.core.internal.ThreadLocalHandleScope;
 
 /**
  * Jdbi manages Handles to allow transaction nesting and extension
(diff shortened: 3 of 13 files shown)


hgschmie:
@anderssv Feedback very much welcome. I am not exactly a Kotlin person so I kind of puzzled this together with some docs, common sense and trial-and-error. This could benefit from review / feedback from people that actually use Jdbi with Kotlin.

hgschmie:
This addresses #2522

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -405,10 +416,10 @@ public <R, X extends Exception> R withHandle(HandleCallback<R, X> callback) thro
             sqlStatements.setAttachAllStatementsForCleanup(sqlStatements.isAttachCallbackStatementsForCleanup());
 
             handleSupplier = ConstantHandleSupplier.of(h);
-            threadHandleSupplier.set(handleSupplier);
+            getHandleSupplierHolder().setHandleSupplier(handleSupplier);
             return decoratedCallback.withHandle(h);
         } finally {
-            threadHandleSupplier.remove();
+            getHandleSupplierHolder().clearHandleSupplier();
```
stevenschlansker, line 422: should you look up the handle supplier holder here again? seems like the lookup can be done once, and you always set / clear on the same instance.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -521,16 +532,16 @@ public <X extends Exception> void useTransaction(final TransactionIsolationLevel
      */
     public <R, E, X extends Exception> R withExtension(Class<E> extensionType, ExtensionCallback<R, E, X> callback)
             throws X {
-        HandleSupplier handleSupplier = threadHandleSupplier.get();
+        HandleSupplier handleSupplier = getHandleSupplierHolder().getHandleSupplier();
```
stevenschlansker, line 535: same here, feels like we could re-use the same handleSupplierHolder

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -332,6 +331,18 @@ public HandleCallbackDecorator getHandleCallbackDecorator() {
...
 
+    /**
+     * Returns an internal handle supplier holder element used by this Jdbi instance.
+     * <br>
+     * <b>This is an internal method and not part of the public API!</b>
+     * @return A handle supplier holder element
+     * @param <T> type parameter
+     */
+    @Alpha
+    public <T extends HandleSupplierHolder> T getHandleSupplierHolder() {
```
stevenschlansker, line 342: Since this is not intended as public API, can we restrict visibility? Also, I think `@Alpha` does not need to be applied to non-public api, since it is always subject to change at any time.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
```diff
@@ -332,6 +331,18 @@ public HandleCallbackDecorator getHandleCallbackDecorator() {
...
+        return (T) handleSupplierHolder.updateAndGet(x -> x != null ? x : config.get(Handles.class).getHandleSupplierHolder());
```
stevenschlansker, line 343: this is a little weird if you ever try to change the handle supplier holder, since this reference will never update.

core/src/main/java/org/jdbi/v3/core/Jdbi.java
hgschmie: can we move the review of the handle supplier holder code to #2523 ?

core/src/main/java/org/jdbi/v3/core/Jdbi.java
hgschmie: Unfortunately not. This is specifically intended to be used by the Kotlin code to allow doing this:

```kotlin
fun Jdbi.attachToContext(): CoroutineContext.Element = this.getHandleSupplierHolder<CoroutineAwareHandleSupplierHolder>().contextElement()
```

This allows attaching the Jdbi object to a coroutine context so that the handles in the internal threadlocal are moved around correctly. 

The extension function above is (in terms of java code) a hidden static function (so basically it is 

```java
public static CoroutineContext.Element attachToContext() {
    return Jdbi<CoroutineAwareHandleSupplierHolder>getHandleSupplierHolder().contextElement();
}
```

 "patched" into the Jdbi class. And that is only possible with public members (because the extension method lives in a different package (org.jdbi.v3.core.kotlin) and is not an actual member of the hierarchy, so neither protected nor package scope would work).

core/src/main/java/org/jdbi/v3/core/Jdbi.java
hgschmie: It allows changing the implementation class to be modified by calling 

```java
Jdbi jdbi = Jdbi.create(...);
jdbi.getConfig(Handles.class).setHandleSupplierHolderClass(MyMagicHandleSupplierHolder.class)
... more code
```

If the implementation would be created when the Jdbi is created, it could not be changed by configuration. If we allow changing at any time, there is a good chance that we might leak handles.

Open for better ideas, this is what I came up with.

hgschmie:
keep this in draft until we have hashed out #2523 and I figured out the sporadically failing tests.

reuben-yolabs:
Hi @hgschmie , I am curious what ended up happening with your efforts here. I can see that this was closed but it wasn't clear to me if kotlin coroutine support was implemented in a different way?

PR #2528, opened 2023-11-06
Spring Jdbi repositories

xfredk:
I wrote some code that scans for sql-object interfaces annotated with @JdbiRepository. These interfaces can be automatically detected by spring and used for autowiring in spring applications. These beans work seamlessly with spring transactions.

Points of notice:
- I changed the JdbiUtil class in a non-backward compatible way! This was because (I think) the original had a bug.
- I added two dependencies to the project:
- - a dependency to jdbi3-sqlobject (The main point of this change)
- - a dependency to spring-context (To enable the detection and processing of the annotations)

I considered making a separate project but thought it was close enough to put it in the spring5 project


hgschmie:
Hi @xfredk!

Thank you for your PR to the Jdbi project! I triggered the CI scripts and I will review in a bit.

spring5/src/main/java/org/jdbi/v3/spring5/JdbiRepositoryRegistrar.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    private void registerJdbiRepositoryFactoryBean(BeanDefinitionRegistry registry, AnnotationMetadata annotationMetadata) {
+        String className = annotationMetadata.getClassName();
+        @SuppressWarnings("rawtypes")
+        Class clazz = ClassUtils.resolveClassName(className, null);
+
+        Map<String, Object> attributes = getAnnotationAttributes(JdbiRepository.class, annotationMetadata);
+        String jdbiQualifier = (String) attributes.get("jdbiQualifier");
```
stevenschlansker, line 56: It's too bad you cannot use the annotation type itself here, like `anno.jdbiQualifier()`, but I guess this is how spring does this?

stevenschlansker (approved):
Thank you for this nice improvement. I am not super familiar with the "Spring way" but the code is nicely written and hopefully others will find this useful too.

spring5/src/main/java/org/jdbi/v3/spring5/JdbiJtaBinder.java
```diff
@@ -0,0 +1,102 @@
...
+        return Proxy.newProxyInstance(extensionType.getClassLoader(), types, handler);
+    }
+
+    private static Object invoke(Object target, Method method, Object[] args) {
+        try {
+            if (Proxy.isProxyClass(target.getClass())) {
+                return Proxy.getInvocationHandler(target)
+                        .invoke(target, method, args);
+            } else {
+                return MethodHandles.lookup().unreflect(method)
```
stevenschlansker, line 93: It is expensive to lookup the method handle every time you invoke the method, but it's probably ok to fix that later when it shows up as an actual performance bottleneck.

stevenschlansker:
@xfredk , it would be nice to add a little bit about this new feature into the docs if you have time.

PR #2543, opened 2023-11-20
Support Kotlin coroutines

hgschmie:
Allows using Jdbi objects (handles and extension objects) with kotlin
coroutines. Objects can be shared between coroutines (even across threads)
as long as the caller can guarantee that not multiple coroutines use them
at the same time.

### DIFF
--- a/kotlin-sqlobject/src/test/kotlin/org/jdbi/v3/sqlobject/kotlin/CoroutineSqlObjectTest.kt
+++ b/kotlin-sqlobject/src/test/kotlin/org/jdbi/v3/sqlobject/kotlin/CoroutineSqlObjectTest.kt
@@ -1,0 +1,141 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.sqlobject.kotlin
+
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.runBlocking
+import org.assertj.core.api.Assertions.assertThat
+import org.jdbi.v3.core.Handle
+import org.jdbi.v3.core.Jdbi
+import org.jdbi.v3.core.kotlin.KotlinPlugin
+import org.jdbi.v3.core.kotlin.withExtensionUnchecked
+import org.jdbi.v3.sqlobject.SqlObject
+import org.jdbi.v3.sqlobject.SqlObjectPlugin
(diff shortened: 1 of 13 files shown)


kotlin-sqlobject/src/test/kotlin/org/jdbi/v3/sqlobject/kotlin/CoroutineSqlObjectTest.kt
```diff
@@ -0,0 +1,131 @@
...
+            val handle2: Handle?
+            val handle3: Handle?
+
+            runBlocking(Dispatchers.IO) {
+                coroutineScope {
+                    // all Jdbi operations will map onto the attached handle in this coroutine context
+                    handle1 = insertSomething(dao)
+
+                    handle2 = selectSomething(dao, 1, "first name")
+                    delay(1000L)
```
stevenschlansker, line 85: 1s is a fair amount of time to wait. Is it possible to write these tests with latches instead of timed sleep?

kotlin/src/main/kotlin/org/jdbi/v3/core/kotlin/CoroutineHandleScope.kt
```diff
@@ -0,0 +1,87 @@
...
+import kotlinx.coroutines.CopyableThreadContextElement
+import kotlinx.coroutines.DelicateCoroutinesApi
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import org.jdbi.v3.core.Handle
+import org.jdbi.v3.core.HandleScope
+import org.jdbi.v3.core.extension.HandleSupplier
+import org.jdbi.v3.core.internal.ThreadLocalHandleScope
+import kotlin.coroutines.CoroutineContext
+
+@OptIn(DelicateCoroutinesApi::class, ExperimentalCoroutinesApi::class)
```
stevenschlansker, line 25: 😬

kotlin/src/main/kotlin/org/jdbi/v3/core/kotlin/CoroutineHandleScope.kt
```diff
@@ -0,0 +1,87 @@
...
+        if (handle == null) {
+            delegate.clear()
+        } else {
+            delegate.set(handle)
+        }
+        return oldState
+    }
+
+    override fun restoreThreadContext(context: CoroutineContext, oldState: HandleSupplier?) {
+        if (oldState == null) {
```
stevenschlansker, line 50: We seem to be checking for a `null` argument to `delegate.set` a lot. Would it make sense to push the null-check into the `ThreadLocalHandleScope.set` method so as not to repeat it when we call in?

stevenschlansker (approved):
I'm not really qualified to review Kotlin, as I've never written a single line, but I did read through everything.

kotlin/src/main/kotlin/org/jdbi/v3/core/kotlin/CoroutineHandleScope.kt
hgschmie: Yes, but a reviewer disagreed. - https://github.com/jdbi/jdbi/pull/2523#discussion_r1379279644

(SCNR :-) )

kotlin-sqlobject/src/test/kotlin/org/jdbi/v3/sqlobject/kotlin/CoroutineSqlObjectTest.kt
hgschmie: I like the latch idea. I will see if I can change this tonight.

PR #2548, opened 2023-11-25
Fix vavr arguments for sqlobjects

### LINKED ISSUE #2529: Vavr arguments do not work with SqlObjects
Using Vavr object like an `Option` as an argument with SqlObjects like
```java
@SqlQuery("select * from something where :name is null or name = :name order by id")
@RegisterBeanMapper(Something.class)
List<Something> selectByOptionName(@Bind Option<String> name);
```
does not work.

Based on the `TestVavrValueArgumentFactoryWithDB` in JDBI Vavr module, I created these test cases for using an Option as an argument. 
```java
public class TestVavrValueArgumentFactoryWithDBSqlObjects {


    private static final Something ERICSOMETHING = new Something(1, "eric");
    private static final Something BRIANSOMETHING = new Something(2, "brian");

    @RegisterExtension
    public JdbiExtension h2Extension = JdbiExtension.h2().withInitializer(TestingInitializers.something()).installPlugins();

    @BeforeEach
    public void createTestData() {
        Handle handle = h2Extension.getSharedHandle();
        handle.createUpdate("insert into something (id, name) values (1, 'eric')").execute();
        handle.createUpdate("insert into something (id, name) values (2, 'brian')").execute();
    }


    @Test
    public void testGetOptionShouldReturnCorrectRowUsingExtension() {
        List<Something> result = h2Extension.getJdbi()
                .withExtension(Dao.class, dao -> dao.selectByOptionName(Option.of("eric")));

        assertThat(result).hasSize(1);
    }

    @Test
    public void testGetOptionEmptyShouldReturnAllRowsUsingExtension() {
        List<Something> result = h2Extension.getJdbi()
                .withExtension(Dao.class, dao -> dao.selectByOptionName(Option.none()));

        assertThat(result).hasSize(2);
    }

    interface Dao {

        @SqlQuery("select * from something where :name is null or name = :name order by id")
        @RegisterBeanMapper(Something.class)
        List<Something> selectByOptionName(@Bind Option<String> name);
    }
```

With a `Some` the query works, but with a `None` it does not.
This is somewhat different than what we see in our codebase. Then the test with a `Some` value also fails with 
```
org.jdbi.v3.core.statement.UnableToExecuteStatementException: org.postgresql.util.PSQLException: ERROR: operator does not exist: integer = bigint[]
  Hint: No operator matches the given name and argument types. You might need to add explicit type casts.
  Position: 50 [statement:"SELECT * FROM our_table  WHERE id = :id ", arguments:{positional:{0:Some(47)}, named:{id:Some(47)}, finder:[]}]
```

While debugging I notice that the `VavrValueArgumentFactory.build` method does not get called when using SqlObject.
It _does_ get called when using the Java Api `.bind("name", Option.none())`.

hgschmie:
Turns out, this has never worked before. To work with SqlObjects,
the argument factory needs to implement ArgumentFactory.Preparable,
otherwise another factory (in this case one that accepts anything
iterable) would take precedence.

Rewrite the factory to support Preparable, add unit tests suggested
by @diversit.

Fixes #2529

### DIFF
--- a/vavr/src/main/java/org/jdbi/v3/vavr/VavrValueArgumentFactory.java
+++ b/vavr/src/main/java/org/jdbi/v3/vavr/VavrValueArgumentFactory.java
@@ -14,10 +14,9 @@
 package org.jdbi.v3.vavr;
 
 import java.lang.reflect.Type;
-import java.util.Arrays;
-import java.util.HashSet;
 import java.util.Optional;
-import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Supplier;
 
 import io.vavr.Lazy;
 import io.vavr.Value;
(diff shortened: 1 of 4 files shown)


vavr/src/main/java/org/jdbi/v3/vavr/VavrValueArgumentFactory.java
```diff
@@ -29,37 +28,62 @@
...
+        for (Class<?> valueClass : VALUE_CLASSES) {
+            if (valueClass.isAssignableFrom(rawType)) {
+                return true;
+            }
+        }
+        return false;
+    }
+    private static Type findGenericType(Type wrapperType, Object nestedValue) {
+        Optional<Type> nestedType = findGenericParameter(wrapperType, Value.class);
+        return nestedType.orElseGet(typeFromValue(nestedValue));
```
stevenschlansker, line 79: I think it would be slightly clearer if the Supplier was pushed up to `() -> typeFromValue(nestedValue)` and then you can just return `Type typeFromValue`. it's ok to leave this if you don't want to change it.

hgschmie: This allowed me to use a constant (the OBJECT_SUPPLIER) for the (pretty common) case where the nestedValue is null. This is the case for any uses with SqlObjects.

Issue #2562, opened 2023-12-14
@Define + optional @BindBean


xak2000:
It is useful to have `@Define` and `@BindBean` on a single **nullable** argument of `@SqlQuery`.

E.g.:

```java
class UserPageToken {
    Direction direction;
    Long userId;
    // getters, setters
}

public interface UserSearchDao {
    @SqlQuery
    @UseFreemarkerEngine
    @UseFreemarkerSqlLocator
    List<User> findAll(
            @BindBean("pageToken") @Define UserPageToken pageToken,
            @Bind int limit
    );
}
```

And SQL query:
```sql
SELECT
    *
FROM
    users u
WHERE
    1 = 1
<#if pageToken??>
        <#if pageToken.direction == "FORWARD">
            AND u.user_id < :pageToken.userId
        <#else>
            AND u.user_id > :pageToken.userId
        </#if>
</#if>
ORDER BY
<#if !pageToken?? || pageToken.direction == "FORWARD">
    u.user_id DESC
<#else>
    u.user_id ASC
</#if>
LIMIT
    :limit
```

The crucial part here is that `pageToken` argument is **optional**. It could be `null` that means it is the first page requested, so the `WHERE` part should not filter by `userId` at all. The caller can't pass a "dummy" `pageToken` bean, because to do this it needs to know last `userId` of the previous returned page (cursor/keyset pagination).

Currently an attempt to pass `null` leads to the exception:

```
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "bean" is null
	at org.jdbi.v3.core.statement.SqlStatement.bindBean(SqlStatement.java:193)
	at org.jdbi.v3.sqlobject.customizer.internal.BindBeanFactory.lambda$createForParameter$0(BindBeanFactory.java:39)
	at org.jdbi.v3.sqlobject.customizer.internal.PojoWarmingCustomizer$1.apply(PojoWarmingCustomizer.java:37)
	at org.jdbi.v3.sqlobject.statement.internal.CustomizingStatementHandler$1.apply(CustomizingStatementHandler.java:118)
	at org.jdbi.v3.sqlobject.statement.internal.CustomizingStatementHandler.lambda$applyCustomizers$11(CustomizingStatementHandler.java:203)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.jdbi.v3.sqlobject.statement.internal.CustomizingStatementHandler.applyCustomizers(CustomizingStatementHandler.java:201)
[...]

PR #2567, opened 2023-12-22
Support null for bean, pojo, field, and method binding

hgschmie:
Instead of throwing a NPE, do nothing. This allows for calling
the various Jdbi methods (and the annotations for sqlobject) to deal
with null objects.

While this is a behavior change, it is acceptable as callers had to
wrap this in `if (bean != null) { ...` conditionals and had to skip
binding for null values. Now the code does skip null values directly.


hgschmie:
This addresses #2562

Issue #2549, opened 2023-11-28
followup on mockito#3037 (dynamic agent loading)


hgschmie:
implement the solution suggested there (using dependency:property)

stevenschlansker:
Potentially relevant issue: https://github.com/mockito/mockito/pull/3137

stevenschlansker:
Dynamic agent is now gone.

PR #2569, opened 2023-12-22
Make build work on java 22-ea

hgschmie:
Revisit the mockito setup (this can be improved once
https://github.com/mockito/mockito/pull/3137 has landed and was
released with mockito). For now, we add the bytebuddy agent directly.

Also update spotbugs and junit, which is needed to pass build under
Java 22-ea.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
           JAVA_HOME: ${{ steps.test_jdk.outputs.path }}
         run: |
           echo "$JAVA_HOME/bin" >> ${GITHUB_PATH}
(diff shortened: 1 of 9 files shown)


stevenschlansker (approved):

stevenschlansker:
Fixes #2549

stevenschlansker:
I have spent a bit more time looking into this, and I think we can avoid needing bytebuddy agent at all (and therefore avoid needing to do all the pom dance here to bring it in for 22-ea).

See #2593

stevenschlansker:
I'll close this for now since we have 22-ea working without the agent altogether.

PR #2576, opened 2023-12-26
Support ResultSet from stored procedures

hgschmie:
Fully support JDBC drivers that return stored procedure results as a
ResultSet. Allow mixing standard out parameters with ResultSet.

Also support returning values from OutParameters.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/result/internal/ResultSetResultIterator.java
+++ b/core/src/main/java/org/jdbi/v3/core/result/internal/ResultSetResultIterator.java
@@ -16,35 +16,56 @@ package org.jdbi.v3.core.result.internal;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.NoSuchElementException;
+import java.util.function.Supplier;
 
+import org.jdbi.v3.core.internal.exceptions.Sneaky;
 import org.jdbi.v3.core.mapper.RowMapper;
 import org.jdbi.v3.core.result.ResultIterator;
 import org.jdbi.v3.core.result.ResultSetException;
 import org.jdbi.v3.core.statement.StatementContext;
 
-import static java.util.Objects.requireNonNull;
-
 class ResultSetResultIterator<T> implements ResultIterator<T> {
     private final ResultSet resultSet;
     private final RowMapper<T> rowMapper;
+
+    private final ResultSetSupplier resultSetSupplier;
     private final StatementContext context;
 
     private volatile boolean alreadyAdvanced = false;
     private volatile boolean hasNext = false;
     private volatile boolean closed = false;
 
-    ResultSetResultIterator(ResultSet resultSet,
-                            RowMapper<T> rowMapper,
-                            StatementContext context) throws SQLException {
-        this.resultSet = requireNonNull(resultSet);
-        this.rowMapper = rowMapper.specialize(resultSet, context);
+    ResultSetResultIterator(Supplier<ResultSet> resultSetSupplier,
+        RowMapper<T> rowMapper,
+        StatementContext context) throws SQLException {
+
         this.context = context;
+
+        if (resultSetSupplier instanceof ResultSetSupplier) {
+            this.resultSetSupplier = (ResultSetSupplier) resultSetSupplier;
+        } else {
+            this.resultSetSupplier = ResultSetSupplier.closingContext(resultSetSupplier, context);
+        }
+
+        this.resultSet = this.resultSetSupplier.get();
+
+        if (resultSet != null) {
+            context.addCleanable(resultSet::close);
+            this.rowMapper = rowMapper.specialize(resultSet, context);
+        } else {
+            close();
+            this.rowMapper = null;
+        }
     }
 
     @Override
     public void close() {
         closed = true;
-        context.close();
+        try {
+            resultSetSupplier.close();
+        } catch (SQLException e) {
+            throw Sneaky.throwAnyway(e);
+        }
     }
 
     @Override
@@ -70,13 +91,13 @@ class ResultSetResultIterator<T> implements ResultIterator<T> {
 
     @Override
     public T next() {
-        if (closed) {
-            throw new IllegalStateException("iterator is closed");
-        }
-
         if (!hasNext()) {
             close();
             throw new NoSuchElementException("No element to advance to");
+        }
+
+        if (closed) {
+            throw new IllegalStateException("iterator is closed");
         }
 
         try {
--- a/core/src/main/java/org/jdbi/v3/core/result/internal/ResultSetResultIterator.java
+++ b/core/src/main/java/org/jdbi/v3/core/result/internal/ResultSetResultIterator.java
@@ -96,10 +96,6 @@ class ResultSetResultIterator<T> implements ResultIterator<T> {
             throw new NoSuchElementException("No element to advance to");
         }
 
-        if (closed) {
-            throw new IllegalStateException("iterator is closed");
-        }
-
         try {
             return rowMapper.map(resultSet, context);
         } catch (SQLException e) {
--- a/core/src/main/java/org/jdbi/v3/core/statement/Call.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/Call.java
@@ -17,6 +17,7 @@ import java.sql.CallableStatement;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
 import java.sql.Types;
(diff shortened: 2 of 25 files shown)


core/src/main/java/org/jdbi/v3/core/result/internal/ResultSetResultIterator.java
```diff
@@ -16,35 +16,56 @@
...
-                            StatementContext context) throws SQLException {
-        this.resultSet = requireNonNull(resultSet);
-        this.rowMapper = rowMapper.specialize(resultSet, context);
+    ResultSetResultIterator(Supplier<ResultSet> resultSetSupplier,
+        RowMapper<T> rowMapper,
+        StatementContext context) throws SQLException {
+
         this.context = context;
+
+        if (resultSetSupplier instanceof ResultSetSupplier) {
```
stevenschlansker, line 44: Do we need this to be a dynamic instanceof check, or would it be good enough to have a static dispatch instead?

```
ResultSetResultIterator(ResultSetSupplier rs, ...) { ... }
ResultSetResultIterator(Supplier<ResultSet>rs, ...) {
    this(ResultSetSupplier.closingContext(rs, ctx);
}

File: core/src/main/java/org/jdbi/v3/core/result/internal/ResultSetResultIterator.java [hunk 2]
```diff
@@ -70,15 +91,15 @@ public boolean hasNext() {
...
-        if (closed) {
-            throw new IllegalStateException("iterator is closed");
-        }
-
         if (!hasNext()) {
             close();
             throw new NoSuchElementException("No element to advance to");
         }
 
+        if (closed) {
```
stevenschlansker, line 99: I think by moving this clause, it is made redundant. If the iterator is closed, `hasNext()` is false, so we already tried to `close(); throw ...;` above.
So if you want to change this, maybe just delete the clause instead?

core/src/main/java/org/jdbi/v3/core/statement/Call.java
```diff
@@ -117,31 +129,84 @@ public void invoke(Consumer<OutParameters> resultConsumer) {
...
+            Supplier<Object> supplier = outparamArgument.supplier((CallableStatement) stmt);
+            // index is 0 based, position is 1 based.
+            out.putValueSupplier(outparamArgument.position - 1, outparamArgument.name, supplier);
+        });
+
+        return resultComputer.apply(out);
+    }
+
+    /**
+     * Specify the fetch size for the call. This should cause the results to be
```
stevenschlansker, line 158: This wording feels a bit stronger than the JDBC spec that backs it, which calls it a "hint" (unlike the other tuneables here)

core/src/main/java/org/jdbi/v3/core/statement/Call.java
```diff
@@ -164,7 +229,14 @@ public void apply(int outPosition, PreparedStatement statement, StatementContext
             this.position = outPosition;
         }
 
-        public Object map(CallableStatement stmt) {
+        public Supplier<Object> supplier(CallableStatement stmt) {
+            return () -> {
+                Object value = map(stmt);
+                return isNull(stmt) ? null : value;
```
stevenschlansker, line 235: if `isNull(stmt)`, does that imply that the `value` we retrieved would always be `null` anyway, which would allow eliding this ternary?

core/src/main/java/org/jdbi/v3/core/statement/OutParameters.java
```diff
@@ -18,339 +18,540 @@
...
 
-        if (obj instanceof java.util.Date) {
-            return ((java.util.Date) obj).getTime();
-        } else {
-            throw new IllegalArgumentException(format("Parameter at %d is a %s, not a Date", pos, obj.getClass()));
+        Object getValue(int index) {
+            if (!map.containsKey(index)) {
+                throw new IllegalArgumentException(format("Parameter #%d does not exist", index));
+            }
+            Supplier<Object> supplier = map.get(index);
```
stevenschlansker, line 545: It is trivially more efficient to call the `get(index)` first, then only check `containsKey` after observing the `supplier != null` which is computed anyway.
(The old code was careful to do this. I am not claiming that this is a big deal.)

core/src/main/java/org/jdbi/v3/core/statement/OutParameters.java
```diff
@@ -18,339 +18,540 @@
...
-    }
 
-    @Nonnull
-    Map<Object, Object> getMap() {
-        return map;
+        Object getValue(String name) {
+            if (!map.containsKey(name)) {
+                throw new IllegalArgumentException(format("Parameter '%s' does not exist", name));
+            }
+            Supplier<Object> supplier = map.get(name);
```
stevenschlansker, line 553: ditto

integration-testing/pom.xml
```diff
@@ -75,5 +101,35 @@
...
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>mssqlserver</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
+
+    <profiles>
+        <profile>
+            <id>no-docker</id>
```
stevenschlansker, line 114: At some point, would it make sense to move the `no-docker` concept to be part of each test case (maybe as an assumption, or a tag, or custom extension) rather than having to maintain lists of test cases in pom?

stevenschlansker (approved):

core/src/main/java/org/jdbi/v3/core/result/internal/ResultSetResultIterator.java
hgschmie: The code calling this will always call `ResultSetResultIterator(Supplier<ResultSet> ...)` independently whether the object is a ResultSetSupplier object or not. The problem is that between creating the ResultSetSupplier and its use there are a number of public APIs that expose `Supplier<ResultSet>`. Most prominently `ResultBearing.of` which is part of the public API.

core/src/main/java/org/jdbi/v3/core/result/internal/ResultSetResultIterator.java
hgschmie: yes. Done.

core/src/main/java/org/jdbi/v3/core/statement/Call.java
hgschmie: This is cut-and-paste code from Query.

core/src/main/java/org/jdbi/v3/core/statement/Call.java
hgschmie: IIRC, there are some drivers that e.g. return "0" for an integer column but isNull == true. This is code that already existed and was just moved around. isNull only implies that the SQL column contains a SQL NULL, not that the return value is null.

integration-testing/pom.xml
hgschmie: no-docker solely exists for you. :-)

core/src/main/java/org/jdbi/v3/core/statement/Call.java
hgschmie: correction, this is true for every driver if e.g. a null is mapped onto an integer so `rs.getInt()` is called. `ResultSet#getInt()` explicitly returns "0" for a SQL NULL (see javadocs) and the `isNull()` call is required to turn it into a `null` value.

core/src/main/java/org/jdbi/v3/core/statement/OutParameters.java
hgschmie: I am not disagreeing but the old code was really hard to understand (I did not understand it. :-) ).  I simplified it.

stevenschlansker: Maybe the code reads a tiny bit simpler, but now every time this functionality is used we will double the number of map looks, for every user, forever :(

The code was written the way it was previously quite intentionally.

integration-testing/pom.xml
stevenschlansker: I still feel it's important to be able to build Jdbi without external daemons.
Docker is even worse, because testcontainers uses ryuk which doesn't work out of the box rootless, so it in fact requires a root daemon to build Jdbi then. (without hacking it to work with podman)

core/src/main/java/org/jdbi/v3/core/statement/OutParameters.java
stevenschlansker: Oh, I see, it is much simpler now. Thanks for fixing this.

Issue #2506, opened 2023-10-04
Remove all mockito tests


hgschmie:
Seems that there is talk about mockito (and other byte modifying agents) to stop working on Java 21+ - https://github.com/mockito/mockito/issues/3037

We already see the warning when building with Java 21.

Mocking is a bad idea for many reasons. We should rip the mocks out.

stevenschlansker:
I think it is unlikely that mockito will go away entirely - too many people rely on this. Some configuration in the build might be necessary.

I am all for using mocks less, although it is arguably the best solution in some cases (stubbing out test interactions with external libraries that obsessively mark their classes as `final`, for example)

hgschmie:
mockito will require, as any other library that wants to install byte code manipulation, the right command line options to the runtime in the future. For now, it is sufficient to add `-XX:+EnableDynamicAgentLoading` to surefire and failsafe. But it *will* go away at some point in the future.

stevenschlansker:
Right, blanket enabling it for all runtime-loaded agents will, but you will still be able to specifically ask to give e.g. bytebuddy the appropriate capabilities.

metaforte:
I see H2DatabaseExtension.withSomething method deprecated with commit comment "Rewrite core tests to use Junit5". Could you please give a sample rewrite for me to get any idea on the intent

stevenschlansker:
I think the intent there is simply that the "something" table is not very useful as a shared test setup, and it makes it hard to evolve the tests since they have this common base that doesn't really add anything.

So the only thing I would take away from that is to define your own table setup that makes sense for the test case, rather than using the `something` table.

stevenschlansker:
All mockito tests that do questionable things have been rewritten. There's a few left where Mockito seems the best approach, like if you want to make a `Connection` that introduces failures in test cases.

PR #2593, opened 2024-01-09
Switch Mockito to proxy-only mode and disable byte-buddy agent

stevenschlansker:
A possible path to get rid of the byte-buddy agent entirely.
Depends on: #2595 
Fixes: #2549
Fixes: #2506 (while this does not remove all tests, the remaining ones I think are reasonably well justified)
Replaces: #2569


hgschmie (approved):
LGTM

PR #2595, opened 2024-01-10
Add a test case showing that even if commit() always fails, Jdbi will commit

stevenschlansker:
I think this might have crept in with the transaction handling rewrite - even if `commit()` always throws, Jdbi will accidentally (?) commit due to skipping `rollback` before `setAutoCommit(false)`
This behavior is pretty surprising to me -- wdyt @hgschmie?


hgschmie:
The test is great but it will break the build. We should keep it here and add a fix to the branch. I might find some time tomorrow in the afternoon to look at this.

stevenschlansker:
I am not proposing to merge this until we figure out how to fix the behavior, or if this is actually correct.

hgschmie:
So I have a potential fix but that currently breaks the TestTransactionsAutoCommit test. Need to think more about this.

PR #2603, opened 2024-01-17
commit throws still commits

hgschmie:
Fix situation where an exception in `Connection#commit()` still called `commit()`.

Fixes #2595


stevenschlansker (approved):
Nice, thanks

Issue #2598, opened 2024-01-11
Add test for non-default locale and timezone


stevenschlansker:
We used to check that Jdbi compiles and tests when you are not using English (different error messages) and not in UTC (possible user timezone bugs). At some point, these disappeared. Let's bring them back. ref: #2596

hgschmie:
fixed by #2604

PR #2604, opened 2024-01-20
Run CI tests with different timezone and locale

hgschmie:
Add CI flows that run all tests with different timezone and locale

- Asia/Colombo was chosen because it has a 5:30 offset
- tr_TR was chosen because it has letters that have no capitalization

addresses #2598

### DIFF
--- a/internal/build/pom.xml
+++ b/internal/build/pom.xml
@@ -134,6 +134,9 @@
         <jdbi.check.skip-ktlint>${jdbi.check.skip-kotlin}</jdbi.check.skip-ktlint>
         <jdbi.check.skip-sortpom>${basepom.check.skip-basic}</jdbi.check.skip-sortpom>
         <jdbi.test.pg-version>15</jdbi.test.pg-version>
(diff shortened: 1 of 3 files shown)


internal/build/pom.xml
```diff
@@ -850,8 +853,11 @@
                         <groups />
                         <!-- junit 4 and 5 -->
                         <excludedGroups>slow,org.jdbi.v3.testing.SlowTests</excludedGroups>
-                        <systemPropertyVariables combine.children="append">
+                        <systemPropertyVariables combine.children="override">
                             <pg-embedded.postgres-version>${jdbi.test.pg-version}</pg-embedded.postgres-version>
+                            <user.timezone>${jdbi.test.timezone}</user.timezone>
```
stevenschlansker, line 858: Any reason not to just set `-Duser.timezone=Asia/Colombo` directly at the cli?

stevenschlansker (approved):

stevenschlansker:
> MySQLJdbiTestContainersExtensionTest » ContainerLaunch Container startup failed for image mysql:latest

That sucks :/

PR #2606, opened 2024-01-23
maven build warnings cleanup

stoyants:
Here is a short summary:
* AssertJ version can be bumped to "3.25.1" Hopefully then CI will not fail even without exclusion of "guava", because guava version is upgraded to "33.0.0-jre" (the same like jdbi)
* HikariCP stop using slf4j-api 2.0.0 since they stop support of Java 8, so no need to exclude it
* Changes in "kotlin" and "kotlin-sqlobject" are recommended by https://kotlinlang.org/docs/maven.html#compile-kotlin-only-source-code
* Creation of "uberjar" in "beanchmark" faces the problem of "merging" of artefacts like "module-info.class", "MANIFEST.MF", etc. and is no way perfect. Now the conflicting once are excluded and it is not any worse (or better) than before. But at least multiple warnings are gone.

About remaining warnings:
* Another warning can be avoided if "<exclude>**/*.MockMaker</exclude>" is added to the respective pom.xml in "basepom" module.
* There are few warning because of disabled test.
* And one warning in "spring5" is caused by missing "jsr305", but as far as I understand "jdbi" is trying to avoit this library.
* There are few more warning coming from "kotlin" modules which I can try to remove.

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -126,6 +126,7 @@
                                 <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
                                     <mainClass>org.openjdk.jmh.Main</mainClass>
                                 </transformer>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer" />
                             </transformers>
                             <filters>
                                 <filter>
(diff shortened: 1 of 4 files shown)


stevenschlansker (approved):
Thanks!

stevenschlansker:
So it turns out the Kotlin source dir changes are not needed. I backed those out. We do it this way:
https://github.com/jdbi/jdbi/blob/master/internal/build/pom.xml#L1084-L1102

stoyants:
> So it turns out the Kotlin source dir changes are not needed. I backed those out. 

Without them the following warnings will "come back" (only in "install" profile, not in "slow-tests"):

[WARNING] Source root doesn't exist: src/main/java 

Both in "kotlin" and "kotlin-sqlobject" modules. They constitute currently ca. 50% of all warnings :)

The "kotlin" documentation recommends these settings especially in the case when module has only "kotlin" sources and no "java" sources. Which is the case in "jdbi" modules.

I tested also the "slow-tests" in my local environment. All seems fine. I have no idea why CI failed, but I doubt that these changes was the reason. Seems to me, that mysql container started very slowly and the test timed-out.

Best Regards, Stoyan

stevenschlansker:
I think @hgschmie looked into this before and Maven made it hard to do dynamic configuration of kotlin source dirs and also not warn on java source dirs. I wonder if we just create the missing source directories with a `.keep` if that is enough.

stevenschlansker:
Yes, the mysql test failure is not related - we are investigating that separately.

stevenschlansker:
I think the simplest fix might be to just create the missing source dirs as empty:
https://github.com/jdbi/jdbi/pull/2608

PR #2608, opened 2024-01-24
kotlin: Add empty missing Java source dirs to prevent warnings

stevenschlansker:


hgschmie (approved):
looks fine; there are a number of ".keepme" files already, maybe rename?

stevenschlansker:
Oh, I added the `.txt` to avoid the license checker warning. Maybe `.keepme` is exempt from license checker already?

PR #2609, opened 2024-01-24
FieldMapper: skip static fields

### LINKED ISSUE #2607: Can not set static final field
If the select returns a column with the same name as a public static final field in the bean, JDBI will try to write the value to that column and will fail with:
```
Caused by: java.lang.IllegalAccessException: Can not set static final com.gitlab.mvysny.jdbiorm.Property field com.gitlab.mvysny.jdbiorm.NestedJoinOutcome.DEPARTMENT_NAME to java.lang.String
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwFinalFieldIllegalAccessException(UnsafeFieldAccessorImpl.java:76)
	at java.base/jdk.internal.reflect.UnsafeFieldAccessorImpl.throwFinalFieldIllegalAccessException(UnsafeFieldAccessorImpl.java:80)
	at java.base/jdk.internal.reflect.UnsafeQualifiedStaticObjectFieldAccessorImpl.set(UnsafeQualifiedStaticObjectFieldAccessorImpl.java:77)
	at java.base/java.lang.reflect.Field.set(Field.java:799)
	at org.jdbi.v3.core.mapper.reflect.FieldMapper$BoundFieldMapper.writeField(FieldMapper.java:285)
	... 57 more
```

In the example above, I have a `SELECT x as department_name from y` mapping values to a bean with a `public static final String DEPARTMENT_NAME = "foo";` field via `FieldMapper`.

Reasoning: in jdbi-orm I'm recommending the users to create a `public static final Property<> ID` fields which allows you to programmatically create Conditions such as `Person.ID.lt(25)`. However, since there is a database column named ID, JDBI's FieldMapper tries to write the value both to the instance field `id` and the static final field `ID` and fails.

I believe that JDBI should ignore static fields in general, or at least static final fields.

Meanwhile, is there a workaround/configuration which would allow me to ignore such fields please?

JDBI 3.43.0

stevenschlansker:
Fixes #2607

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,7 @@
 
 - fix edge condition when calling `Connection#commit()` threw an Exception, we called commit() again. Now the
   code explicitly calls `Connection#rollback()`. Fixes #2595
+- FieldMapper: skip static fields (#2607, reported by @mvysny)
 
 # 3.43.0
 
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/FieldMapper.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/FieldMapper.java
@@ -15,6 +15,7 @@ package org.jdbi.v3.core.mapper.reflect;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.ArrayList;
@@ -143,7 +144,7 @@ public final class FieldMapper<T> implements RowMapper<T> {
         for (Class<?> aType = type; aType != null; aType = aType.getSuperclass()) {
             for (Field field : aType.getDeclaredFields()) {
                 Nested nested = field.getAnnotation(Nested.class);
-                if (!JdbiAnnotations.isMapped(field)) {
+                if (Modifier.isStatic(field.getModifiers()) || !JdbiAnnotations.isMapped(field)) {
                     continue;
                 }
 
--- a/core/src/test/java/org/jdbi/v3/core/mapper/reflect/FieldMapperTest.java
+++ b/core/src/test/java/org/jdbi/v3/core/mapper/reflect/FieldMapperTest.java
@@ -334,4 +334,20 @@ public class FieldMapperTest {
         @Nullable
         String s;
     }
+
+    @Test
+    public void staticFieldIgnored() {
+        Handle handle = h2Extension.getSharedHandle();
+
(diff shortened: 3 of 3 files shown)


core/src/main/java/org/jdbi/v3/core/mapper/reflect/FieldMapper.java
```diff
@@ -143,7 +144,7 @@ private Optional<RowMapper<T>> createSpecializedRowMapper(StatementContext ctx,
         for (Class<?> aType = type; aType != null; aType = aType.getSuperclass()) {
             for (Field field : aType.getDeclaredFields()) {
                 Nested nested = field.getAnnotation(Nested.class);
-                if (!JdbiAnnotations.isMapped(field)) {
+                if (Modifier.isStatic(field.getModifiers()) || !JdbiAnnotations.isMapped(field)) {
```
hgschmie, line 147: are there any other modifiers that we should think of?

RELEASE_NOTES.md
```diff
@@ -2,6 +2,7 @@
 
 - fix edge condition when calling `Connection#commit()` threw an Exception, we called commit() again. Now the
   code explicitly calls `Connection#rollback()`. Fixes #2595
+- FieldMapper: skip static fields (#2607)
```
hgschmie, line 5: I started to explicitly call out the people that reported bugs. This gives them a mention on the release page (e.g. here: https://github.com/jdbi/jdbi/releases/tag/v3.43.0) so they can take that back to their managers to justify contributing to Jdbi in their work time. It's a tiny thing for us but I know that some organizations track metrics for that.

hgschmie (approved):

core/src/main/java/org/jdbi/v3/core/mapper/reflect/FieldMapper.java
stevenschlansker: I don't think so? I suppose you could argue `transient` but I don't want to potentially break user code without at least some justification. Are you thinking of any specific modifier?

RELEASE_NOTES.md
stevenschlansker: I am happy to give collaborators attribution where due, but doing it for the corporate metrics seems like the wrong reason to do it to me... we would encourage tiny pointless contributions to boost your metrics

stevenschlansker: image

core/src/main/java/org/jdbi/v3/core/mapper/reflect/FieldMapper.java
stevenschlansker: We can always add more exclusions in future changes.

PR #2612, opened 2024-01-26
Update testcontainers and derby

hgschmie:
tc 1.19.4 contains the mysql fix for mysql 8.3.x
derby 1.16 runs on JDK 17+ only

### DIFF
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -16,7 +16,12 @@ jobs:
 
     strategy:
       matrix:
-        testcontainer-version: [ 1.18.3, 1.19.3 ]
+        testcontainer-version: [ 1.19.4 ]
+        mysql-version: [ "mysql:latest" ]
+        # 1.18.3 does not work with mysql:latest (8.3.x).
+        include:
+          - testcontainer-version: 1.18.3
+            mysql-version: "mysql:8.0"
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/slow-tests.yml
```diff
@@ -16,7 +16,12 @@ jobs:
 
     strategy:
       matrix:
-        testcontainer-version: [ 1.18.3, 1.19.3 ]
+        testcontainer-version: [ 1.19.4 ]
+        mysql-version: [ "mysql:latest" ]
+        # 1.18.3 does not work with mysql:latest (8.3.x).
+        include:
+          - testcontainer-version: 1.18.3
```
stevenschlansker, line 23: Why do we keep testing against old, buggy versions of the library?
It is nice to be able to leave bugs in the past at some point.

stevenschlansker (approved):

.github/workflows/slow-tests.yml
hgschmie: tc 1.18 is still the "previous minor". These tests exist mainly to ensure that when we adopt a new version of a library, we do not accidentally break users that are stuck on that version because e.g. their test framework requires it. Gives a somewhat smoother path to adoption. 

We leave that version behind as soon as tc 1.20 is released.

hgschmie:
it's the same BTW with derby. We *could* upgrade to 1.17 and just run the test when Java 21 is used. Right now, we run the test when 17+ is used (and stick with 1.16) because we support compiling the codebase with 17 and 21. Once we drop 17 for compilation, we can upgrade derby.

I also resolved the dependabot issue, we are not actually using the LDAP piece that was in the CVE.

PR #2614, opened 2024-01-26
update setup-java action version

hgschmie:


stevenschlansker (approved):
Sucks that this configuration has to be repeated so many times. For all of Maven's problems, every other system ends up seeming worse!

hgschmie:
we have many pipelines. that is how github actions are configured. I don't think there is a good way to unify this. TBH, this was a single command: `rg 'setup-java' .github | perl -pi -e 's|setup-java\@v3|setup-java\@v4|g' ` :-)

Issue #2590, opened 2024-01-03
Investigate japi for apidiffing 


hgschmie:
(looking at the work that @spannm  did in #2276 as well) 

japi is used by apache flink

stoyants:
@hgschmie Do you mean this tool / plugin [japicmp](https://siom79.github.io/japicmp/) ?

hgschmie:
shipped in 3.44.0

PR #2616, opened 2024-01-28
Introduce japicmp for API diffing

hgschmie:
- Update to basepom 58
- Introduce japicmp for API diffing


hgschmie:
Resolves #2590

spannm (approved):
I checked out the PR and tried to test but cannot get past unit test failures due to "Could not download artifact for Zonky Postgres 15" in core
Other than that LGTM. It will be interesting to see how this changes work on this project. Hopefully it won't be too intrusive...

hgschmie:
Is that a general download failure? What error do you see?

spannm:
Hi @hgschmie, this occurred behind a network proxy and with a Nexus artifact manager in between in the office today. Most probably this can be ignored ...

PR #2621, opened 2024-01-30
SqlObject: add `@Definition` annotation for defining computed constants

stevenschlansker:


hgschmie:
Given that we already have a `@Define` annotation, I am not a huge fan of the name. 

Also, this needs to be documented, please add some documentation (more than one sentence) to the index.adoc

hgschmie (approved):

stevenschlansker:
Do you have a suggestion for a better name? It's intentionally similar since it has the same effect under the hood.
I did add documentation, including limitations and an example use case.

stevenschlansker:
I left this as `@Beta` so if we come up with a better name I'm all ears. I thought of `@DefineConstant` or `@DefineAttribute` but neither of them felt compellingly better to me.

PR #2625, opened 2024-02-05
Testcontainers DB2; resolves #1654

### LINKED ISSUE #1654: DB2 Blob is returned as a com.ibm.db2.jcc.am.c6 object
Hi,
inserting a byte array into a blob column works fine, but when reading from the DB you receive a IBM c6 object (form IBM jdbc driver).

It seems it is possible to convert to a byte array with the below code. 
I'm not sure how could I use this with ArgumentFactory or ColumnMapper?

```java
private static byte[] db2Blob(final c6 blob)
	{
		byte[] byteArr = null;
		try
		{
			commpressedFile = new byte[blob.getBinaryStream().available()];
			blob.getBinaryStream().read(byteArr);
		}
		catch (final SQLException | IOException e)
		{
			e.printStackTrace();
		}

		return byteArr;
	}
```

stoyants:
DB2 Testcontainer setup.
A test fetching blob as "byte[]" added.

### DIFF
--- a/testcontainers/pom.xml
+++ b/testcontainers/pom.xml
@@ -33,6 +33,7 @@
         <!-- docker tests run long -->
         <basepom.test.timeout>600</basepom.test.timeout>
         <dep.clickhouse.version>0.6.0</dep.clickhouse.version>
+        <dep.db2.version>11.5.9.0</dep.db2.version>
         <dep.oracle-xe.version>23.3.0.23.09</dep.oracle-xe.version>
         <dep.trino.version>435</dep.trino.version>
         <dep.yugabyte.version>42.3.5-yb-4</dep.yugabyte.version>
@@ -65,6 +66,12 @@
                 <groupId>io.trino</groupId>
                 <artifactId>trino-jdbc</artifactId>
                 <version>${dep.trino.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.ibm.db2</groupId>
+                <artifactId>jcc</artifactId>
+                <version>${dep.db2.version}</version>
             </dependency>
         </dependencies>
     </dependencyManagement>
@@ -157,6 +164,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.ibm.db2</groupId>
+            <artifactId>jcc</artifactId>
+            <scope>test</scope>
+        </dependency>
+
         <!-- containers -->
 
         <dependency>
@@ -210,6 +223,12 @@
         <dependency>
             <groupId>org.testcontainers</groupId>
             <artifactId>mssqlserver</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>db2</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/testcontainers/src/main/java/org/jdbi/v3/testing/junit5/tc/TestcontainersDatabaseInformation.java
+++ b/testcontainers/src/main/java/org/jdbi/v3/testing/junit5/tc/TestcontainersDatabaseInformation.java
@@ -66,6 +66,8 @@ public final class TestcontainersDatabaseInformation {
     private static final TestcontainersDatabaseInformation MSSQL =
         of("sa", null, null, (catalogName, schemaName) -> format("CREATE DATABASE %s", catalogName));
 
+    private static final TestcontainersDatabaseInformation DB2 =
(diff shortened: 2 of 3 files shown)


stevenschlansker (approved):
Thanks @stoyants ! Looks good except you need to run `mvn license:format` to add the silly license header that everyone seems to insist that gets applied everywhere ;)

stoyants:
@stevenschlansker Sorry, missed this one with the license. Now is fixed.

stevenschlansker:
```
 Caused by: java.lang.IllegalStateException: Failed to verify that image 'icr.io/db2_community/db2:11.5.9.0' is a compatible substitute for 'ibmcom/db2'. This generally means that you are trying to use an image that Testcontainers has not been designed to use. If this is deliberate, and if you are confident that the image is compatible, you should declare compatibility in code using the `asCompatibleSubstituteFor` method. For example:
   DockerImageName myImage = DockerImageName.parse("icr.io/db2_community/db2:11.5.9.0").asCompatibleSubstituteFor("ibmcom/db2");
and then use `myImage` instead.
	at org.testcontainers.utility.DockerImageName.assertCompatibleWith(DockerImageName.java:279)
	at org.testcontainers.containers.Db2Container.<init>(Db2Container.java:45)
	at org.jdbi.v3.testing.junit5.tc.DB2JdbiTestContainersExtensionTest.<clinit>(DB2JdbiTestContainersExtensionTest.java:32)
	... 17 more
```

Hm.

stoyants:
@stevenschlansker 
Indeed, "Hm" :( 
Never seen this before, while testing localy.
I will have a look tomorrow.
Sorry for this, everything was realy running slow, but smoothly, on my machine.

stevenschlansker:
Thanks, there is no hurry! Maybe @hgschmie  has seen this sort of thing before, I have never developed with Testcontainers myself.

stoyants:
@stevenschlansker 
I just double checked, it runs without any problems in my environment. (WSL Ubuntu under Windows 10 on Intel CPU)
[This log](https://github.com/jdbi/jdbi/actions/runs/7796040931/job/21259986944?pr=2625#step:5:3551) also shows successful execution of the tests.
I'm happy to add suggested `.asCompatibleSubstituteFor("ibmcom/db2")`, but am also sceptical that it will solve the problem you got.
Can you give me some hint about the environment where do you get this exception? (OS, CPU, Docker)
I was suspecting that it might be Mac-related and I have to add `@EnabledOnOs(architectures = {"x86_64"})`, but IBM claims in this document [Installing the Db2 Community Edition Docker image on macOS systems](https://www.ibm.com/docs/en/db2/11.5?topic=system-macos) that their Comunity Edition is available for MacOS as well.
Also, "testcontaners" support both docker repos:
image

stoyants:
@stevenschlansker 
I just noticed something else, in your stacktrace "assertCompatibleWith" invocatino is at line 45 of Db2Container.java.
Is it (somehow) possible that you  are useing "testconatiners" in version older than 1.19.0?
Because the support of the new IBM repo, was introduced only in version 1.19.0
```
commit 9e8b18b72d9c8e61e737840b2db69fcceb9c2d2e
Author: Eddú Meléndez Gonzales <eddu.melendez@gmail.com>
Date:   Fri Aug 4 11:15:17 2023 -0600

    Support icr.io/db2_community/db2 as a compatible image (#7397)

    `ibmcom/db2` is deprecated. New image location is `icr.io/db2_community/db2`

```

stevenschlansker:
Thanks, the environment with the failure is the CI: https://github.com/jdbi/jdbi/actions/runs/7796040931/job/21259987695?pr=2625

stoyants:
@stevenschlansker Yes, I see. GitHub CI/CD pipeline forces older testcontainers version (because of "mysql"). I was able to reproduce the problem with the older version.  And I set the suggested setting for backward compatibility, it runs fine for me now. I hope it will be fine also in GitHub CI.

stevenschlansker:
@hgschmie , here is an example where we already run into confusing bugs with old testcontainers :) I hope we can only support the latest version soon.

@stoyants , thank you for tracking this down, and for the contribution.

hgschmie:
Wohoo, DB2! :-) Thank you @stoyants!

@stevenschlansker I hear you. We need to accept that we have users that can not upgrade all the time, so if we want to support features that are only available in a specific version of e.g. tc, we need a conditional. 

Glad that it worked out in this case. Again, wohoo DB2. :-)

hgschmie:
(btw, adding this to the docs (index.adoc, release notes) would be great. Otherwise I will do it next time I am working on docs.

PR #2630, opened 2024-02-11
BasePOM update + reproducible build

hgschmie:
- Update to BasePOM 59
- bump inline plugin to 1.4.0
- add necessary properties to support reproducible builds

### DIFF
--- a/internal/root/pom.xml
+++ b/internal/root/pom.xml
@@ -18,7 +18,7 @@
     <parent>
         <groupId>org.basepom</groupId>
         <artifactId>basepom-oss</artifactId>
-        <version>58</version>
+        <version>59</version>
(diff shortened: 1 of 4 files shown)


internal/root/pom.xml
```diff
@@ -18,7 +18,7 @@
     <parent>
         <groupId>org.basepom</groupId>
         <artifactId>basepom-oss</artifactId>
-        <version>58</version>
```
stevenschlansker, line 21: MR says 60, but this says 59

stevenschlansker (approved):

internal/root/pom.xml
hgschmie: typo. 59 is correct.

PR #2632, opened 2024-02-12
Prevent database container to shutdown while statements are still executed

stoyants:


hgschmie (changes requested):
We need to talk more about this. Please don't merge right away.

testcontainers/src/main/java/org/jdbi/v3/testing/junit5/tc/TestcontainersDatabaseInformationSupplier.java
```diff
@@ -124,6 +132,9 @@ public TestcontainersDatabaseInformation get() {
     private void executeStatements(final List<String> statements) throws SQLException {
         try (Connection c = dataSource.getConnection()) {
             for (String statement : statements) {
+                if (closed.get()) {
```
hgschmie, line 135: this piece makes sense.

testcontainers/src/main/java/org/jdbi/v3/testing/junit5/tc/TestcontainersDatabaseInformationSupplier.java
```diff
@@ -41,6 +42,7 @@ final class TestcontainersDatabaseInformationSupplier implements Supplier<Testco
     private final SynchronousQueue<TestcontainersDatabaseInformation> nextSchema = new SynchronousQueue<>();
     private final AtomicBoolean closed = new AtomicBoolean();
     private final CountDownLatch stopped = new CountDownLatch(1);
+    private final ReentrantLock mutex = new ReentrantLock();
```
hgschmie, line 45: you don't need the mutex. The CountDownLatch right next door does the exact same thing...

testcontainers/src/main/java/org/jdbi/v3/testing/junit5/tc/TestcontainersDatabaseInformationSupplier.java
```diff
@@ -62,6 +64,7 @@ void start(DataSource dataSource) {
     public void close() {
         LOG.info("Shutdown initiated...");
         if (!this.closed.getAndSet(true)) {
+            mutex.lock();
             executor.shutdownNow();
             try {
                 if (!stopped.await(10, TimeUnit.SECONDS)) {
```
hgschmie, line 70: ... just change this line to `if (!stopped.await()) {`

hgschmie (changes requested):
So this is not the right fix. here is why:

- firstly this introduces a potential infinite hang. If, for whatever reason the statement execution would not make progress, it holds the mutex indefinitely and in turn the main thread would be infinitely blocked. This is a bad outcome for any test that is using this code.
- secondly, all the mutex does is emulate the effect of the countdown latch that is already in there. The main thread *waits* for the database creation thread to do its thing and finish. Once it finishes, it calls `stopped.countDown()` to signal that it has finished and exited. The main thread waits for this in the `if (!stopped.await(10, TimeUnit.SECONDS)) {` statement. 

You can accomplish the exact same effect by removing the timeout. In that case, the main thread will wait indefinitely for the database creation thread to finish. 

Interrupting the execution is the right thing to do (which the main thread does). Adding the `if (!closed.get())` is useful to shortcut statement execution but ultimately simply not logging the exceptions when shutting down is most likely the right thing to do.

stoyants:
@hgschmie 
I created a simple test, which shows how OJDBC reacts on thread interruption: https://github.com/stoyants/ojdbc-test
It is only 2 files, pom.xml and one test class, and uses testcontainers + ojdbc, but no HikariCP or JDBI.
Upon execution test will generate nearly the same log as I got before in jdbi3-oracle12.

Best Regards, Stoyan

Issue #2629, opened 2024-02-11
Exceptiona are thrown after tests execution, if  the database test container is closed prematurely.


stoyants:
Hi guys,
I found out that after execution of TestScript in the "jdbi3-oracle12" some Exceptions pop-at the end.
Tests are fine, they are all successful, but these Exceptions are confusing and annoying.
An example can be found in the attached file.
[TestScript.log](https://github.com/jdbi/jdbi/files/14233909/TestScript.log)

I have dig a little bit and found out, that the problem happen as follows:
In the TestcontainersDatabaseInformationSupplier class the main thread "run()" method has an "infinite" loop, which is creating a new TestcontainersDatabaseInformation (a new schema and/or user, depending on the database) as soon as one is consumed via there "get()" method. This includes also execution of the list of statements for schema/user initialization in the private executeStatements(....) method.
So, the problem happens if the test is so short and fast, that it finishes, before the next schema / user is initialised completely.
The main testing thread invokes the close() method, it does not check are any statements are currently executed and then container shutdown is initiated. Then statements can't be executed anymore in the "[database-schema-creator" and the java.sql.SQLRecoverableException is thrown.

I also have a suggestion how this can be avoided.
1. as first and simple step, the loop executing initializing statement can check is the TestcontainersDatabaseInformationSupplier is already close. like this:

```java
    private void executeStatements(final List<String> statements) throws SQLException {
        try (Connection c = dataSource.getConnection()) {
            for (String statement : statements) {
                if (closed.get()) {
                    break;
                }
                try (Statement stmt = c.createStatement()) {
                    stmt.executeUpdate(statement);
                }
            }
        }
    }
```

2. But the above is still not sufficient, we have to protect the invocation of the "executeStatements" using ReentrantLock mutex.

```java
    private final ReentrantLock mutex = new ReentrantLock();

    @Override
    public void close() {
        LOG.info("Shutdown initiated...");
        if (!this.closed.getAndSet(true)) {
            mutex.lock();
            executor.shutdownNow();
            ...
        }
    }

    @Override
    public void run() {
        while (!closed.get()) {
            try {
                ......
                try {
                    mutex.lock();
                    executeStatements(databaseInformation.getCreationScript());
                } finally {
                    mutex.unlock();
                }
                ......
            }
        }
        stopped.countDown();
    }

```

Of course, will be glad to see any any other suggestions how to avoid this Exception.

If you agree to review my suggestion, I can open a MR.

Best Regards, Stoyan

[...]

PR #2634, opened 2024-02-15
Make jdbi testcontainers shutdown configurable

hgschmie:
Allow setting the timeout for shutting down the JdbiTestContainersExtension. This addresses the
issue that the database may shut down before init scripts are completely executed which results
in spurious error messages.

Addresses #2629

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,7 @@
 
 - Test DB2 basic integration (#2625, thanks @stoyants)
 - create CycloneDX SBOM files for release versions
+- add `JdbiTestContainersExtension#setShutdownWaitTime(int)` to control waiting for extension shutdown if a database is very slow. Addresses #2629 (thanks @stoyants).
 
 # 3.44.1
 
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -6187,7 +6187,7 @@ The `dbContainer` object is registered as class level field for sharing between 
 
 The link:{jdbidocs}/testing/junit5/tc/JdbiTestcontainersExtension.html[JdbiTestcontainersExtension^] has built-in support to isolate each test method in a test class. The container can be declared as a static class member which speeds up test execution significantly.
 
-There is built-in support for MySQL, MariaDB, TiDB, PostgreSQL, CockroachDB, YugabyteDB, ClickHouseDB, Oracle XE, Oracle free, Trino and MS SQLServer. It also supports the various compatible flavors (such as PostGIS for PostgreSQL).
+There is built-in support for MySQL, MariaDB, TiDB, PostgreSQL, CockroachDB, YugabyteDB, ClickHouseDB, Oracle XE, Oracle free, Trino, MS SQLServer and DB2. It also supports the various compatible flavors (such as PostGIS for PostgreSQL).
 
 By default, Jdbi tries to run every test with its own schema instance in a single schema. For database engines that do not support schemas, it will use separate catalogs. For database engines that do not support catalogs or schemas (e.g. MySQL), it will use separate databases.
 
@@ -6203,6 +6203,7 @@ For each database, Jdbi can provide custom user, catalog and schema names and re
 | Oracle XE, Oracle free | Schema | `system` | (default catalog) | (random)
 | Trino | Schema | (default) | `memory` | (random)
 | MS SQLServer | Catalog | `sa` | (random) |
+| DB2 | Schema | (default) | `test` | (random) |
 |=====
 
 - Database user _(default)_: Jdbi uses the default user as returned by the testcontainer class.
@@ -6249,6 +6250,20 @@ private static final TestcontainersDatabaseInformation CUSTOM_MYSQL =
 
 
 The Javadoc for the link:{jdbidocs}/testing/junit5/tc/TestcontainersDatabaseInformation.html[TestcontainersDatabaseInformation^] class has additional details on how to create custom link:{jdbidocs}/testing/junit5/tc/JdbiTestcontainersExtension.html[JdbiTestcontainersExtension^] instances.
+
+==== Controlling the extension shutdown timeout
+
+Some testcontainer based database may take a long time to create a new schema so when tests run very fast, shutting down the link:{jdbidocs}/testing/junit5/tc/JdbiTestcontainersExtension.html[JdbiTestcontainersExtension^] at the end of a test may interrupt this process and log spurious messages in the logs. Those messages are benign and can be safely ignored. If it is necessary (e.g. because there is a "no messages" policy in place at your organization), the shutdown time can be extended by setting the wait time to a higher value than the default of 10 seconds:
+
+[source,java,indent=0]
+----
+@RegisterExtension
+JdbiExtension extension = JdbiTestcontainersExtension.instance(CUSTOM_MYSQL, dbContainer)
+    .setShutdownWaitTimeInSeconds(20); // set shutdown wait time to 20 seconds
+----
+
+[WARNING]
+Using a wait time of 0 seconds will wait until all of the internal state has been shutdown completely. This may result in infinite hangs of the tests when shutting down.
 
 
 == Third-Party Integration
--- a/testcontainers/src/main/java/org/jdbi/v3/testing/junit5/tc/JdbiTestcontainersExtension.java
+++ b/testcontainers/src/main/java/org/jdbi/v3/testing/junit5/tc/JdbiTestcontainersExtension.java
@@ -17,6 +17,7 @@ import javax.sql.DataSource;
 
 import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
+import org.jdbi.v3.meta.Alpha;
 import org.jdbi.v3.meta.Beta;
 import org.jdbi.v3.testing.junit5.JdbiExtension;
 import org.testcontainers.containers.JdbcDatabaseContainer;
@@ -45,7 +46,7 @@ public final class JdbiTestcontainersExtension extends JdbiExtension {
      * @return An initialized {@link JdbiExtension} instance that uses the database container.
      * @throws IllegalArgumentException If the provided container class is not supported.
      */
-    public static JdbiExtension instance(JdbcDatabaseContainer<?> jdbcDatabaseContainer) {
+    public static JdbiTestcontainersExtension instance(JdbcDatabaseContainer<?> jdbcDatabaseContainer) {
         TestcontainersDatabaseInformation databaseInformation = TestcontainersDatabaseInformation.forTestcontainerClass(jdbcDatabaseContainer.getClass());
 
         if (databaseInformation == null) {
@@ -62,7 +63,7 @@ public final class JdbiTestcontainersExtension extends JdbiExtension {
      * @param jdbcDatabaseContainer A {@link JdbcDatabaseContainer} instance.
      * @return An initialized {@link JdbiExtension} instance that uses the database container.
      */
-    public static JdbiExtension instance(TestcontainersDatabaseInformation databaseInformation, JdbcDatabaseContainer<?> jdbcDatabaseContainer) {
+    public static JdbiTestcontainersExtension instance(TestcontainersDatabaseInformation databaseInformation, JdbcDatabaseContainer<?> jdbcDatabaseContainer) {
         return new JdbiTestcontainersExtension(databaseInformation, jdbcDatabaseContainer);
     }
 
@@ -70,6 +71,19 @@ public final class JdbiTestcontainersExtension extends JdbiExtension {
         this.jdbcDatabaseContainer = jdbcDatabaseContainer;
         this.databaseInformation = databaseInformation;
         this.instanceProvider = new TestcontainersDatabaseInformationSupplier(databaseInformation);
+    }
+
+    /**
+     * Sets the maximum wait time for shutting down the extension.
+     * @see TestcontainersDatabaseInformation#setShutdownWaitTimeInSeconds(int)
+     *
+     * @since 3.45.0
+     */
+    @Alpha
+    public JdbiTestcontainersExtension setShutdownWaitTimeInSeconds(int seconds) {
+        this.databaseInformation.setShutdownWaitTimeInSeconds(seconds);
+
+        return this;
     }
 
     @Override
--- a/testcontainers/src/main/java/org/jdbi/v3/testing/junit5/tc/TestcontainersDatabaseInformation.java
+++ b/testcontainers/src/main/java/org/jdbi/v3/testing/junit5/tc/TestcontainersDatabaseInformation.java
@@ -21,6 +21,7 @@ import java.util.Optional;
 import java.util.StringJoiner;
 import java.util.function.BiFunction;
 
+import org.jdbi.v3.meta.Alpha;
 import org.jdbi.v3.meta.Beta;
 import org.testcontainers.containers.JdbcDatabaseContainer;
 
@@ -123,6 +124,8 @@ public final class TestcontainersDatabaseInformation {
     private final String schema;
     private final BiFunction<String, String, List<String>> createStatement;
 
+    private volatile int shutdownWaitTimeInSeconds = 10; // default shutdown wait time is 10 seconds
+
     /**
      * Creates a new database information instance that describes a database.
      *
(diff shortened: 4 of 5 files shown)


hgschmie:
@stoyants This is an alternative change to #2632. I am reluctant to call it a fix because there is nothing broken, just some spurious messages. Calling

```java
@RegisterExtension
    JdbiExtension extension = JdbiTestcontainersExtension.instance(...)
        .setShutdownWaitTimeInSeconds(0);
```

has the same effect as the proposed change that adds the mutex.

Please let me know if that works for you. However, the messages from the connection pool where it complains that a connection "is broken" because either the server shut down or the client was interrupted will not change, as those are thrown from within the pool code.

stoyants:
@hgschmie Thanks for spending time on this issue.
Pity, I can't find such method "setShutdownWaitTimeInSeconds(int)" neither in JdbiExtension, nor in OracleContainer.
But I did some tests and they show that in fact container is fine, I can establish a new connection to the Oracle even after test are over and the exceptions get thrown.

I also did some tests on the same computer, but under "vanilla" Linux - exception was thrown only once in 50 runs.

I'm still "hunting" it. For the moment  my theory is that "executor.shutdownNow();"  causes "java.io.InterruptedIOException: Socket read interrupted"

Look, the problem does not deserve so much of your time. I can take back the Pull Request (close it) if you prefer.
But please don't close the issue, I will work on it. If I find anything new or have some new idea, will share it here.

Best regards, Stoyan

hgschmie:
Well, yes, the `setShutdownWaitTimeInSeconds` is what we will have, once that PR goes in. :-) 

If you want to try it out, do a `gh pr checkout 2634` in the project, then build the tree.

stevenschlansker:
> Please let me know if that works for you. However, the messages from the connection pool where it complains that a connection "is broken" because either the server shut down or the client was interrupted will not change, as those are thrown from within the pool code.

I want to push back against this being seen as not-a-bug. The code should correctly tear down its resources, which includes making sure that the connection pool is terminated before the database is shut down. Unless I'm mistaken, this used to work very reliably before we introduced all the extension abstractions, so it would be nice to at least maintain "no spurious errors" as a goal.

I will try to review this later today.

hgschmie:
> > Please let me know if that works for you. However, the messages from the connection pool where it complains that a connection "is broken" because either the server shut down or the client was interrupted will not change, as those are thrown from within the pool code.
> 
> I want to push back against this being seen as not-a-bug. The code should correctly tear down its resources, which includes making sure that the connection pool is terminated before the database is shut down. Unless I'm mistaken, this used to work very reliably before we introduced all the extension and testcontainers abstractions, so it would be nice to at least maintain "no spurious errors" as a goal.
> 

Understandably. However, the testcontainer code has always worked that way. 

When we shut down the testcontainer extension and terminate the TestcontainersDatabaseInformationSupplier, we do two things:
- we set the "closed" bit so the database supplier thread will exit the next time it goes through its loop
- we call "executor.shutDownNow()" which interrupts all threads, especially the database supplier thread.

Then the code will wait for the countdown latch being counted down which is the thread signalling that it exited.

Now one of two things happens:

*Either*

The interrupt makes it all the way through, e.g. to a client that is sitting in an I/O wait because it waits for a response from the database. In that case, e.g. oracle throws a `java.io.InterruptedIOException: Socket read interrupted` which results in a `java.sql.SQLRecoverableException: ORA-18730: Interrupted IO error.: Socket read interrupted`. That would not be a problem, except that the connection pool that we use will catch this exception in its proxy statement code and loudly complain about it:

```
[database-schema-creator] WARN com.zaxxer.hikari.pool.ProxyConnection - jdbi-template-pool (*.*) - Connection oracle.jdbc.driver.T4CConnection@4886b6bd marked as broken because of SQLSTATE(08006), ErrorCode(18730)
java.sql.SQLRecoverableException: ORA-18730: Interrupted IO error.: Socket read interrupted
```

The annoying part is that the pool catches the exception, logs the stack trace on WARN level and then rethrows it. So even if we expect and suppress logging on our side, the connection pool will still log it - https://github.com/brettwooldridge/HikariCP/issues/2172

*OR* 

the client keeps waiting (the interrupt gets lost or something) so at some point, the wait for the countdown latch times out, there is a 

```
[WARN} Could not shut down database creation thread within 10 seconds
```

log message and then the main thread continues on to the testcontainer extension, shuts that down at which point the client will lose connection to the server and probably log another error.

----

If you are willing to take some performance penalty for the sake of 'this needs to not log a spurious "we are shutting down, please ignore" exception' (this is *NOT* a bug. There is no error condition or anything. This is a distributed system shutting down forcefully (which is what the `executor.shutDownNow()` is). There is *NOTHING WRONG* or any test failure or anything).

The database creator thread will sit in I/O in various situations:
- when executing a database command to the server
- when trying to put the next database information object into the synchronous queue

The first situation is fine, those commands will finish at some point and the thread will leave IOWait. So the shutdown/await combo work (there needs to be a `if (closed.get()) break;` between the `executeStatements(...)` and the `nextSchema.put(...)` but that is simple.

However, if the thread sits in the `nextSchema.put(...)`, it will wait forever until another test takes the database info object from the queue. And there are no more tests because we are shutting down. So the thread needs to be interrupted. We could retain the `Future<?>` returned from the `submit` to the pool and then cancel the future, but that has the same effect as the `shutDownNow()` - it interrupts the thread and everything described above will happen the same way. 


Using the mutex makes no difference. In fact, what would happen is that the database creator thread would take the mutex, then execute its database creation commands (which may run a long time) and *refuse* (because it blocks the thread that is trying to shut down the database creator thread on the mutex) to exit earlier, again leading to long, uncontrollable hangs. And interrupting it, will cause the same warn messages to crop up.

This code has been used forever (its lineage goes back to the otj-pg code through the embedded pg code, see https://github.com/hgschmie/pg-embedded/blob/master/src/main/java/de/softwareforge/testing/postgres/embedded/DatabaseManager.java#L214-L218). The reason why this is not visible here is because the connection pool (or lack thereof) does not log the interrupted exceptions. HikariCP is trying to be helpful here by logging a condition as "WARN" that is expected. If I could suppress it, I would simply add code to make it invisible. But https://github.com/brettwooldridge/HikariCP/issues/2171 is blocking this. 

So this boils down to: We could change the code to look like this:

```java
private volatile Future<?> threadFuture = null;

void start(DataSource dataSource) {
        this.dataSource = dataSource;
        this.threadFuture = this.executor.submit(this);
    }

    @Override
    public void close() {
        LOG.info("Shutdown initiated...");
        if (!this.closed.getAndSet(true)) {
            executor.shutdown();
            if (threadFuture != null) {
                threadFuture.cancel(true);
            }
            try {
                if (!(executor.awaitTermination(... ) && stopped.await(...))) {
                    LOG.warn("Could not shut down database creation thread");
                }
            } catch (InterruptedException e) {
                Thread.currentThread().interrupt();
            }
            LOG.info("Shutdown completed.");
        }
    }
```

This works, but I can still trivially get this (from mysql):

```
[database-schema-creator] WARN com.zaxxer.hikari.pool.ProxyConnection - jdbi-template-pool (*.*) - Connection com.mysql.cj.jdbc.ConnectionImpl@54067fdc marked as broken because of SQLSTATE(08S01), ErrorCode(0)
com.mysql.cj.jdbc.exceptions.CommunicationsException: Communications link failure

The last packet successfully received from the server was 149 milliseconds ago. The last packet sent successfully to the server was 156 milliseconds ago.
```

by exceeding the deadline in the "awaitTermination" and "await". `DO SLEEP(...)` in MySQL helps me debugging this. :-)  


I am not sure, what you would do. I am open to any suggestions, but as this is not even a problem (it is just a spurious log line), I feel the easiest is to change our slf4j logging setup to suppress the `com.zaxxer.hikari.pool.ProxyConnection` warn messages...

stoyants:
@hgschmie 

First, please forget what I said some time ago, that exceptions pop-up, because the database container is shutdown. It is shutdown, but later and it is fine. Container side is not involved at all here. It is only the client (JDBI code + Hikari CP + Oracle JDBC). With a small effort I can make the exceptions pop-up in the middle of the tests, just repeat the fast test, which also shows, that we have to exclude the "container is shutdown too early" hypothesis.

Second, a general observation - because of the infinite loop in run() method of "database-schema-creator" when "close()" is invoked it will find the thread wither waiting on "nextSchema.put(..)" or in "executeStatements(...)". The former happen "nearly always"  and is well handled, but later happen also, like in my case, it is not handled and some JDBC drivers don't accept it very well.

> - we set the "closed" bit so the database supplier thread will exit the next time it goes through its loop
> - we call "executor.shutDownNow()" which interrupts all threads, especially the database supplier thread.

Yes, this is exactly what happen. But it is a little bit like a policeman waiting just behind the speed limit sign to issue you a ticket. I mean, we say "you are requested to stop" and then immediately initiate the shutdown process, not giving any chance to the thread, creating the next schema to finish grecefully.

The problem is that Oracle JDBC is very "sensitive" to interruptions. They (literally) have a code like:
```
... if (Thread.interrupted()) {
	throw new InterruptedIOException("Socket read interrupted");
}
```
And seems that "shutdownNow()" is doing exactly this - it sends interrupt to the Thread.

Then Oracle JDBC catch, log and re-throw - no way to stop them.
The same happen in HikariCP - catch, log and re-throw - no way to stop them. (I saw the request to HikariCP)
We can catch and supress the Exception, but it will hide only 1 of 3 logs.

One thing we all agree is to have "if (closed.get()) break;" in "executeStatements(...)"
But we have to avoid calling "shutdownNow()" while in "stmt.executeUpdate(statement)"

I see at least 2 options:
1. just give some time to the "database-schema-creator" thread to notice close request:
```
        if (!this.closed.getAndSet(true)) {
            try {TimeUnit.SECONDS.sleep(1);} catch (InterruptedException e) { } // swallow the exception, it is intended
            executor.shutdownNow();
```
The problem here is that it has to wait, even when not necessary. And it never knows is the wait long enough.

2. use the mutex, but with timeout (tryLock instead of lock) - if it can wait 10 seconds after shutdownNow(), it should be fine to wait 1 second before it - usualy the second is not used, few milliseconds are enough, but to be on the safe side.
```
    public void close() {
        LOG.info("Shutdown initiated...");
        if (!this.closed.getAndSet(true)) {
            try {
                if (!mutex.tryLock(1, TimeUnit.SECONDS)) {
                    LOG.warn("Could not stop database creation statements within 1 second");
                }
            } catch (InterruptedException e) {
                LOG.warn("Interrupt ignored");
            }
            executor.shutdownNow();
            try {
                if (!stopped.await(10, TimeUnit.SECONDS)) {
                    LOG.warn("Could not shut down database creation thread within 10 seconds");
                }
            } catch (InterruptedException e) {
                Thread.currentThread().interrupt();
            }
            LOG.info("Shutdown completed.");
        }
    }
```
Funny enough, processing the Exception (logging) can take much longer (up to a second) then waiting for the statement to finish gracefully (less than 200 milliseconds).

I have updated the PR with the above version.

I tried to keep the change small, maybe what @hgschmie is suggesting will also work, but it seems a bigger change, with more unknown parameters.

stoyants:
@hgschmie and @stevenschlansker 
I thin this issue (#2634) and (#2632) are not related at all. The name of the issue, which I open #2632 is wrong and misleading - it should be "Prevent 'database-schema-creator'  to be shutdown while schema creation statements are still executed"

hgschmie:
> @hgschmie
> 
> First, please forget what I said some time ago, that exceptions pop-up, because the database container is shutdown. It is shutdown, but later and it is fine. Container side is not involved at all here. It is only the client (JDBI code + Hikari CP + Oracle JDBC). With a small effort I can make the exceptions pop-up in the middle of the tests, just repeat the fast test, which also shows, that we have to exclude the "container is shutdown too early" hypothesis.

Correct. As I wrote above, it is one of two conditions:

- interrupting the client while it sits in I/O Wait through the `excecutor.shutDownNow()` or `Future.cancel()`
- timing out on the countdown latch await, thus logging a warning and shutting down the test container extension

The first one is much more common and most likely what you observe.


> Second, a general observation - because of the infinite loop in run() method of "database-schema-creator" when "close()" is invoked it will find the thread wither waiting on "nextSchema.put(..)" or in "executeStatements(...)". The former happen "nearly always" and is well handled, but later happen also, like in my case, it is not handled and some JDBC drivers don't accept it very well.

And that is where we disagree. All JDBC drivers handle it well (they receive an interrupt, they cease to do I/O wait and unblock the thread (by throwing an interrupted exception or some other exception). *That is the correct behavior to an exceptional event*. What is not handled well is the drivers (and the connection pool for that matter) spout all sorts of nonsense into the log files. This is the job of the caller who receives the exception, not the code that throws it.

This is part of where Java gets its bad rep. "Overreporting" standard operating conditions. There is this assumption that 'every exception is bad and signals an error'. Nothing could be further from the truth. Most code bases throw exceptions all the time. They get caught, handled, life goes on. In this case, both the driver and the connection pool feel the need to log a perfectly normal condition when shutting down infrastructure. Sucks IMHO but no reason to worry about it or go to great length "to avoid it".

Interrupting another thread is a normal operation in concurrent code. I don't understand why there would be a need to build these machinery which (best case scenario) does not change anything or (worst case scenario) leads to very long all the way to infinite hangs.

Can you explain more why you feel so strongly about this? Is there some driving need or is that just the urge of "my log files should never contain any exception under any circumstance"?

> Yes, this is exactly what happen. But it is a little bit like a policeman waiting just behind the speed limit sign to issue you a ticket. I mean, we say "you are requested to stop" and then immediately initiate the shutdown process, not giving any chance to the thread, creating the next schema to finish grecefully.

There is no need for graceful shutdown. If crashing the server would be faster, we should crash it. This is all about performance for unit tests. We don't preserve state, nuking and paving over is a fine strategy here.

> The problem is that Oracle JDBC is very "sensitive" to interruptions. They (literally) have a code like:
> 
> ```java
> ... if (Thread.interrupted()) {
> 	throw new InterruptedIOException("Socket read interrupted");
> }
> ```

It is not a problem. That is what I am trying to explain. What the driver does *is the right thing to do*. Logging that exception on the way to the caller who will catch it and deal with which creates unnecessary noise is the problem.


> And seems that "shutdownNow()" is doing exactly this - it sends interrupt to the Thread.

Yes. It is *literally* its job. see https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ExecutorService.html#shutdownNow()

> Then Oracle JDBC catch, log and re-throw - no way to stop them. The same happen in HikariCP - catch, log and re-throw - no way to stop them. (I saw the request to HikariCP) We can catch and supress the Exception, but it will hide only 1 of 3 logs.

So why bother? 

> 
> One thing we all agree is to have "if (closed.get()) break;" in "executeStatements(...)" But we have to avoid calling "shutdownNow()" while in "stmt.executeUpdate(statement)"

Why do you say "we have to" ? 

> 
> I see at least 2 options:
> 
> 1. just give some time to the "database-schema-creator" thread to notice close request:
> 
> ```java
>         if (!this.closed.getAndSet(true)) {
>             try {TimeUnit.SECONDS.sleep(1);} catch (InterruptedException e) {Thread.currentThread().interrupt();}
>             executor.shutdownNow();

Adding arbitrary sleeps never solves anything. If you sleep one second, something that takes two seconds will cause a problem. So you sleep two seconds, now something that takes three seconds is the problem. And so on. That is what the synchronization primitives are for. That the driver (and the pool) chose to log a normal condition is unfortunate but I still see no reason to make any code changes.

> 2. use the mutex, but with timeout (tryLock instead of lock) - if it can wait 10 seconds after shutdownNow(), it should be fine to wait 1 second before it - usualy the second is not used, few milliseconds are enough, but to be on the safe side.

So we introduce another synchronization mechanism on top of the two we already have (the countdown latch and Thread.interrupt())? 

I am at my end here. Unless there is a compelling reason beyond "any exception is bad and needs to be avoided", I see no reason to change code that works fine. I will work with the Hikari folks to be able to get rid of the exception logging in the pool code and for the driver, my recommendation is to tune the log level here (the oracle driver actually has quite a bit of infrastructure to tune the logging in specific situations).

hgschmie:
I reworked the code a bit and I think this is a workable solution:

- set the closed flag
- initiate excutor shutdown
- see if there is a database pending (so the creator thread sits in the put operation)
- if no, wait for the latch to count down. If the creation operation ends within the allotted timeout (or unlimited if it was set to zero), fine. The thread has finished, all good.

- otherwise, do shutDownNow. The thread would be either sitting in the put or has already exited (because the close flag was set), wait again (this time logging a possible warning)


I think that is the best best that I can come up with. This should allow of a configurable time to finish possible database creation scripts without interrupting but also allows fast shutdown for databases that don't need it. 

Note, that the *only* wait to avoid exceptions is to do 
```java
JdbiExtension extension = JdbiTestcontainersExtension.setShutdownWaitTimeInSeconds(0)
    .instance(getDbContainer())
    ...
```
(setting the shutdown time to "infinite") Otherwise, there are always scenarios where the shutting down the threadpool will interrupt the creator thread. But this can also cause infinite hangs in uni tests, something that we *will* avoid by default. If a caller chooses to do so, they do at their own risk.

Issue #2633, opened 2024-02-14
Thread-based nesting behavior for managed handles is surprising and not adequately documented


jpallas:
It's possible I've misunderstood or overlooked something, but as best I can tell:

The [documentation on managed transactions](https://jdbi.org/releases/3.41.2/#_managed_transactions) says  

> The [inTransaction](https://jdbi.org/releases/3.41.2/apidocs/org/jdbi/v3/core/Jdbi.html#inTransaction(org.jdbi.v3.core.HandleCallback)) and [useTransaction](https://jdbi.org/releases/3.41.2/apidocs/org/jdbi/v3/core/Jdbi.html#useTransaction(org.jdbi.v3.core.HandleConsumer)) methods on the [Jdbi](https://jdbi.org/releases/3.41.2/apidocs/org/jdbi/v3/core/Jdbi.html) object create a new [Handle](https://jdbi.org/releases/3.41.2/apidocs/org/jdbi/v3/core/Handle.html), open a transaction and pass it to the callback code.

> The [inTransaction](https://jdbi.org/releases/3.41.2/apidocs/org/jdbi/v3/core/Handle.html#inTransaction(org.jdbi.v3.core.HandleCallback)) and [useTransaction](https://jdbi.org/releases/3.41.2/apidocs/org/jdbi/v3/core/Handle.html#useTransaction(org.jdbi.v3.core.HandleConsumer)) methods on [Handle](https://jdbi.org/releases/3.41.2/apidocs/org/jdbi/v3/core/Handle.html) objects open a transaction on the handle itself and then pass it to the callback code. By default, if these methods are called while the handle is already in an open transaction, the existing transaction is reused (no nested transaction is created).

That says nothing about the behavior of non-transactional handle methods, and it doesn't mention reusing handles by the managed handle methods on `Jdbi`.

The [section on obtaining a managed handle](https://jdbi.org/releases/3.41.2/#_obtaining_a_managed_handle) does *not* mention anything about nesting behavior.  Reading it gives no reason to think that these methods might not return a new handle, and more importantly, might not return an auto-commit handle if the calling thread is inside the callback for a transactional handle.

But, in the implementation, it appears that this reuse of managed handles actually applies to both transactional handles and non-transactional handles.  I had code that indirectly called `Jdbi.useHandle` nested within a call to `Jdbi.inTransaction`, and the nested call did not get an autocommit handle, which caused much confusion and led to a deadlock situation.

[...]

PR #2635, opened 2024-02-16
Add additional documentation for nested callbacks

hgschmie:
@jpallas pointed out in #2633 that our documentation about nesting
callbacks with managed handles is incomplete and confusing.

Add a chapter about nested callback, call out the limitations, add
references to the managed handle and managed transaction chapters.


hgschmie:
@jpallas LMK if this the level of documentation that works for you. I still would be interested in possible lockup / error scenarios.

stevenschlansker (approved):

jpallas:
This is a really helpful improvement.  Thank you!  The only suggestion I have is to make some mention of handle reuse in the Javadoc for these methods.  IDEs will display Javadoc but don't know about the external documentation.

Issue #2623, opened 2024-02-04
JDBI can't find the registered value mapper for a Map<String, MyClass<SomeOtherClass>> return type


grssam:
Suppose a class

```java
public class MyClass<T> {
  ...
  T data;
  ...
}
```

using this in declarative-style approach.. This works:

```java
@SqlQuery("select * from my_table where condition=:value;")
@RegisterBeanMapper(MyClass.java)
List<MyClass> getList(String value);
```

but this fails with an exception "No row mapper registered for map value com.package.MyClass<T>"

```java
@SqlQuery("select m.name as name, m.* from my_table m where condition=:value;")
@RegisterBeanMapper(MyClass.java)
@KeyColumn("name")
MultiMap<String, MyClass> getMap(String value);
```

fails with `Map<String, MyClass>`  or `Map<String, MyClass<SomeOtherClass>>` or `MultiMap<String, MyClass<SomeOtherClass>>` as well.

I tried to dig through the code. Looks like that in the fluent api approach, this can be solved by providing the actual `Type` of the map value, but no such way possible in declarative style as annotation arguments need to be compile time constants.

hgschmie:
I think you are running against the limitations of what can be expressed with annotations. 
The main problem is that your "MyClass<T>" is parameterized but you use the raw type ("MyClass") in the Sql object. In addition, the `@RegisterBeanMapper` annotation does registers  a bean mapper for the bean in question but there is no type information on what to do with the "T" that is part of the type definition.

You basically have two options:

a) 

have a concrete type for the row. Instead of `MyClass<T>` have an `MyIntClass extends MyClass<Integer>` that resolves the `T` to an Integer (or whatever type you want to use). 

Now this works:

```java
@SqlQuery("select m.name as name, m.* from my_table m where condition=:value;")
@KeyColumn("name")
@RegisterConstructorMapper(MyIntClass.class)
Multimap<String, MyIntClass> getIntMap(String value);
```

Note that Jdbi still needs to know how to construct the Multimap. If this is a Guava multimap, you need to load the Guava plugin.

b) You do *not* have a concrete type but want to use a generic:

```java
@SqlQuery("select m.name as name, m.* from my_table m where condition=:value;")
@KeyColumn("name")
Multimap<String, MyClass<Integer>> getMap(int id);
```

Note that you still need to use a concrete type in the method. Java does not have dynamic typing, you need to provide an actual type information. 

[...]

PR #2636, opened 2024-02-16
Sample code for Multimap and generics

hgschmie:
Addresses the questions in #2623 with sample code.

### DIFF
--- a/e2e/src/test/java/org/jdbi/v3/e2e/TestIssue2623.java
+++ b/e2e/src/test/java/org/jdbi/v3/e2e/TestIssue2623.java
@@ -1,0 +1,120 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.e2e;
+
+import java.sql.ResultSet;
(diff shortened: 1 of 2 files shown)


e2e/src/test/java/org/jdbi/v3/e2e/TestIssue2623.java
```diff
@@ -0,0 +1,120 @@
...
+
+        @SqlQuery("select name, id from users where id = :id")
+        @KeyColumn("id")
+        @RegisterConstructorMapper(IntUser.class)
+        Multimap<Integer, IntUser> getIntUsers(int id);
+    }
+
+    public record User<T>(String name, T id) {}
+
+    public static class UserIntegerMapper implements RowMapper<User<Integer>> {
```
stevenschlansker, line 79: This might really neatly be replaced with a `ConstructorMapper`, unless showing this mapper is part of the point of the example.

e2e/src/test/java/org/jdbi/v3/e2e/TestIssue2623.java
```diff
@@ -0,0 +1,120 @@
...
+    public static class IntUser {
+        private final String name;
+        private final Integer id;
+
+        public IntUser(String name, Integer id) {
+            this.name = name;
+            this.id = id;
+        }
+
+        public String getName() {
```
stevenschlansker, line 95: This could be a `record` too to save some boilerplate hashCode / equals?

stevenschlansker (approved):

e2e/src/test/java/org/jdbi/v3/e2e/TestIssue2623.java
hgschmie: showing the mapper is the point of the example.

e2e/src/test/java/org/jdbi/v3/e2e/TestIssue2623.java
hgschmie: again, same goal. Show the actual implementation. Correct, it could be a record, too.

PR #2648, opened 2024-02-20
Handle constructing record classes

elonazoulay:
Get paramater types for records from RecordComponent. Record constructors lose generic type information. This is due to https://bugs.openjdk.org/browse/JDK-8320575.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
@@ -14,6 +14,13 @@
 package org.jdbi.v3.core.mapper.reflect;
 
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.Parameter;
+import java.lang.reflect.Type;
+import java.util.Arrays;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
 
 import org.jdbi.v3.core.internal.exceptions.Unchecked;
 
@@ -22,9 +29,17 @@ import static java.util.Objects.requireNonNull;
 class ConstructorInstanceFactory<T> extends InstanceFactory<T> {
     private final Constructor<T> constructor;
 
+    private final Supplier<Type[]> typeSupplier;
+
     ConstructorInstanceFactory(Constructor<T> constructor) {
         super(constructor);
         this.constructor = requireNonNull(constructor, "constructor is null");
+        this.typeSupplier = getTypeSupplier(constructor, super::getTypes);
+    }
+
+    @Override
+    Type[] getTypes() {
+        return typeSupplier.get();
     }
 
     @Override
@@ -35,5 +50,43 @@ class ConstructorInstanceFactory<T> extends InstanceFactory<T> {
     @Override
     public String toString() {
         return constructor.toString();
+    }
+
+    // Note: this can be removed once https://bugs.openjdk.org/browse/JDK-8320575 is resolved
+    private static <T> boolean isGenericInformationLost(Constructor<T> factory) {
+        // If there is more than one constructor, ensure that the factory constructor is the canonical constructor.
+        // If so it will need to get the type parameters from the declared fields.
+        if (factory.getParameters().length != getFields(factory)
+            .count()) {
+            return false;
+        }
+        boolean lossDetected = false;
+        for (int i = 0; i < factory.getParameters().length; i++) {
+            Parameter parameter = factory.getParameters()[i];
+            Field field = factory.getDeclaringClass().getDeclaredFields()[i];
+            if (!parameter.getName().equals(field.getName()) || !parameter.getType().equals(field.getType())) {
+                return false;
+            }
+            // Check if the generic type information is lost, due to jdk21 record constructor: https://bugs.openjdk.org/browse/JDK-8320575
+            if (parameter.getType().getTypeParameters().length > 0
+                && !parameter.getParameterizedType().equals(field.getGenericType())) {
+                lossDetected = true;
+            }
+        }
+        return lossDetected;
+    }
+
+    private static <T> Supplier<Type[]> getTypeSupplier(Constructor<T> constructor, Supplier<Type[]> defaultSupplier) {
+        if (isGenericInformationLost(constructor)) {
+            return () -> getFields(constructor)
+                .map(Field::getGenericType)
+                .toArray(Type[]::new);
+        }
+        return defaultSupplier;
+    }
+
+    private static <T> Stream<Field> getFields(Constructor<T> constructor) {
+        return Arrays.stream(constructor.getDeclaringClass().getDeclaredFields())
+            .filter(field -> !Modifier.isStatic(field.getModifiers()));
     }
 }
--- a/core/src/test/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapperTest.java
+++ b/core/src/test/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapperTest.java
@@ -14,6 +14,7 @@
 package org.jdbi.v3.core.mapper.reflect;
 
 import java.beans.ConstructorProperties;
+import java.util.Optional;
 
 import jakarta.annotation.Nullable;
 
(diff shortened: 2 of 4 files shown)


elonazoulay:
Context:
Constructors for record classes lose generic type information due to the jdk 21 bug mentioned above. Instead, get the type information from the record component.
The jdbi target jdk was updated to 17 since `java.lang.Class.isRecord()` and `getRecordComponents()` were introduced in java 16.

If you would like to keep the target jdk at 11 this can still be done:
Candidate for a record class is that all constructor and declared field parameters and `getType()` match.
If `parameter[i].getParametrizedType()` is an instance of `Class<?>` but `field[i].getGenericType()` is an instance of `ParametrizedType` then this implies the parameters types should be obtained via the fields.

Happy to do either way, whatever works. Also I will add tests: locally I was using h2 database to test, will get more familiar with how tests are done (this is my first pr to jdbi).

elonazoulay:
> Hi @elonazoulay! Thank you for submitting a PR for Jdbi! Please be aware that we still fully support JDK11, so any patch needs to compile and work with JDK11+. Afaik, the isRecord() method is JDK21+.

Hey @hgschmie! :) Sure, I can try the alternative approach above. Another idea: should I submit as a contrib RowMapperFactory that has `targetJdk` set to `21`?  If everything needs to support jdk11, I can go with the other idea, whatever works. Thanks!

elonazoulay:
Updated - lmk if this approach is ok.

stevenschlansker (approved):
Unfortunate that this is necessary. This can be backed out after the bug is fixed upstream, right? Maybe leave a comment to that effect?

thank you for the contribution :)

stevenschlansker:
Looks like Checkstyle is not happy:
```
[INFO] There are 4 errors reported by Checkstyle 10.12.7 with policy/checkstyle.xml ruleset.
Error:  src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java:[41,5] (blocks) LeftCurly: '{' at column 5 should be on the previous line.
Error:  src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java:[56,5] (blocks) LeftCurly: '{' at column 5 should be on the previous line.
Error:  src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java:[72,68] (whitespace) OperatorWrap: '&&' should be on a new line.
Error:  src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java:[81,5] (blocks) LeftCurly: '{' at column 5 should be on the previous line.
```

stevenschlansker (changes requested):
Oh, and some tests would be nice.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
```diff
@@ -36,4 +51,36 @@ T newInstance(Object... params) {
...
+        boolean candidate = false;
+        for (int i = 0; i < factory.getParameters().length; i++) {
+            Parameter parameter = factory.getParameters()[i];
+            Field field = factory.getDeclaringClass().getDeclaredFields()[i];
+            if (!parameter.getName().equals(field.getName()) || !parameter.getType().equals(field.getType())) {
+                return false;
+            }
+            // Check if the generic type information is lost, due to jdk21 record constructor: https://bugs.openjdk.org/browse/JDK-8320575
+            if (parameter.getType().getTypeParameters().length > 0 &&
+                !parameter.getParameterizedType().equals(field.getGenericType())) {
```
elonazoulay, line 73: The second condition could be simplified to `!(parameter.getParametrizedType() instanceof ParametrizedType)`, wdyt?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
```diff
@@ -36,4 +51,36 @@ T newInstance(Object... params) {
     public String toString() {
         return constructor.toString();
     }
+
+    private static <T> boolean isRecordConstructor(Constructor<T> factory)
+    {
+        // If there is more than one constructor, ensure that the factory constructor is the canonical constructor.
+        // If so it will need to get the type parameters from the declared fields.
+        if (factory.getParameters().length != Arrays.stream(factory.getDeclaringClass().getDeclaredFields())
+            .filter(field -> !Modifier.isStatic(field.getModifiers()))
```
elonazoulay, line 60: Should `isStatic` be statically imported?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
```diff
@@ -22,9 +28,18 @@
 class ConstructorInstanceFactory<T> extends InstanceFactory<T> {
     private final Constructor<T> constructor;
 
+    private final Supplier<Type[]> typeSupplier;
+
     ConstructorInstanceFactory(Constructor<T> constructor) {
         super(constructor);
         this.constructor = requireNonNull(constructor, "constructor is null");
+        this.typeSupplier = isRecordConstructor(constructor) ? getRecordParameterTypeSupplier(constructor) : super::getTypes;
```
elonazoulay, line 36: was thinking to use inheritance but then this method would have to be hoisted up, more code would change, but I can do that if it's preferred.

stevenschlansker: I think keeping the change small is good - unless I misunderstand, it would be nice to remove this code once the upstream bug is fixed.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
stevenschlansker: I usually wouldn't, unless it is used many times and the readability would improve.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
stevenschlansker: I worry a bit about the behavior if you have different types for your parameter and field, like

```
class Test {
    T field;
    Test(Supplier<T> field) {
        this.field = field.get();
    }
}
```

Should we make the check stronger, like checking `GenericTypes.isSuperType(parameter.getType(), field.getGenericType())`?

elonazoulay: Update: tried that and the `superType` parameter is erased in the test. The existing comparision catches this. Also for the example above, since Supplier<T> is a different class than T, it wouldn't even get that far since the types differ.

lmk if that's ok.

elonazoulay:
> Oh, and some tests would be nice.

I was able to test locally using java21 and record classes (using the h2 jdbi extension). Should I create a separate module that uses targetjdk of 21 and put tests there? lmk, I can do that, or anything else you suggest.

elonazoulay:
> Unfortunate that this is necessary. This can be backed out after the bug is fixed upstream, right? Maybe leave a comment to that effect?
> 
> thank you for the contribution :)

yep, I added a comment in the code to back out when the issue is resolved. Thanks for the review!

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
stevenschlansker: ok, that's fine - what I would like to see is a unit test showing that it works when it should and a negative test showing it doesn't match when it shouldn't

stevenschlansker:
For unit tests, is adding some general tests that run with the "current" jdk good enough, since we test with Java 21 in CI? Or do we need to set the target jdk explicitly to 21 to get this behavior?

elonazoulay:
> For unit tests, is adding some general tests that run with the "current" jdk good enough, since we test with Java 21 in CI? Or do we need to set the target jdk explicitly to 21 to get this behavior?

I was able to add tests to simulate the behavior: created a constructor with raw types. lmk if that works.

core/src/test/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapperTest.java
```diff
@@ -468,4 +469,83 @@ static MultipleStaticFactoryMethodsBean two(String s) {
             return new MultipleStaticFactoryMethodsBean();
         }
     }
+
+    // Note: the tests with ClassWithGeneric* are to simulate the behavior which triggers https://bugs.openjdk.org/browse/JDK-8320575
+    // A record constructor block will lose it's generic type information.
+    static class ClassWithGenericThing {
```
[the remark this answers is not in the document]
stevenschlansker: I don't think the "Thing" naming is a pattern we need to copy, but it's also ok to use :)

elonazoulay:
Thanks for the review! This will allow us to get onto jdk21, really helpful. I'll keep an eye on the jdk issue and submit an anti-pr that will revert this one (except for the tests) :)

hgschmie:
@elonazoulay  thank you so much for the patch and tests! wohoo!

What worries me a bit is that there are some assumptions that may or may not be true:

The "isGenericInformationLost" assumes that
- c'tor parameters and field count are the same
- parameter order and field order are the same
- parameter name and field name are the same

I guess that this matches the current implementation of records; is this set in stone (spec) somewhere or is that just heuristics.

Assuming that this matches, computing this information is expensive and done every time a record is constructed. That may lead to surprising performance impacts. Would it be possible to cache this information? 

Minor nit: `Note: this can be removed...` - "this" is all the added code. I doubt we can ever remove it because there will always be people with older revisions of Java 21 that would see a regression if we remove that in a future version. Maybe move the static methods into a small, package private helper (e.g. Jdk8320575Helper) or so? 

Generally speaking, I think that change is useful but it may have some unexpected impact on performance, especially when it is not actually needed (e.g. a record with no generic fields). Some benchmarking may be useful. This can most likely be alleviated with caching (which could even be just a static map, as the information will never change once computed.

Another approach, which would be fantastic (IMHO) would be to add a JDK 21 version of the ConstructorMapper in a multi-version jar. But I understand that we have not actually shipped a multi-version jar (yet), and PMD seems to be stalled in 7.0-rc4-hell, so we can not move source code compat to JDK 21 for some modules (because PMD 6 chokes on Java 21 classes). 

anyway, ship it. Until we release the next versions, we can always benchmark it and figure out the impact.

stevenschlansker:
I do think once the JDK is fixed, we can revert this. If users choose to not upgrade the JDK and suffer bugs at that point, that is their choice.

elonazoulay:
I will try the caching approach and see if I can put some benchmarks up as well. The fix is supposed to be backported to jdk21, I can try to special case it to the broken versions (including jdk22 - fix is targeted for jdk23) and then move it to the multi version jar. I will confirm about the constraints for record classes as well, will be interesting to find all this out. Thanks!

dalibor-petras:
@stevenschlansker:
Oracle released build  21.0.6-oracle resolving the root cause https://bugs.openjdk.org/browse/JDK-8334608.  New PR  with reversal could be tested.
Open source builds of v21 are yet to backport see https://github.com/openjdk/jdk21u-dev/pull/1026

PR #2650, opened 2024-02-22
Cache types in ConstructorMapper

elonazoulay:
Cache types to avoid expensive calculations in ConstructorInstanceFactory.
Following up on suggestions from #2648.

The constructor lookup should be quick: `hashCode` is class name and `equals` compares parameters using object equality.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
@@ -13,38 +13,50 @@
  */
 package org.jdbi.v3.core.mapper.reflect;
 
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Parameter;
 import java.lang.reflect.Type;
 import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.WeakHashMap;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 
 import org.jdbi.v3.core.internal.exceptions.Unchecked;
 
+import static java.util.Collections.synchronizedMap;
 import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toUnmodifiableList;
 
 class ConstructorInstanceFactory<T> extends InstanceFactory<T> {
-    private final Constructor<T> constructor;
+    private static final MethodHandles.Lookup LOOKUP = MethodHandles.lookup();
+    private static final Map<Constructor, ConstructorHandleAndTypes> CONSTRUCTOR_CACHE = synchronizedMap(new WeakHashMap<>());
 
-    private final Supplier<Type[]> typeSupplier;
+    private final Constructor<T> constructor;
+    private final List<Type> types;
+    private final MethodHandle constructorHandle;
 
     ConstructorInstanceFactory(Constructor<T> constructor) {
         super(constructor);
         this.constructor = requireNonNull(constructor, "constructor is null");
-        this.typeSupplier = getTypeSupplier(constructor, super::getTypes);
+        ConstructorHandleAndTypes constructorHandleAndTypes = getConstructorHandleAndTypes(constructor, super::getTypes);
+        this.types = constructorHandleAndTypes.getTypes();
+        this.constructorHandle = constructorHandleAndTypes.getConstructorHandle();
     }
 
     @Override
-    Type[] getTypes() {
-        return typeSupplier.get();
+    List<Type> getTypes() {
+        return types;
     }
 
     @Override
     T newInstance(Object... params) {
-        return Unchecked.<Object[], T>function(constructor::newInstance).apply(params);
+        return (T) Unchecked.<Object[], Object>function(constructorHandle::invokeWithArguments).apply(params);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
```diff
@@ -75,6 +75,9 @@ public final class ConstructorMapper<T> implements RowMapper<T> {
             + "that your result set has the columns expected, annotate the "
             + "parameter names explicitly with @ColumnName, or annotate nullable parameters as @Nullable";
 
+
+    static final Map<Constructor, Type[]> CONSTRUCTOR_CACHE = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
elonazoulay: I'll try that out, thanks!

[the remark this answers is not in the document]
elonazoulay: updated. Originally I was going to use a guava cache but I see that guava is only imported in test scope. lmk if you think it's a good idea to add a size limit, cache stats, ability to clear cache, disable cache (maybe for benchmarking), etc.

hgschmie:
Looks good. Lmk when you move it from draft to ready.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
```diff
@@ -27,19 +29,18 @@
 import static java.util.Objects.requireNonNull;
 
 class ConstructorInstanceFactory<T> extends InstanceFactory<T> {
-    private final Constructor<T> constructor;
+    private static final Map<Constructor, Type[]> CONSTRUCTOR_CACHE = new ConcurrentHashMap<>();
```
stevenschlansker, line 32: This will prevent the JVM from unloading any class that the `ConstructorMapper` ever considers, since it keeps a static strong reference to the `Constructor`.

Would the change work just as well with a `synchronizedMap(new WeakHashMap<>())` or something like that? Then, the JVM would be able to unload classes since we do not strongly retain their Constructors.

stevenschlansker (commented):
One note about inadvertent retention of classes. Otherwise LGTM

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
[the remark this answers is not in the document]
stevenschlansker: We do not depend on Guava from Jdbi core, since it is a huge dep to push onto our users. Additionally, there are numerous known cache bugs that seemingly will never get fixed. The Guava `CacheBuilder` javadocs tell you to use Caffeine instead to avoid such bugs.

I think for a simple use case like this, a synchronized weak hash map is good enough. A map that caches JVM entities like constructors and classes will not grow without bound, so there is no need to size it.

stevenschlansker:
This is a broader change than the scope that you're trying to accomplish here, but I think the performance would be even better if we construct and cache a `MethodHandle` that invokes the constructor or factory method, and then cache that.

stevenschlansker:
One last thought: it would be nice if we could see the improvement in a benchmark in `benchmarks/`. I think the improvement is clear enough from reading that this is not strictly required, but I think it's a good habit for us to get into to prove the benefit of such changes through repeatable benchmarks.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
elonazoulay: sounds good, will try all of these ideas out and update, thanks!

elonazoulay:
Updated, will run the benchmarks on the commit before the record constructor workaround, before the this commit and then after this commit and post the results to the pr. Thanks for all the ideas!

elonazoulay:
Just working through the ci failures, will update.

elonazoulay:
I ran the constructor benchmarks to compare pre-constructor change, pre-caching change and this pr. They all performed pretty similarly, lmk :
Pre-constructor change from:
```
Benchmark                                  Mode  Cnt   Score   Error  Units
BeanBindingBenchmark.batchJdbc            thrpt    5  15.336 ± 0.504  ops/s
BeanBindingBenchmark.batchJdbiBean        thrpt    5  13.993 ± 0.484  ops/s
BeanBindingBenchmark.batchJdbiMap         thrpt    5  11.432 ± 0.258  ops/s
BeanBindingBenchmark.batchJdbiNamed       thrpt    5  11.986 ± 0.217  ops/s
BeanBindingBenchmark.batchJdbiPositional  thrpt    5  11.194 ± 0.448  ops/s
BeanBindingBenchmark.oneJdbc              thrpt    5  46.277 ± 0.694  ops/s
BeanBindingBenchmark.oneJdbi              thrpt    5  22.616 ± 0.759  ops/s
```

After adding the constructor pr:
```
Benchmark                                  Mode  Cnt   Score   Error  Units
BeanBindingBenchmark.batchJdbc            thrpt    5  16.315 ± 0.404  ops/s
BeanBindingBenchmark.batchJdbiBean        thrpt    5  14.184 ± 0.508  ops/s
BeanBindingBenchmark.batchJdbiMap         thrpt    5  11.130 ± 0.325  ops/s
BeanBindingBenchmark.batchJdbiNamed       thrpt    5  12.300 ± 0.374  ops/s
BeanBindingBenchmark.batchJdbiPositional  thrpt    5  11.194 ± 1.009  ops/s
BeanBindingBenchmark.oneJdbc              thrpt    5  46.310 ± 6.561  ops/s
BeanBindingBenchmark.oneJdbi              thrpt    5  22.879 ± 0.744  ops/s
```

This pr:
```
Benchmark                                  Mode  Cnt   Score   Error  Units
BeanBindingBenchmark.batchJdbc            thrpt    5  16.280 ± 0.507  ops/s
BeanBindingBenchmark.batchJdbiBean        thrpt    5  13.609 ± 0.405  ops/s
BeanBindingBenchmark.batchJdbiMap         thrpt    5  11.018 ± 0.477  ops/s
BeanBindingBenchmark.batchJdbiNamed       thrpt    5  11.698 ± 0.290  ops/s
BeanBindingBenchmark.batchJdbiPositional  thrpt    5  11.020 ± 0.295  ops/s
BeanBindingBenchmark.oneJdbc              thrpt    5  46.451 ± 7.482  ops/s
BeanBindingBenchmark.oneJdbi              thrpt    5  23.296 ± 0.781  ops/s
```

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
```diff
@@ -13,38 +13,50 @@
...
 
 import org.jdbi.v3.core.internal.exceptions.Unchecked;
 
+import static java.util.Collections.synchronizedMap;
 import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toUnmodifiableList;
 
 class ConstructorInstanceFactory<T> extends InstanceFactory<T> {
-    private final Constructor<T> constructor;
+    private static final MethodHandles.Lookup LOOKUP = MethodHandles.lookup();
```
elonazoulay, line 37: lmk if this shouldn't be cached.

elonazoulay: Let me know if you want the weak hashmap added back vs caching in the constructor instance factory (saved the previous version, so I can quickly update).

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
```diff
@@ -207,14 +207,14 @@ private Optional<RowMapper<T>> createSpecializedRowMapper(StatementContext ctx,
                                                List<String> unmatchedColumns) {
         final int count = factory.getParameterCount();
         final Parameter[] parameters = factory.getParameters();
-        final Type[] types = factory.getTypes();
+        final List<Type> types = factory.getTypes();
```
elonazoulay, line 210: Had to convert to list due to error about exposing an array. Thinking to change the api to store the array internally and change the signature to `getType(index)`, wdyt?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/InstanceFactory.java
```diff
@@ -44,10 +46,10 @@ Parameter[] getParameters() {
         return executable.getParameters();
     }
 
-    Type[] getTypes() {
+    List<Type> getTypes() {
```
elonazoulay, line 49: same here - should it be `Type getType(int index)`?

elonazoulay:
Update: the pg benchmark failed for one of the cases (on master commit), other than that there was pretty similar performance. Will try again.

elonazoulay:
lmk if you prefer this version or the weak hashmap version (static cache w synchronized weak hashmap).

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
stevenschlansker: That sounds ok

core/src/main/java/org/jdbi/v3/core/mapper/reflect/InstanceFactory.java
stevenschlansker: OK by me

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorInstanceFactory.java
stevenschlansker: So I think there is a bit of a misunderstanding here. I didn't mean to cache the lookup factory instance itself, but the `constructorHandle` instance itself.

Something like `SynchronizedMap<WeakHashMap<ConstructorHandleAndTypes, MethodHandle>>>` (referring to the key type from the other PR)

Sorry I wasn't clear, and sorry for taking so long to respond! Been a busy couple weeks at work.

stevenschlansker:
Merging #2657 . Thanks!

PR #2657, opened 2024-03-13
Cache types weakhm

elonazoulay:
This is the weak hashmap version of #2650 just for viz. The benchmarks are posted in that pr. 
lmk which one you prefer. I can push this back to the original pr branch. Happy to do anything, code changes, more benchmarks, etc. - including just waiting for the jdk fix:)


elonazoulay:
Here's the jdk pr that fixes the issue: https://github.com/openjdk/jdk/pull/17070

stevenschlansker (approved):
Thanks! This is looking good

PR #2675, opened 2024-06-03
Raise exception if a positional param is missing

### LINKED ISSUE #2674: Accidental one-based arguments fail silently
As a beginner to JDBI, I mistakenly assumed arguments are one-based. It seems that JDBI's current behaviour in this situation is to make the first argument implicitly a NULL, so this mistake confused me for a while as to why it wasn't finding the row:

```kotlin
val row = handle.createQuery("SELECT * FROM foo WHERE id = ?")
  .bind(1, fooID)
  .mapToMap()
  .one()
```

Would you be open to a PR to throw an explicit error in this situation? I'd be happy to make the PR myself; looks like `ArgumentBinder.bindPositional` is where it wants to go, just after it has checked the arg count is correct

angusholder:
Fixes #2674

I made it match the wording of the existing `missingNamedParameter` exception: https://github.com/jdbi/jdbi/blob/18958cb2ca00def74f1bedc29363a66ca1db98b5/core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java#L170-L172

The only concern I have is that this could be a breaking change for any existing code relying on the current implicit-null behaviour. However, I can't imagine why anyone would intentionally have a SQL param that's always NULL, and given that named params already error if any are missing, I think this change is justifiable. Any code it breaks is probably buggy?

I moved the param count check after this check because missing param seems a more useful error message, so we check it first.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/ArgumentBinder.java
@@ -71,11 +71,10 @@ class ArgumentBinder {
     }
 
     void bindPositional(Binding binding) {
-        boolean moreArgumentsProvidedThanDeclared = binding.positionals.size() != params.getParameterCount();
-        if (moreArgumentsProvidedThanDeclared && !ctx.getConfig(SqlStatements.class).isUnusedBindingAllowed()) {
-            throw new UnableToCreateStatementException("Superfluous positional param at (0 based) position " + params.getParameterCount(), ctx);
-        }
         for (int index = 0; index < params.getParameterCount(); index++) {
+            if (!binding.positionals.containsKey(index)) {
+                throw new UnableToCreateStatementException(format("Missing positional parameter %d in binding:%s", index, binding), ctx);
+            }
             QualifiedType<?> type = typeOf(binding.positionals.get(index));
             try {
                 argumentFactoryForType(type)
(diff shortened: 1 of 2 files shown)


stevenschlansker:
Yeah, I have the same concern. We can always make the behavior configurable if anyone finds it breaks their code badly.

stevenschlansker:
Thanks @angusholder  for the contribution. I think the tests are failing. Maybe you updated the message but not the tests?

angusholder:
> Thanks @angusholder for the contribution. I think the tests are failing. Maybe you updated the message but not the tests?

You're right, I've fixed it now, thanks!

My tests are now failing because JdbiLeakChecker says they've leaked a cleanable, but I'm not sure what's wrong - I'm using try-with-resources on the Handle - can you help me?

```
[ERROR] org.jdbi.v3.core.statement.TestArgumentBinder.testOneBasedIndexingThrowsError -- Time elapsed: 0.274 s <<< FAILURE!
java.lang.AssertionError: Found 1 leaked cleanable objects in 2 contexts
        at org.jdbi.v3.core.junit5.JdbiLeakChecker.checkForLeaks(JdbiLeakChecker.java:160)
        at org.jdbi.v3.core.junit5.PgDatabaseExtension.afterEach(PgDatabaseExtension.java:144)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)

[ERROR] org.jdbi.v3.core.statement.TestArgumentBinder.testMissingParamThrowsError -- Time elapsed: 0.227 s <<< FAILURE!
java.lang.AssertionError: Found 1 leaked cleanable objects in 2 contexts
        at org.jdbi.v3.core.junit5.JdbiLeakChecker.checkForLeaks(JdbiLeakChecker.java:160)
        at org.jdbi.v3.core.junit5.PgDatabaseExtension.afterEach(PgDatabaseExtension.java:144)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

angusholder:
@stevenschlansker I switched those tests from try-with-resources to Jdbi.useHandle, and it fixed the resource leak. Can you approve to re-run the CI?

stevenschlansker (approved):
Thank you @angusholder !

PR #2677, opened 2024-06-23
Replace JdbiOptionals.stream() with Optional.stream()

angusholder:
Optional.stream() requires Java 9+, but since 3.40 JDBI has required Java 11+, so this compatibility code is no longer needed


stevenschlansker (approved):
Thanks @angusholder !

PR #2685, opened 2024-07-24
Fix handling multiple commits on same handle when autocommit disabled

mhib:
The issue was introduced with 3.41.1 while fixing https://github.com/jdbi/jdbi/issues/2478

Always recalculating `handlerState` like in the `BoundLocalTransactionHandler` constructor, seems to fix it.

### DIFF
--- a/core/src/test/java/org/jdbi/v3/core/TestMultipleCommitsWithoutAutocommit.java
+++ b/core/src/test/java/org/jdbi/v3/core/TestMultipleCommitsWithoutAutocommit.java
@@ -1,0 +1,41 @@
+package org.jdbi.v3.core;
+
+import org.assertj.core.api.Assertions;
+import org.jdbi.v3.core.junit5.H2DatabaseExtension;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import java.sql.Connection;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/jdbi/v3/core/TestMultipleCommitsWithoutAutocommit.java
```diff
@@ -0,0 +1,41 @@
...
+public class TestMultipleCommitsWithoutAutocommit {
+    @RegisterExtension
+    public H2DatabaseExtension h2Extension = H2DatabaseExtension.instance();
+    @BeforeEach
+    public void startUp() {
+        Assertions.setMaxStackTraceElementsDisplayed(100);
+    }
+
+    @Test
+    public void testAfterCommitThrowsRollback() throws Exception {
```
stevenschlansker, line 23: Is this test method name reflecting what is going on here? I don't see any throws or rollback happening.

stevenschlansker (approved):
Thanks for this contribution @mhib . One comment on the test name which doesn't seem to reflect what is being tested here.

core/src/test/java/org/jdbi/v3/core/TestMultipleCommitsWithoutAutocommit.java
mhib: Fixed

Issue #2518, opened 2023-10-24
Java 21, Virtual Threads and ThreadLocal


hgschmie:
Jdbi at this point has one relevant ThreadLocal instance, which is a central one:

We maintain a reference to the "current handle supplier" in Jdbi#threadHandleSupplier. This enables the ability to bind extension objects in `Jdbi#withExtension` and Handles in `Jdbi#withHandle` and transparently keep using the same handle even in recursive calls as long as these operations are done on the same thread. 

We do not use the "threadlocal as cache" scenario as described in JEP 429/446, so it is not clear whether there is real pressure to remove the thread local. Even with thousands of virtual threads, the number of concurrent instances will be limited by e.g. the number of concurrent database connections supported, which is a pretty limited resource. A ThreadLocal with a few hundred instance s that all get removed don't seem to be a big deal. The code cleans out the ThreadLocal instances by calling `#remove()` so the ThreadLocal itself should not grow unbounded.

JEP 446 offers scoped values as an alternative. However, as of Java 21 LTS, scoped values have not been shipped, so we would need to build a multi-version jar to support it. 

There actually is a second ThreadLocal, in the StringTemplate loader/cache mechanism; this should be easily replaced with a cache using the core cache facility.

stevenschlansker:
ThreadLocal is supposed to be fully supported by virtual threads, so I agree that since we limit it by number of DB connections, it should be fine. Scoped values sound cool but I think it's too early to consider them.

stevenschlansker:
We hope that it just works, but I haven't personally tested it yet. If you do run it and find any problems please let us know and they should get fixed.

stevenschlansker:
Thanks for giving this a try - unfortunately, the stack trace provided isn't enough to diagnose the problem. Any chance you could post your sample project?
What are the other connections that this thread failed to acquire doing at the time - are they blocked? A full thread stack dump might help.

stevenschlansker:
And it is expected that right now Jdbi would pin carrier threads potentially - we do use `synchronized`.
If there are a few hotspots we can fix individually, we would consider it.

If the project turns into rewriting all `synchronized` blocks to use `Lock`, we might instead wait for Project Loom to fix synchronized to work properly with threads.

stevenschlansker:
Hi @vszakd , I downloaded and imported your project, but I cannot figure out how to run it. None of the classes have a `main` method. Which class am I supposed to run?

stevenschlansker:
I created:
[...]

PR #2597, opened 2024-01-10
Introduce java21 artifact with initial test showing Virtual Threads at least basically work

stevenschlansker:

PR #2686, opened 2024-07-24
Use Lock instead of synchronized block in MemoizingSupplier to prevent pinning

mhib:
The issue was caused by virtual threads waiting for the connection, as JVM was unable to unpin them while in `synchronized` block.

I've read this discussion: https://github.com/jdbi/jdbi/issues/2518#issuecomment-1875999057

But I believe `MemoizingSupplier` qualifies as a hotspot.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/internal/MemoizingSupplier.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/MemoizingSupplier.java
@@ -13,6 +13,7 @@
  */
 package org.jdbi.v3.core.internal;
 
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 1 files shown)


stevenschlansker:
Hi @mhib , thanks for proposing this change.
The latest Loom EA builds promise:

> This build improves the implementation of Java monitors (synchronized methods) to work better with virtual threads.

which I hope would remove the need to change `synchronized` blocks out for locks.

Is this potential pin causing a specific problem for you that is solved with this change? I would expect the pin to be of a very short duration that would be unlikely to harm throughput, since it is only a field read, except for the first initialization.

Otherwise, we're hoping that Loom will advance to the point where we do not have to change our code to consider its implementation limitations.

mhib:
@stevenschlansker It fixes the deadlock from https://github.com/jdbi/jdbi/pull/2597. I use jdbi with virtual threads quite a lot, but without using Jdbi plugins, so I've never encountered this scenario.

stevenschlansker:
OK, good enough. Let's give it a spin. Thanks for looking into this.

stevenschlansker (approved):

mhib:
@stevenschlansker Note that the taget branch of this pull request was set to the branch from this: https://github.com/jdbi/jdbi/pull/2597, not the `master` branch. I'm sorry If I should have underlined it more.

stevenschlansker:
Oops, sorry I totally missed that. Will look again this week.

PR #2706, opened 2024-10-04
mssql 2017-latest

stevenschlansker:
maybe mssql will be more stable with a newer patch? :shrug:


hgschmie (approved):

hgschmie:
#2708 does similar things and the build does pass over there.

PR #2707, opened 2024-10-07
sqlite: use named memory database so contents are shared

### LINKED ISSUE #2688: JdbiSqliteExtension does not work with SQLite memory databases
`JdbiSqliteExtension` uses a JDBC URL of `jdbc:sqlite::memory:`.

From the SQLite memory database documentation, https://www.sqlite.org/inmemorydb.html

> The database ceases to exist as soon as the database connection is closed. Every :memory: database is distinct from every other. So, opening two database connections each with the filename ":memory:" will create two independent in-memory databases.

This means that every connection via `JdbiSqliteExtension` gets a new database, which will not include any data from previous queries, including `JdbiExtensionInitializer` efforts.

Multiple connections can use the same memory database if they use shared memory, by specifying `cache=shared` on the JDBC URL. Further, multiple shared memory databases can be used if the URL names a file and indicates `mode=memory`, eg:

`jdbc:sqlite:file:memory?mode=memory&cache=shared`

Please see the "In-memory Databases And Shared Cache" section of the linked documentation.

I do not believe that `JdbiSqliteExtension` can work as intended in its current form.

I have been able to work around this locally by extending the extension.

```
public class MyJdbiSqliteExtension extends JdbiSqliteExtension {
    @Override
    public String getUrl() {
       return "jdbc:sqlite:file:example?mode=memory&cache=shared";
    }
}

@RegisterExtension
public static JdbiExtension extension = new MyJdbiSqliteExtension()
            .withPlugin(new SQLitePlugin())
            .withInitializer((ds, handle) -> { ... });
```

Probably unrelated, but for some reason I can't determine, I'm not able to use `@RegisterExtension` and `JdbiExtension` on an instance variable, only as a class/static variable.

stevenschlansker:
fixes #2688

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - geantyref 2.0.0
+- sqlite: fix sqlite extension memory database usage to share between different connection (#2688)
 
 # 3.45.4
 
--- a/testing/src/main/java/org/jdbi/v3/testing/junit5/JdbiSqliteExtension.java
+++ b/testing/src/main/java/org/jdbi/v3/testing/junit5/JdbiSqliteExtension.java
@@ -13,6 +13,8 @@
  */
 package org.jdbi.v3.testing.junit5;
 
+import java.util.UUID;
+
 import javax.sql.DataSource;
 
 import org.sqlite.SQLiteDataSource;
@@ -51,25 +53,31 @@ import org.sqlite.SQLiteDataSource;
  */
 public class JdbiSqliteExtension extends JdbiExtension {
 
-    private static final String URL = "jdbc:sqlite::memory:";
+    private static final String JDBC_URL_TEMPLATE = "jdbc:sqlite:file:%s?mode=memory&cache=shared";
+
+    private final String url;
 
     static JdbiExtension instance() {
         return new JdbiSqliteExtension();
     }
 
-    public JdbiSqliteExtension() {}
+    public JdbiSqliteExtension() {
+        this(String.format(JDBC_URL_TEMPLATE, UUID.randomUUID().toString()));
+    }
+
+    public JdbiSqliteExtension(String url) {
+        this.url = url;
+    }
 
     @Override
     public String getUrl() {
-        return URL;
+        return url;
     }
 
     @Override
     protected DataSource createDataSource() {
         SQLiteDataSource ds = new SQLiteDataSource();
         ds.setUrl(getUrl());
-
         return ds;
     }
-
 }
--- a/testing/src/test/java/org/jdbi/v3/testing/junit5/JdbiSqliteExtendWithTest.java
+++ b/testing/src/test/java/org/jdbi/v3/testing/junit5/JdbiSqliteExtendWithTest.java
@@ -19,6 +19,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatNoException;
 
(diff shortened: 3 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - geantyref 2.0.0
+- sqlite: fix sqlite extension memory database usage to share between different connectsion (#2688)
```
hgschmie, line 4: typo

testing/src/main/java/org/jdbi/v3/testing/junit5/JdbiSqliteExtension.java
```diff
@@ -51,7 +53,9 @@
  */
 public class JdbiSqliteExtension extends JdbiExtension {
 
-    private static final String URL = "jdbc:sqlite::memory:";
+    private static final String URL_FMT = "jdbc:sqlite:file:%s?mode=memory&cache=shared";
```
hgschmie, line 56: nt use n acrnym? (e.g. `JDBC_TEMPLATE`)?

testing/src/test/java/org/jdbi/v3/testing/junit5/JdbiSqliteExtendWithTest.java
```diff
@@ -36,4 +36,14 @@ public void testHandleIsAlive(Handle handle) {
 
         assertThat(one).isOne();
     }
+
+    @Test
+    public void dataSourceSharesData(Jdbi jdbi) {
+        jdbi.useHandle(h -> {
```
hgschmie, line 42: what is this testing? I am not sure what that change does and this test has no assertions.

hgschmie (changes requested):
I am happy to approve this if I can understand what it does. :-) There seem to be at least some assertions missing.

testing/src/main/java/org/jdbi/v3/testing/junit5/JdbiSqliteExtension.java
```diff
@@ -61,7 +65,7 @@ public JdbiSqliteExtension() {}
 
     @Override
     public String getUrl() {
-        return URL;
+        return String.format(URL_FMT, id);
```
hgschmie, line 68: I think a better fix for this would be e.g. what we do with the `GenericJdbiExtension` and add a `withUrl()` method or a c'tor that takes an URL. So that you can do 

```java
@RegisterExtension
    public JdbiExtension sqlLite = new JdbiSqliteExtension("... specific URL ...");
```

or
```java
@RegisterExtension
    public JdbiExtension sqlLite = new JdbiSqliteExtension()
        .withUrl("... specific URL ...");
```

See e.g. JdbiGenericExtensionTest.

hgschmie (changes requested):

testing/src/test/java/org/jdbi/v3/testing/junit5/JdbiSqliteExtendWithTest.java
stevenschlansker: The test currently (before this change) fails - as the original reporter posts, every time you `useHandle` right now you get a *different* database, which is nearly useless.

stevenschlansker: I added an `assertNoExceptionThrownBy` to make this more explicit

stevenschlansker:
Did you read the reporting issue? The problem is that a new database is opened each time you get a handle, so it is not possible to actually use the extension in any meaningful way.

stevenschlansker:
I added an `assertNoExceptionThrownBy` to make the test more explicit.

hgschmie (approved):

PR #2708, opened 2024-10-11
Move to most current MSSQL Server

hgschmie:
Also test the 1.20 test container version in the slow tests

### DIFF
--- a/.github/workflows/slow-tests.yml
+++ b/.github/workflows/slow-tests.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        testcontainer-version: [ 1.19.4 ]
+        testcontainer-version: [ 1.19.8, 1.20.2 ]
         mysql-version: [ "mysql:latest" ]
         # 1.18.3 does not work with mysql:latest (8.3.x).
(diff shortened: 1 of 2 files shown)


.github/workflows/slow-tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        testcontainer-version: [ 1.19.4 ]
+        testcontainer-version: [ 1.19.8, 1.20.2 ]
```
stevenschlansker, line 19: Why do we keep testing against 1.19? As far as I can see there is no support for it anymore, if there is a bug the only recourse is to move to 1.20.x anyway

stevenschlansker (approved):

.github/workflows/slow-tests.yml
hgschmie: I tend to keep a few "older latest versions" around for testing. It is a good indicator if we were to silently introduce newer API use that would break older builds when people upgrade without upgrading deps. This caught a few issues with flyway. 

Obviously we don't *have* to. :-)

PR #2709, opened 2024-10-11
Fix x86_64 only EnableOnOs annotations

hgschmie:
Ensure that all "x86 only" tests use both x86_64 and amd64 as selectors.


stevenschlansker:
I am surprised it is necessary to specify both `x86_64` and `amd64`. As far as I understand, these are different names for the same thing.

I also don't quite understand how this ends up fixing the test, since it seems that this would only add additional test runs, not remove or change any existing ones?

stevenschlansker (approved):

PR #2713, opened 2024-10-12
Update github actions

hgschmie:
- always fetch depth 0 and tags
- set ids for setup-java, test runs
- enable enforcer for install-fast


hgschmie:
this stacks on top of #2714 and #2715 and will turn master green.

stevenschlansker (approved):

PR #2727, opened 2024-10-22
Fix native Spring compile with `@JdbiRepository`

### LINKED ISSUE #2694: "@JdbiRepository" annotated repositories are not supported in AOT mode.
I have a repository annotated with @JdbiRepository. All works fine in non-AOT mode.
When I am trying to build application for native mode with maven I am having error:

```
Exception in thread "main" java.lang.IllegalStateException: Error processing bean with name 'com.mycompany.MyRepository': instance supplier is not supported
        at org.springframework.beans.factory.aot.DefaultBeanRegistrationCodeFragments.getTarget(DefaultBeanRegistrationCodeFragments.java:83)
        at org.springframework.beans.factory.aot.BeanDefinitionMethodGenerator.generateBeanDefinitionMethod(BeanDefinitionMethodGenerator.java:85)
        at org.springframework.beans.factory.aot.BeanRegistrationsAotContribution.lambda$generateRegisterBeanDefinitionsMethod$2(BeanRegistrationsAotContribution.java:90)
        at java.base/java.util.LinkedHashMap.forEach(LinkedHashMap.java:986)
        at org.springframework.beans.factory.aot.BeanRegistrationsAotContribution.generateRegisterBeanDefinitionsMethod(BeanRegistrationsAotContribution.java:88)
        at org.springframework.beans.factory.aot.BeanRegistrationsAotContribution.lambda$applyTo$1(BeanRegistrationsAotContribution.java:73)
        at org.springframework.aot.generate.GeneratedMethod.<init>(GeneratedMethod.java:54)
        at org.springframework.aot.generate.GeneratedMethods.add(GeneratedMethods.java:112)
        at org.springframework.aot.generate.GeneratedMethods.add(GeneratedMethods.java:89)
        at org.springframework.beans.factory.aot.BeanRegistrationsAotContribution.applyTo(BeanRegistrationsAotContribution.java:72)
        at org.springframework.context.aot.BeanFactoryInitializationAotContributions.applyTo(BeanFactoryInitializationAotContributions.java:78)
        at org.springframework.context.aot.ApplicationContextAotGenerator.lambda$processAheadOfTime$0(ApplicationContextAotGenerator.java:58)
        at org.springframework.context.aot.ApplicationContextAotGenerator.withCglibClassHandler(ApplicationContextAotGenerator.java:67)
        at org.springframework.context.aot.ApplicationContextAotGenerator.processAheadOfTime(ApplicationContextAotGenerator.java:53)
        at org.springframework.context.aot.ContextAotProcessor.performAotProcessing(ContextAotProcessor.java:106)
        at org.springframework.context.aot.ContextAotProcessor.doProcess(ContextAotProcessor.java:84)
        at org.springframework.context.aot.ContextAotProcessor.doProcess(ContextAotProcessor.java:49)
        at org.springframework.context.aot.AbstractAotProcessor.process(AbstractAotProcessor.java:82)
        at org.springframework.boot.SpringApplicationAotProcessor.main(SpringApplicationAotProcessor.java:80)
```

hpoettker:
Resolves #2694.

For every SQL object annotated with `JdbiRepository`, the `JdbiRepositoryRegistrar` currently instantiates a `JdbiRepositoryFactoryBean` itself and then passes a reference to its `getObject` method to the bean definition registry to be called at a later time by the application context.

This works fine in the standard JVM but not with AOT and GraalVM where building the native image fails if the feature is used. The problem is that in AOT mode Spring generates the bean definitions at compile time to write corresponding Java code that GraalVM can then compile. But when it tries to do so for the bean definitions generated by `JdbiRepositoryRegistrar`, it can't as it cannot determine how the passed instance of the `JdbiRepositoryFactoryBean` has been built. Compilation then stops with the error message `instance supplier is not supported`.

The PR fixes this by letting `JdbiRepositoryRegistrar` register bean definitions for the `JdbiRepositoryFactoryBean`s instead that Spring AOT knows how to generate explicit code for.

With this fix, a native image can be compiled (unless of course there are other issues in the application), but it does not fix the issue that the native image will fail to run unless native hints are provided for some of the JDBI classes during compilation.

### DIFF
--- a/spring5/src/main/java/org/jdbi/v3/spring5/JdbiRepositoryFactoryBean.java
+++ b/spring5/src/main/java/org/jdbi/v3/spring5/JdbiRepositoryFactoryBean.java
@@ -52,6 +52,7 @@ public class JdbiRepositoryFactoryBean implements FactoryBean<Object>, Applicati
     /**
      * The object type of the repository.
      */
+    @SuppressWarnings("unused")
     public void setObjectType(Class<?> objectType) {
         this.objectType = objectType;
     }
@@ -61,6 +62,7 @@ public class JdbiRepositoryFactoryBean implements FactoryBean<Object>, Applicati
      * @param jdbiQualifier The name of the jdbi bean to bind the repository to.
      *                      if <code>null</code> then no name will be specified during resolution.
      */
+    @SuppressWarnings("unused")
     public void setJdbiQualifier(@Nullable String jdbiQualifier) {
         this.jdbiQualifier = jdbiQualifier;
     }
--- a/spring5/src/main/java/org/jdbi/v3/spring5/JdbiRepositoryRegistrar.java
+++ b/spring5/src/main/java/org/jdbi/v3/spring5/JdbiRepositoryRegistrar.java
@@ -17,12 +17,11 @@ import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.Set;
 
-import org.springframework.beans.factory.BeanFactory;
 import org.springframework.beans.factory.annotation.AnnotatedBeanDefinition;
 import org.springframework.beans.factory.annotation.AnnotatedGenericBeanDefinition;
 import org.springframework.beans.factory.config.BeanDefinition;
-import org.springframework.beans.factory.support.BeanDefinitionBuilder;
 import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.beans.factory.support.RootBeanDefinition;
 import org.springframework.context.annotation.ClassPathScanningCandidateComponentProvider;
 import org.springframework.context.annotation.ImportBeanDefinitionRegistrar;
 import org.springframework.core.type.AnnotationMetadata;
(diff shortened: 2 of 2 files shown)


stevenschlansker (approved):
Thanks for this fix!

hpoettker:
Thanks for the quick response for adding the change to the release notes.

Regarding the other unreleased change. 😄 
Spring 5.x and Spring 6.0 have fallen out of Spring's open-source support by now. And they have recently introduced a policy change that any fixes for unsupported versions are only released for customers with commercial support. So 5.3.40 and 5.3.41 won't be published on Maven Central.

I think you should feel free to ignore any nagging CVE checkers about this. The scenario that people pull in Spring transitively through `jdbi3-spring5` seems a bit far-fetched to me.
The alternative would be build against 6.1.x, which should also be fine as changes to the module are rare.

hgschmie:
wow, spring is weird.

stevenschlansker:
Gotta monetize somehow! 😎

stevenschlansker:
fixed in 3.47.0

PR #2733, opened 2024-11-04
add support for @Nested on optional fields

WellingR:
The `@Nested` annotation previously failed to map any `@Nested` fields/properties when the type was wrapped in a java `Optional`.

I implemented a change to include support for `@Nested Optional<NestedBean>` in all three types of reflective mappers we have.

In addition, I included a change to automatically map to Optional.empty in case all fields in the `@Nested` object are null in the result set.


core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java
```diff
@@ -89,8 +89,13 @@ protected PojoProperties<T> getProperties(ConfigRegistry config) {
         return (PojoProperties<T>) BeanPropertiesFactory.propertiesFor(type, config);
     }
 
-    @Override
+    @Deprecated // preserved for binary compatibility
```
WellingR, line 92: This gives me the error
```
WARNING: Incompatibility detected: Requires semantic version level	 MINOR
```
Should I bump the version as part of this MR? Since this is a feature and not a bugfix I would expect the minor version to be bumped anyways

hgschmie: good question. That is a japi warning; does the build still succeed? We might sort that out at release time.

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
```diff
@@ -44,10 +47,7 @@
 import static java.lang.String.format;
 
 import static org.jdbi.v3.core.mapper.reflect.JdbiConstructors.findFactoryFor;
-import static org.jdbi.v3.core.mapper.reflect.ReflectionMapperUtil.addPropertyNamePrefix;
-import static org.jdbi.v3.core.mapper.reflect.ReflectionMapperUtil.anyColumnsStartWithPrefix;
-import static org.jdbi.v3.core.mapper.reflect.ReflectionMapperUtil.findColumnIndex;
-import static org.jdbi.v3.core.mapper.reflect.ReflectionMapperUtil.getColumnNames;
+import static org.jdbi.v3.core.mapper.reflect.ReflectionMapperUtil.*;
```
hgschmie, line 50: please fix that. :-) We don't do wildcard imports. There are code style file for the most common IDEs in the ide-support folder (or at https://github.com/jdbi/jdbi/tree/master/ide-support)

hgschmie:
First, thank you for the PR. I will need to look through this a bit more, there are some changes in there that may need a bit of more intense reviewing (the innards of jdbi are complex and some interactions are non-obvious). 

If you find the time to reformat the code according to our code style, this would help but is not necessary for us to consider/accept the patch. I will reformat it accordingly. :-)

hgschmie:
(also we don't do conventional commit messages, so I will rewrite the subject :-) )

hgschmie:
So I looked at the code a bit:

The general idea seems to be that if you wrap a nested bean into an optional and that bean is absent, then insert an `Optional.empty()` instead of a null. That piece is fine.

But then there is a test that contains

```java
public static class OptionalBean {
        @JdbiConstructor
        public OptionalBean(OptionalInt intValue, Optional<String> name) { ... }
}

public static class OptionalBeanWithNestedOptionals {
        @JdbiConstructor
        public OptionalBeanWithNestedOptionals(@Nested Optional<OptionalBean> bean) { ... }
}

and the test assumes that if the SQL query returns 'null', 'null' for both intValue and name, then this will return an OptionalBeanWithNestedOptionals where the `bean` value is an Optional.empty().

This is surprising to me. Because I would have expected this to be a `Optional.of(OptionalBean)` where the `intValue` is an `OptionalInt.empty()` and `name` is an `Optional.empty()`.

Jdbi does not apply any magic. This seems to be magic for me. 

@stevenschlansker  options?

**Comment by WellingR** [comment]
> the test assumes that if the SQL query returns 'null', 'null' for both intValue and name, then this will return an OptionalBeanWithNestedOptionals where the `bean` value is an Optional.empty().
>
> This is surprising to me. Because I would have expected this to be a `Optional.of(OptionalBean)` where the `intValue` is an `OptionalInt.empty()` and `name` is an `Optional.empty()`.

I specifically added a bit of documentation on this. If we want to avoid this magic then I think there may be another option:

* I could change the default behaviour to just return `Optional.of(...)` where the fields are empty optionals.
* I could look into a new use case for the `@PropagateNull` annotation (or perhaps a new annotation, or argument of the existing annotation) where we could place it on the class level of the nested bean or on the type parameter of the optional. To enable the behaviour where we automatically propagate null if all fields in the result set were null

File: core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java (hunk 1 above)
Comment (WellingR) [reply core/src/main/java/org/jdbi/v3/core/mapper/reflect/BeanMapper.java:92]: `make install` fails and skips a lot of jobs. However `make tests` runs without any issues

File: core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java (hunk 2 above)
Comment (WellingR) [reply core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java:50]: Yes I guess my IDE did this. I will have a look at this later

Comment (WellingR) [reply core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java]: Fixed

**Comment by WellingR** [comment]
> [...] This is surprising to me. Because I would have expected this to be a Optional.of(OptionalBean) where the intValue is an OptionalInt.empty() and name is an Optional.empty().
> 
> Jdbi does not apply any magic. This seems to be magic for me.

I added a commit that removed this magic. On second thought I agree that if we want this functionality, then we should probably enable it though some sort of annotation. I will give this some thought and might create a separate followup MR for that later.

**Comment by WellingR** [comment]
@hgschmie A few months ago I pushed a commit that should have addressed your concern.

Now that https://github.com/jdbi/jdbi/pull/2738 has been merged. I can use that feature to add some code that allows me to work around the problem in a nice way.

I would appreciate another review

**Review by stevenschlansker — APPROVED** [review]
Thank you for your persistence and patience @WellingR .
I took a look over this MR and it seems like it should be good to go to me.
If you have some time, a quick update to the documentation describing how this works would be appreciated.
I'll give @hgschmie a few days to add any objections and, save for that, merge and build a release later this week.

File: core/src/main/java/org/jdbi/v3/core/mapper/reflect/FieldMapper.java [hunk 3]
```diff
@@ -135,10 +137,10 @@ && anyColumnsStartWithPrefix(unmatchedColumns, prefix, columnNameMatchers)) {
         return mapper;
     }
 
-    private Optional<RowMapper<T>> createSpecializedRowMapper(StatementContext ctx,
+    private <R> Optional<RowMapper<R>> createSpecializedRowMapper(StatementContext ctx,
                                                List<String> columnNames,
                                                List<ColumnNameMatcher> columnNameMatchers,
-                                               List<String> unmatchedColumns) {
+                                               List<String> unmatchedColumns, org.jdbi.v3.core.mapper.reflect.internal.RowMapperFieldPostProcessor<T, R> postProcessor) {
```
stevenschlansker, line 143: same here, unnecessarily fully qualified?

core/src/main/java/org/jdbi/v3/core/mapper/reflect/internal/RowMapperFieldPostProcessor.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Optional;
+
+/**
+ * A function that can post-process a row mapper field.
+ * This is specifically used to post process fields annotated with {@link Nested}.
+ *
+ * @param <T> the type of the object being processed
+ * @param <R> the type of the result
+ */
+public interface RowMapperFieldPostProcessor<T, R> {
```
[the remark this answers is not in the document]
WellingR: After looking at this again I no longer think this adds a lot of value. I removed it in https://github.com/jdbi/jdbi/pull/2733/commits/47a4e4a179ec81ea62456750bd5bfbc7b5e5449c

Having the extra class might make things just a bit easier to understand, then again `Function.identity()` and `Optional::ofNullable` are the two behaviours that we have, so it is actually pretty simple.

stevenschlansker:
Ok @WellingR please check out 3.49.0 :)

Issue #2736, opened 2024-11-08
JDBI no longer usable on JREs/JDKs without Java Flight Recorder classes


PascalSchumacher:
After updating jdbi3-core from `3.45.4` to `3.47.0` we are getting:

```
java.lang.NoClassDefFoundError: jdk.jfr.Event
	at java.base/java.lang.ClassLoader.defineClassImpl(Native Method) ~[?:?]
	at java.base/java.lang.ClassLoader.defineClassInternal(ClassLoader.java:472) ~[?:?]
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:433) ~[?:?]
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150) ~[?:?]
	at java.base/java.net.URLClassLoader.defineClass(URLClassLoader.java:657) ~[?:?]
	at java.base/java.net.URLClassLoader$ClassFinder.run(URLClassLoader.java:1061) ~[?:?]
	at java.base/java.net.URLClassLoader$ClassFinder.run(URLClassLoader.java:1046) ~[?:?]
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:784) ~[?:?]
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:572) ~[?:?]
	at java.base/java.lang.ClassLoader.loadClassHelper(ClassLoader.java:1204) ~[?:?]
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:1119) ~[?:?]
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:1102) ~[?:?]
	at org.jdbi.v3.core.statement.SqlStatement.<clinit>(SqlStatement.java:76) ~[jdbi3-core-3.47.0.jar:3.47.0]
	at org.jdbi.v3.core.Handle.createQuery(Handle.java:432) ~[jdbi3-core-3.47.0.jar:3.47.0]
	at org.jdbi.v3.core.Handle.createQuery(Handle.java:444) ~[jdbi3-core-3.47.0.jar:3.47.0]
	at ourorg.OurClass
	at org.jdbi.v3.core.Jdbi.withHandle(Jdbi.java:422) ~[jdbi3-core-3.47.0.jar:3.47.0]
```

I think the problem is that we are using OpenJ9 Java 21 JDK, which does not include the Java Flight Recorder classes.

By the way: Thank you very much for developing JDBI!

[...]

PR #2739, opened 2024-11-19
Add semeru dist for testing

hgschmie:
This is the distribution reported in #2736 as not containing jfr.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/internal/JfrSupport.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/JfrSupport.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/jdbi/v3/core/internal/JfrSupport.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        return Optional.of(new JfrWrapper());
+    }
+
+    public static void attachJfrEvent(Optional<JfrWrapper> optWrapper, StatementContext ctx) {
+        if (!PRESENT) {
+            return;
+        }
+        optWrapper.ifPresent(wrapper -> {
```
stevenschlansker, line 51: Since this is already guarded by `PRESENT`, I would ditch the `Optional` wrapper and just store `null` if the feature is not present.

core/src/main/java/org/jdbi/v3/core/internal/JfrSupport.java
```diff
@@ -0,0 +1,82 @@
...
+                evt.parameters = ctx.getBinding().describe(stmtConfig.getJfrParamMaxLength());
+                ctx.addCleanable(() -> {
+                    evt.rowsMapped = ctx.getMappedRows();
+                    evt.commit();
+                });
+            }
+        });
+    }
+
+    public static final class JfrWrapper {
```
stevenschlansker, line 71: Could this wrapper object be an interface the event implements, to avoid a wrapper allocation?

stevenschlansker (approved):

hgschmie:
Superseded by #2757

PR #2744, opened 2024-12-20
Fix the link to sample code

armandino:
Currently, the "Jdbi sample code" link points to a [README.md](https://github.com/jdbi/jdbi/blob/master/examples/README.md) that doesn't have any code samples. I updated the link to point to the parent directory instead.


stevenschlansker (approved):

stevenschlansker:
Thanks!

Issue #2762, opened 2025-04-11
Kotlin plugin lacks required runtime dependency kotlin-reflect


simonolander:
It seems that `ResultBearing.mapTo(kClass: KClass<T>)` in [Extensions.kt](https://github.com/jdbi/jdbi/blob/11365a64e5bd5ba9b5d648e7b623e811985fc915/kotlin/src/main/kotlin/org/jdbi/v3/core/kotlin/Extensions.kt#L70) requires kotlin-reflect to be on the classpath.

Trying to use `mapTo`, e.g. like so:

```kotlin
jdbiHandle.createQuery("SELECT * FROM my_table").mapTo<MyKotlinEntity>().list()
```
causes the following exception to be thrown:
```
kotlin.jvm.KotlinReflectionNotSupportedError: Kotlin reflection implementation is not found at runtime. Make sure you have kotlin-reflect.jar in the classpath
	at kotlin.jvm.internal.ClassReference.error(ClassReference.kt:88)
	at kotlin.jvm.internal.ClassReference.getConstructors(ClassReference.kt:21)
	at org.jdbi.v3.core.kotlin.KotlinMapperKt.findConstructor(KotlinMapper.kt:368)
	at org.jdbi.v3.core.kotlin.KotlinMapperKt.access$findConstructor(KotlinMapper.kt:1)
	at org.jdbi.v3.core.kotlin.KotlinMapper.<init>(KotlinMapper.kt:50)
	at org.jdbi.v3.core.kotlin.KotlinMapper.<init>(KotlinMapper.kt:61)
	at org.jdbi.v3.core.kotlin.KotlinMapper.<init>(KotlinMapper.kt:61)
	at org.jdbi.v3.core.kotlin.KotlinMapperFactory.build(KotlinMapperFactory.kt:41)
	at org.jdbi.v3.core.mapper.RowMappers.findFor(RowMappers.java:171)
	at org.jdbi.v3.core.mapper.Mappers.findFor(Mappers.java:109)
	at org.jdbi.v3.core.statement.StatementContext.findMapperFor(StatementContext.java:234)
	at org.jdbi.v3.core.result.ResultBearing.lambda$mapTo$1(ResultBearing.java:130)
	at org.jdbi.v3.core.result.ResultBearing$1.scanResultSet(ResultBearing.java:58)
	at org.jdbi.v3.core.statement.Query.scanResultSet(Query.java:61)
[...]

PR #2763, opened 2025-04-14
kotlin: add kotlin-reflect to runtime classpath (#2762)

stevenschlansker:


hgschmie (commented):
Weird. The code uses kotlin.reflect.full and kotlin.reflect.jvm in the KotlinMapper and according to the docs (https://kotlinlang.org/api/core/kotlin-reflect/), it is in kotlin-reflect. I am not sure how that compiles with a runtime dependency.

for 1.6 this shows:
```
[INFO] org.jdbi:jdbi3-kotlin:jar:3.49.1-SNAPSHOT
[...]
[INFO] +- org.jetbrains.kotlin:kotlin-reflect:jar:1.6.21:compile
```

as a compile dependency. 

and the same for 2.1

Can we verify first whether the bug report is actually correct?

hgschmie:
So, this is definitely not the right fix. We are using kotlin-reflect in compile scope through a profile in the build pom. I guess it was too smart. I will send a PR.

hgschmie:
#2768 is a better fix.

PR #2764, opened 2025-04-14
kotlin: Drop explicit support for EOL versions < 1.9

stevenschlansker:


internal/build/pom.xml
```diff
@@ -151,9 +151,8 @@
         <jdbi.test.pg-version>15</jdbi.test.pg-version>
         <jdbi.test.region>US</jdbi.test.region>
         <jdbi.test.timezone>UTC</jdbi.test.timezone>
-        <kotlin.compiler.apiVersion>1.6</kotlin.compiler.apiVersion>
```
stevenschlansker, line 154: This property seems unused entirely?

.github/workflows/versions-check.yml
```diff
@@ -386,7 +386,7 @@ jobs:
 
     strategy:
       matrix:
-        test-version: [ 1.6.21, 1.7.21, 1.8.22, 1.9.22 ]
```
hgschmie, line 389: I get it that Kotlin has a "you better upgrade or good luck" versioning scheme but there will be people that are stuck on older versions. I'd suggest keeping 1.9 and 2.0 for now.

RELEASE_NOTES.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Drop explicit support for unsupported Kotlin versions < 2.1 (#2764)
```
hgschmie, line 3: unsupported sounds like it would not work technically. Maybe obsolete instead.

.github/workflows/versions-check.yml
stevenschlansker: Okay. From looking at the Kotlin site, I would be extremely uncomfortable running an older version as there seems to be basically no support other than upgrading (no new patches on old minor versions seem to ever happen). But to each their own :)

PR #2770, opened 2025-04-22
Fix invoking subclass methods with on-demand extensions

### LINKED ISSUE #2769: Invoking subclass methods with Jdbi.onDemand
### Discussed in https://github.com/jdbi/jdbi/discussions/2765

<div type='discussions-op-text'>

<sup>Originally posted by **djsstarburst** April 20, 2025</sup>
Hi Henning and Steven,

Thanks for your work on the wonderful Jdbi library - - I use it every day.

I'm porting a system from a different database to Postgres for a new product.  The changes for the new database must be strictly isolated from the DAO code run for the original database.  For each DAO with methods that must be different for Postgres, I've created a subclass of the DAO with implementations of the differing methods.  I've created a Jdbi `ExtensionFactory` which accepts an `extensionType` if it has a Postgres subclass, and which attaches to the Postgres subclass if it exists.  The `ExtensionFactory` is below.

This works very well for all ways that a `Handle` gets attached _except_ when the DAO is attached using `Jdbi.onDemand`.  In that case, invoking methods fails with the backtrace below.

I'm guessing the problem I'm trying to solve isn't new.  What's the best way to ensure that all the ways that a handle can be attached to a DAO, including `Jdbi.onDemand`, call the subclass method if it exists?

Thanks very much!
David Stryker

```
java.lang.NullPointerException: Cannot invoke "org.jdbi.v3.core.extension.ExtensionMetadata$ExtensionHandlerInvoker.invoke(Object[])" because the return value of "java.util.Map.get(Object)" is null

	at org.jdbi.v3.core.extension.ExtensionFactoryDelegate.lambda$attach$0(ExtensionFactoryDelegate.java:118)
	at org.jdbi.v3.core.internal.OnDemandExtensions.invoke(OnDemandExtensions.java:85)
	at org.jdbi.v3.core.internal.OnDemandExtensions.lambda$createProxy$2(OnDemandExtensions.java:71)
	at org.jdbi.v3.core.Jdbi.callWithExtension(Jdbi.java:544)
	at org.jdbi.v3.core.Jdbi.withExtension(Jdbi.java:531)
	at org.jdbi.v3.core.internal.OnDemandExtensions.lambda$createProxy$3(OnDemandExtensions.java:71)
	at io.starburst.stargate.database.$Proxy78.getInterfaceName(Unknown Source)
	at io.starburst.stargate.database.TestPostgresDaoPicker.testOnDemandOverloading(TestPostgresDaoPicker.java:96)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
```

```
public class JdbiDaoPickerExtensionFactory
        implements ExtensionFactory
{
    private final PostgresDaoPicker postgresDaoPicker;

    public JdbiDaoPickerExtensionFactory(PostgresDaoPicker postgresDaoPicker)
    {
        this.postgresDaoPicker = requireNonNull(postgresDaoPicker, "postgresDaoPicker is null");
    }

    @Override
    public boolean accepts(Class<?> extensionType)
    {
        // Only accept the extensionType if we're running Postgres, and there is a Postgres overloading
        return postgresDaoPicker.getOverloadingSubclass(extensionType).isPresent();
    }

    @Override
    public Set<FactoryFlag> getFactoryFlags()
    {
        return EnumSet.of(FactoryFlag.DONT_USE_PROXY);
    }

    @Override
    public <E> E attach(Class<E> extensionType, HandleSupplier handleSupplier)
    {
        return handleSupplier.getHandle().attach(postgresDaoPicker.pickDao(extensionType));
    }
}
```</div>

stevenschlansker:
Right now we build a Map with Method keys. The Method refers to the declaring type, but we don't actually care if it is some super or subtype, we just want to call the handler either way. We already know there's only one possible entry due to checks in the extension metadata builder

Fixes #2769

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Fix onDemand extension handling of method invocations from a subtype (#2769, thanks @djsstarburst!)
 - Move all kotlin deps into the module poms (fixes #2762, thanks @simonolander)
 - Drop explicit support for obsolete Kotlin versions < 1.9 (#2764)
 
--- a/core/src/main/java/org/jdbi/v3/core/extension/ExtensionFactoryDelegate.java
+++ b/core/src/main/java/org/jdbi/v3/core/extension/ExtensionFactoryDelegate.java
@@ -13,7 +13,6 @@
  */
 package org.jdbi.v3.core.extension;
 
-import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
 import java.util.Collection;
 import java.util.Collections;
@@ -25,6 +24,7 @@ import java.util.stream.Collectors;
 import org.jdbi.v3.core.config.ConfigRegistry;
 import org.jdbi.v3.core.extension.ExtensionMetadata.Builder;
 import org.jdbi.v3.core.extension.ExtensionMetadata.ExtensionHandlerInvoker;
+import org.jdbi.v3.core.internal.JdbiClassUtils.MethodKey;
 
 import static java.lang.String.format;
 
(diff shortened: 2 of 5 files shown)


djsstarburst:
Indeed, a build of branch ondemand-subtypes fixes the problem I saw with selecting a subtype's methods when using `Jdbi.onDemand`.  Thanks Steven!

djsstarburst (approved):

hgschmie (approved):

PR #2779, opened 2025-04-24
xlint changes

hgschmie:
- **fix most Xlint warnings**
- **backports of deprecation related changes**

### DIFF
--- a/cache/caffeine-cache/src/main/java/org/jdbi/v3/cache/caffeine/CaffeineCache.java
+++ b/cache/caffeine-cache/src/main/java/org/jdbi/v3/cache/caffeine/CaffeineCache.java
@@ -44,6 +44,7 @@ public final class CaffeineCache<K, V> implements JdbiCache<K, V> {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public CacheStats getStats() {
         return cache.stats();
     }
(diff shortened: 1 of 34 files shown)


core/src/main/java/org/jdbi/v3/core/mapper/GetObjectColumnMapperFactory.java
```diff
@@ -61,12 +61,13 @@ public static ColumnMapperFactory forClasses(Set<Class<?>> types) {
...
 
     @Override
+    @SuppressWarnings("unchecked")
     public Optional<ColumnMapper<?>> build(Type type, ConfigRegistry config) {
         return Optional.of(type)
             .filter(Class.class::isInstance)
             .map(Class.class::cast)
             .filter(supportedTypes::contains)
-            .map(GetObjectColumnMapper<Object>::new);
+            .map(GetObjectColumnMapper::new);
```
stevenschlansker, line 70: This change looks like a step backward. We should avoid raw types if possible

stevenschlansker (approved):

core/src/main/java/org/jdbi/v3/core/mapper/GetObjectColumnMapperFactory.java
hgschmie: the compiler marked that as gray ("not needed").

hgschmie: reverted

PR #2799, opened 2025-05-12
upgrade pg-embedded

hgschmie:


stevenschlansker (approved):
The change should be fine, but it seems the artifact was not uploaded yet at time of build.

hgschmie:
yes, the new central upload process is slow as molasses

PR #2800, opened 2025-05-14
Add kotlin value class mappers for column and arguments

hgschmie:
This rounds out value class support for Kotlin. Note that this is still
very much experimental.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- Add experimental support for Kotlin Value types as columns in the KotlinMapper.
+- Add experimental support for Kotlin Value types as arguments and columns (#2790, suggested and aided by @SpyMachine)
 
 # 3.49.3
 
--- a/core/src/main/java/org/jdbi/v3/core/argument/internal/ObjectPropertyNamedArgumentFinder.java
+++ b/core/src/main/java/org/jdbi/v3/core/argument/internal/ObjectPropertyNamedArgumentFinder.java
@@ -29,7 +29,7 @@ import org.jdbi.v3.core.statement.UnableToCreateStatementException;
  * optional prefix.
  */
 public abstract class ObjectPropertyNamedArgumentFinder implements NamedArgumentFinder {
-    private final String prefix;
+    protected final String prefix;
     protected final Object obj;
 
     private final Map<String, Optional<NamedArgumentFinder>> childArgumentFinders = new ConcurrentHashMap<>();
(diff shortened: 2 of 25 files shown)


SpyMachine:
Awesome, thank you for throwing this together!

I believe this may have broken the behavior of java collection types automatically mapping to arguments. My application is now failing to insert with:

> No argument factory registered for type [java.util.List] for element [listOfThings]...

hgschmie:
thanks for catching the regression. Looking into it.

hgschmie:
So I added a number of tests around kotlin and java lists of value types and data classes that hold record types but I was unable to reproduce this.

This might be something specific to your code; do you have a small reproducer (ideally a unit test) that shows that regression?

SpyMachine:
@hgschmie yep here ya go...

```
    val h2Extension: JdbiExtension = JdbiExtension.h2().withPlugin(KotlinPlugin(installKotlinMapperFactory = false))
        .withInitializer { ds, handle -> handle.execute("CREATE TABLE something (id integer, first text, other text, things text array)") }
    ....
    data class EvenMoreThings(val id: Int, val first: MagicType, val other: SpaceType, val justAListOfStrings: List<String>)

    @Test
    fun testListsWork() {
        handle.registerArgument(KotlinValueClassArgumentFactory())
        handle.registerColumnMapper(KotlinValueClassColumnMapperFactory())
        handle.registerRowMapper(KotlinMapperFactory())

        val expected = MagicType("does this work?")
        val other = SpaceType("no it does not.")
        val ktBean = EvenMoreThings(1, expected, other, listOf("a", "b", "c"))

        handle.createUpdate("INSERT INTO something(id, first, other, things) VALUES(:id, :first, :other, :justAListOfStrings)")
            .bindKotlin(ktBean)
            .execute()

        val result = handle.createQuery("SELECT id, first, other, things from something")
            .mapTo<MoreThings>()
            .single()

        assertThat(result).isEqualTo(MoreThings(1, expected, other))
```

hgschmie:
Thank you. I was able to reproduce it and I have a fix that I need to test and add. 

On a sidenote: The "List<String>" argument works by accident. It is picked up by the SqlArrayArgumentFactory and turned into a string. This is not exactly intentional but that is how it works.

The current code incorrectly strips out the generic (so it is a "List", not a "List<String>" and that is why the factory no longer picks it up and you see the error. I would argue that the factory not picking this up is actually the correct behavior.

hgschmie:
@SpyMachine this is probably the version that gets merged. It passes your test and all the value class tests. Please try. I am planning to push 3.49.4 out before the weekend.

SpyMachine:
Hi, I can can confirm it’s working in my app!

PR #2805, opened 2025-05-19
Mark spotbugs-annotations as optional so it does not end up in transitive dependencies

stevenschlansker:
We don't need to include spotbugs deps at runtime, and it can interfere with build configuration downstream


hgschmie (approved):

hgschmie:
We have more annotation-only jars, we should make all of them optional and provided.

PR #2813, opened 2025-05-30
SerializableTransactionRunner: handle serialization failures that happen in batches

stevenschlansker:
the failing sqlstate might be in a 'next' exception of a BatchUpdateException


core/src/main/java/org/jdbi/v3/core/transaction/SerializableTransactionRunner.java
```diff
@@ -104,17 +104,25 @@ public TransactionHandler specialize(Handle handle) throws SQLException {
...
     protected boolean isSqlState(String expectedSqlState, Throwable throwable) {
-        Throwable t = throwable;
-
-        do {
-            if (t instanceof SQLException) {
-                String sqlState = ((SQLException) t).getSQLState();
+        Deque<Throwable> exceptions = new ArrayDeque<>();
+        exceptions.add(throwable);
+        while (!exceptions.isEmpty()) {
+            Throwable next = exceptions.pollFirst();
```
hgschmie: use getFirst. This is protected by the !exceptions.isEmpty, so the NoSuchElement should never trigger, but its use feels more natural to me than the poll which may return null.

core/src/main/java/org/jdbi/v3/core/transaction/SerializableTransactionRunner.java
```diff
@@ -104,17 +104,25 @@ public TransactionHandler specialize(Handle handle) throws SQLException {
...
+            if (next instanceof SQLException) {
+                SQLException sqlExn = (SQLException) next;
```
hgschmie: use `e`, `exception` or `sqlException`. We will not run out of vowels any time soon. :-)

core/src/main/java/org/jdbi/v3/core/transaction/SerializableTransactionRunner.java
```diff
@@ -104,17 +104,25 @@ public TransactionHandler specialize(Handle handle) throws SQLException {
      * @return True if Throwable or one of its causes is an SQLException whose SQLState begins with the given state.
      */
     protected boolean isSqlState(String expectedSqlState, Throwable throwable) {
```
hgschmie, line 108:
```java
protected boolean isSqlState(String expectedSqlState, Throwable throwable) {
    Exception e = throwable;
    do {
        if (e instanceof SQLException s) {
            var sqlState = s.getSQLState();
            if (sqlState != null && sqlState.startsWith(expectedSqlState)) {
                return true;
            }
            e = s.getNextException();
            if (e == null) {
                e = e.getCause();
            }
        } else {
            e = e.getCause();
        }
    } while (e != null);
    return false;
}
```

not sure that needs a deque. The only reason I can see if q SQLException contains both a "nextException" and a "cause". In this case, the "cause" might win (it implies a more system level exception). SQL Exceptions often represent a semantic failure (e.g. updating 10 rows in a batch and some fail) and they are cascading. 

In this specific case, we only seem to care whether there is a SQL exception in the chain that represents a specific SQL State, my assumption would be that in case of a major fault (IO e.g.), nextException would be empty and getCause would return an IOException.

stevenschlansker: I updated this to use `SQLException.iterator()` which removes the need for a deque.

core/src/main/java/org/jdbi/v3/core/transaction/SerializableTransactionRunner.java
```diff
@@ -103,18 +104,26 @@ public TransactionHandler specialize(Handle handle) throws SQLException {
...
-        Throwable t = throwable;
-
-        do {
-            if (t instanceof SQLException) {
-                String sqlState = ((SQLException) t).getSQLState();
-
-                if (sqlState != null && sqlState.startsWith(expectedSqlState)) {
-                    return true;
+        Throwable cause = throwable;
```
hgschmie: cause is still a terrible name. :-)

hgschmie (approved):

core/src/main/java/org/jdbi/v3/core/transaction/SerializableTransactionRunner.java
stevenschlansker: Ok, it is now `t`.

PR #2814, opened 2025-06-02
Update Language support docs to include warning about annotation copying

### LINKED ISSUE #1685: Json column type does not work as BindBean field
Env: Postgres 12, Kotlin 1.3.70, JDBI 3.10.1 (core, kotlin, gson2, kotlin-sqlobject)

```
jdbi.installPlugin(SqlObjectPlugin())
jdbi.installPlugin(KotlinPlugin())
jdbi.installPlugin(KotlinSqlObjectPlugin())

data class Test(var name: String, @Json var details: JsonData) {}
data class JsonData(val name: String, val value: Int) {}

CREATE TABLE Test (
    id            SERIAL PRIMARY KEY,
    name          TEXT NOT NULL UNIQUE,
    details       JSONB
);
```

**Following code DOES work**
```
@SqlUpdate("INSERT INTO Test(name, details) VALUES (:name, CAST(:details AS JSONB))")
@GetGeneratedKeys("id")
fun insertWithDetails(name: String, @Json details: JsonData): Int
```

```
@SqlQuery("SELECT details FROM Test WHERE name = :name")
@Json
fun getDetails(name: String): JsonData?	
```

```
@SqlQuery("SELECT * FROM Test WHERE name = :name")
fun getTest(name: String): Test?

```

**Following code DOES NOT work**
```
fun test() {
	dao.insertBatch(listOf(
			Test("A", JsonData("foo", 1)),
			Test("B", JsonData("bar", 2)),
			Test("C", JsonData("foo", 3))
	).iterator())
}

//DAO-class
@SqlBatch("INSERT INTO Test(name, details) VALUES (:name, CAST(:details AS JSONB)")
@BatchChunkSize(1000)
fun insertBatch(@BindBean items: Iterator<Test>)

@SqlUpdate("INSERT INTO Test(name, details) VALUES (:name, CAST(:details AS JSONB)")
fun insertTest(@BindBean item: Test)
```

> Request processing failed; nested exception is org.jdbi.v3.core.statement.UnableToCreateStatementException: No argument factory registered for type [test.JsonData] for element [details] on [Test(name=A, details=JsonData(name=foo, value=1))] [statement:"INSERT INTO Test(name, details) VALUES (:name, CAST(:details AS JSONB)", arguments:{positional:{}, named:{}, finder:[{lazy bean property arguments "Test(name=A, details=JsonData(name=foo, value=1))"}]}]
org.jdbi.v3.core.statement.UnableToCreateStatementException: No argument factory registered for type [test.JsonData] for element [details] on [Test(name=A, details=JsonData(name=foo, value=1))] [statement:"INSERT INTO Test(name, details) VALUES (:name, CAST(:details AS JSONB)", arguments:{positional:{}, named:{}, finder:[{lazy bean property arguments "Test(name=A, details=JsonData(name=foo, value=1))"}]}]

Error-message is suggesting I am missing some configuration but I don't know where from? I did try Json-annotation in front of JsonData-class but it did nothing.

stevenschlansker:
Fixes #1685

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -314,13 +314,20 @@ The guava caching module is considered experimental.
 
 Jdbi can be used from any language and technology running on the JVM that can use Java code (Kotlin, Scala, Clojure, JRuby etc).
 
-While Java is a first class citizen (and will be for the foreseeable future), we provide modules for some languages that allow more idiomatic access to Jdbi functionality.
+While Java is the primary target for Jdbi, we provide modules for some languages that allow more idiomatic access to Jdbi functionality.
 
 [glossary]
 jdbi3-kotlin::
     Automatically map Kotlin data classes.
 jdbi3-kotlin-sqlobject::
     Kotlin support for the SQL Object extension.
+
+[CAUTION]
+Other languages don't always emit equivalent idiomatic Java code in the way you expect.
+Especially, by default annotations on Kotlin or Lombok properties may not be copied to the Java property.
+This means you may need to configure Lombok annotation copying like
+`lombok.copyableAnnotations += org.jdbi.v3.json.Json`
+or make sure Kotlin attaches annotations to generated Java properties like `@get:Json`.
 
 ==== External modules
 


docs/src/adoc/index.adoc
```diff
@@ -314,14 +314,21 @@ The guava caching module is considered experimental.
 
 Jdbi can be used from any language and technology running on the JVM that can use Java code (Kotlin, Scala, Clojure, JRuby etc).
 
-While Java is a first class citizen (and will be for the foreseeable future), we provide modules for some languages that allow more idiomatic access to Jdbi functionality.
```
stevenschlansker, line 317: I removed this because it's kinda weird to even imply we might remove Java support at this point.

hgschmie: I think the implication was that "Java is a first class citizen but there could be a future where multiple languages are treated equal". I agree though, that this ship has sailed.

docs/src/adoc/index.adoc
```diff
@@ -314,14 +314,21 @@ The guava caching module is considered experimental.
...
 jdbi3-kotlin::
     Automatically map Kotlin data classes.
 jdbi3-kotlin-sqlobject::
     Kotlin support for the SQL Object extension.
 
+[CAUTION]
+Other languages don't always emit equivalent idiomatic Java code in the way you expect.
+Especially, by default annotations on Kotlin or Lombok properties may not be copied to the Java property.
+This means you may need to configure Lombok annotation copying like
+`lombok.copyableAnnotations += org.jdbi.v3.json.Json`
```
hgschmie, line 329: this is a subset of the docs here: https://jdbi.org/#_lombok

It might make sense to link that part for lombok. Is there a specific reason to add this?

The specific example is listed at the end of the lombok chapter as necessary when you want to use the Json functionality.

hgschmie: Organizing docs is difficult. I don't want to drop small bits and pieces across the docs that will make it harder to keep it consistent. I this case, I think there is more value to add a link to the lombok docs that contain everything in depth than just adding a sentence that shows an incomplete lombok example and only hints at what Kotlin users should do.

PR #2817, opened 2025-06-03
TestPreparedBatchGenerateKeysPostgres: reduce dependency on current time

stevenschlansker:
Currently this test uses database time. It's more reliable to mock out time for unit tests. Also, prefer `timestamptz` instead of `timestamp` due to weirdness with timezones in the database drivers.

(This came out of GraalVM work to improve native-test reliability)

### DIFF
--- a/core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatchGenerateKeysPostgres.java
+++ b/core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatchGenerateKeysPostgres.java
@@ -13,8 +13,10 @@
  */
 package org.jdbi.v3.core.statement;
 
+import java.time.Instant;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.List;
 
 import de.softwareforge.testing.postgres.junit5.EmbeddedPgExtension;
@@ -36,7 +38,7 @@ public class TestPreparedBatchGenerateKeysPostgres {
 
     @RegisterExtension
     public PgDatabaseExtension pgExtension = PgDatabaseExtension.instance(pg).withInitializer(
-        handle -> handle.execute("create table something (id serial, name varchar(50), create_time timestamp default now())")
+        handle -> handle.execute("create table something (id serial primary key, name varchar(50), create_time timestamptz)")
     );
 
     @Test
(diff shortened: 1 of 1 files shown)


core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatchGenerateKeysPostgres.java
```diff
@@ -85,29 +88,34 @@ public void testBatchResultBearing() {
...
+                    r.getObject("create_time", OffsetDateTime.class)))
                 .list();
 
-            PreparedBatch batch2 = h.prepareBatch("update something set create_time = now() where name like :name returning id, name, create_time");
+            var now = OffsetDateTime.ofInstant(Instant.ofEpochSecond(1234567), ZoneOffset.UTC);
+
+            PreparedBatch batch2 = h.prepareBatch("update something set create_time = :now where name like :name returning id, name, create_time");
 
             batch2.bind("name", "Brian%")
+                .bind("now", now)
```
hgschmie, line 102: if you bind this explicitly, I would lose the `default now()` in the table definition (replace with `not null`).

core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatchGenerateKeysPostgres.java
```diff
@@ -85,29 +88,34 @@ public void testBatchResultBearing() {
             List<IdCreateTime> ids = batch1.executePreparedBatch("id", "create_time")
                 .map((r, ctx) -> new IdCreateTime(
                     r.getInt("id"),
-                    r.getTimestamp("create_time")))
+                    r.getObject("create_time", OffsetDateTime.class)))
                 .list();
 
-            PreparedBatch batch2 = h.prepareBatch("update something set create_time = now() where name like :name returning id, name, create_time");
+            var now = OffsetDateTime.ofInstant(Instant.ofEpochSecond(1234567), ZoneOffset.UTC);
```
hgschmie, line 97: consider using `OffsetDateTime.now(clock)` with a clock that has a constant time.

hgschmie (approved):

core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatchGenerateKeysPostgres.java
```diff
@@ -74,7 +77,7 @@ public void testBatchInsertWithKeyGenerationAndExplicitSeveralColumnNames() {
     }
 
     @Test
-    public void testBatchResultBearing() {
+    public void testBatchResultBearing() throws ParseException {
```
hgschmie: sonar claims this is unnecessary.

core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatchGenerateKeysPostgres.java
```diff
@@ -85,29 +88,34 @@ public void testBatchResultBearing() {
             List<IdCreateTime> ids = batch1.executePreparedBatch("id", "create_time")
```
hgschmie, line 87: sonar claims this is unused.

core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatchGenerateKeysPostgres.java
stevenschlansker: The result is the same either way, doesn't save any chars, and uses another interface. I think I'll leave it as is.

Issue #2749, opened 2025-01-23
Feature Request: Support batch chunk size directly in PreparedBatch implementation


alkamo:
The SQL Object interface supports simple configuration of how many rows to process at a time via the `@BatchChunkSize` annotation. I think it would be beneficial to support a batch size argument for `PreparedBatch` directly, so that the user could feed in any number of bind parameter sets, but when `execute` is called, multiple calls would be made based on the chunk size.

Currently, we have to do this manually when using `PreparedBatch` outside of the SQL Object interface, which tends to result in boilerplate code that looks something like this:

```
int counter = 0;
try(PreparedBatch insertQuery = getHandle().prepareBatch("insert :value into some_table")
{
   for (Integer values : valueList)
   {
      insertQuery.bind("value", value).add();
      counter++;
      if (counter >= 10000)
      {
         insertQuery.execute();
         counter = 0;
      }
   }
   if (counter > 0)
   {
      insertQuery.execute();
  }
}
```

If `PreparedBatch` supported a chunk size, we could make this much more concise:

```
try(PreparedBatch insertQuery = getHandle().prepareBatch("insert :value into some_table")
{
   insertQuery.setBatchChunkSize(10000);
   values.forEach(value -> insertQuery.bind("value", value).add());
   insertQuery.execute();
}
```

stevenschlansker:
Thanks for the suggestion - seems like a good feature to support.

jang-namu:
Can i work on this issue?

stevenschlansker:
Please feel free to give it a go :)

JScodesconcise:
@hgschmie the requested tests and documentation seems like it's incomplete. Can I finish those off?

PR #2818, opened 2025-06-10
Add batch chunk size support to PreparedBatch

jang-namu:
## Summary

This PR implements batch chunk size functionality for `PreparedBatch` as requested in issue #2749. This feature allows users to control how many bindings are processed in a single batch execution, which is useful for very large batch operations to avoid excessive memory usage or database limitations.

### Core Implementation
- **PreparedBatch.java**: Added `batchChunkSize` field with getter/setter methods
- **internalBatchExecute()**: Refactored to support chunked execution by splitting large batches into smaller chunks
- **executeChunk()**: New method to handle execution of individual chunks

### Testing
- **TestPreparedBatch.java**: Added comprehensive test suite covering:
  - Default batch chunk size behavior (`Integer.MAX_VALUE`)
  - Setting and getting batch chunk size
  - Invalid batch chunk size validation
  - Chunked execution with various scenarios (multiple chunks, single chunk, large batches)
  - Fluent interface support
  - Empty batch handling

### New Methods
```java
public PreparedBatch setBatchChunkSize(int batchChunkSize)
public int getBatchChunkSize()
````

### CODE REVIEW

File: core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java [hunk 1]
```diff
@@ -239,35 +240,50 @@ private ExecutedBatch internalBatchExecute() {
...
-                throw new UnableToExecuteStatementException("Exception while binding parameters", e, ctx);
+            int[] totalModifiedRows = new int[0];
+            for (int chunkStart = 0; chunkStart < bindings.size(); chunkStart += batchChunkSize) {
+                int chunkEnd = Math.min(chunkStart + batchChunkSize, bindings.size());
+
+                beforeExecution();
+                int[] chunkModifiedRows = executeChunk(parsedParameters, chunkStart, chunkEnd);
+                int originalLength = totalModifiedRows.length;
+                int newLength = originalLength + chunkModifiedRows.length;
+                int[] newTotalModifiedRows = new int[newLength];
```
stevenschlansker: We expect to get back exactly as many modified rows as we submit batches of parameters, right? Would it make sense to allocate the `int[] modifiedRows = new int[bindings.size()];` outside the for loop and copy results into the same array, rather than re-allocating every step?

stevenschlansker (approved):
Thank you for contributing this! I have one question about repeated copying of the results array - please consider it, and then we can merge this in for the next release.

core/src/main/java/org/jdbi/v3/core/statement/PreparedBatch.java
stevenschlansker: Or, if we are concerned about variable length, collect the `int[]` results into an `ArrayList<int[]>` and then assemble the final results array in the end once the final length is known. This costs slightly more temporary memory but also avoids N^2 behavior with repeated copies.

jang-namu: Thanks for the detailed feedback! I've implemented method 2 as suggested. This approach:

1. Handles variable - length results gracefully - As you mentioned, this approach naturally accommodates any unexpected variations in chunk result sizes without requiring upfront size calculations
2. Defers final array assembly - The final flattening happens only once at the end using stream().flatMapToInt(Arrays::stream).toArray()

If we could guarantee that modified rows always equals the number of parameters, we could use method 1 like this
```java
int[] totalModifiedRows = new int[bindings.size()];
for (int chunkStart = 0; chunkStart < bindings.size(); chunkStart += batchChunkSize) {
    int chunkEnd = Math.min(chunkStart + batchChunkSize, bindings.size());

    beforeExecution();

    int[] chunkModifiedRows = executeChunk(parsedParameters, chunkStart, chunkEnd);
    System.arraycopy(chunkModifiedRows, 0, totalModifiedRows, chunkStart, chunkModifiedRows.length);
}
```

core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatch.java
```diff
@@ -275,6 +275,130 @@ public void testBindNull() {
...
+            fail("Expected IllegalArgumentException for chunk size 0");
+        } catch (IllegalArgumentException e) {
+            assertThat(e.getMessage()).contains("Batch chunk size must be greater than zero");
+        }
+        try {
+            batch.setBatchChunkSize(-1);
+            fail("Expected IllegalArgumentException for negative chunk size");
+        } catch (IllegalArgumentException e) {
+            assertThat(e.getMessage()).contains("Batch chunk size must be greater than zero");
+        }
```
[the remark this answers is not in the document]
hgschmie: > [nitpick] Consider using assertThrows instead of try-catch blocks for exception testing to improve test readability and conciseness.

Yes, this should be addressed. We have plenty of tests that use assertThrows

core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatch.java
```diff
@@ -275,6 +275,130 @@ public void testBindNull() {
...
+        batch.setBatchChunkSize(3);
+        // Add 7 items (will require 3 chunks: 3, 3, 1)
+        for (int i = 0; i < 7; i++) {
+            batch.bind("id", i).bind("name", "Name" + i).add();
+        }
+        int[] results = batch.execute();
+        // Should get 7 update counts (one for each row)
+        assertThat(results).hasSize(7);
+        // All should be successful inserts
+        assertThat(results).containsOnly(1);
```
hgschmie, line 327: those should be chained 

```java
assertThat(results).hasSize(7).containsOnly(1);
```

core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatch.java
```diff
@@ -275,6 +275,130 @@ public void testBindNull() {
...
+        // Verify all data was inserted
+        List<Something> records = h.createQuery("select * from something order by id").mapToBean(Something.class).list();
+        assertThat(records).hasSize(7);
+        assertThat(records).extracting(Something::getName)
```
hgschmie, line 331: can be chained.

core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatch.java
```diff
@@ -275,6 +275,130 @@ public void testBindNull() {
...
+        PreparedBatch batch = h.prepareBatch("insert into something (id, name) values (:id, :name)");
+        // Set chunk size larger than number of items
+        batch.setBatchChunkSize(10);
+        // Add only 3 items
+        for (int i = 0; i < 3; i++) {
+            batch.bind("id", i).bind("name", "SingleChunk" + i).add();
+        }
+        int[] results = batch.execute();
+        assertThat(results).hasSize(3);
+        assertThat(results).containsOnly(1);
```
hgschmie, line 347: can be chained.

core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatch.java
```diff
@@ -275,6 +275,130 @@ public void testBindNull() {
...
+        // Verify all data was inserted
+        List<Something> records = h.createQuery("select * from something order by id").mapToBean(Something.class).list();
+        assertThat(records).hasSize(3);
+        assertThat(records).extracting(Something::getName)
```
hgschmie, line 351: can be chained. and so on. there are a few more.

core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatch.java
```diff
@@ -275,6 +275,130 @@ public void testBindNull() {
...
+
+    @Test
+    public void testBatchChunkSizeWithEmptyBatch() {
+        Handle h = h2Extension.getSharedHandle();
+        PreparedBatch batch = h.prepareBatch("insert into something (id, name) values (:id, :name)");
+        batch.setBatchChunkSize(5);
+        int[] results = batch.execute();
+        assertThat(results).isEmpty();
+        assertThat(batch.getBatchChunkSize()).isEqualTo(5);
+    }
```
hgschmie, line 401: please add a blank line after this (or reformat with the Jdbi formatting rules loaded).

core/src/test/java/org/jdbi/v3/core/statement/TestPreparedBatch.java
```diff
@@ -275,6 +275,130 @@ public void testBindNull() {
...
+            fail("Expected IllegalArgumentException for negative chunk size");
+        } catch (IllegalArgumentException e) {
+            assertThat(e.getMessage()).contains("Batch chunk size must be greater than zero");
+        }
+    }
+
+    @Test
+    public void testBatchChunkSizeExecution() {
+        Handle h = h2Extension.getSharedHandle();
+        PreparedBatch batch = h.prepareBatch("insert into something (id, name) values (:id, :name)");
```
hgschmie, line 316: I know we have tons of examples that don't but this should technically be in a try-with-resources block.

hgschmie (changes requested):

hgschmie:
This change is great, don't get me wrong, but what I am missing is an actual test that validates that setting a chunk size of "5" and executing 13 batch operations, it gets split into 3 chunks. All the tests only ensure that the operation succeeds (iaw, all the batch chunks are written and can be read) but not whether there is actual chunking. 

This may require some modification (e.g. return the number of chunks executed somewhere) or an internal, package-private chunk counter that we can read in an unit test to verify that the chunking really works.

Is this still ensuring that all batch operations are executed in the same transaction?

hgschmie:
Oh, and adding some information to the documentation would be really great (`docs/src/adoc/index.adoc` is the source for the jdbi.org site).

JScodesconcise:
@hgschmie I made the changes you requested in PR: #2952.

Let me know if you'd like me to change anything.

Issue #2824, opened 2025-07-04
Kotlin: Do not filter out unknown annotations from fields


unoexperto:
Hi folks,

We have custom json serializer for certain types. They are being serialized using Moshi. Here is its definition

```
@Retention(AnnotationRetention.RUNTIME)
@Target(AnnotationTarget.FIELD, AnnotationTarget.VALUE_PARAMETER, AnnotationTarget.PROPERTY)
@JsonQualifier
annotation class WithBpTypes(val typeMap: KClass<out WithDescriptors>)
```

Please note @JsonQualifier frm Moshi. That makes it stick in runtime. The problem in jdbi is that this annotation never reaches moshi json adapter because it's filtered by your `getQualifiers()` extension method in `KotlinMapper.resolveConstructorParameterMapper()`.

Could you please remove this filter AND also use different `Moshi.adapter()` in your `MoshiJsonMapper` that accepts additional argument `Set<Annotations>` ? I'm willing to contribute PR if you agree **in principle** that it won't be breaking change. 

@hgschmie what do you think?

UPD: Hilarious but it turned out I was original author of MoshiJsonMapper.
UPD2: Also my attempt to extend `JsonColumnMapperFactory` from `QualifiedColumnMapperFactory` so that build() can receive more type information is not going so well because there is infinite recursion in `ColumnMappers.findFor()` because it re-enters factories' build() function in line 217. That needs to be addressed as well.

hgschmie:
Will take a look.

unoexperto:
@hgschmie Thank you! I added PR that prepares jdbi for these changes. Could you please take a look ? https://github.com/jdbi/jdbi/pull/2827

PR #2827, opened 2025-07-06
Added Set<Annotation> argument to JsonMapper.forType()

unoexperto:
Added `Set<Annotation>` argument to `JsonMapper.forType()` method to allow json serialize to customize its behavior based on annotations of type/field that jdbi uses as argument or parses from result set. Related to issue #2824

FYI: @hgschmie

For this to work **in future** we'll need to inherit `JsonArgumentFactory` from `QualifiedArgumentFactory` and `JsonColumnMapperFactory` from `QualifiedColumnMapperFactory` so that builder methods can see annotations in qualified type.

For now I'd love to have this in main branch so I can new `JsonArgumentFactory` and `JsonColumnMapperFactory` locally.

---

Here is how I use it

```
// Annotation that contains reference to schema of Map<String, Value<?>> so that it can be serialized to proper json (e.g. Instant -> epoch/iso string)
@Retention(AnnotationRetention.RUNTIME)
@Target(AnnotationTarget.FIELD, AnnotationTarget.VALUE_PARAMETER, AnnotationTarget.PROPERTY)
@JsonQualifier
@org.jdbi.v3.core.qualifier.Qualifier
annotation class WithBpTypes(val typeMap: KClass<out WithDescriptors>)

interface WithDescriptors {
    val fieldTypes: Map<String, ValueType>
}
```

and I have modified argument and column factories which we later can generalize to generic json classes that jdbi currently has. Mine look like this (I use moshi)

```
// Fixed moshi adapter so that it gets list of annotations from jdbi
class MoshiJsonMapper : JsonMapper {

    override fun forType(type: Type, annotations: Set<Annotation>, config: ConfigRegistry): TypedJsonMapper? {

        return object : TypedJsonMapper {
            private val adapter = config.get(org.jdbi.v3.moshi.MoshiConfig::class.java).moshi.adapter<Any>(type, annotations)

            override fun toJson(value: Any?, config: ConfigRegistry): String? {
                return adapter.toJson(value)
            }

            override fun fromJson(json: String, config: ConfigRegistry?): Any? {
                try {
                    return adapter.fromJson(json);
                } catch (e: IOException) {
                    throw UnableToProduceResultException(e);
                }
            }
        }
    }
}

class ValueMapJsonArgumentFactory : QualifiedArgumentFactory {

    override fun build(qt: QualifiedType<*>, value: Any?, config: ConfigRegistry): Optional<Argument> {
        return if (qt.qualifiers.any { it is WithBpTypes }) {

            val mapper = config.get(JsonConfig::class.java).jsonMapper.forType(qt.type, qt.qualifiers, config)
            val a = config.get<Arguments>(Arguments::class.java)

            // look for specialized json support first, revert to simple String binding if absent
            var bindJson = a.prepareFor(ENCODED_JSON).orElse(null)
                ?: a.prepareFor(String::class.java).orElse(null)
                ?: throw UnableToCreateStatementException(JSON_NOT_STORABLE)

            // json null -> sql null
            val json = value?.let { mapper.toJson(it, config).takeIf { it != "null" } }
            Optional.ofNullable(bindJson.apply(json))
        } else
            Optional.empty<Argument>()
    }

    companion object {
        @JvmStatic
        private val ENCODED_JSON: QualifiedType<String?> = QualifiedType.of<String?>(String::class.java).with(EncodedJson::class.java)

        @JvmStatic
        private val JSON_NOT_STORABLE = String.format(
            "No argument factory found for `@%s String` or 'String'",
            EncodedJson::class.java.getSimpleName()
        )
    }
}

class ValueMapJsonColumnMapperFactory : QualifiedColumnMapperFactory {

    override fun build(qualifiedType: QualifiedType<*>, config: ConfigRegistry): Optional<ColumnMapper<*>> {

        // If we don't check for presence of special annotation we'll end up with infinite recursive call in `cm.findFor()`
        return if (qualifiedType.qualifiers.any { it is WithBpTypes }) {

            val cm = config.get<ColumnMappers>(ColumnMappers::class.java)

            // look for specialized json support first, revert to simple String mapping if absent
            val jsonStringMapper =
                cm.findFor<String>(QualifiedType.of<String>(String::class.java).with(EncodedJson::class.java)).getOrNull()
                    ?: cm.findFor<String>(String::class.java).getOrNull()
                    ?: throw UnableToProduceResultException(JSON_NOT_RETRIEVABLE)

            val mapper = config.get<JsonConfig?>(JsonConfig::class.java).jsonMapper.forType(qualifiedType.type, qualifiedType.qualifiers, config)

            Optional.of<ColumnMapper<*>>(ColumnMapper { rs: ResultSet, columnNumber: Int, ctx: StatementContext? ->
                mapper.fromJson(
                    jsonStringMapper.map(rs, columnNumber, ctx) ?: "null", // sql null -> json null
                    config
                )
            })
        } else
            Optional.empty()
    }

    companion object {
        private val JSON_NOT_RETRIEVABLE = String.format("No column mapper found for '@%s String', or 'String'", Json::class.java.getSimpleName())
    }
} 
```

### DIFF
--- a/gson2/src/main/java/org/jdbi/v3/gson2/GsonJsonMapper.java
+++ b/gson2/src/main/java/org/jdbi/v3/gson2/GsonJsonMapper.java
@@ -14,7 +14,9 @@
 package org.jdbi.v3.gson2;
 
 import java.io.IOException;
+import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import java.util.Set;
 
 import com.google.gson.TypeAdapter;
 import com.google.gson.reflect.TypeToken;
@@ -24,7 +26,7 @@ import org.jdbi.v3.json.JsonMapper;
 
 class GsonJsonMapper implements JsonMapper {
     @Override
-    public TypedJsonMapper forType(Type type, ConfigRegistry config) {
+    public TypedJsonMapper forType(Type type, Set<? extends Annotation> annotations, ConfigRegistry config) {
         return new TypedJsonMapper() {
             @SuppressWarnings("rawtypes")
             private final TypeAdapter adapter = config.get(Gson2Config.class)
--- a/jackson2/src/main/java/org/jdbi/v3/jackson2/JacksonJsonMapper.java
+++ b/jackson2/src/main/java/org/jdbi/v3/jackson2/JacksonJsonMapper.java
@@ -14,7 +14,9 @@
 package org.jdbi.v3.jackson2;
 
 import java.io.IOException;
+import java.lang.annotation.Annotation;
 import java.lang.reflect.Type;
+import java.util.Set;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JavaType;
@@ -27,7 +29,7 @@ import org.jdbi.v3.json.JsonMapper;
 
 class JacksonJsonMapper implements JsonMapper {
     @Override
-    public TypedJsonMapper forType(Type type, ConfigRegistry config) {
+    public TypedJsonMapper forType(Type type, Set<? extends Annotation> annotations, ConfigRegistry config) {
         return new TypedJsonMapper() {
             private final ObjectMapper mapper = config.get(Jackson2Config.class).getMapper();
             private final JavaType mappedType = mapper.constructType(type);
--- a/json/src/main/java/org/jdbi/v3/json/JsonMapper.java
+++ b/json/src/main/java/org/jdbi/v3/json/JsonMapper.java
@@ -13,7 +13,10 @@
  */
 package org.jdbi.v3.json;
 
+import java.lang.annotation.Annotation;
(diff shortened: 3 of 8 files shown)


gson2/src/main/java/org/jdbi/v3/gson2/GsonJsonMapper.java
```diff
@@ -24,7 +26,7 @@
 
 class GsonJsonMapper implements JsonMapper {
     @Override
-    public TypedJsonMapper forType(Type type, ConfigRegistry config) {
```
hgschmie, line 27: these are all non-backwards compatible changes which would require us to bump major (!) version. They can not be applied in the current shape.

You need to retain backwards compatibility if you want to add additional methods. 

See https://github.com/jdbi/jdbi/actions/runs/16097134400/job/45421456804

hgschmie (changes requested):

PR #2828, opened 2025-07-11
ExtensionHandler: since target does not vary, pull out the method handle binding

stevenschlansker:
This also flattens the call stack on every SqlObject invocation by inlining an Unchecked usage

This is a breaking change, but the extension interface is still declared as `@Alpha` so I don't feel too bad about it.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
+++ b/core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
@@ -18,9 +18,8 @@ import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Method;
 
 import org.jdbi.v3.core.config.ConfigRegistry;
-import org.jdbi.v3.core.internal.exceptions.Unchecked;
+import org.jdbi.v3.core.internal.exceptions.Sneaky;
 import org.jdbi.v3.meta.Alpha;
-import org.jdbi.v3.meta.Beta;
 
 import static java.lang.String.format;
 
@@ -35,34 +34,12 @@ import static java.lang.String.format;
 @Alpha
 public interface ExtensionHandler {
 
-    /** Implementation for the {@link Object#equals(Object)} method. Each object using this handler is only equal to itself. */
-    @SuppressWarnings("PMD.CompareObjectsWithEquals")
-    ExtensionHandler EQUALS_HANDLER = (handleSupplier, target, args) -> target == args[0];
-
-    /** Implementation for the {@link Object#hashCode()} method. */
-    ExtensionHandler HASHCODE_HANDLER = (handleSupplier, target, args) -> System.identityHashCode(target);
-
-    /** Handler that only returns null independent of any input parameters. */
-    ExtensionHandler NULL_HANDLER = (handleSupplier, target, args) -> null;
-
     /**
-     * Gets invoked to return a value for the method that this handler was bound to.
-     * @param handleSupplier A {@link HandleSupplier} instance for accessing the handle and its related objects
-     * @param target The target object on which the handler should operate
-     * @param args Optional arguments for the handler
-     * @return The return value for the method that was bound to the extension handler. Can be null
-     * @throws Exception Any exception from the underlying code
+     * Attach this extension handler to a target instance.
+     * @param config the configuration at time of attach
+     * @param target the target object on which the handler should operate
      */
-    Object invoke(HandleSupplier handleSupplier, Object target, Object... args) throws Exception;
-
-    /**
-     * Called after the method handler is constructed to pre-initialize any important
-     * configuration data structures.
-     *
-     * @param config the method configuration to use for warming up
-     */
-    @Beta
-    default void warm(ConfigRegistry config) {}
+    AttachedExtensionHandler attachTo(ConfigRegistry config, Object target);
 
     /**
      * Returns a default handler for missing functionality. The handler will throw an exception when invoked.
@@ -70,7 +47,7 @@ public interface ExtensionHandler {
      * @return An {@link ExtensionHandler} instance
      */
     static ExtensionHandler missingExtensionHandler(Method method) {
-        return (target, args, handleSupplier) -> {
+        return (ExtensionHandler.Simple) (handleSupplier, args) -> {
             throw new IllegalStateException(format(
                     "Method %s.%s has no registered extension handler!",
                     method.getDeclaringClass().getSimpleName(),
@@ -110,11 +87,32 @@ public interface ExtensionHandler {
      * @return An {@link ExtensionHandler}
      */
     static ExtensionHandler createForMethodHandle(MethodHandle methodHandle) {
-        return (handleSupplier, target, args) -> {
-            if (target == null) {
-                throw new IllegalStateException("no target object present, called from a proxy factory?");
+        return new ExtensionHandler() {
+            @Override
+            public AttachedExtensionHandler attachTo(ConfigRegistry config, Object target) {
+                if (target == null) {
+                    throw new IllegalStateException("no target object present, called from a proxy factory?");
+                }
+                return new AttachedExtensionHandler() {
+                    final MethodHandle boundMh = methodHandle.bindTo(target);
+                    @Override
+                    public Object invoke(HandleSupplier handleSupplier, Object... args) throws Exception {
+                        try {
+                            return boundMh.invokeWithArguments(args);
+                        } catch (Throwable e) {
+                            throw Sneaky.throwAnyway(e);
+                        }
+                    }
+                };
             }
-            return Unchecked.<Object[], Object>function(methodHandle.bindTo(target)::invokeWithArguments).apply(args);
         };
+    }
+
+    @FunctionalInterface
+    interface Simple extends ExtensionHandler, AttachedExtensionHandler {
+        @Override
+        default AttachedExtensionHandler attachTo(ConfigRegistry config, Object target) {
+            return this;
+        }
     }
 }
--- a/kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/KotlinDefaultMethodHandlerFactory.kt
+++ b/kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/KotlinDefaultMethodHandlerFactory.kt
@@ -13,6 +13,7 @@
  */
 package org.jdbi.v3.sqlobject.kotlin
 
+import org.jdbi.v3.core.extension.AttachedExtensionHandler
 import org.jdbi.v3.core.extension.ExtensionHandler
 import org.jdbi.v3.core.extension.ExtensionHandlerFactory
 import org.jdbi.v3.core.kotlin.isKotlinClass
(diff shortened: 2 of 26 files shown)


core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
```diff
@@ -110,11 +111,48 @@ static ExtensionHandler createForSpecialMethod(Method method) throws IllegalAcce
...
-        return (handleSupplier, target, args) -> {
-            if (target == null) {
-                throw new IllegalStateException("no target object present, called from a proxy factory?");
+        return new ExtensionHandler() {
+            @Override
+            public Invoker createInvoker(Object target) {
+                if (target == null) {
+                    throw new IllegalStateException("no target object present, called from a proxy factory?");
+                }
+                MethodHandle boundMh = methodHandle.bindTo(target);
```
stevenschlansker: change #1: avoid re-binding the methodhandle on every invocation

core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
```diff
@@ -110,11 +111,48 @@ static ExtensionHandler createForSpecialMethod(Method method) throws IllegalAcce
...
+                return new Invoker() {
+                    @Override
+                    public Object invoke(HandleSupplier handleSupplier, Object... args) throws Exception {
+                        try {
+                            return boundMh.invokeWithArguments(args);
```
stevenschlansker, line 101: change #2: inline Unchecked here to avoid extra stack frames cluttering user code paths

core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
```diff
@@ -52,25 +52,26 @@ public interface ExtensionHandler {
      * @param args Optional arguments for the handler
      * @return The return value for the method that was bound to the extension handler. Can be null
      * @throws Exception Any exception from the underlying code
+     * @deprecated call {@link #createInvoker(Object)} instead
      */
-    Object invoke(HandleSupplier handleSupplier, Object target, Object... args) throws Exception;
+    @Deprecated(forRemoval = true)
```
stevenschlansker: I left this only to try to satisfy `japicmp` but it still seems to be unhappy and I do not understand why.

I am tempted to delete this entirely and just ignore the japicmp failure since this is `@Alpha`.

core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
```diff
@@ -37,13 +37,13 @@ public interface ExtensionHandler {
 
     /** Implementation for the {@link Object#equals(Object)} method. Each object using this handler is only equal to itself. */
     @SuppressWarnings("PMD.CompareObjectsWithEquals")
-    ExtensionHandler EQUALS_HANDLER = (handleSupplier, target, args) -> target == args[0];
+    ExtensionHandler EQUALS_HANDLER = target -> (handleSupplier, args) -> target == args[0];
```
hgschmie: uh, that looks like black magic now. ;-)

core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
hgschmie: https://github.com/siom79/japicmp/issues/421

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlScriptsHandler.java
```diff
@@ -16,6 +16,7 @@
 import java.lang.reflect.Method;
 
 import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.config.ConfigRegistry;
```
hgschmie: this is not needed / fails the build.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlUpdateHandler.java
```diff
@@ -18,6 +18,7 @@
 
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.extension.HandleSupplier;
```
hgschmie: this is not needed and fails the build.

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java
```diff
@@ -188,125 +182,141 @@ Type getParameterType(Parameter parameter) {
...
+            public void warm(ConfigRegistry config) {
+                superInvoker.warm(config);
+                resultReturner.warm(config);
+            }
+        };
+    }
 
-        return resultReturner.mappedResult(iterable, result.getContext());
+    @Override
+    protected void warm(ConfigRegistry config) {
```
hgschmie: [ERROR] /Users/henning/code/jdbi/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/SqlBatchHandler.java:319: Tighten empty braces. [RegexpMultiline]

sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/CustomizingStatementHandler.java
```diff
@@ -181,19 +176,33 @@ private static SqlStatementCustomizerFactory instantiateFactory(Annotation annot
...
+
+            @Override
+            public void warm(ConfigRegistry config) {
+                statementCustomizers.forEach(s -> s.warm(config));
+                CustomizingStatementHandler.this.warm(config);
+            }
+        };
+    }
+
+    protected void warm(ConfigRegistry config) {
```
hgschmie: [ERROR] /Users/henning/code/jdbi/sqlobject/src/main/java/org/jdbi/v3/sqlobject/statement/internal/CustomizingStatementHandler.java:205: Tighten empty braces. [RegexpMultiline]

hgschmie:
```diff
diff --git i/sqlobject/pom.xml w/sqlobject/pom.xml
index 90d70108c..2df22fa30 100644
--- i/sqlobject/pom.xml
+++ w/sqlobject/pom.xml
@@ -128,4 +128,20 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>com.github.siom79.japicmp</groupId>
+                <artifactId>japicmp-maven-plugin</artifactId>
+                <configuration>
+                    <parameter>
+                        <excludes combine.children="append">
+                            <exclude>org.jdbi.v3.sqlobject.Handler#invoke(org.jdbi.v3.core.extension.HandleSupplier, java.lang.Object,java.lang.Object[])</exclude>
+                        </excludes>
+                    </parameter>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
 </project>
```

fixes the japicmp problem. 

We should be able to remove this for the next build.

core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
hgschmie: You can remove this and add the exclude statement to the pom. This is temporarily and can go away once we have done another release.

core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
```diff
@@ -52,25 +52,26 @@ public interface ExtensionHandler {
...
-     * Called after the method handler is constructed to pre-initialize any important
-     * configuration data structures.
-     *
-     * @param config the method configuration to use for warming up
+     * Bind this extension handler to a target instance and return an Invoker.
+     * @param target The target object on which the handler should operate
      */
-    @Beta
-    default void warm(ConfigRegistry config) {}
+    ExtensionHandler.Invoker createInvoker(Object target);
```
hgschmie: here you use the qualified name.

core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
```diff
@@ -110,11 +111,48 @@ static ExtensionHandler createForSpecialMethod(Method method) throws IllegalAcce
      * @return An {@link ExtensionHandler}
      */
     static ExtensionHandler createForMethodHandle(MethodHandle methodHandle) {
-        return (handleSupplier, target, args) -> {
-            if (target == null) {
-                throw new IllegalStateException("no target object present, called from a proxy factory?");
+        return new ExtensionHandler() {
+            @Override
+            public Invoker createInvoker(Object target) {
```
hgschmie: here you don't use the qualified name. This make the code harder to read than necessary. I prefer the "ExtensionHandler.Invoker" name everywhere but I would also ask whether it would make sense to simply factor this out into its own class and not use a nested interface.

core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
```diff
@@ -110,11 +111,48 @@ static ExtensionHandler createForSpecialMethod(Method method) throws IllegalAcce
...
+                        }
+                    }
+                };
             }
-            return Unchecked.<Object[], Object>function(methodHandle.bindTo(target)::invokeWithArguments).apply(args);
         };
     }
+
+    /**
+     * Extension handler with a target object binding.
```
hgschmie: I would not associate something called "Invoker" with "extension handler with a target object binding". I would have called that "BoundExtensionHandler" or something.

core/src/main/java/org/jdbi/v3/core/extension/ExtensionHandler.java
hgschmie: given the comment below, wouldn't "bindExtensionHandler" or similar match more closely what this code does?

hgschmie (changes requested):
All of the meat seems to be in the extension handler changes, so I focused on that, most of the changes further down seem to be simply mechanical (which is good!). 

Fixing the build errors (japicmp and checkstyle) is necessary, please consider renaming the nested invoker interface, I find the name not matching what the thing represents. 

Otherwise, the change is pretty neat. I still don't understand enough about method handles to have been come up with a similar change myself.

I am curious: Did you benchmark the change? It should speed up a bit, should it not?

kotlin-sqlobject/src/main/kotlin/org/jdbi/v3/sqlobject/kotlin/KotlinDefaultMethodHandlerFactory.kt
```diff
@@ -32,13 +33,15 @@ class KotlinDefaultMethodHandlerFactory : ExtensionHandlerFactory {
...
 
         return Optional.of(
-            ExtensionHandler { _, t, a ->
-                @Suppress("SwallowedException")
-                try {
-                    @Suppress("SpreadOperator")
-                    implementation.invoke(null, t, *a)
-                } catch (e: InvocationTargetException) {
-                    throw e.targetException
+            ExtensionHandler { c, t ->
```
hgschmie: Kotlin still hates you: 

`Use "_" instead of this unused lambda parameter "c".`

:-)

hgschmie (approved):

PR #2832, opened 2025-07-24
Add support for PGvector extension

### LINKED ISSUE #2830: ArgumentFactory for PGvector
### Discussed in https://github.com/jdbi/jdbi/discussions/2829

<div type='discussions-op-text'>

<sup>Originally posted by **heuermh** July 24, 2025</sup>
On postgres, https://github.com/pgvector/pgvector-java provides support for the `vector` type for the extension https://github.com/pgvector/pgvector.

Running with straight JDBC works well but through Jdbi results in
```
Exception in thread "main" org.jdbi.v3.core.statement.UnableToCreateStatementException: No argument factory
registered for '[0.005163866,...,0.01420848]' of qualified type com.pgvector.PGvector [statement:"INSERT INTO
data (id, embedding) VALUES (:id, :embedding)", arguments:{named:{id:1552927,embedding:[0.005163866,[...]]
	at org.jdbi.v3.core.statement.ArgumentBinder.factoryNotFound(ArgumentBinder.java:200)
	at org.jdbi.v3.core.statement.ArgumentBinder.lambda$argumentFactoryForType$0(ArgumentBinder.java:167)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at org.jdbi.v3.core.statement.ArgumentBinder.lambda$argumentFactoryForType$1(ArgumentBinder.java:167)
	at org.jdbi.v3.core.statement.ArgumentBinder.lambda$argumentFactoryForType$2(ArgumentBinder.java:168)
	at org.jdbi.v3.core.statement.ArgumentBinder.bindNamed(ArgumentBinder.java:122)
	at org.jdbi.v3.core.statement.ArgumentBinder.bind(ArgumentBinder.java:68)
	at org.jdbi.v3.core.statement.SqlStatement.internalExecute(SqlStatement.java:1812)
	at org.jdbi.v3.core.result.ResultProducers.lambda$returningUpdateCount$0(ResultProducers.java:49)
	at org.jdbi.v3.core.statement.Update.execute(Update.java:79)
	at org.jdbi.v3.core.statement.Update.execute(Update.java:58)
	at bio.network.vocabulary.vector.VectorSearch.lambda$createEmbedding$0(VectorSearch.java:92)
	at org.jdbi.v3.core.HandleConsumer.lambda$asCallback$0(HandleConsumer.java:32)
	at org.jdbi.v3.core.Jdbi.withHandle(Jdbi.java:422)
	at org.jdbi.v3.core.Jdbi.useHandle(Jdbi.java:438)
```

I am not sure if the issue is with the way I'm trying to use the PGvector-configured connection in Jdbi
```java
try (Connection connection = DriverManager.getConnection(jdbcUrl)) {
  PGvector.addVectorType(connection);
  Jdbi jdbi = Jdbi.create(connection);

  float[] embedding = ...;

  jdbi.useHandle(handle ->
    handle
      .createUpdate(locate("insert"))
      .bind("id", id)
      .bind("embedding", new PGvector(embedding))
      .execute());
```

and/or if I need to create an additional Jbdi ArgumentFactory for the `vector` or com.pgvector.PGvector type.</div>

stevenschlansker:
Fixes #2830

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -208,6 +208,11 @@
         </dependency>
 
         <dependency>
+            <groupId>com.pgvector</groupId>
+            <artifactId>pgvector</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>com.h2database</groupId>
             <artifactId>h2</artifactId>
         </dependency>
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -7694,6 +7694,9 @@ link:{jdkdocs}/java.base/java/time/Instant.html[Instant^] or link:{jdkdocs}/java
 It also configures SQL array type support for `int`, `long`, `float`, `double`,
 `String`, and `UUID`.
 
+If link:https://github.com/pgvector/pgvector-java[PGvector-java^] is available, Jdbi will register bindings
+and mappers for `PGvector`, `PGhalfvec`, `PGsparsevec`, and `PGbit`.
+
 See the link:{jdbidocs}/postgres/package-summary.html[javadoc^] for an exhaustive list.
 
 TIP: Some Postgres operators, for example the `?` query operator, collide with Jdbi or JDBC specific special characters.
--- a/internal/build/pom.xml
+++ b/internal/build/pom.xml
@@ -118,14 +118,13 @@
         <dep.oracle-xe.version>23.8.0.25.04</dep.oracle-xe.version>
         <dep.otj-pg-embedded.version>1.1.1</dep.otj-pg-embedded.version>
         <dep.pg-embedded.version>5.3.0</dep.pg-embedded.version>
+        <dep.pgvector.version>0.1.6</dep.pgvector.version>
         <dep.plugin.asciidoctor.version>3.2.0</dep.plugin.asciidoctor.version>
         <dep.plugin.cyclonedx.version>2.9.1</dep.plugin.cyclonedx.version>
         <dep.plugin.detekt.version>1.23.8</dep.plugin.detekt.version>
         <dep.plugin.dokka.version>2.0.0</dep.plugin.dokka.version>
         <dep.plugin.flatten.version>1.7.1</dep.plugin.flatten.version>
         <dep.plugin.inline.version>1.5.1</dep.plugin.inline.version>
-        <!-- remove after next basepom release (63) -->
-        <dep.plugin.jacoco.version>0.8.13</dep.plugin.jacoco.version>
         <dep.plugin.japicmp.version>0.23.1</dep.plugin.japicmp.version>
         <dep.plugin.kotlin.version>${dep.kotlin.version}</dep.plugin.kotlin.version>
         <dep.plugin.ktlint.version>3.5.0</dep.plugin.ktlint.version>
@@ -335,6 +334,12 @@
                 <groupId>org.postgresql</groupId>
                 <artifactId>postgresql</artifactId>
                 <version>${dep.postgresql.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>com.pgvector</groupId>
+                <artifactId>pgvector</artifactId>
+                <version>${dep.pgvector.version}</version>
             </dependency>
 
             <dependency>
--- a/postgres/src/main/java/org/jdbi/v3/postgres/PostgresPlugin.java
+++ b/postgres/src/main/java/org/jdbi/v3/postgres/PostgresPlugin.java
@@ -14,13 +14,17 @@
 package org.jdbi.v3.postgres;
 
 import java.sql.Connection;
+import java.sql.SQLException;
 import java.util.Map;
 
+import com.pgvector.PGbit;
+import com.pgvector.PGvector;
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.core.argument.ArgumentFactory;
 import org.jdbi.v3.core.generic.GenericType;
 import org.jdbi.v3.core.internal.JdbiClassUtils;
+import org.jdbi.v3.core.internal.UtilityClassException;
 import org.jdbi.v3.core.internal.exceptions.Unchecked;
 import org.jdbi.v3.core.spi.JdbiPlugin;
 import org.jdbi.v3.postgres.internal.BitStringEnumSetArgumentFactory;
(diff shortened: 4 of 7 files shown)


docs/pom.xml
```diff
@@ -207,6 +207,11 @@
             <artifactId>postgresql</artifactId>
         </dependency>
 
+        <dependency>
```
hgschmie, line 210: is this needed? I don't see any actual code references in the doc folder.

docs/src/adoc/index.adoc
```diff
@@ -7694,6 +7694,9 @@ link:{jdkdocs}/java.base/java/time/Instant.html[Instant^] or link:{jdkdocs}/java
 It also configures SQL array type support for `int`, `long`, `float`, `double`,
 `String`, and `UUID`.
 
+If link:https://github.com/pgvector/pgvector-java[PGvector-java^] is available, Jdbi will register bindings
+and mappers for `PGvector`, `PGhalfvec`, and `PGsparsevec`.
```
hgschmie: why not PGbit? 

Consider linking https://javadoc.io/doc/com.pgvector/pgvector/latest/com/pgvector/ pages for javadocs.

internal/build/pom.xml
```diff
@@ -337,6 +337,12 @@
                 <version>${dep.postgresql.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>com.pgvector</groupId>
+                <artifactId>pgvector</artifactId>
+                <version>0.1.6</version>
```
hgschmie: consider using a property `${dep.pgvector.version}`.

postgres/src/main/java/org/jdbi/v3/postgres/PostgresPlugin.java
```diff
@@ -14,8 +14,10 @@
 package org.jdbi.v3.postgres;
 
 import java.sql.Connection;
+import java.sql.SQLException;
 import java.util.Map;
 
+import com.pgvector.PGvector;
```
hgschmie, line 21: Is there a test that confirms this is still working with the dependency absent?

postgres/src/main/java/org/jdbi/v3/postgres/PostgresPlugin.java
```diff
@@ -163,12 +167,24 @@ public void customizeJdbi(Jdbi jdbi) {
...
         PGConnection pgConnection = Unchecked.supplier(() -> conn.unwrap(PGConnection.class)).get();
         return handle.configure(PostgresTypes.class, pt -> {
             pt.addTypesToConnection(pgConnection);
             pt.setLobApi(new PgLobApiImpl(conn));
         });
     }
+
+    static final class VectorEnabler {
+        private VectorEnabler() {
+            throw new RuntimeException();
```
hgschmie: please throw UtilityClassException as we do in many other places.

postgres/src/test/java/org/jdbi/v3/postgres/TestVectorExtension.java
```diff
@@ -0,0 +1,88 @@
...
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Tag("slow")
+@Testcontainers
+public class TestVectorExtension {
+    @Container
+    static JdbcDatabaseContainer<?> dbContainer = new PgVectorContainerProvider().newInstance("pg17");
```
hgschmie, line 36: consider pulling jdbi.test.pg-version using System.getProperty, similar as we do for the MySQL tests and add it to "pg" to not fully hardcode the version here.

hgschmie (approved):
The "throw RuntimeException" statement should be changed, everything else is make the code nicer. It would be good to have a test that checks that the plugin still works with PGVector absent but that would need to be an integration test.

hgschmie:
Also, it seems that running the build actually re-sorts some deps; it moves the postgresql artifact behind pgvector.

docs/src/adoc/index.adoc
stevenschlansker: Because it's not mentioned anywhere in the pgvector readme, so I didn't realize it even exists.

hgschmie:
> Also, it seems that running the build actually re-sorts some deps; it moves the postgresql artifact behind pgvector.

This is why the reproducible build fails. In the first run, it stores the `Git-Commit-Id-Describe: v3.49.5-12-ga9fe6fa` in the manifest that is used for the checkout. When running the compare, main has a change, so it stores `v3.49.5-12-ga9fe6fa-dirty`and that makes it not reproducible.

It actually works correctly. ;-)

docs/pom.xml
stevenschlansker: Strange, I had a build failure about docs project dependencies at some point, but I don't see it now.

stevenschlansker:
The failure message around sorting poms and reproducible builds is not easy to unpack if you don't already know what's going on...

docs/pom.xml
stevenschlansker: @hgschmie Yes, it is, otherwise docs cannot build:

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.11.2:aggregate (aggregate) on project jdbi3-docs: An error has occurred in Javadoc report generation: 
Error:  Exit code: 1
Error:  /home/runner/work/jdbi/jdbi/postgres/src/main/java/org/jdbi/v3/postgres/PostgresPlugin.java:20: error: package com.pgvector does not exist
Error:  import com.pgvector.PGbit;
```

postgres/src/main/java/org/jdbi/v3/postgres/PostgresPlugin.java
stevenschlansker: `postgis`, `moshi`, `jackson`, `gson2` artifacts all depend on `jdbi3-postgresql but not `pgvector` and work ok, so I think that's sufficient? It is a little bit accidental but it does cover this case.

postgres/src/test/java/org/jdbi/v3/postgres/TestVectorExtension.java
stevenschlansker: We don't actually care about the server version at all here, as long as pgvector works, so I think I'll leave it as is for now to keep locality.

PR #2839, opened 2025-08-18
sqlite: update url tests to not possibly access the internet

stevenschlansker:
this is why these classes are deprecated

### DIFF
--- a/sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
+++ b/sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
@@ -14,6 +14,7 @@
 package org.jdbi.v3.sqlite3;
 
 import java.net.MalformedURLException;
+import java.net.URI;
 import java.net.URL;
 
 import org.assertj.core.api.Assertions;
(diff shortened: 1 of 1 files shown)


sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
```diff
@@ -45,15 +46,15 @@ public void tearDown() {
 
     @Test
     public void testInsertUrlSuccessful() throws MalformedURLException {
-        String googleString = "http://www.google.com";
-        URL googleUrl = new URL(googleString);
+        String urlString = "file:/my-url";
+        URL url = new URL(urlString);
```
hgschmie: all the c'tors are deprecated for 20+; it seems that URI is the new hotness.

Maybe replace with `URI.create(...).toURL()` ?

sqlite/src/test/java/org/jdbi/v3/sqlite3/TestUrls.java
```diff
@@ -45,15 +46,15 @@ public void tearDown() {
 
     @Test
     public void testInsertUrlSuccessful() throws MalformedURLException {
-        String googleString = "http://www.google.com";
```
hgschmie, line 48: if google.com is the problem, use example.com ?

stevenschlansker: `example.com` still triggers a DNS request that might fail, as far as I can tell. the problem is that merely observing a URL (e.g. `hashCode()` ) can trigger a DNS request. and, graal doesn't include any http(s) url protocol handlers unless you explicitly request it, which fails here.

PR #2840, opened 2025-08-18
split derby and hsqldb tests

stevenschlansker:
derby doesn't do well in the native-image environment

### DIFF
--- a/testing/src/test/java/org/jdbi/v3/testing/junit5/JdbiDerbyExtensionTest.java
+++ b/testing/src/test/java/org/jdbi/v3/testing/junit5/JdbiDerbyExtensionTest.java
@@ -19,6 +19,7 @@ import java.util.UUID;
 import org.jdbi.v3.core.statement.Query;
 import org.jdbi.v3.testing.junit5.internal.TestingInitializers;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledInNativeImage;
 import org.junit.jupiter.api.condition.EnabledForJreRange;
 import org.junit.jupiter.api.condition.JRE;
 import org.junit.jupiter.api.extension.RegisterExtension;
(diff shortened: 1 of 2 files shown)


testing/src/test/java/org/jdbi/v3/testing/junit5/JdbiDerbyExtensionTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.DisabledInNativeImage;
+import org.junit.jupiter.api.condition.EnabledForJreRange;
+import org.junit.jupiter.api.condition.JRE;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@EnabledForJreRange(min = JRE.JAVA_17)
+@DisabledInNativeImage // Derby does not support native-image
```
hgschmie: we use an old derby because we are stuck on 17; does that still apply for 21?

testing/src/test/java/org/jdbi/v3/testing/junit5/JdbiHsqldbExtensionTest.java
```diff
@@ -26,29 +26,13 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 @EnabledForJreRange(min = JRE.JAVA_17)
```
hgschmie, line 28: this might not be needed for hsqldb.

testing/src/test/java/org/jdbi/v3/testing/junit5/JdbiDerbyExtensionTest.java
stevenschlansker: they dynamically generate code with a custom classloader
https://issues.apache.org/jira/browse/DERBY-7108

testing/src/test/java/org/jdbi/v3/testing/junit5/JdbiHsqldbExtensionTest.java
stevenschlansker: The build seems to fail on 11

PR #2842, opened 2025-08-29
make annotation jars available in all modules

hgschmie:
We use some of the annotation jars in modules but currently this needs to be added to each module specifically.

As we make the jars provided and optional, we can pull them into the build pom which makes them available everywhere evenly.

This also requires a small change to the BOM generation which would currently pull compile dependencies in (so we turn those off).

### DIFF
--- a/cache/caffeine-cache/pom.xml
+++ b/cache/caffeine-cache/pom.xml
@@ -43,13 +43,6 @@
         </dependency>
 
         <dependency>
-            <groupId>com.github.spotbugs</groupId>
-            <artifactId>spotbugs-annotations</artifactId>
-            <scope>provided</scope>
-            <optional>true</optional>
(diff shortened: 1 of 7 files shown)


hgschmie:
> oracle12/pom.xml	Removed test dependencies (assertj-core, junit-jupiter-api)
> Comments suppressed due to low confidence (1)
> **oracle12/pom.xml:1**
> 
> * The removal of test dependencies (assertj-core and junit-jupiter-api) from oracle12/pom.xml doesn't align with the PR's stated purpose of centralizing annotation jars. These are test dependencies, not annotation libraries, and their removal should be justified or moved to a separate change.

Truth! I though I could pull a fast one but Copilot figured it out 😎

stevenschlansker (approved):

PR #2856, opened 2025-10-02
remove some of the oracle version tests

hgschmie:


stevenschlansker (approved):
`full` is better than `slim` somehow?

PR #2863, opened 2025-10-13
bump basepom to 65

hgschmie:


docs/pom.xml
```diff
@@ -387,6 +387,9 @@
             <properties>
                 <basepom.check.skip-checkstyle>true</basepom.check.skip-checkstyle>
                 <basepom.check.skip-pmd>true</basepom.check.skip-pmd>
+
+                <!-- javadoc title -->
+                <basepom.javadoc.title>Jdbi3 ${project.version} API</basepom.javadoc.title>
```
stevenschlansker, line 392: the `3` in `jdbi3` seems a little redundant here

hgschmie: If we choose to bump to 4.0.0, it is not (Jdbi flavor 3, version 4.0.0). It is also what we always had as title.

PR #2865, opened 2025-10-13
Turn off URL inheritance in the poms

hgschmie:
One never stops learning about Maven. It is actually possible to
stop it from rewriting the urls for connections and sites.

### DIFF
--- a/internal/root/pom.xml
+++ b/internal/root/pom.xml
@@ -12,7 +12,10 @@
 ~   See the License for the specific language governing permissions and
 ~   limitations under the License.
 -->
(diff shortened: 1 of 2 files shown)


internal/root/pom.xml
```diff
@@ -56,7 +59,9 @@
     <!-- do not remove URL and SCM, otherwise OSS refuses to accept the release -->
     <url>https://jdbi.org/</url>
 
-    <scm>
+    <scm child.scm.connection.inherit.append.path="false"
+         child.scm.developerConnection.inherit.append.path="false"
```
stevenschlansker, line 63: Wow, now this is obscure

hgschmie: I plan to add it as default for basepom-66

PR #2868, opened 2025-10-15
Restore retention policy on @Alpha and @Beta

stevenschlansker:


hgschmie:
LGTM but why was that removed in the first place?

stevenschlansker:
Reverting to how it was in the last release from intermediate work on the Graal PR that got merged.

PR #2869, opened 2025-10-15
Remove redundant dep and plugin

stevenschlansker:


stevenschlansker:
@hgschmie I tried to remove this, but it seems to be necessary. It is because Spring Boot provides the AOT processing that takes the usual spring context dynamic subclassing, and does it ahead of time so Graal native-image can inspect the generated classes.

It is unfortunate but so is Spring...

PR #2870, opened 2025-10-15
Draft: ci: try building graal native-image in parallel

stevenschlansker:


internal/build/pom.xml
```diff
@@ -1004,6 +1004,7 @@
                             <enabled>true</enabled>
                         </metadataRepository>
                         <buildArgs>
+                            <buildArg>-J-Xmx10g</buildArg>
```
hgschmie, line 1007: this seems random. And it makes building the code hard and penalizes everyone even though most people will not need/use native

stevenschlansker: This is a draft, I am just experimenting.

stevenschlansker: And it is not random. By default the native-image builder will use something like 80% of available mem, which makes it not possible to run them in parallel, since you immediately overcommit memory.

This experiment is to *reduce* the amount of memory native-image uses per process, and run multiple in parallel.

But as far as I can tell, this does not help, so far.

PR #2871, opened 2025-10-15
For 3.50.0, move baseline to Java 17!

stevenschlansker:


internal/build/pom.xml
```diff
@@ -46,7 +46,6 @@
         <dep.asciidoctorj-diagram-ditaamini.version>1.0.3</dep.asciidoctorj-diagram-ditaamini.version>
         <dep.asciidoctorj-diagram.version>3.0.1</dep.asciidoctorj-diagram.version>
         <dep.asciidoctorj.version>3.0.0</dep.asciidoctorj.version>
-        <!-- leave at 3.x for now, 4.0.0 needs Java 17 -->
```
hgschmie, line 49: well, if we don't update to 4.x.x, leave the comment in place so we remember to do so.

internal/build/pom.xml
```diff
@@ -57,7 +56,6 @@
         <dep.derby.version>10.16.1.1</dep.derby.version>
         <dep.detekt.version>1.23.8</dep.detekt.version>
         <dep.errorprone.version>2.39.0</dep.errorprone.version>
-        <!-- last flyway version to support java 11 -->
```
hgschmie, line 60: same here. That still holds true. And 11.x needs additional changes.

hgschmie (approved):

PR #2875, opened 2025-10-30
JDBI3 - JDK 17 compatibility changes

hgschmie:
This is the full set of changes to release a JDK 17 version of JDBI3.


stevenschlansker:
This is mixing in a ton of "remove deprecated code" changes, which are great, but means this probably cannot be merged until Jdbi 4. Please consider #2871 in the meantime and we can work on this bigger change over time.

PR #2876, opened 2025-10-31
update graalvm CI to use 21 and 25

hgschmie:


Eng-Fouad:
It seems that GraalVM 21 does not recognize the new config file `reachability-metadata.json` ([introduced since GraalVM 23](https://www.graalvm.org/jdk23/reference-manual/native-image/metadata/)). It recognizes the old config files such as: `reflect-config.json`, `proxy-config.json`, etc.

PR #2880, opened 2025-11-04
java17 changes

hgschmie:
Collection of code changes with Java 17:

- **Use Java 17 constructs to make code more readable**
- **move java 17 only test classes to their respective modules**

### DIFF
--- a/benchmark/src/main/java/org/jdbi/v3/benchmark/BeanBindingBenchmark.java
+++ b/benchmark/src/main/java/org/jdbi/v3/benchmark/BeanBindingBenchmark.java
@@ -22,7 +22,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.jdbi.v3.core.internal.exceptions.Unchecked;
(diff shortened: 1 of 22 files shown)


java21/src/test/java/org/jdbi/v3/java21/telemetry/InMemorySpanExporter.java
```diff
@@ -23,31 +23,28 @@
 import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 
-import static io.opentelemetry.sdk.common.CompletableResultCode.ofFailure;
-import static io.opentelemetry.sdk.common.CompletableResultCode.ofSuccess;
```
stevenschlansker, line 27: why are you changing all these static imports? I don't care much one way or the other, but it seemed fine before, and this seems to mix functional updates with kinda random style changes

hgschmie: this may be a "Intellij reformat" setting. What should we be doing? I am not consciously making changes.

stevenschlansker: I don't have a super strong opinion either way, but would like to avoid IDE reformat wars.

PR #2881, opened 2025-11-04
fix graal ci

hgschmie:
- **update graalvm CI action**
- **Updates for native build**

### DIFF
--- a/.github/workflows/graal-ci.yml
+++ b/.github/workflows/graal-ci.yml
@@ -14,10 +14,9 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  ci:
-    id: native-build
-    name: GraalVM native build
+  native-build:
     runs-on: ubuntu-latest
+    name: GraalVM native build
     strategy:
       matrix:
         java-version: [ 25 ]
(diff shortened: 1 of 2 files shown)


docs/src/adoc/index.adoc
```diff
@@ -9318,23 +9318,6 @@ Configure the `maven-compiler-plugin` in your POM:
 * Java -> Compiler
 * Under "Classfile Generation," check the option "Store information about method parameters (usable via reflection)."
 
-=== GraalVM Native Image
```
stevenschlansker, line 9321: moving this block messes up the diff - was this intentional? seems like unnecessary shuffling.

.github/workflows/graal-ci.yml
```diff
@@ -9,20 +9,59 @@ on:
...
         with:
-          java-version: 25
+          java-version: ${{ matrix.java-version }}
           cache: maven
 
-      - name: build
+      - name: build code
+        id: install
         env:
-          MAVEN_ARGS: "-B -Dbasepom.check.skip-enforcer=false -Dbasepom.check.skip-all=true -Djdbi.native=true"
```
stevenschlansker, line 26: I intentionally skipped these checks because the build is already long and all these checks are run elsewhere

hgschmie: this shaves ~ 30-40 seconds off a 1:20h build. I can turn it off again, but it seems to be not that significant. 

We do not run any of the checkers in native mode in any other pipeline. I expect that not to be different but one never knows.

docs/src/adoc/index.adoc
hgschmie: yes. the "advanced topics" chapters are a bit jumbled; we should keep them in order of relevance / importance to users. I felt that native image is less important than generics, config, how to write a plugin, logging etc. I moved it right next to the generator and before the decorator thing that is incredibly obscure.

PR #2887, opened 2025-11-05
make `make install-fast` even faster

stevenschlansker:
M3 MBP before: 2m after: 20s

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ install-notests:: install
 install-nodocker:: MAVEN_ARGS += -Dno-docker=true
 install-nodocker:: install
 
-install-fast:: MAVEN_ARGS += -Pfast
+install-fast:: MAVEN_ARGS += -Pfast -T2C
 install-fast:: install
 
 compare-reproducible:: MAVEN_ARGS += -Dbasepom.test.skip=true -Djdbi.check.skip-japicmp=true


hgschmie:
throws warnings like this:

```
[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but this         *
[WARNING] * project contains the following plugin(s) that have goals not  *
[WARNING] * marked as thread-safe to support parallel execution.          *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against Apache Maven.                           *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked as thread-safe in jdbi3 - internal - documentation:
[WARNING]   org.jetbrains.kotlin:kotlin-maven-plugin:2.2.21
[WARNING]
[WARNING] Enable debug to see precisely which goals are not marked as thread-safe.
[WARNING] *****************************************************************
```

stevenschlansker:
Confusingly, the issue in the Kotlin tracker is marked as "fixed" in 1.1.3...
https://youtrack.jetbrains.com/issue/KT-10028/Support-parallel-builds-in-maven

hgschmie:
Well, the fix is "don't do that" :-) There is no actual code resolution. "Fixed" is misleading here, it should probably be "resolved, won't fix"

stevenschlansker:
That's incredibly lame. Kotlin feels like a second-class language more and more where they don't bother to actually fix problems...

PR #2892, opened 2025-11-06
spring: replace spring-jcl adapter with jcl-over-slf4j

stevenschlansker:
spring-jcl tries to autodetect log4j which is pointless here, and it has duplicate classes to jcl-over-slf4j, so the two cannot mix.

jcl-over-slf4j is a dependency of maven-loader, coming in through pg-embedded.

this means you cannot mix the `jdbi3-spring` and `jdbi3-postgres` `test-jar`s together, or the duplicate finder complains

so, standardize on jcl-over-slf4j + slf4j-simple, and nuke spring-jcl

### DIFF
--- a/internal/build/pom.xml
+++ b/internal/build/pom.xml
@@ -236,6 +236,12 @@
                 <groupId>org.springframework</groupId>
                 <artifactId>spring-core</artifactId>
                 <version>${dep.spring.version}</version>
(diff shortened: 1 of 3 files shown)


stevenschlansker:
It's hard for me to imagine a Spring application that doesn't itself bring in `spring-core` :)

PR #2893, opened 2025-11-06
PostgresPlugin: unnecessary use of Unchecked

stevenschlansker:


hgschmie (approved):
This has been in the code since forever; do we know what the original intention was?

stevenschlansker:
Probably the original author didn't have `throws SQLException` in the plugin method signature at the time, but since it's allowed to throw that, no need to unwrap anything.

PR #2899, opened 2025-11-22
feat: Support TYPE_USE @Nullable annotations (e.g., jspecify)

protocol7:
Add support for TYPE_USE @Nullable annotations in ConstructorMapper, enabling compatibility with modern nullability frameworks like JSpecify.

Background on Type Annotations and JSpecify:
---------------------------------------------

Traditional nullable annotations use @Target(PARAMETER), appearing directly on method/constructor parameters:
  void method(@Nullable String param)

Type-use annotations (@Target(TYPE_USE)) appear on types themselves, introduced in Java 8 (JSR 308). They provide more precise nullability information:
  void method(@Nullable String param)  // nullable applies to String type

JSpecify is a modern nullability annotation framework that uses TYPE_USE annotations with RUNTIME retention. This allows:
- More precise nullability checking by static analysis tools
- Better IDE support and null-safety guarantees
- Runtime reflection of nullability information

The Problem:
-----------

JDBI's ConstructorMapper.isNullable() only checked parameter.getAnnotations(), which returns annotations with @Target(PARAMETER). This missed TYPE_USE annotations like JSpecify's @Nullable, causing constructor parameter mapping to fail even when parameters were correctly annotated as nullable.

Example failure:
  @JdbiConstructor
  public Bean(@Nullable String optional, String required) {...}

When 'optional' column was missing from result set, JDBI would throw: "parameter '[optional]' has no matching columns in the result set" despite the @Nullable annotation being present.

The Fix:
--------

Modified isNullable() to check both:
1. parameter.getAnnotations() - for traditional @Target(PARAMETER)
2. parameter.getAnnotatedType().getAnnotations() - for TYPE_USE

This is a minimal, backward-compatible change that maintains existing behavior while adding support for TYPE_USE nullable annotations.

Benefits:
- Full JSpecify compatibility
- Supports other TYPE_USE nullable annotations
- No breaking changes to existing code

Testing:
--------

Added comprehensive tests using a TYPE_USE @Nullable test annotation:
- testTypeUseNullableParameterPresent: nullable param with column present
- testTypeUseNullableParameterAbsent: nullable param with column absent
- testTypeUseNonNullableAbsent: non-nullable param requires column

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
@@ -358,7 +358,9 @@ public final class ConstructorMapper<T> implements RowMapper<T> {
 
     private boolean isNullable(Parameter parameter) {
         // Any annotation named @Nullable is honored. We're nice that way.
-        return Stream.of(parameter.getAnnotations())
+        // Check both parameter annotations and type-use annotations (e.g., jspecify @Nullable)
+        return Stream.of(parameter.getAnnotations(), parameter.getAnnotatedType().getAnnotations())
+            .flatMap(Stream::of)
             .map(Annotation::annotationType)
             .map(Class::getSimpleName)
             .anyMatch("Nullable"::equals);
--- a/core/src/test/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapperTest.java
+++ b/core/src/test/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapperTest.java
@@ -605,4 +605,45 @@ public class ConstructorMapperTest {
             .hasMessageContaining("or specify it with @JdbiConstructor");
     }
 
+    // Test TYPE_USE @Nullable annotations (e.g., jspecify)
+    @Test
+    public void testTypeUseNullableParameterPresent() {
+        TypeUseNullableParameterBean bean = handle
+            .registerRowMapper(ConstructorMapper.factory(TypeUseNullableParameterBean.class))
+            .select("select s, i from bean")
+            .mapTo(TypeUseNullableParameterBean.class)
+            .one();
+        assertThat(bean.s).isEqualTo("3");
+        assertThat(bean.i).isEqualTo(2);
+    }
+
+    @Test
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
```diff
@@ -358,7 +358,12 @@ private OptionalInt locatePropagateNullColumnIndex(List<String> columnNames, Lis
 
     private boolean isNullable(Parameter parameter) {
         // Any annotation named @Nullable is honored. We're nice that way.
+        // Check both parameter annotations and type-use annotations (e.g., jspecify @Nullable)
         return Stream.of(parameter.getAnnotations())
```
stevenschlansker: It looks like this change is repeating the mapping to annotation type and getSimpleName.
Is there a reason to prefer this approach over something slightly less redundant, like
```
Stream.of(parameter.getAnnotations(), parameter.getAnnotatedType().getAnnotations())
        .flatMap(Stream::of)
        .map(Annotation::annotationType)
        .map(Class::getSimpleName)
        ...
```

stevenschlansker (approved):
Thank you for reporting this issue and providing a fix. I have one comment about redundant code that I'd like to see addressed before merging, unless there's a reason not to.

stevenschlansker:
Additionally, the style checkers have some minor requested changes (tighten empty braces)

core/src/main/java/org/jdbi/v3/core/mapper/reflect/ConstructorMapper.java
protocol7: Great point, I've fixed this now.

protocol7:
> Additionally, the style checkers have some minor requested changes (tighten empty braces)

Fixed!

stevenschlansker:
Thank you for the contribution!

protocol7:
Thanks! Any idea when this might go into a release?

stevenschlansker:
There's a couple of open PRs and one issue I'd like to get fixed first, then we can make a new patch release. So hopefully in the next few days to a week or so.

protocol7:
👍🏻

PR #2901, opened 2025-12-01
Allow configuring log level for Slf4JSqlLogger

phinjensen:
This is a small, non-breaking change that makes it easy to configure the log level of messages generated by `Slf4JSqlLogger`. I have several projects which spew out vast numbers of `DEBUG` messages, and this makes it really easy to change the level to something easier to see, like INFO.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
@@ -27,19 +27,18 @@ import org.slf4j.event.Level;
  */
 public class Slf4JSqlLogger implements SqlLogger {
     private final Logger log;
-    private Level level = Level.DEBUG;
+    private final Level level;
 
     public Slf4JSqlLogger() {
-        this(LoggerFactory.getLogger("org.jdbi.sql"));
+        this(LoggerFactory.getLogger("org.jdbi.sql"), Level.DEBUG);
     }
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jdbi/v3/core/statement/Slf4JSqlLogger.java
```diff
@@ -26,6 +27,7 @@
  */
 public class Slf4JSqlLogger implements SqlLogger {
     private final Logger log;
+    private Level level = Level.DEBUG;
```
stevenschlansker: It would be nice to declare this field `final` which I think is possible by changing the constructors such that the no-arg constructor calls `this(Level.DEBUG, LoggerFactory.getLogger(...))` rather than the other way around.

stevenschlansker:
@phinjensen , thank you for the contribution. I have one requested change if you could address that.

phinjensen:
@stevenschlansker Good idea. I just pushed a commit making that change.

stevenschlansker (approved):
Thank you for the contribution @phinjensen !

stevenschlansker:
Requires #2902 or a similar change before this can be merged.

hgschmie:
Adding this PR *will* break compatibility to previous versions as it will require slf4j 2.x. Anyone who is still using slf4j 1.7.x must upgrade or they can not use this version of Jdbi.

This needs to be stated very clearly in the docs.

PR #2902, opened 2025-12-01
ci: remove slf4j version 1.x test

stevenschlansker:
slf4j 1.x is no longer supported, and we'd like to use new 2.x features)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,8 @@
 # Unreleased
 
 - Add new `jackson3` artifact for Jackson 3 ( #2878 )
+- Update to SLF4J 2, to allow for configurable log levels in SqlLogger (#2902)
+
 - Deprecate for removal installPlugins plugin discovery. It's too easy to get yourself into trouble.
 - Support TYPE_USE `@Nullable` annotations like JSpecify ( #2899, thanks @protocol7 ! )
 
--- a/.github/workflows/versions-check.yml
+++ b/.github/workflows/versions-check.yml
@@ -508,7 +508,7 @@ jobs:
 
     strategy:
       matrix:
-        test-version: [ 2.1.0-alpha1, 1.7.36 ]
+        test-version: [ 2.1.0-alpha1 ]
 
     steps:
       - uses: actions/checkout@v4


hgschmie:
The docs (https://www.slf4j.org/manual.html#compatibility) state that "Client code compiled with slf4j-api-N.jar will run perfectly fine with slf4j-api-M.jar for any N and M."

Here is my rationale: people are stuck on older versions. Especially slf4j is one of those libraries where upgrading a large project is really hard. While we compile against 2.0, we actually do not use any of the newer (not backwards compatible) APIs in slf4j. So you can drop Jdbi into any project that uses slf4j, even if it is an older version (usually 1.7.x which is ubiquitous). 

This test ensures that if you do that, all of Jdbi will still work. Basically, it ensures that we do not use a 2.0 API by accident. 

The user still has to use the same versions for the API jar and the provider jar but that is out of scope for us.


If you feel very strongly about removing this test, let me know; I think testing compatibility is important, especially for a (relatively mature) library such as Jdbi.

hgschmie:
And that: https://github.com/jdbi/jdbi/actions/runs/19836159325/job/56844122963  is exactly why that test exists. Applying this change will break Jdbi for anyone that is using slf4j 1.7.x. I am very reluctant to force people to upgrade such a central dependency. I'd rather have us go back to requiring 1.7.x

hgschmie:
IFF we make this change, we should drop 1.7.36 as a test version but keep the action.

stevenschlansker:
Hm, I didn't realize 1.x compatibility is so important. I thought all major logging frameworks would implement the 2.x api by now, but I only have experience with logback, (and it looks like log4j is no problem either).

RELEASE_NOTES.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Remove support for slf4j 1.x, which is no longer maintained (#2902)
```
hgschmie: This is ok, but I would have liked something like "Starting with this release, Jdbi uses features from slf4j 2.x and some parts will no longer work with version 1.x. If you are still using Jdbi with slf4j, you are strongly recommended to upgrade"

We are not dropping support because 1.7.x is no longer maintained, we are dropping support because we want to use newer 2.x features.

hgschmie (approved):

RELEASE_NOTES.md
stevenschlansker: Updated

PR #2903, opened 2025-12-01
Initial Jackson 3 support

### LINKED ISSUE #2878: Add support for jackson 3

stevenschlansker:
Fixes #2878

### DIFF
--- a/internal/build/pom.xml
+++ b/internal/build/pom.xml
@@ -68,7 +68,9 @@
         <dep.hikari-cp.version>7.0.2</dep.hikari-cp.version>
         <dep.hsqldb.version>2.7.4</dep.hsqldb.version>
         <dep.immutables.version>2.11.6</dep.immutables.version>
+        <dep.jackson-annotations.version>2.20</dep.jackson-annotations.version>
         <dep.jackson2.version>2.18.5</dep.jackson2.version>
+        <dep.jackson3.version>3.0.3</dep.jackson3.version>
         <dep.jakarta-annotation-api.version>3.0.0</dep.jakarta-annotation-api.version>
         <dep.jakarta-inject-api.version>2.0.1.MR</dep.jakarta-inject-api.version>
         <dep.javax-inject.version>1</dep.javax-inject.version>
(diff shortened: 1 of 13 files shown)


.github/workflows/versions-check.yml
```diff
@@ -142,7 +142,7 @@ jobs:
 
     strategy:
       matrix:
-        test-version: [ 2.20.1, 2.19.4, 2.17.3, 2.16.2 ]
```
stevenschlansker, line 145: 2.19, 2.17, 2.16 are all EOL

internal/build/pom.xml
```diff
@@ -68,7 +68,8 @@
         <dep.hikari-cp.version>7.0.2</dep.hikari-cp.version>
         <dep.hsqldb.version>2.7.4</dep.hsqldb.version>
         <dep.immutables.version>2.11.6</dep.immutables.version>
-        <dep.jackson2.version>2.18.5</dep.jackson2.version>
+        <dep.jackson2.version>2.20.1</dep.jackson2.version>
```
stevenschlansker: jackson3 depends on jackson2 2.20.1, so upgrading this dep here is the simplest fix for now. 2.18.5 compatibility is still checked via the ci versions step.

We can raise an issue with Jackson if we consider compiling to 2.18 api as essential.

stevenschlansker:
I filed https://github.com/FasterXML/jackson-databind/issues/5454

internal/build/pom.xml
pjfanning: jackson 3 does not depend on jackson 2.20.1
jackson-annotations is no longer on the same versioning scheme as the other jackson jars
jackson 3.0.x depends on jackson-annotations 2.20 - nothing else from jackson 2

jackson 2.21.0 and jackson 3.1.0 will likely depend on jackson-annotations 2.21

.github/workflows/versions-check.yml
hgschmie: Yes. But we still maintain compatibility with them, do we not? The point is that people are using Jdbi with those versions.

.github/workflows/versions-check.yml
```diff
@@ -142,7 +142,7 @@ jobs:
...
+        test-version: [ 2.20.1 ]
```
hgschmie: I am ok with this change but I feel we do people that are stuck on older versions a disservice.

Issue #2651, opened 2024-02-23
[MSSQL] Snapshot transaction isolation level cannot be used with `inTransaction` methods


mjadczak:
The Microsoft SQL Server JDBC driver [defines a custom value](https://github.com/microsoft/mssql-jdbc/blob/cfb018a62ebc0cf4a7fadac6ffaa282467fc44dd/src/main/java/com/microsoft/sqlserver/jdbc/ISQLServerConnection.java#L24) for the [SNAPSHOT transaction level](https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/sql/snapshot-isolation-in-sql-server) which is unique to SQL Server.

Because the methods on JDBI and Handle objects which allow setting an isolation level, like `inTransaction`, require a parameter of type `TransactionIsolationLevel`, which is an enum defined directly by JDBI, this special value cannot be passed in, and we must instead set the default isolation level on the DataSource, or manually manage the handle's lifetime and use the `setTransactionLevel` override which allows specifying an `int`, neither of which are ideal.

Some immediate ideas for solutions are:
- add an override for these methods which takes a raw `int` directly
- change `TransactionIsolationLevel` to admit extension in some way - whether ad-hoc by the user, or by registering some sort of config or extension with the JDBI instance

stevenschlansker:
Thanks for reporting this.

stevenschlansker:
Hi @mjadczak , do you have some time to look at https://github.com/jdbi/jdbi/pull/2684 and confirm if it would help?

mjadczak:
Thank you for looking at this! Just looking at the API-level change, I agree this solves the issue.

I will try and build locally to verify this works sometime this week, but I note that the CI build is failing on the PR, so I'm not sure if I'll be able to?

stevenschlansker:
Yes, I need to sort that out. This is technically a breaking change since the isolation level is no longer an `Enum`. I am hoping that the real-world breakage is small, since I doubt many / anyone has used custom isolation levels before.

@hgschmie , do you have thoughts on this?

stevenschlansker:
Darn, this might be harder to sort out than I thought. In order to be a valid annotation it must remain an Enum type. However, and Enum type can never be extended. So I think the current approach will not work :(

mjadczak:
Is there a particular reason to enforce that the isolation level is strongly typed, when the underlying JDBC call is not?

[...]

PR #2907, opened 2025-12-10
Add test demonstrating DelegatingTransactionHandler bug with @Transaction annotation

hgschmie:
## Issue Description

This PR adds a comprehensive unit test that demonstrates a long-standing bug with how `DelegatingTransactionHandler` and related classes work.

## The Problem

When users create a custom transaction handler by extending `DelegatingTransactionHandler`, they typically override only the 2-arg `inTransaction(Handle, HandleCallback)` method, expecting all transaction logic to flow through it for housekeeping, logging, etc.

However, the `@Transaction` annotation (and `Handle.inTransaction` with isolation level) calls the 3-arg `inTransaction(Handle, TransactionIsolationLevel, HandleCallback)` version. The base `DelegatingTransactionHandler` directly delegates this to the wrapped handler, **bypassing the custom 2-arg override entirely**.

This creates a subtle bug where methods annotated with `@Transaction` don't go through the custom transaction handling logic, even though developers expect all transactions to be intercepted.

## Test Details

The test file `TestDelegatingTransactionHandlerBug.java` contains three test methods:

1. **`testThreeArgInTransactionBypassesCustomTwoArgOverride()`** - **EXPECTED TO FAIL** ✗
   - Demonstrates the bug by showing that when the 3-arg `inTransaction` method is called, the custom logic in the 2-arg override is bypassed
   - Expected: Both calls should increment the counter to 2
   - Actual: Only the 2-arg direct call increments it to 1 (the bug)
   - **This test will pass once the bug is fixed**

2. **`testTwoArgInTransactionCallsCustomLogic()`** - PASSES ✓
   - Shows that direct calls to the 2-arg version work correctly

3. **`testWorkaroundByOverridingBothMethods()`** - PASSES ✓
   - Demonstrates the workaround where users must override BOTH methods

## Expected Behavior

The 3-arg `inTransaction` method in `DelegatingTransactionHandler` should delegate to the 2-arg version (after setting isolation level), similar to how `LocalTransactionHandler.BoundLocalTransactionHandler` implements it.

## Current Workaround

Users must override BOTH methods in their `DelegatingTransactionHandler` subclasses, with the 3-arg version delegating to the 2-arg version after setting the isolation level.

## Test Results

```
Tests run: 3, Failures: 1
✓ testTwoArgInTransactionCallsCustomLogic - PASSED
✓ testWorkaroundByOverridingBothMethods - PASSED  
✗ testThreeArgInTransactionBypassesCustomTwoArgOverride - FAILED (expected - demonstrates bug)
```

The failing test output:
```
org.opentest4j.AssertionFailedError: 
[Both 2-arg and 3-arg inTransaction calls should go through the custom 2-arg override] 
expected: 2
 but was: 1
```

This proves that the 3-arg version bypasses the custom logic.

## Related Code

- `/core/src/main/java/org/jdbi/v3/core/transaction/DelegatingTransactionHandler.java`
- `/sqlobject/src/main/java/org/jdbi/v3/sqlobject/transaction/internal/TransactionDecorator.java` (calls the 3-arg version)
- `/core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java` (BoundLocalTransactionHandler shows the correct pattern)

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/transaction/AbstractDelegatingTransactionHandler.java
+++ b/core/src/main/java/org/jdbi/v3/core/transaction/AbstractDelegatingTransactionHandler.java
@@ -17,17 +17,12 @@ import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.HandleCallback;
 
 /**
- * Simple delegating subclass that just invokes its delegate.
- *
- * @deprecated Should not be used for subclassing outside the main code base. Use {@link AbstractDelegatingTransactionHandler} instead, which ensures that both
- * {@link TransactionHandler#inTransaction(Handle, HandleCallback)} and
- * {@link TransactionHandler#inTransaction(Handle, TransactionIsolationLevel, HandleCallback)} will call custom transaction logic.
+ * Delegating {@link TransactionHandler} subclass intended for extension.
  */
-@Deprecated(forRemoval = true)
-public class DelegatingTransactionHandler implements TransactionHandler {
+public class AbstractDelegatingTransactionHandler implements TransactionHandler {
     private final TransactionHandler delegate;
 
-    public DelegatingTransactionHandler(TransactionHandler delegate) {
+    protected AbstractDelegatingTransactionHandler(TransactionHandler delegate) {
         this.delegate = delegate;
     }
 
@@ -72,14 +67,23 @@ public class DelegatingTransactionHandler implements TransactionHandler {
 
     @Override
     public <R, X extends Exception> R inTransaction(Handle handle,
-                                                    HandleCallback<R, X> callback) throws X {
+        HandleCallback<R, X> callback) throws X {
         return delegate.inTransaction(handle, callback);
     }
 
     @Override
-    public <R, X extends Exception> R inTransaction(Handle handle,
-                                                    TransactionIsolationLevel level,
-                                                    HandleCallback<R, X> callback) throws X {
-        return delegate.inTransaction(handle, level, callback);
+    public final <R, X extends Exception> R inTransaction(Handle handle,
+        TransactionIsolationLevel level,
+        HandleCallback<R, X> callback) throws X {
+
+        final TransactionIsolationLevel initial = handle.getTransactionIsolationLevel();
+        try {
+            handle.setTransactionIsolationLevel(level);
+            return inTransaction(handle, callback);
+        } finally {
+            if (level != TransactionIsolationLevel.UNKNOWN) {
+                handle.setTransactionIsolationLevel(initial);
+            }
+        }
     }
 }
--- a/core/src/test/java/org/jdbi/v3/core/transaction/TestIssue2907.java
+++ b/core/src/test/java/org/jdbi/v3/core/transaction/TestIssue2907.java
@@ -1,0 +1,108 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.transaction;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.HandleCallback;
+import org.jdbi.v3.core.junit5.H2DatabaseExtension;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TestIssue2907 {
+
+    @RegisterExtension
+    public H2DatabaseExtension h2Extension = H2DatabaseExtension.instance();
+
+    @Test
+    public void testCustomHandler() {
+        var transactionHandler = new CustomTransactionHandler(LocalTransactionHandler.binding());
+        h2Extension.getJdbi().setTransactionHandler(transactionHandler);
+
+        try (Handle handle = h2Extension.openHandle()) {
+            // Custom code gets called if two args version is used
+            handle.inTransaction(h -> "result1");
+            assertThat(transactionHandler.getCustomCallCount()).isEqualTo(1);
+
+            // Custom code gets called if three args version is used
+            handle.inTransaction(TransactionIsolationLevel.READ_COMMITTED, h -> "result2");
+
+            assertThat(transactionHandler.getCustomCallCount())
+                .as("Both 2-arg and 3-arg inTransaction calls should go through the custom 2-arg override")
+                .isEqualTo(2);
+        }
+    }
+
+    /**
+     * Ensure that two args method calls custom logic.
+     */
+    @Test
+    public void testTwoArgInTransactionCallsCustomLogic() {
+        var customHandler = new CustomTransactionHandler(LocalTransactionHandler.binding());
+        h2Extension.getJdbi().setTransactionHandler(customHandler);
+
+        try (Handle handle = h2Extension.openHandle()) {
+            // Direct call to 2-arg version works as expected
+            handle.inTransaction(h -> "result");
+            assertThat(customHandler.getCustomCallCount()).isEqualTo(1);
+        }
+    }
+
+    /**
+     * Ensure that three args method calls custom logic.
+     */
+    @Test
+    public void testThreeArgInTransactionCallsCustomLogic() {
+        var customHandler = new CustomTransactionHandler(LocalTransactionHandler.binding());
+        h2Extension.getJdbi().setTransactionHandler(customHandler);
+
+        try (Handle handle = h2Extension.openHandle()) {
+            // Direct call to 3-arg version works as expected
+            handle.inTransaction(TransactionIsolationLevel.READ_COMMITTED, h -> "result");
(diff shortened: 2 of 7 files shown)


hgschmie:
@Randgalt this branch now has a potential fix which make overriding the two-argument method in DelegateTransactionHandler sufficient (it also no longer allows overriding the three argument method, so any "workaround" classes may break. That part can still be changed.

core/src/main/java/org/jdbi/v3/core/transaction/DelegatingTransactionHandler.java
```diff
@@ -67,14 +67,23 @@ public void releaseSavepoint(Handle handle, String name) {
...
+        HandleCallback<R, X> callback) throws X {
         return delegate.inTransaction(handle, callback);
     }
 
     @Override
-    public <R, X extends Exception> R inTransaction(Handle handle,
-                                                    TransactionIsolationLevel level,
-                                                    HandleCallback<R, X> callback) throws X {
-        return delegate.inTransaction(handle, level, callback);
+    public final <R, X extends Exception> R inTransaction(Handle handle,
```
hgschmie: I am somewhat torn making this method final. Not making it final allows people to do actual different implementation for two and three args methods but it also looks like a major footgun if someone were to override *only* the three argument method.

core/src/test/java/org/jdbi/v3/core/transaction/TestDelegatingTransactionHandlerBug.java
```diff
@@ -0,0 +1,133 @@
...
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.HandleCallback;
+import org.jdbi.v3.core.junit5.H2DatabaseExtension;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test that demonstrates the bug where custom DelegatingTransactionHandler subclasses that only override the 2-arg inTransaction method are bypassed when the
```
hgschmie: I will remove most of that AI slop before committing. :-)

core/src/main/java/org/jdbi/v3/core/transaction/DelegatingTransactionHandler.java
stevenschlansker: Isn't this considered a breaking change? If we really want to do this, we can, but I suspect existing custom transaction handler subclasses might not link if we add `final` here

Randgalt: You could deprecate `DelegatingTransactionHandler` and add a new one that does this. Though, this requires everyone to adopt it.

core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
```diff
@@ -259,7 +259,9 @@ public <R, X extends Exception> R inTransaction(Handle handle,
                 handle.setTransactionIsolationLevel(level);
                 return inTransaction(handle, callback);
             } finally {
-                handle.setTransactionIsolationLevel(initial);
+                if (level != TransactionIsolationLevel.UNKNOWN) {
```
Randgalt, line 262: Note: the implementation in `SerializableTransactionRunner` doesn't do this check.

core/src/main/java/org/jdbi/v3/core/transaction/DelegatingTransactionHandler.java
stevenschlansker: Yes, deprecate and replace was more in line with what I'd imagined. Then we can clean up the deprecated way in 4.x

Maybe the new interface would only have one method, with a nullable parameter, to avoid the problem entirely.

stevenschlansker: It's worth noting that reworking the interface here would also give us a chance to fix https://github.com/jdbi/jdbi/issues/2651 at the same time.

core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
hgschmie: This is fixed with this change.

core/src/main/java/org/jdbi/v3/core/Handle.java
```diff
@@ -947,16 +947,20 @@ public int hashCode() {
...
+        SetTransactionIsolation(TransactionIsolationLevel level) {
+            this.level = level;
+            this.initial = getTransactionIsolationLevel();
+            setTransactionIsolationLevel(level);
         }
 
         @Override
         public void close() {
-            setTransactionIsolationLevel(prevLevel);
+            if (this.level != TransactionIsolationLevel.UNKNOWN) {
```
stevenschlansker, line 961: I don't think I understand this change. Then again, I'm not really sure how `UNKNOWN` behaves at all in the first place.

core/src/main/java/org/jdbi/v3/core/transaction/AbstractDelegatingTransactionHandler.java
```diff
@@ -0,0 +1,89 @@
...
+ */
+package org.jdbi.v3.core.transaction;
+
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.HandleCallback;
+
+/**
+ * Delegating {@link TransactionHandler} subclass intended for extension.
+ */
+public class AbstractDelegatingTransactionHandler implements TransactionHandler {
```
stevenschlansker, line 22: It's a bit inconsistent to have `Abstract` in the name but the class is not actually declared `abstract`.

core/src/main/java/org/jdbi/v3/core/transaction/AbstractDelegatingTransactionHandler.java
```diff
@@ -0,0 +1,89 @@
...
+    private final TransactionHandler delegate;
+
+    protected AbstractDelegatingTransactionHandler(TransactionHandler delegate) {
+        this.delegate = delegate;
+    }
+
+    protected TransactionHandler getDelegate() {
```
stevenschlansker, line 29: I'm somewhat a fan of dropping the `get` prefix and just using the property name `delegate()` but this is of course fine as-is.

core/src/main/java/org/jdbi/v3/core/transaction/AbstractDelegatingTransactionHandler.java
```diff
@@ -0,0 +1,89 @@
...
+        return delegate;
+    }
+
+    @Override
+    public void begin(Handle handle) {
+        delegate.begin(handle);
```
stevenschlansker, line 35: The `getDelegate()` method is virtual, should all these call `getDelegate().begin(handle)` for consistency, or declare `getDelegate()` as `final`?

core/src/main/java/org/jdbi/v3/core/transaction/AbstractDelegatingTransactionHandler.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public <R, X extends Exception> R inTransaction(Handle handle,
+        HandleCallback<R, X> callback) throws X {
+        return delegate.inTransaction(handle, callback);
+    }
+
+    @Override
+    public final <R, X extends Exception> R inTransaction(Handle handle,
```
stevenschlansker, line 75: So, this fixes the problem as reported, but introduces a different pitfall.

If an implementation has custom behavior for the three-arg method that uses the `level` argument, then that behavior is totally hidden when you wrap in this new `AbstractDelegatingTransactionHandler`, since the wrapper will by default redirect the three-arg method to the two-arg method. Then the inner impl never gets any chance to inspect the `level` argument for its custom behavior.

Maybe in practice this pitfall is okay, and having a less-sharp base class is worth the tradeoff.
But I also think this is a potential reason to prefer consolidating these two methods, and deprecating-for-removal the two-arg method entirely, and instead always use the three-arg method with a `null` isolation level instead of the two-arg form.

core/src/test/java/org/jdbi/v3/core/transaction/TestIssue2907.java
```diff
@@ -0,0 +1,108 @@
...
+        public <R, X extends Exception> R inTransaction(Handle handle,
+            HandleCallback<R, X> callback) throws X {
+            // Custom housekeeping code that users expect to run on every transaction
+            customCallCount.incrementAndGet();
+            return super.inTransaction(handle, callback);
+        }
+
+        // NOTE: We intentionally DO NOT override the 3-arg version, expecting that
+        // it will delegate to the 2-arg version as it does in LocalTransactionHandler.
+        // This is the bug - when @Transaction annotation is used, it calls the 3-arg version
```
stevenschlansker, line 101: I don't know if this is actually a bug. Per my comments above, this expectation breaks the ability of the delegate to implement the three-arg method, since the AbstractDelegating base class implements the isolation level logic and rewrites the call away from the three-arg method.

core/src/main/java/org/jdbi/v3/core/transaction/AbstractDelegatingTransactionHandler.java
hgschmie: Correct. Will fix.

core/src/main/java/org/jdbi/v3/core/Handle.java
hgschmie: when calling setTransactionIsolationLevel, it does a check whether the level is "UNKNOWN" and if yes, it does not change the level. 

The current code has an actual bug that in this case, the transaction level is reset but was never set in the first place. This fixes it. I can pull this to a separate commit / PR.

core/src/main/java/org/jdbi/v3/core/transaction/AbstractDelegatingTransactionHandler.java
hgschmie: fair point.

PR #2908, opened 2025-12-10
post release fixes

hgschmie:
- **Various fixes, found during the 3.51 release**
- **change deprecated systemProperty config entries**

### DIFF
--- a/cache/caffeine-cache/src/it/test-caffeine-cache/pom.xml
+++ b/cache/caffeine-cache/src/it/test-caffeine-cache/pom.xml
@@ -125,9 +125,9 @@
                     <dependenciesToScan>
                         <dependency>org.jdbi:jdbi3-core:*:tests</dependency>
                     </dependenciesToScan>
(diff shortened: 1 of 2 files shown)


native-tests/pom.xml
```diff
@@ -88,6 +71,12 @@
             <groupId>org.jdbi</groupId>
             <artifactId>jdbi3-guice</artifactId>
             <scope>test</scope>
+            <exclusions>
```
stevenschlansker, line 74: Hm. When I tried this before, I got an error that exclusions must be removed from managed dependencies. Why is this now valid for you?!

stevenschlansker (approved):

native-tests/pom.xml
stevenschlansker: Maybe now we skip the build check that catches this.

hgschmie: I turned off the checker in line 31. :-)

PR #2912, opened 2025-12-12
Fix name of parameter in Spring Annotation Example.

michael-simons:


michael-simons:
Thank you folks for a very nice library with great docs :)

PR #2922, opened 2026-02-11
Bump basepom to 68

hgschmie:
- remove no longer needed PMD.UnnecessaryCast suppressions
- turn off "PublicMemberInNonPublicType" warnings, changes here may affect our public API


stevenschlansker:
Oh! the page only partly loaded so I wasn't seeing all changes - refreshing fixed it. Looks good.

PR #2924, opened 2026-02-14
Fix for Issue-1003 [Add an Oracle plugin for untyped nulls]

### LINKED ISSUE #1003: Add an Oracle plugin for untyped nulls
Per SO question: https://stackoverflow.com/questions/48254280/why-does-jdbi-bind-fail-with-function-as-parameter/48255296#48255296

Oracle does not allow `Types.OTHER` for untyped nulls, but it does allow `Types.NULL` (even though it's not a data type constant). We should add an Oracle plugin that calls `Arguments.setUntypedNullArgument(new NullArgument(Types.NULL))`.

hgschmie:
## Summary
- Adds `OraclePlugin` class in the `oracle12` module that configures `Arguments.setUntypedNullArgument` to use `new NullArgument(Types.NULL)` instead of the default `Types.OTHER`
- Oracle does not allow `Types.OTHER` for untyped null values but accepts `Types.NULL`
- Follows the pattern of other database plugins (H2DatabasePlugin, SQLitePlugin, etc.) by extending `JdbiPlugin.Singleton`
- Includes unit tests verifying the configuration

Fixes #1003

## Test plan
- [ ] Unit test verifies that installing OraclePlugin changes the untyped null argument from Types.OTHER to Types.NULL
- [ ] Unit test verifies singleton equality behavior
- [ ] Verify no regressions in oracle12 module tests

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/oracle12/src/test/java/org/jdbi/v3/oracle12/TestOraclePlugin.java
+++ b/oracle12/src/test/java/org/jdbi/v3/oracle12/TestOraclePlugin.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.oracle12;
+
+import java.sql.Types;
+
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.argument.Arguments;
+import org.jdbi.v3.core.argument.NullArgument;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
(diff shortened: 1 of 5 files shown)


oracle12/src/test/java/org/jdbi/v3/oracle12/TestOraclePlugin.java
```diff
@@ -0,0 +1,51 @@
...
+
+        // before installing the plugin, the default is Types.OTHER
+        NullArgument defaultNull = (NullArgument) jdbi.getConfig(Arguments.class).getUntypedNullArgument();
+        assertThat(defaultNull.getSqlType()).isEqualTo(Types.OTHER);
+
+        jdbi.installPlugin(new OraclePlugin());
+
+        // after installing the plugin, the untyped null uses Types.NULL
+        NullArgument oracleNull = (NullArgument) jdbi.getConfig(Arguments.class).getUntypedNullArgument();
+        assertThat(oracleNull.getSqlType()).isEqualTo(Types.NULL);
```
stevenschlansker, line 41: While this tests that the oracle plugin behaves as requested, it does not seem to actually verify the assertion that the type must be `Types.NULL`. I assume this assertion is correct?

stevenschlansker (approved):

oracle12/src/test/java/org/jdbi/v3/oracle12/TestOraclePlugin.java
hgschmie: I will ask it to write an integration test with the oracle container.

Issue #1436, opened 2019-01-25
Document how to create custom qualifying annotations in developer guide


qualidafial:
- Defining the qualifying annotation (`@Qualifier`, `@Target`, `@Retention(RUNTIME)`)
- Implement QualifiedColumnMapperFactory or QualifiedArgumentFactory to create qualifier-agnostic composable factories. e.g. a QualifiedColumnMapperFactory for arbitrary `Optional<T>`, when asked for a mapper of type `@Foo Optional<Bar>`, would extract the `Bar` type parameter, and look up a column column mapper for type `@Foo Bar` and delegate to that.

PR #2931, opened 2026-02-16
Document how to create custom qualifying annotations

hgschmie:
## Summary
- Adds a "Creating custom qualifiers" subsection to the Qualified Types section
- Shows the three steps: define the `@Qualifier` annotation, implement qualified `ArgumentFactory` and `ColumnMapper`, register and use with `QualifiedType`
- Uses a `@Reversed` string example consistent with existing test code in `TestCustomQualifier`

Addresses #1436.

## Test plan
- [x] `make install-fast` passes (AsciiDoc compiles to `docs/target/generated-docs/index.html`)
- [ ] Review rendered HTML for correct formatting

[Generated by AI]

### DIFF
--- a/docs/src/adoc/index.adoc
+++ b/docs/src/adoc/index.adoc
@@ -27,6 +27,7 @@
 :coreexampledir: ../../../core/src/test/java/org/jdbi/v3/core/mapper
 :guavaexampledir: ../../../guava/src/test/java/org/jdbi/v3/guava
 :sqlobjectexampledir: ../../../sqlobject/src/test/java/org/jdbi/v3/sqlobject
+:coreexamples: ../../../core/src/test/java
 
 [preface]
 image::logo.svg[Jdbi logo]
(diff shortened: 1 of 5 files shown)


docs/src/adoc/index.adoc
```diff
@@ -5756,6 +5756,73 @@ link:{jdbidocs}/core/statement/SqlStatement.html#bindBean(java.lang.Object)[bind
...
+public @interface Reversed {}
+----
+
+To handle qualified values, create an link:{jdbidocs}/core/argument/ArgumentFactory.html[ArgumentFactory^] and/or a link:{jdbidocs}/core/mapper/ColumnMapper.html[ColumnMapper^] annotated with the qualifier.
+Jdbi matches factories to values by their qualifier annotations:
+
+[source,java,indent=0]
+----
+@Reversed // <1>
+public class ReversedStringArgumentFactory extends AbstractArgumentFactory<String> {
```
stevenschlansker: Can we refer to the test code we already have, so we know it compiles and runs?

(Can you set your AI to always write external tests?)

stevenschlansker (approved):

docs/src/adoc/index.adoc
hgschmie: I literally asked it to pull an example from the code base. We can not include code from outside the doc module in the documentation so it is either duplicating the code (and run tests that we know already run in another place) or simply copy it as an example into the codebase. I can totally ask claude to write an external test.

docs/src/adoc/index.adoc
```diff
@@ -27,6 +27,7 @@
 :coreexampledir: ../../../core/src/test/java/org/jdbi/v3/core/mapper
 :guavaexampledir: ../../../guava/src/test/java/org/jdbi/v3/guava
 :sqlobjectexampledir: ../../../sqlobject/src/test/java/org/jdbi/v3/sqlobject
+:coreexamples: ../../../core/src/test/java
```
hgschmie, line 30: I asked the agent to introduce a new variable. I want to replace coreexampledir with this, once the PR went in. This makes the references more versatile and we don't need a variable for each package.

hgschmie:
@stevenschlansker does that look better for you? I asked the agent to include the existing code and write a new test.

PR #2934, opened 2026-02-17
Add INOUT parameter support for Call statements

### LINKED ISSUE #1606: Does jdbi support in/out parameter in procedure(for Oracle)?
I have a procedure in Oracle like this:

```sql
procedure foo(v1 in number, v2 in char, v3 in out number );
```
and my java code is:

```java
handle.bind("v1", 42)
           .bind("v2", "Y")
           .bind("v3", 99)
           .registerOutParameter("v3", Types.INTEGER)
```
But when I get result for "v3", the value is always 0, and I follow the source code(not very carefully), I think jdbi will overwrite the parameter, when `bind` and `registerOutParameter` use the same name.

So is there any solution for invoking a procedure, when procedure has a paremeter is both in and out ?

hgschmie:
## Summary

- When `registerOutParameter()` is called for a parameter that was previously bound with `bind()`, the existing input value is captured and applied to the `CallableStatement` after registering the out parameter
- This enables stored procedure INOUT parameters which require both `setXxx()` and `registerOutParameter()` on the same position
- Constraint: `bind()` must be called before `registerOutParameter()` for INOUT parameters

Fixes #1606

## Test plan

- [x] Oracle integration tests with INOUT parameters (named, positional, mixed IN+INOUT, null)
- [x] PostgreSQL integration tests with INOUT parameters (named, positional, mixed, null)
- [x] Existing `TestCallable` tests pass (no regressions)
- [x] Existing `TestOutparameterCursor` tests pass (no regressions)

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -5,6 +5,7 @@
 - Retire Apache Derby integration (#2866)
 - Run test suite against Spring Framework 7 (#2919)
 - Add OraclePlugin that sets untyped null argument to Types.NULL for Oracle compatibility (#1003)
+- Support INOUT parameters for stored procedure Call statements (#1606)
 
 # 3.51.0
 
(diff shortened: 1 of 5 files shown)


hgschmie:
We have never supported IN OUT parameters, so this change is actually pretty clever and it seems to work for oracle and postgres.

stevenschlansker (approved):

PR #2937, opened 2026-02-17
jdbi4: Remove JUnit 4 and JUnit Vintage runner

stevenschlansker:


docs/src/adoc/index.adoc
```diff
@@ -218,11 +218,11 @@ SQL Object extension for declarative database access.
 [glossary]
 jdbi3-testing::
 Testing framework support.
-Currently, only supports JUnit4 and JUnit5.
+Currently, only supports JUnit6.
 
 jdbi3-testcontainers::
 Support for arbitrary databases running in https://testcontainers.org/[Testcontainers^].
-Currently, only supports JUnit 5.
+Currently, only supports JUnit 6.
```
hgschmie: Different spelling than above. :-)

docs/src/adoc/index.adoc
```diff
@@ -5945,43 +5945,12 @@ If the underlying cache library exposes per-cache statistics, these can be acces
 
 [NOTE]
 The official test support from Jdbi is in the `jdbi-testing` package.
-There are a number of additional JUnit 5 extensions in the `jdbi-core` test artifact.
+There are a number of additional JUnit 6 extensions in the `jdbi-core` test artifact.
```
hgschmie: I wonder if we should just say "modern JUnit (5 or better); I am pretty sure that junit 5 still works if you use a recent JDK.

docs/src/adoc/index.adoc
```diff
@@ -5945,43 +5945,12 @@ If the underlying cache library exposes per-cache statistics, these can be acces
...
-    <artifactId>otj-pg-embedded</artifactId>
-    <version>1.0.2</version>
-    <scope>test</scope>
-</dependency>
-----
-
-=== JUnit 5
-
-The `jdbi3-testing` artifact provides link:{jdbidocs}/testing/junit5/JdbiExtension.html[JdbiExtension^] for JUnit 5 based tests.
+The `jdbi3-testing` artifact provides link:{jdbidocs}/testing/junit/JdbiExtension.html[JdbiExtension^] for JUnit 5 based tests.
```
hgschmie: 6. :-)

docs/src/adoc/index.adoc
```diff
@@ -5945,43 +5945,12 @@ If the underlying cache library exposes per-cache statistics, these can be acces
...
 
 It supports both the https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/RegisterExtension.html[@RegisterExtension^] and https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/extension/ExtendWith.html[@ExtendWith^] annotations.
```
hgschmie, line 5955: the links still point at junit 5

hgschmie:
those: https://sonarcloud.io/project/issues?id=jdbi_jdbi&pullRequest=2937&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true might have popped up again after we silenced them but the unused import at least should be addressed.

docs/src/adoc/index.adoc
stevenschlansker: I will start deleting all of these references to specific releases I think... it is so annoying to update them, and there is no check that you have caught them all.

docs/src/adoc/index.adoc
stevenschlansker: I left us as supporting 5 for now, so will leave docs to point at 5 as well.

oracle12/src/main/java/org/jdbi/oracle12/OracleDatabaseExtension.java
```diff
@@ -20,41 +20,46 @@
 
 import oracle.jdbc.pool.OracleDataSource;
 import org.jdbi.core.Jdbi;
-import org.jdbi.testing.JdbiRule;
+import org.jdbi.testing.junit.JdbiExtension;
 import org.testcontainers.containers.OracleContainer;
 
 /**
  * JUnit 4 Helper for a database managed by testcontainers.
```
hgschmie: This is porting this to "modern" JUnit, so the comment should be changed.

Do we need to port this? I don't think this is used anywhere. Just delete it.

stevenschlansker: I thought this was provided to end users as a convenient way to stand up a testing Oracle instance, not only for Jdbi itself. Are you sure you want to just delete it? As much as I detest Oracle, that seems a bit rash? Let's do that in a followup PR at least.

stevenschlansker: Hm, I guess it does have our weird "something" table baked in, so maybe this isn't actually useful to anyone as is.

PR #2938, opened 2026-02-18
jdbi4: Remove auto-install plugins feature, finally

stevenschlansker:


hgschmie:
I don't have any particularly strong feelings about that feature but I would also be ok with just removing the test and plugin integration and leave the config files in place and offer users the ability to either write code that uses the service loader or some utility method that they can call in place.

docs/src/adoc/index.adoc
```diff
@@ -9531,15 +9531,9 @@ E.g. RowMappers registry delegates to ColumnMappers registry, if it does not hav
 JdbiPlugin can be used to bundle bulk configuration.
 Plugins should be installed via `Jdbi.installPlugin(JdbiPlugin)`.
 
-Jars may provide a file in `META-INF/services/org.jdbi.core.spi.JdbiPlugin`
-containing the fully qualified class name of your plugin.
-
 In general, Jdbi's separate artifacts each provide a single relevant plugin (e.g. `jdbi3-sqlite`), and such modules will be auto-loadable.
```
hgschmie: I think this paragraph needs to go as well or needs to be rewritten.

PR #2940, opened 2026-02-18
jdbi4: jdk21 baseline

stevenschlansker:
raise jdk baseline to 21 for jdbi4


hgschmie:
not sure if kotlin is ready for java 25.

stevenschlansker:
Yeah, Kotlin is not ready for 25. So I backed off to 21 and now I think this is good to go.

e2e/src/test/java/org/jdbi/e2e/vt/TestVirtualThreads.java
```diff
@@ -11,7 +11,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.jdbi.java21;
+package org.jdbi.e2e.vt;
```
hgschmie: .virtualthreads. ? `.vt.` is vry shrthnd.

README.md
```diff
@@ -24,18 +24,11 @@ Also check out the code examples in the [Examples](https://github.com/jdbi/jdbi/
 
 ## Prerequisites
 
-Jdbi requires Java 17 or better to run.
-
-We run CI tests against Java 17, 21, and 25.
```
hgschmie, line 29: Why lose that line? We can still run with Java 21 and Java 25 (we can not yet compile to 25 but we can run on it).

hgschmie (approved):

README.md
stevenschlansker: Because it's already in the docs and updating it twice is a pain. I can put it back.

PR #2941, opened 2026-02-19
Add option to not export bindings with OpenTelemetry

gmellemstrand:
For our system the data in bindings is sensitive and should not be shown in telemetry data. The Java OpenTelemetry library does not seem to include any option to filter out attributes.

This PR lets you change a setting in SqlStatements.java in order to not add the binding attributes. I have also added a test to verify that no binding data is being exported.


hgschmie:
LGTM, @gmellemstrand if you could add a line to the RELEASE_NOTES.md file, that would be great.

Issue #2445, opened 2023-07-19
Multiple Constructor Mappers of same type


JoshDoug:
Is it not possible to have multiple ConstructorMappers registered with the same type, even if they have different prefixes.

```Java
@RegisterConstructorMapper(value = Example.class)
@RegisterConstructorMapper(value = ExampleSub.class, prefix = "first")
@RegisterConstructorMapper(value = ExampleSub.class, prefix = "second")
@SqlQuery("""
                ...
            """)
@UseRowReducer(ExampleReducer.class)
Example findById(@Bind("id") int id);
```

The reducer the RowView only has the second constructor mapper of the ExampleSub class with the prefix second.

Is this intended behaviour or a bug?

stevenschlansker:
I don't think this is intentionally not supported, but neither has it been tested to work. We'd love to add this feature.

PR #2942, opened 2026-02-21
Multiple prefixed mappers for a single type

phinjensen:
In any case where a table has two foreign key references to the same table and you have a query that joins with both, it's not possible to use a RowMapper for both of those objects because, even if you register a mapper twice with different prefixes, the `rowView.getObject(Class<?>)` method will just use the last registered mapper for the type. #2445 is a ticket someone opened for the same problem, and this PR allows users to access row mappers by type _and_ prefix, making this possible.

The basic idea is to make sure mapper prefixes are accessible and even part of the mapper cache key, and add new `RowView` overload methods that allow `String prefix` as a second argument; if that argument is present, it will only return use a mapper that was created with that prefix.

I've provided a single test in `TestBeanMapper` that gives a good idea of what this looks like in a RowReducer. I realize more tests are probably necessary, but I just wanted to create the pull request and get some feedback on implementation before I go write more tests. How does this look? Is my implementation good, or should I make some changes?

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/internal/PrefixedMapperKey.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/PrefixedMapperKey.java
@@ -1,0 +1,19 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.internal;
+
+import java.lang.reflect.Type;
+
+public record PrefixedMapperKey(Type type, String prefix) {
+}
--- a/core/src/main/java/org/jdbi/v3/core/internal/PrefixedMapperKey.java
+++ b/core/src/main/java/org/jdbi/v3/core/internal/PrefixedMapperKey.java
@@ -15,5 +15,4 @@ package org.jdbi.v3.core.internal;
 
 import java.lang.reflect.Type;
 
-public record PrefixedMapperKey(Type type, String prefix) {
-}
+public record PrefixedMapperKey(Type type, String prefix) {}
--- a/core/src/main/java/org/jdbi/v3/core/mapper/PrefixedRowMapper.java
+++ b/core/src/main/java/org/jdbi/v3/core/mapper/PrefixedRowMapper.java
@@ -1,0 +1,18 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.mapper;
+
+public interface PrefixedRowMapper<T> extends RowMapper<T> {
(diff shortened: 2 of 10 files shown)


hgschmie:
Hi,

Thank you for submitting a PR. I will take a look over the weekend.

hgschmie:
First, thank you again for submitting a PR. This addresses a real problem and we should be able to handle this case.

The PR adds a lot of methods to some very core pieces of Jdbi code, namely the RowMapper and the related factories. At this point, there is a single, indirect test that would exercise that functionality. At the very least, we will need to bring the test suite up to par with the regular findRowMapper tests to ensure that adding these new methods will extend functionality to the factory (and I can see a number of use cases).

Prefixing a collection of columns into a single object (which is what BeanMapper, ConstructorMapper, FieldMapper do) is a valuable effort. The "generic prefixing" is best kept in those classes. I really want to avoid penalizing all our row mapper code (which is some of the hottest paths in all of Jdbi) by adding prefixes that are used only in a few use cases.

I will need to think about this problem some more. This change will most likely not make it into the next Jdbi release, but I will address this in the release after that.

phinjensen:
Hi @hgschmie, sorry for the late reply. I definitely agree that there need to be many more tests, and I'll try to find time to write some soon. I didn't want to spend too much time on them if the whole implementation was fundamentally flawed, or if you thought the API should change drastically.

> Prefixing a collection of columns into a single object (which is what BeanMapper, ConstructorMapper, FieldMapper do) is a valuable effort. The "generic prefixing" is best kept in those classes. I really want to avoid penalizing all our row mapper code (which is some of the hottest paths in all of Jdbi) by adding prefixes that are used only in a few use cases.

I tried to make sure that performance wouldn't be impacted for calls using the existing APIs. What part of the code do you feel would penalize the row mapper code that doesn't use prefixes? The most significant part I can think of is where it creates the new PrefixedMapperKey type for the cache, which would use a few extra bytes for the cache key, but I'm not sure I understand how that would affect performance.

There's probably something I'm missing, and I'd love to get the PR up to snuff, so if you could help me understand that better, I'd greatly appreciate it.

MikeNeilson:
> Prefixing a collection of columns into a single object (which is what BeanMapper, ConstructorMapper, FieldMapper do) is a valuable effort. The "generic prefixing" is best kept in those classes. I really want to avoid penalizing all our row mapper code (which is some of the hottest paths in all of Jdbi) by adding prefixes that are used only in a few use cases.

Given that most of the prefixes would be null, I would suspect performance impact on the lookup would be rather minimal, if really anything. Doesn't java `record` automatically implement a hashCode and equals that Map would use to actually do the look up?
There's a slight overhead of creating the key, which could add up.
(Though certain does make sense to do some measurements and show how big of an impact there actually is.)

Perhaps a comprise though for those of us who thing the performance penalty (if any) is worth it: Create separate `PrefixRowView` and `...PrefixRowReducer` interfaces?  I'll grant not ideal either, as it adds some development overhead that needs to be kept in sync. And definitely would only make sense to do *if* there was a large enough performance difference.

Another compromise, though it would be a breaking change for everyone, would be for RowView to take the PrefixMapperKey instead of just the type. Users could then create the key in the Reducer Once and the the hash lookup would likely be identical to just the class, or at least barely measurable. Or at least add it as an option that people could switch to over time. or to reduce memory allocations maybe PrefixMapperKey (or the RowViewImpl) could have a static method to create them and have it's own hash map of keys. Still has to do the lookup, but that should be cheaper than a `new` every time.

I'm currently working on a project where being able to share the use the same row mapper twice would be helpful. For the curious: https://github.com/opendcs/opendcs/pull/2112) (needs to be used in two places, the Data Type itself and a Data Type that references it).
My current alternative is to just use the Accumulator so I have access to the Context and ResultSet  instance and can just pass the mappers in the constructor. While that works, the RowReducer interface design is just seems cleaner all around and easier to maintain. I will likely have to go the accumulator route given the timing of my current work, but it would be nice to improve it later with such a feature.


Sorry, that was rather long. This is important enough I'm also willing to help, though baring any measured issues with performance, I would argue that @phinjensen 's implementation is quite reasonable.

stevenschlansker:
Thanks @phinjensen for starting this work, and @hgschmie and @MikeNeilson for the review and feedback.
I took the review feedback and performance question, and had my bot build a regression test, and indeed this approach currently introduces a small regression on the hot path for an uncommon case. And, agree the API surface does not need to be so large. I opened a followup PR, #3015, that tries to build on this approach - preserve the hot path, api cleanup and null prefix semantics clarification, and add the missing tests.

stevenschlansker:
Merged as #3015

PR #2947, opened 2026-03-05
Add new SqlExceptionHandler to allow mapping database exceptions

### LINKED ISSUE #566: Exception Handling Improvements?
Hi,

I am a fond user of JDBI, but a pain point is in the way that all database exceptions are wrapped into a generic `UnableToExecuteStatementException` and then the underlying exception cause has to be inspected, usually to determine the underlying database specific error that was thrown and then mapping that to an appropriate client-facing response in a REST service.

For example, one might have a `UNIQUE` constraint on a `Person` object:

```
    @POST
    public Person addPerson(@NotNull @Valid Person newPerson) {
        try {
            Person p = personDao.create(newPerson);
            return p;
        } catch (Exception e) {
            // hmm, well this is awkward...
            if (e.getCause() instanceof PSQLException) {
                if (((PSQLException)e.getCause()).getSQLState().equals("23505")) {
                    // return an HTTP 409
                    throw new WebApplicationException("Duplicate person object.", CONFLICT);
                }
            }
            throw new WebApplicationException(e.getMessage(), INTERNAL_SERVER_ERROR);
        }
    }
}
```

There was an old thread on the google groups on this topic, here:
https://groups.google.com/forum/#!topic/jdbi/L_OEd3mCWc8

Essentially, the endeavor would be to map different common SQL implementations' exceptions to normalized exceptions, e.g. `DuplicateKeyException`. I recognize that there is some overhead to maintaining such a mapping, but there are many other database access libraries that do just that.

A less onerous approach would be to just provide the normalized exception classes and provide APIs to plug in a companion library to provide such functionality for a particular database tech (e.g. MySQL, Postgresql, etc.).

Is anything like this being considered or is already in the works?

stevenschlansker:
Fixes #566

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlExceptionHandler.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlExceptionHandler.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.core.statement;
+
+import java.sql.SQLException;
+
+import jakarta.annotation.Nullable;
+
+import org.jdbi.v3.meta.Beta;
+
+/**
+ * Handler for exceptions thrown while executing SQL statements.
+ */
+@Beta
+@FunctionalInterface
+public interface SqlExceptionHandler {
+    /**
+     * Take action based on a SQLException thrown during statement execution.
+     * If you return a replacement exception, before throwing it Jdbi will automatically
+     * attach the original exception as "suppressed", so don't do that.
+     * @param ex the exception thrown
+     * @return a replacement exception to throw, or null to defer
+     */
+    @Nullable
+    Throwable handle(SQLException ex);
+}
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
@@ -1786,6 +1786,7 @@ public abstract class SqlStatement<This extends SqlStatement<This>> extends Base
         return String.format("%s[sql=%s, bindings=%s]", getClass().getSimpleName(), sql, getContext().getBinding());
     }
 
+    @SuppressWarnings("PMD.ExceptionAsFlowControl")
     PreparedStatement internalExecute() {
         final StatementContext ctx = getContext();
         final OptionalEvent evt = JfrSupport.newStatementEvent();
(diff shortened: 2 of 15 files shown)


core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -470,6 +486,19 @@ String preparedRender(String template, StatementContext ctx) {
         }
     }
 
+    UnableToExecuteStatementException handleException(SQLException e, StatementContext ctx) {
+        var handlerIter = exceptionHandlers.descendingIterator();
+        while (handlerIter.hasNext()) {
+            SqlExceptionHandler handler = handlerIter.next();
+            Throwable rewritten = handler.handle(e);
+            if (rewritten != null) {
+                rewritten.addSuppressed(e);
```
hgschmie: that part makes no sense to me. If the exception (e) gets rewritten into the "rewritten" Throwable, why add it as a suppressed exception. That should be the job of the "handler.handle(e)" method. It can decide whether it wants to attach the original exception or not. But unconditionally doubling the exception is not the right thing. Consider

```java
var foo = new SqlExceptionHandler() {
  @Override
  public Throwable handle(SQLException e) {
    return e;
  }
}
```

sqlobject/src/main/java/org/jdbi/v3/sqlobject/config/internal/RegisterSqlExceptionHandlersImpl.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.jdbi.v3.core.config.ConfigRegistry;
+import org.jdbi.v3.core.extension.SimpleExtensionConfigurer;
+import org.jdbi.v3.core.internal.JdbiClassUtils;
+import org.jdbi.v3.core.statement.SqlExceptionHandler;
+import org.jdbi.v3.core.statement.SqlStatements;
+import org.jdbi.v3.sqlobject.config.RegisterSqlExceptionHandler;
+import org.jdbi.v3.sqlobject.config.RegisterSqlExceptionHandlers;
+
+public class RegisterSqlExceptionHandlersImpl extends SimpleExtensionConfigurer {
```
hgschmie: as you make every variable under the sun final, can we make these classes final as well?

core/src/test/java/org/jdbi/v3/core/HandleExceptionTest.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @AfterEach
+    public void tearDown() {
+        h.close();
+    }
+
+    @Test
+    void handleException() {
+        h.getConfig(SqlStatements.class).addExceptionHandler(new SqlExceptionHandler() {
```
hgschmie: maybe have a test with multiple handlers like the one below?

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -470,6 +486,19 @@ String preparedRender(String template, StatementContext ctx) {
...
+                throw Sneaky.throwAnyway(rewritten);
+            }
+        }
+        throw new UnableToExecuteStatementException(e, ctx);
```
hgschmie, line 501: That matches the old behavior, right?

hgschmie:
There are 9 places in the code where an UnableToExecuteStatementException is thrown as consequence of a SQLException (and two where there is no underlying exception). You cover three. A change like this should cover all the places.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
stevenschlansker: Correct, there is no behavior change, unless you opt in with a custom handler. None are registered by default.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
```diff
@@ -1795,35 +1796,45 @@ PreparedStatement internalExecute() {
...
+            } catch (SQLException e) {
+                throw stmtConfig.handleException(e, ctx);
+            }
 
-        afterExecution();
+            afterExecution();
 
-        return stmt;
+            return stmt;
+        } catch (Exception e) {
```
stevenschlansker, line 1830: Without closing the statement here, the leak checker fails the unit test for `Call.invoke()` throwing an exception.

core/src/main/java/org/jdbi/v3/core/statement/SqlExceptionHandler.java
```diff
@@ -0,0 +1,36 @@
...
+ * Handler for exceptions thrown while executing SQL statements.
+ */
+@Beta
+@FunctionalInterface
+public interface SqlExceptionHandler {
+    /**
+     * Take action based on a SQLException thrown during statement execution.
+     * Jdbi will automatically attach the original exception as "suppressed" so don't do that.
+     * @param ex the exception thrown
+     * @return a replacement exception to throw, or null to defer
```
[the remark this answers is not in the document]
hgschmie: Copilot is right, the comment is no longer correct.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatement.java
hgschmie: totally possible. Amazing that this was not reported / discovered earlier.

core/src/test/java/org/jdbi/v3/core/junit5/JdbiLeakChecker.java
```diff
@@ -147,17 +149,17 @@ public void checkForLeaks() {
...
             Set<Cleanable> leakedCleanables = context.leakedElements();
             if (!leakedCleanables.isEmpty()) {
-                leakedCleanablesCount += leakedCleanables.size();
+                leakedCleanableList.addAll(leakedCleanables);
             }
         }
 
-        if (leakedCleanablesCount > 0) {
-            fail(format("Found %d leaked cleanable objects in %d contexts", leakedCleanablesCount, contextElements.size()));
+        if (!leakedCleanableList.isEmpty()) {
```
hgschmie: I assume this is some debug code to find the call leak. I seem to remember that most of the cleanables don't have a toString method, so the value may be limited. But I am fine with it if it helps pinpointing problems.

hgschmie (approved):
Ship it, but please fix the comment for SqlExceptionHandler

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -57,6 +60,7 @@ public final class SqlStatements implements JdbiConfig<SqlStatements> {
     private volatile boolean attachCallbackStatementsForCleanup = true;
     private volatile boolean scriptStatementsNeedSemicolon = true;
     private final Collection<StatementCustomizer> customizers;
+    private final Deque<SqlExceptionHandler> exceptionHandlers;
```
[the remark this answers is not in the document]
hgschmie: I mean, it does have a point here...

core/src/test/java/org/jdbi/v3/core/junit5/JdbiLeakChecker.java
stevenschlansker: Yeah, I was trying to get a list of what actually leaked, but since it's a lambda it's not very helpful after all. So I'll remove this.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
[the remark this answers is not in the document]
stevenschlansker: Well, the whole config thing is not supposed to be concurrently edited. And despite any attempts to provide thread safety here, we still do not support re-configuring a Handle et al from a different thread. This is part of why I want to rip the whole current config paradigm out and replace it.

xak2000:
@stevenschlansker I'm not sure if github sends notifications about comments that are left directly to commit's code. Please pay attention to my comment above.

My 2 main points:

1. The javadocs to `SqlExceptionHandler` is very confusing to say the least. (sorry :D) I would try to improve them myself but I have so many questions about how this method actually works and what is the contract/api of this method. So I don't want to describe the behavior if I'm not sure my description is correct.
2. Uncoditionally adding original `SQLException` as suppressed exception is not flexible enough way of doing this. With this implementation the original problem described in my comment https://github.com/jdbi/jdbi/issues/566#issuecomment-1407005408 is not solved as stack traces will be a mess. But I believe this can be improved in a new commit if we do this using default interface method (this will keep the interface byte code compatible, I think).

stevenschlansker:
Sorry this didn't immediately fit your needs. Let's follow up on the original issue, which I'm happy to reopen.
This PR does work for us, but maybe our needs aren't the same as yours.

PR #2960, opened 2026-04-03
add support for ZoneOffset

hgschmie:


core/src/test/java/org/jdbi/v3/core/TestJavaTimeZones.java
```diff
@@ -0,0 +1,115 @@
...
+    static final ZoneId TEST_ZONE_ID;
+
+    static {
+        var testZoneOffset = ZoneOffset.of("-02:30");
+        if (ZoneId.systemDefault().getRules().getOffset(Instant.now()).equals(testZoneOffset)) {
+            testZoneOffset = ZoneOffset.of("-09:30");
+        }
+
+        var testZoneId = ZoneId.of("Asia/Katmandu"); // +05:45 ...
+        if (ZoneId.systemDefault().getId().equals(testZoneId.getId())) {
```
stevenschlansker, line 39: I'm not really understanding why the system default comes into play here.  Wouldn't this all be more reliable / understandable if the tests did not rely on the system timezone at all?

hgschmie: this is to ensure that the test zone is not the same as the system default. I chose a strange locale but in case someone in Asia/Katmandu runs the tests, they would end up with the test zone be the same as the system default so the test could not differentiate them.

PR #2964, opened 2026-04-09
SqlExceptionHandler: add StatementContext parameter, remove confusing return value

### LINKED ISSUE #566: Exception Handling Improvements?

stevenschlansker:
Fixes #566 based on user feedback

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlExceptionHandler.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlExceptionHandler.java
@@ -15,23 +15,23 @@ package org.jdbi.v3.core.statement;
 
 import java.sql.SQLException;
 
-import jakarta.annotation.Nullable;
-
-import org.jdbi.v3.meta.Beta;
+import org.jdbi.v3.meta.Alpha;
 
 /**
  * Handler for exceptions thrown while executing SQL statements.
  */
-@Beta
+@Alpha
 @FunctionalInterface
 public interface SqlExceptionHandler {
     /**
      * Take action based on a SQLException thrown during statement execution.
-     * If you return a replacement exception, before throwing it Jdbi will automatically
-     * attach the original exception as "suppressed", so don't do that.
+     * If you would like to replace the thrown exception with a new one, you
+     * may {@code throw} from this method. If the method returns normally,
+     * exception handling proceeds to the next oldest registered handler.
+     * If no handler opts to {@code throw}, the original exception will propagate
+     * wrapped by an {@link UnableToExecuteStatementException}.
      * @param ex the exception thrown
-     * @return a replacement exception to throw, or null to defer
+     * @param ctx the statement context
      */
-    @Nullable
-    Throwable handle(SQLException ex);
+    void handle(SQLException ex, StatementContext ctx);
 }
--- a/core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
+++ b/core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
@@ -38,7 +38,6 @@ import org.jdbi.v3.core.cache.JdbiCacheBuilder;
 import org.jdbi.v3.core.cache.JdbiCacheLoader;
 import org.jdbi.v3.core.cache.internal.DefaultJdbiCacheBuilder;
 import org.jdbi.v3.core.config.JdbiConfig;
-import org.jdbi.v3.core.internal.exceptions.Sneaky;
 import org.jdbi.v3.meta.Beta;
 
 /**
@@ -489,14 +488,7 @@ public final class SqlStatements implements JdbiConfig<SqlStatements> {
     UnableToExecuteStatementException handleException(SQLException e, StatementContext ctx) {
         var handlerIter = exceptionHandlers.descendingIterator();
         while (handlerIter.hasNext()) {
-            SqlExceptionHandler handler = handlerIter.next();
-            Throwable rewritten = handler.handle(e);
-            if (rewritten != null) {
-                if (!e.equals(rewritten)) {
-                    rewritten.addSuppressed(e);
-                }
-                throw Sneaky.throwAnyway(rewritten);
-            }
+            handlerIter.next().handle(e, ctx);
         }
         throw new UnableToExecuteStatementException(e, ctx);
     }
--- a/sqlobject/src/test/java/org/jdbi/v3/sqlobject/config/RegisterExceptionHandlersTest.java
+++ b/sqlobject/src/test/java/org/jdbi/v3/sqlobject/config/RegisterExceptionHandlersTest.java
@@ -18,6 +18,7 @@ import java.sql.SQLException;
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.junit5.H2DatabaseExtension;
 import org.jdbi.v3.core.statement.SqlExceptionHandler;
+import org.jdbi.v3.core.statement.StatementContext;
 import org.jdbi.v3.sqlobject.SqlObjectPlugin;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 import org.junit.jupiter.api.AfterEach;
@@ -92,18 +93,15 @@ class RegisterExceptionHandlersTest {
 
     public static class HandlerOne implements SqlExceptionHandler {
         @Override
-        public Throwable handle(SQLException ex) {
-            return null;
-        }
+        public void handle(SQLException ex, StatementContext ctx) {} // chaining test
     }
 
     public static class HandlerTwo implements SqlExceptionHandler {
         @Override
-        public Throwable handle(SQLException ex) {
+        public void handle(SQLException ex, StatementContext ctx) {
             if ("23505".equals(ex.getSQLState())) {
-                return new RuntimeException("Wahoo");
+                throw new RuntimeException("Wahoo");
             }
-            return null;
         }
     }
 }
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Add StatementContext parameter to SqlExceptionHandler and remove return value
+
 # 3.52.1
 
 - fix regression for `java.time.Instant` mapping from 3.52.0 (#2955, reported by @Eng-Fouad and @toadzky)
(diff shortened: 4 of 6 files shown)


core/src/main/java/org/jdbi/v3/core/statement/SqlExceptionHandler.java
```diff
@@ -27,11 +27,10 @@
...
-     * If you return a replacement exception, before throwing it Jdbi will automatically
-     * attach the original exception as "suppressed", so don't do that.
+     * If you would like to replace the thrown exception with a new one, you
+     * may throw from this method.
      * @param ex the exception thrown
-     * @return a replacement exception to throw, or null to defer
      */
     @Nullable
-    Throwable handle(SQLException ex);
+    void handle(SQLException ex, StatementContext ctx);
```
hgschmie, line 36: there is a justified comment on sonar for one implementation having an empty handle() method. Is that a reasonable implementation (it simply seems to swallow the exception)? If yes, should this be an empty default method? If not, maybe add the comment as suggested here: https://sonarcloud.io/project/issues?id=jdbi_jdbi&pullRequest=2964&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

hgschmie: The change also reconfirms that this should have been marked as @Alpha, not @Beta. I think we should change that as well.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
```diff
@@ -489,14 +488,7 @@ String preparedRender(String template, StatementContext ctx) {
...
         while (handlerIter.hasNext()) {
-            SqlExceptionHandler handler = handlerIter.next();
-            Throwable rewritten = handler.handle(e);
-            if (rewritten != null) {
-                if (!e.equals(rewritten)) {
-                    rewritten.addSuppressed(e);
-                }
-                throw Sneaky.throwAnyway(rewritten);
-            }
+            handlerIter.next().handle(e, ctx);
```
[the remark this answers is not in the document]
stevenschlansker: The feedback was specifically to not add the suppression by default, which is why I removed it.

sqlobject/src/test/java/org/jdbi/v3/sqlobject/config/RegisterExceptionHandlersTest.java
```diff
@@ -92,18 +93,15 @@ public interface ExceptionClassDao {
 
     public static class HandlerOne implements SqlExceptionHandler {
         @Override
-        public Throwable handle(SQLException ex) {
-            return null;
-        }
+        public void handle(SQLException ex, StatementContext ctx) {}
```
hgschmie: can you add a comment here? The sonar feedback does make sense.

core/src/main/java/org/jdbi/v3/core/statement/SqlStatements.java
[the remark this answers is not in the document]
hgschmie: yes, but copilot did not know that. So the feedback is valid and expected

core/src/main/java/org/jdbi/v3/core/statement/SqlExceptionHandler.java
hgschmie: Change this to @Alpha?

xak2000:
@stevenschlansker Hey! Sorry for the delay.

I did a review. It looks like this implementation should solve my use case.

Though, as I previously said, I'm strongly against using `throw` as a handling mechanism. Despite the slightly confusing name `handle` the interface is still just an exception translator. You cannot "handle" the exception in any other way except (optionally) translate it to another exception. You cannot return a fallback value, you cannot initiate a retry, etc. Therefore, the contract that just returns the translated exception looks much more traditional and expected to me.

Also, by using "throw" contract we make it impossible for the handler implementation to fail unexpectedly. E.g. if handling logic was unable to handle the exception because `IOException` is happened. Or an external API returned some `TimeoutException`, etc. I.e. basically a bug in the handler, or an outage of a part of the system.

Hmm.. Actually, I'm wrong. It is still possible. The implementation just needs to throw an exception, and this exception will be thrown by JDBI. But the original SQL exception will be lost. I'm not sure if it is the best logic, but maybe it is.

Now when I'm thinking about it more, I think I see what are you trying to achieve. `handle` is not just `translate`. While `handle` can't really "handle" the exception (i.e. return an alternative result, retry, etc.), it can still excecute a side-effect, without translating the original exception at all. E.g. it could be logging. Maybe something else.. I can't think of other use cases except logging at the moment, but who knows.. Is that why you want to `throw` instead of `return`?

Otherwise, LGTM.

The javadoc of the `handle` interface method is still slightly unclear, IMHO. Can I do a suggestion for it?

Specifically it doesn't say what will happen if the implementation will not throw any exception. Based on the code: the original `SQLException` will be wrapped in `UnableToExecuteStatementException` and then thrown. I believe it would be good to explicitly document this, so the person who will try to implement this contract will understand all the corner cases.

Also, the javadoc doesn't mention at all the handlers **chain**. If I understand correctly, you can add multiple handlers and all of them will be executed in order of their addition. I believe this is important information for a developer who will implement the contract. Especially because as soon as one of the handlers will throw an exception, all further handlers will not be executed at all. It could be very unexpected and  confusing ("Why my handler wasn't executed?!"). :)

---

I also looked at the tests and I see that now the usage of the interface in a lambda looks worse. I'm just thinking what can be done to improve it. Maybe it is worth to add a default method and a method with a single param, like this:

```java
default void handle(SQLException ex, StatementContext ctx) {
    handle(ex);
}

void handle(SQLException ex);
```

This way it would be possible to be used in lambda like this `.addExceptionHandler(ex -> {});`. On the other hand, then the main method of `@FunctionalInterface` will be the method with single param, which is bad. So, probably, not worth it. Forget it. Bad idea. :D

stevenschlansker:
Thank you for taking a look @xak2000 . 

> Though, as I previously said, I'm strongly against using throw as a handling mechanism. Despite the slightly confusing name handle the interface is still just an exception translator. You cannot "handle" the exception in any other way except (optionally) translate it to another exception. You cannot return a fallback value, you cannot initiate a retry, etc. Therefore, the contract that just returns the translated exception looks much more traditional and expected to me.

So, the reason I did it this way was in response to your feedback that it was confusing to have multiple different possible actions -- when should you `return` versus when should you `throw`?  I take this point to mean that it would be best if there is only one possible option. Since we can't / shouldn't prevent the user from `throw` ing some random exception, then it follows that this should be the blessed path, and eliminate the other path -- therefore, no return type.

This leaves it up to the user to `handle` the exception - that might be logging, or it might be throwing a different exception. What I'd imagined you'd do is implement your translation logic and simply throw a new exception if translation finds a match.

> Hmm.. Actually, I'm wrong. It is still possible. The implementation just needs to throw an exception, and this exception will be thrown by JDBI. But the original SQL exception will be lost. I'm not sure if it is the best logic, but maybe it is.

Well, isn't that up to you? If you want to retain the original exception, you simply do `throw new SomethingWentWrongException("Bad!", originalSqlException)` and then it's not lost. Or, you `addSuppressed` it. Whatever you like. This way, it's your choice, rather than up to Jdbi to either decide for you (which you pushed back against) or hide behind another abstraction (which I think is not justified when you can just `throw`).

stevenschlansker:
The chaining was mentioned on the `addExceptionHandler` method, but I added another note to the interface itself. I also updated the doc to describe what happens if you decline to throw a new exception.

xak2000:
> I did it this way was in response to your feedback that it was confusing to have multiple different possible actions -- when should you `return` versus when should you `throw`? I take this point to mean that it would be best if there is only one possible option.

Ah, I see. No, that is not what I meant. :D I meant the method documentation should clearly say if the exception should be thrown or returned. I was totally fine with the returning. It is just the docs was very confusing.



> Well, isn't that up to you? If you want to retain the original exception, you simply do `throw new SomethingWentWrongException("Bad!", originalSqlException)` and then it's not lost.

Correct, but that means that every `handle` implementation should have a global `try/catch(Throwable t)` block around the whole code, otherwise some unexpected runtime exception could still be thrown by some 3rd party library used in the translation logic.

But I agree: this is probably the most flexible way. The handler has full control.

> I also updated the doc to describe what happens if you decline to throw a new exception.

For some reason I don't see the update.

xak2000:
> For some reason I don't see the update.

I can see it now. Look good.

Btw, it also reveals an important detail:
> exception handling proceeds to the next oldest registered handler

I totally missed that in the original code. I was sure that handlers are executed in order of their registration, not in the reverse order as this line claims.

stevenschlansker:
I think I closed my laptop too fast and the push didn't finish until I woke it back up, oops...

hgschmie (approved):

core/src/main/java/org/jdbi/v3/core/statement/SqlExceptionHandler.java
stevenschlansker: Why would we want an empty default method? The whole point of this interface is to implement it...
I do not care Alpha vs Beta, it is all the same to me, a distinction without a difference.

Issue #2979, opened 2026-06-05
Annotation for pre-flight statements on SqlObject methods


ltanguy:
Hello,

It would be great to have the possibility to add arbitrary preflight statements on a @SqlQuery/@SqlUpdate/... method.
This could be useful for example to define scoped variables before executing a query.

As an example, to have a non-default threshold for a trigram search in Postgres, I ended up with the following working solution on current JDBI versions:
```java
@SqlUpdate("set-search-threshold")
void setSearchThreshold();

@SqlQuery("find-user")
@RegisterFieldMapper(User.class)
List<User> findUserWithoutThreshold(String search);

default List<User> findUser(String search) {
    setSearchThreshold();
    return findUserWithoutThreshold(search);
}
```

I think it could be expressed more elegantly as such:

```java
@SqlPreflight("set pg_trgm.word_similarity_threshold=0.4")
@SqlQuery("find-user")
@RegisterFieldMapper(User.class)
List<User> findUser(String search);
```

stevenschlansker:
Hi @ltanguy , thank you for this interesting feature suggestion. Given the simple workaround I don't think we'd immediately prioritize maintainers working on this feature, but it could be an interesting contribution from the community.

stevenschlansker:
Released in 3.54.0

PR #2981, opened 2026-06-17
Add @SqlPreflight annotation for SQL Object methods (#2979)

ltanguy:
﻿Implements #2979.

## What

Adds `@SqlPreflight`, an annotation that runs a literal SQL statement on the **same `Handle`**, immediately before a SQL Object method's main statement. The primary use case is setting a scoped/session variable before a query (e.g. a non-default Postgres trigram similarity threshold) without the boilerplate of a separate method plus a `default` wrapper.

```java
@SqlPreflight("select set_config('pg_trgm.word_similarity_threshold', :threshold, true)")
@SqlQuery("select name from users where name %> :search")
@RegisterFieldMapper(User.class)
List<User> findUser(@Bind("search") String search, @Bind("threshold") String threshold);
```

## Details

- The method's arguments are bound to the preflight statement, honoring `@Bind`/`@BindBean` and default by-name binding. Arguments the preflight does not reference are ignored, so a static preflight works on a method that declares arguments.
- Repeatable (multiple preflights run in declaration order) and placeable on a method or on the SQL Object type (type-level runs before method-level).
- Implemented as an `ExtensionHandlerCustomizer` (registered via `@UseExtensionHandlerCustomizer`), not the deprecated `HandlerDecorator`.
- Argument binding is reused via a small internal `ParameterBinder`, deliberately applying only *parameter* customizers — the full statement-handler chain is not reused because query-only customizers such as `@FetchSize` cast to `Query` and would fail on the preflight `Update`.
- The customizer is idempotent: when `@SqlPreflight` is present on both the type and the method, the extension framework applies the customizer twice, so it collapses duplicate wrapping to run each preflight once.
- The preflight statement sets `setUnusedBindingAllowed(true)` on its own (child) config so binding all method arguments never trips the unused-binding check; the main statement keeps strict checking.

## Scope notes

- Value is **literal SQL only** — not resolved through the configured `SqlLocator`. Documented.
- Marked `@Alpha`.

## Tests

`TestSqlPreflight` (H2, no mocks) covers: runs-before/same-handle, argument binding, multiple-in-order, type-level, type+method single execution, static-preflight-with-args (no superfluous-binding error), coexistence with `@FetchSize`, and `@SqlUpdate` as the main statement.

## Docs

New `@SqlPreflight` section in the reference docs and a `RELEASE_NOTES.md` entry.


docs/src/adoc/index.adoc
```diff
@@ -4090,6 +4090,35 @@ By default, methods annotated with link:{jdbidocs}/sqlobject/statement/SqlScript
...
+
+The annotation is repeatable--multiple preflight statements run in declaration order--and may be placed on a method or on the SQL Object type, in which case it applies to all of the type's methods (type-level preflights run before method-level ones).
+
+Unlike link:{jdbidocs}/sqlobject/statement/SqlQuery.html[@SqlQuery^] and link:{jdbidocs}/sqlobject/statement/SqlUpdate.html[@SqlUpdate^], the `@SqlPreflight` value is always used as literal SQL; it is not resolved through the configured link:{jdbidocs}/sqlobject/locator/SqlLocator.html[SqlLocator^], so external `.sql` files and link:{jdbidocs}/sqlobject/locator/UseClasspathSqlLocator.html[@UseClasspathSqlLocator^] have no effect on it.
+
+Because the preflight runs on the same connection as the main statement, scoped variables set this way are visible to the query.
+When connection pooling is in use, prefer transaction-scoped settings (for example `SET LOCAL` in Postgres) together with link:{jdbidocs}/sqlobject/transaction/Transaction.html[@Transaction^], so the setting does not leak to later users of the connection.
```
stevenschlansker, line 4116: Unfortunately, stacking `@SqlPreflight` and `@Transaction` doesn't seem to guarantee correct ordering behavior. While the current OpenJDK implementation returns annotations in their source-declared order, it is not guaranteed either in the Java language spec nor the getAnnotations spec. So I am going to tweak this advice a bit to suggest explicitly declaring ordering.

stevenschlansker (approved):

docs/src/adoc/index.adoc
stevenschlansker: Upon further review, I think it is possible to simplify the implementation and avoid this hazard at the same time. I will propose a followup PR.

stevenschlansker:
Merged as #2988

PR #2982, opened 2026-06-22
Cache derived attach configurations per source registry

ulmetrs:
# Cache derived attach configurations per source registry

## Summary

Deriving an extension's configuration on `attach` performs a full
`ConfigRegistry.createCopy()` for the instance configuration **plus one copy per
extension method**. `Jdbi#onDemand(Class)` (and `Jdbi#withExtension`) re-attaches
the extension on **every** method call against the same, stable `Jdbi`-level
configuration, so without caching it repeats `1 + methodCount` registry copies on
every single call. For on-demand-heavy workloads those copies dominate allocation.

This PR introduces `ExtensionConfigCache`, a per-`ConfigRegistry` `JdbiConfig`
(placed in `org.jdbi.v3.core.extension.internal`, mirroring
`org.jdbi.v3.core.config.internal.ConfigCaches`) that memoizes the derived
instance and method configurations, keyed by extension type and by method.

## Why this is safe

The derived configuration depends on the **identity of the source registry**, so
the cache is deliberately **not** shared across copies — `createCopy()` returns a
fresh, empty instance. This gives two clean behaviors:

- **One-shot attach sources** (e.g. the per-`Handle` configuration used by
  `Handle#attach(Class)`) cache nothing reusable and behave **as before — no
  meaningful regression** (see benchmark below).
- **Long-lived attach sources** (the `Jdbi` root configuration used by on-demand)
  amortize the `1 + methodCount` copies to a **one-time cost**.

Entries live and die with their owning registry, so there is no unbounded growth.

I verified the on-demand path actually hits the cache:
`onDemand` → proxy → `Jdbi#withExtension` → `LazyHandleSupplier.getConfig()` returns
`jdbi.getConfig()` (the **stable root registry**, same instance on every call) →
`ExtensionFactoryDelegate.attach()` derives from it. The instance key
(`extensionType`) and method key (`Method`) cannot collide because
`Extensions#findMetadata` memoizes one `ExtensionMetadata` per type, so a given
source registry yields exactly one derived configuration per type/method.

## Behavior change worth calling out

Because the derived configuration is now frozen at first use, **reconfiguring the
`Jdbi` root configuration after an on-demand extension has already been used will
no longer be picked up** by that extension. Previously each call re-copied the
live root config, so a late `jdbi.registerRowMapper(...)` between two on-demand
calls would have been observed.

This matches the documented configuration contract — "a new context inherits a
copy of its parent configuration **at the time of creation**" and Jdbi
configuration is expected to be finalized before use. The shared, Jdbi-level
`ConfigCaches` already makes the same "do not respect reconfiguration" assumption.

## Related

This is the same hotspot discussed in #1103 (monitor contention in
`ConfigRegistry.createCopy()` under load), where lazily/once deriving config from
the parent registry to "amortize the cost of cloning" was suggested.

## Tests & benchmark

- `TestExtensionConfigCache` — unit tests for per-type / per-method memoization
  and the fresh-on-copy semantics, plus a wiring test through `ExtensionMetadata`
  asserting that repeated attaches against the same source registry reuse the
  derived configuration while a different source registry derives its own.
- `OnDemandAttachBenchmark` (JMH) — exercises the attach configuration-derivation
  cost with a many-method DAO (so the `1 + methodCount` copies dominate). The
  `onDemandAttach` benchmark drives the cached path; `handleAttach` opens a fresh
  `Handle` per call (a one-shot attach source, mirroring open-handle-per-unit-of-work)
  to confirm no regression on the path the cache cannot help.

## Benchmark results

`OnDemandAttachBenchmark` (JMH, H2, 10-method DAO), allocation via `-prof gc`:

| Benchmark | before (no cache) | after (cache) | allocation | throughput |
| --- | --- | --- | --- | --- |
| `onDemandAttach` (stable config, cache hit) | 90,574 B/op @ 42.0 ops/ms | 34,425 B/op @ 83.7 ops/ms | **−62%** | **~2×** |
| `handleAttach` (fresh handle/call, cache miss) | 258,366 B/op @ 12.6 ops/ms | 261,814 B/op @ 12.0 ops/ms | +1.3% | within noise |

The on-demand path allocates **~62% less per call** and roughly doubles throughput.
The cache-miss path shows no meaningful regression: the ~1% allocation delta is the
empty cache holder allocated on each fresh registry, dwarfed by the per-call
connection open (~258 KB here), and it turns positive for any handle reused more
than once.

(Reduced JMH settings — 1 fork × 3 warmup + 4 measurement iterations — for
turnaround; allocation norms were stable to within ±200 B/op. A full default run
reproduces the same ratios.)

## Notes for reviewers

- `ExtensionConfigCache` is intentionally in an `.internal` package so it is not
  part of the public API (excluded from japicmp / Javadoc), consistent with
  `ConfigCaches`.

### DIFF
--- a/benchmark/src/main/java/org/jdbi/v3/benchmark/OnDemandAttachBenchmark.java
+++ b/benchmark/src/main/java/org/jdbi/v3/benchmark/OnDemandAttachBenchmark.java
@@ -1,0 +1,175 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jdbi.v3.benchmark;
+
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+
+import org.jdbi.v3.core.Handle;
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.mapper.RowMapper;
+import org.jdbi.v3.core.statement.StatementContext;
+import org.jdbi.v3.sqlobject.SqlObjectPlugin;
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.TearDown;
+import org.openjdk.jmh.annotations.Warmup;
+
+/**
+ * Exercises the per-{@code attach} configuration derivation cost.
+ *
+ * <p>{@link Jdbi#onDemand(Class)} (and {@link Jdbi#withExtension}) re-attaches the extension on every
+ * method call against the same, stable {@code Jdbi}-level configuration. Each attach derives an
+ * instance configuration plus one configuration per extension method, each a full
+ * {@code ConfigRegistry.createCopy()}. The {@code ManyMethodDao} below has many methods so that the
+ * {@code 1 + methodCount} copies dominate, making the effect of caching the derived configurations
+ * visible. {@link #handleAttach()} attaches against a per-{@link Handle} configuration (a one-shot
+ * attach source that does not benefit from the cache) and is included to confirm there is no
+ * regression on that path.
+ */
+@State(Scope.Thread)
+@BenchmarkMode(Mode.Throughput)
+@Measurement(time = 5, iterations = 5)
+@Warmup(time = 5, iterations = 5)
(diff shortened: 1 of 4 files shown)


ulmetrs:
## Validated in a production-like service deploy

Beyond the JMH benchmark, I deployed this patch into a real on-demand-heavy service (Java 25, with the patched `jdbi3-core` pinned in) and profiled allocations under live traffic with async-profiler (`-e alloc`):

- **`ConfigRegistry.createCopy` dropped to 0.11% of allocation**, versus **~38%** in the unpatched baseline — the per-attach config copies are eliminated.
- The extension **attach path was still active** (~9% of allocation in jdbi frames) and `ExtensionConfigCache` was present on the hot path — so the reduction is the cache hitting under real traffic, not an idle window.
- The residual attach-path allocation is SQL parsing (`warm`), result mapping, and collection churn — **not** config copies.

This matches the JMH results above (−62% allocation / ~2× throughput on the on-demand path).

stevenschlansker:
Thank you for taking an interest in this problem. We're interested in getting this fixed. I will be on vacation for the next two weeks and unable to review - just wanted to let you know that you're not being ignored, we are just busy, but we will eventually get to this proposal.

stevenschlansker:
Thank you again for this work, and for the JMH and async-profiler numbers. The hotspot is real, it is the same one as #1103, and we want the win. We reviewed the approach in depth, and the freeze-at-first-use semantics is the blocker. Two supported patterns break silently (both confirmed with differential tests against this branch and its merge base):

1. Configuration added to the `Jdbi` after the first on-demand use is ignored forever. `jdbi.onDemand(Dao.class)`, one call, then `jdbi.registerRowMapper(...)`: the next call throws `NoSuchMapperException`.
2. A handle's config registry is created once per handle, not once per attach, so the "one-shot attach source" premise for `Handle#attach` does not hold. Repeated attaches on one handle share the frozen entry and ignore registrations made on the handle between them.

There is also a sharing hazard: the cached method configuration becomes the handle's live configuration during invocation. A DAO default method that calls `getHandle().define(...)` or `registerRowMapper(...)` writes into a registry shared across threads and across future attaches. Invalidation cannot repair this, because config mutations are plain setters on `JdbiConfig` objects and the registry cannot observe them.

The entire cost you measured is the eager deep copy inside `ConfigRegistry.createCopy()`, so we reworked the fix one level down in #3018. That PR keeps your commit and your benchmark, removes the cache, and makes `createCopy()` materialize each config object on its first access. Each attach still derives a private configuration, so nothing freezes, and the same change also removes the eager copy cost per `Handle` and per statement. On your benchmark (full settings, our machine) it keeps most of the on-demand win: -41% allocation and 1.51x throughput on `onDemandAttach` versus master. It also improves `handleAttach` (-14% allocation, 1.18x), the path the cache could not reach.

Please take a look at #3018. A re-run of your async-profiler validation on that branch would be a very useful cross-check.

stevenschlansker:
Merged as #3018

PR #2983, opened 2026-06-27
ConfigCaches: run cache computers without holding the bin monitor to fix ABBA deadlock

### LINKED ISSUE #2980: Deadlock in ConfigCaches: recursive computeIfAbsent during mapper warming holds bin monitor
### Affected version
jdbi `3.53.0` (the relevant code is unchanged on `master`). Reproduced on Temurin JDK 21.0.11 and 25.0.3.

### Summary
Under concurrent load, request threads permanently deadlock inside `ConfigCaches`. `ConfigCaches.get` resolves a cache miss with `ConcurrentHashMap.computeIfAbsent`, and the mapping function (`computer.apply`) runs **while `computeIfAbsent` holds the monitor on the target bin**. In real usage the mapping function is a mapper/qualifier factory whose resolution recursively asks the cache for *other* entries (resolving a row mapper consults the column-mapper and qualifier caches). Two threads can therefore each hold one bin's monitor and then block acquiring the bin the other holds — an ABBA deadlock across two `ConcurrentHashMap` bins.

https://github.com/jdbi/jdbi/blob/v3.53.0/core/src/main/java/org/jdbi/v3/core/config/internal/ConfigCaches.java#L60-L67

```java
public V get(K key, ConfigRegistry config) {
    return (V) config.get(ConfigCaches.class).caches
        .computeIfAbsent(this, x -> new ConcurrentHashMap<>())
        .computeIfAbsent(keyNormalizer.apply(key), x -> computer.apply(config, key)); // computer runs under the bin lock
}
```

This became a production hang after upgrading from `3.28.0`. The cache code is the same in `3.28.0`; what changed is that the newer extension framework eagerly **warms** every SqlObject method's mappers at attach time (`ExtensionMetadata.createExtensionHandlerInvoker` → `...warm()` → `Mappers.findFor` → `ConfigCaches.get`), so this recursive resolution now runs concurrently across many request threads during cold attaches — reliably triggering the latent hazard.

### Thread dump (sanitized excerpt)
~100 threads were `BLOCKED` waiting to lock a single `ConcurrentHashMap` bin owned by a thread that is itself `BLOCKED` in the same `computeIfAbsent`:

```
"holder" BLOCKED (on object monitor)
    at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(java.base@21.0.11/ConcurrentHashMap.java:1726)
    - locked <0x...> (a java.util.concurrent.ConcurrentHashMap$Node)
    at org.jdbi.v3.core.config.internal.ConfigCaches$1.get(ConfigCaches.java:64)
    at org.jdbi.v3.core.qualifier.Qualifiers.findFor(Qualifiers.java:66)
    at org.jdbi.v3.core.mapper.QualifiedColumnMapperFactory.lambda$adapt$0(QualifiedColumnMapperFactory.java:47)
    at org.jdbi.v3.core.mapper.ColumnMappers.findFor(ColumnMappers.java:217)
    ... <application RowMapperFactory> ...
    at org.jdbi.v3.core.mapper.RowMappers.findFor(RowMappers.java:171)
    at org.jdbi.v3.core.mapper.Mappers.findFor(Mappers.java:109)
    at org.jdbi.v3.sqlobject.statement.internal.ResultReturner.warm(ResultReturner.java:144)
    at org.jdbi.v3.sqlobject.statement.internal.CustomizingStatementHandler$3.warm(CustomizingStatementHandler.java:201)
    at org.jdbi.v3.core.extension.ExtensionMetadata$ExtensionHandlerInvoker.<init>(ExtensionMetadata.java:323)
    at org.jdbi.v3.core.extension.ExtensionMetadata.createExtensionHandlerInvoker(ExtensionMetadata.java:135)
    at org.jdbi.v3.core.extension.ExtensionFactoryDelegate.attach(ExtensionFactoryDelegate.java:136)
    at org.jdbi.v3.core.extension.Extensions.findFor(Extensions.java:177)
    at org.jdbi.v3.core.Jdbi.withExtension(Jdbi.java:545)
    at org.jdbi.v3.core.internal.OnDemandExtensions.lambda$createProxy$0(OnDemandExtensions.java:72)
    ... <request-handling frames elided> ...
```

### Reproduction
A deterministic JUnit reproduction (two threads, two bins, a latch to force both to hold their bin monitor before recursing) plus a negative control that shows the same setup does **not** deadlock when the loader does not re-enter the cache:

https://github.com/jaredstehler/jdbi/blob/f92d23c1c093f692c25964d2fb55f2cc349147cc/core/src/test/java/org/jdbi/v3/core/config/internal/TestConfigCachesDeadlock.java

Run with `mvn -pl core -Dtest=TestConfigCachesDeadlock test`. The test asserts via `ThreadMXBean.findDeadlockedThreads()` that the JVM detects the deadlock on the `ConfigCaches` → `computeIfAbsent` path.

### Suggested direction
The mapping function passed to `computeIfAbsent` violates the `ConcurrentHashMap` contract ("the mapping function must not modify this map"). Resolving the value outside the `computeIfAbsent` lock (e.g. compute-then-`putIfAbsent`, accepting occasional duplicate computation of these constant-pool-keyed values) would remove the lock held across recursive cache access.

stevenschlansker:
Claude says:

ConcurrentHashMap.computeIfAbsent holds the bin monitor for the duration of the
mapping function. Our cache computers routinely recurse into other caches (and
other bins of the same map) to resolve qualifiers and column mappers, so two
threads warming such mappers concurrently could take bin monitors in opposite
order and deadlock (ABBA).

Install an incomplete CompletableFuture-backed entry via an atomic putIfAbsent,
then run the computer with no lock held; late arrivals join the winner's result.
The computer runs exactly once per key. Failures are not memoized: the entry is
dropped so a later call retries, and the cause is handed to any joining threads.
A computer that re-enters the cache for the key it is itself computing describes
an unsatisfiable self-dependency, so we reject it rather than self-deadlock on
the join, mirroring computeIfAbsent's recursive-update behavior.

Null results are rejected rather than cached. The old computeIfAbsent silently
never stored null and recomputed every call; no production computer returns null,
so caching one is not a needed feature. Rejecting it at the computer fails loudly
there instead of NPE-ing downstream.

Fixes https://github.com/jdbi/jdbi/issues/2980

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/config/internal/ConfigCaches.java
+++ b/core/src/main/java/org/jdbi/v3/core/config/internal/ConfigCaches.java
@@ -105,7 +105,8 @@ public final class ConfigCaches implements JdbiConfig<ConfigCaches> {
                 final Object computed;
                 try {
                     computed = computer.apply(config, key);
-                } catch (RuntimeException | Error e) {
+                } catch (Throwable e) {
+                    // Clean up on ANY throwable, since otherwise we leave waiting threads blocked forever
                     fail(cache, normalizedKey, entry, e);
                     throw e;
                 }
@@ -117,7 +118,7 @@ public final class ConfigCaches implements JdbiConfig<ConfigCaches> {
                     fail(cache, normalizedKey, entry, e);
                     throw e;
                 }
-                entry.value.complete(computed);
+                entry.complete(computed);
                 return (V) computed;
             }
         };
@@ -129,16 +130,31 @@ public final class ConfigCaches implements JdbiConfig<ConfigCaches> {
      */
     private static void fail(Map<Object, Object> cache, Object normalizedKey, CacheEntry entry, Throwable cause) {
         cache.remove(normalizedKey, entry);
-        entry.value.completeExceptionally(cause);
+        entry.completeExceptionally(cause);
     }
 
     /**
      * A cache slot. {@code value} is completed exactly once by the thread that installed the entry;
      * {@code owner} is that thread, retained only to detect a computer re-entering the cache for the key
      * it is itself computing (an unsatisfiable self-dependency we must reject rather than deadlock on).
+     * The owner reference is meaningful only while the computation is in flight, so it is cleared on
+     * completion: the cache is unbounded and lives for the life of the Jdbi instance, and a retained
+     * Thread would pin that thread (and its thread-locals and context classloader) from collection.
      */
     private static final class CacheEntry {
-        final Thread owner = Thread.currentThread();
+        // No synchronization needed: a thread always sees its own writes, and any other thread that reads a
+        // stale owner fails the identity check and falls through to join() — the correct path.
+        Thread owner = Thread.currentThread();
         final CompletableFuture<Object> value = new CompletableFuture<>();
+
+        void complete(Object computed) {
+            owner = null;
+            value.complete(computed);
+        }
+
+        void completeExceptionally(Throwable cause) {
+            owner = null;
+            value.completeExceptionally(cause);
+        }
     }
 }
--- a/core/src/main/java/org/jdbi/v3/core/config/internal/ConfigCaches.java
+++ b/core/src/main/java/org/jdbi/v3/core/config/internal/ConfigCaches.java
@@ -14,6 +14,7 @@
 package org.jdbi.v3.core.config.internal;
 
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiFunction;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


hgschmie:
I am surprised that we have to step so deeply into JVM logic for a simple cache. Isn't that something where we can just use an existing library (or copy code from there), such as caffeine?

hgschmie:
There is one open sonarqube issue (catching Error) that I actually agree with; there is IMHO no point catching error here because you are probably microseconds from crashing anyway and trying to update the cache makes not sense. If the program really carries on after an error was thrown (because some other part decides to catch it), there is no point to do some minor correctness fix. 

Other that that, while convoluted (and the wall of text was *clearly* generated by AI), this looks ok.

hgschmie (approved):
Please remove the "catch error", catching RuntimeExceptions is ok.

stevenschlansker:
> I am surprised that we have to step so deeply into JVM logic for a simple cache. Isn't that something where we can just use an existing library (or copy code from there), such as caffeine?

I would have loved to use caffeine here but @hgschmie removed it from core https://github.com/jdbi/jdbi/pull/2246 ;) 
If we think this would be reason to overturn that decision, we could re-add caffeine to core and remove the cache pluggability and use the async cache here instead of a hand-rolled one.

stevenschlansker:
@hgschmie , I pushed back on Claude re: the `catch (Error)` change, and Claude insists it's justified, and it's strictly better to at least try to release threads waiting on a `.join()` instead of sentencing them to block forever.

```
  The likely reviewer instinct, and why it's wrong here. "Don't catch Error" is a good default rule — you
  generally shouldn't handle Error, because the JVM may be in an unrecoverable state. But this code isn't
  handling it. It catches, runs cleanup that other threads depend on, and rethrows. The Error still propagates to
  the caller exactly as it would have. We're not pretending to recover; we're just not leaving a poisoned,
  never-completed entry behind that strands other threads.

  What I'd do. Keep the RuntimeException | Error catch and reply to the reviewer explaining why, since the
  reasoning is non-obvious. Better still, make the intent explicit in the code so the next reader (and the next
  reviewer) doesn't trip on the same thing. The cleanest form is catch (Throwable e) with a comment, which also
  covers the theoretical non-Error/non-RuntimeException Throwable:

  } catch (Throwable e) {
      // Run cleanup for ANY throwable, including Error: fail() releases threads blocked on
      // existing.value.join() and drops the un-memoized entry. If we let an Error escape without
      // this, the CacheEntry's future is never completed and every joiner hangs forever — the
      // exact deadlock class this code exists to prevent (issue #2980). We rethrow, so the
      // Error still propagates; we are not handling it, only not stranding other threads.
      fail(cache, normalizedKey, entry, e);
      throw e;
  }

  One caveat worth being honest about: this means fail() runs even on OutOfMemoryError, where allocation inside
  cache.remove/completeExceptionally could itself fail. But that path is no worse than today — if cleanup can't
  run under OOM, the joiners hang either way, and the alternative (not catching) guarantees they hang. So
  catching is strictly better.
```
 
I'm inclined to take the Throwable form - leaving `join` blocked forever is not great, and this covers all cases. It's like a "finally" block but only for exceptions, not normal return. Do you still feel strongly that we should remove it?

hgschmie:
> > I am surprised that we have to step so deeply into JVM logic for a simple cache. Isn't that something where we can just use an existing library (or copy code from there), such as caffeine?
> 
> I would have loved to use caffeine here but @hgschmie removed it from core #2246 ;) If we think this would be reason to overturn that decision, we could re-add caffeine to core and remove the cache pluggability.

We removed caffeine from core when we removed the optional caching pieces. It was never used in the core proper. I have no problem with introducing a dependency if it provides value for us.

hgschmie:
> @hgschmie , I pushed back on Claude re: the `catch (Error)` change, and Claude insists 

[...]

> I'm inclined to take the Throwable form - leaving `join` blocked forever is not great, and this covers all cases. It's like a "finally" block but only for exceptions, not normal return. Do you still feel strongly that we should remove it?

I'm not really into arguing with a LLM. I can live with catch Throwable but IMHO there is no point catching errors because the JVM is usually borked anyway. But the `RuntimeException | Error` is not correct as any checked exception would not be caught.

stevenschlansker:
Ok, I can try swapping caffeine back in tomorrow maybe.

hgschmie:
Personally, I think that Sonar has a point here but it is minor enough that I see no reason spending cycles on it:

```
Why is that an issue?

Throwable is the superclass of all errors and exceptions in Java. Error is the superclass of all errors, which are not meant to be caught by applications.

Catching either Throwable or Error will also catch OutOfMemoryError and InternalError, from which an application should not attempt to recover.
```

stevenschlansker:
Right: the intent here is not to "recover", but to avoid leaving hanging threads that might block jvm exit.
Regardless, if we use caffeine, the entire thing probably becomes like 5 lines.

stevenschlansker:
Ok, I have investigated using Caffeine asyncCache as the backing store for this ConfigCaches data.
Caffeine local async-cache is not an appropriate fix for this issue, because the implementation is a very fancy delegate (implementing eviction, weak key / value, etc...) on top of a ConcurrentHashMap - `AsyncCache.get` has the same "a value mapper may not itself modify the cache" restriction.

Therefore to fix this bug I think we need to take this complexity for now, unfortunately.
The entire approach to caching in Jdbi currently is quite broken, I am excited to get this removed in v4.

PR #3006, opened 2026-07-28
Move GraalVM CI to matrix (25.0, 25.2) and native-build-tools to 1.1.5

stevenschlansker:
Change Graal native-image to a matrix build and track both LTS and "Innovation" streams

Update native plugin to 1.1.5. 1.1.6 is skipped: it cannot run under Maven 3.9 at all (native-build-tools#1000).

Three registrations are needed on top of the version bumps, each verified load-bearing by removing it and watching the suite fail:

- Jdbi ships core/src/test/resources/mockito-extensions/org.mockito. plugins.MockMaker to select the proxy mock maker, so Jdbi declares the resource that makes it readable in the image. The metadata repository covers Mockito's own jar, not files we add to its extension directory. Without it all 150 Mockito tests fail.
- java.sql.Timestamp[] for TestSqlArrays and ConcurrentHashMap's constructor for MapCollectorFactoryTest, both reached reflectively.
- hsqldb formats dates through Locale.UK, so en-GB has to be in the image explicitly or HsqlDateTime's initialiser dies on a CLDR bundle.

### DIFF
--- a/.github/workflows/graal-ci.yml
+++ b/.github/workflows/graal-ci.yml
@@ -24,10 +24,22 @@ permissions:
 jobs:
   native-build:
     runs-on: ubuntu-latest
-    name: GraalVM native build
+    name: GraalVM ${{ matrix.graalvm-version }} ${{ matrix.distribution }} native build
     strategy:
+      fail-fast: false
+      # Spelled out rather than a cross product, because the cells are not
+      # interchangeable: only the interim releases can run exact reachability
+      # metadata, so that will be enabled per cell rather than globally.
       matrix:
-        java-version: [ 25 ]
+        include:
+          - graalvm-version: '25.0'
+            distribution: 'graalvm'
+          - graalvm-version: '25.0'
+            distribution: 'graalvm-community'
+          - graalvm-version: '25.2'
+            distribution: 'graalvm'
+          - graalvm-version: '25.2'
+            distribution: 'graalvm-community'
 
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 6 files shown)


native-tests/pom.xml
```diff
@@ -445,6 +445,12 @@
                 <groupId>org.graalvm.buildtools</groupId>
                 <artifactId>native-maven-plugin</artifactId>
                 <extensions>true</extensions>
+                <configuration>
+                    <buildArgs combine.children="append">
+                        <!-- hsqldb formats dates through Locale.UK, so en-GB bundles must be in the image -->
+                        <buildArg>-H:IncludeLocales=en-GB</buildArg>
```
hgschmie, line 451: lol

hgschmie (approved):
shipit

.github/workflows/graal-ci.yml
```diff
@@ -24,10 +24,22 @@ permissions:
...
+      # Spelled out rather than a cross product, because the cells are not
+      # interchangeable: only the interim releases can run exact reachability
+      # metadata, so that will be enabled per cell rather than globally.
       matrix:
-        java-version: [ 25 ]
+        include:
+          - graalvm-version: '25.0'
+            distribution: 'graalvm'
+          - graalvm-version: '25.0'
+            distribution: 'graalvm-community'
```
hgschmie, line 38: are those substantially different?

stevenschlansker: There are very minor differences I tripped over once. If we don't find any differences going forward, we can probably pick only the community build.

PR #3007, opened 2026-07-28
native-image: check reachability metadata exactly

stevenschlansker:
Adds a native-exact-metadata profile passing --exact-reachability-metadata for jdbi's packages, so a missing registration fails the build rather than degrading quietly: jdbi ignores ReflectiveOperationException in places, so a gap otherwise changes behaviour without reporting anything.

The flag matches the calling code's package and takes neither wildcards nor a package hierarchy, so all 95 packages are named. ExactMetadataPackagesTest scans the source tree and fails on any drift, so a package added later cannot silently stop being checked.

Ten registrations were missing, each traced to a reported failure:

  - three SQL Object types need field queries, because DefinitionsFactory.configureDefinitions queries declared fields for every extension type
  - Mockito's plugin loader needs the MockResolver proxy, and Spring's SpringMockResolver constructor
  - six types are allocated without a constructor: unsafeAllocated
  - TypeConversionHelper needs allDeclaredClasses so kotlin-reflect can resolve a nested Java class. GraalVM reports nothing for this one: Warn mode is silent and the symptom is KotlinReflectionInternalError, "Unresolved class ... (kind = null)"

ExactMetadataEnforcedTest keeps a green run meaningful. It looks up an unregistered type in a checked package and requires the image to answer as the build says it should, rejected with the flag on and absent with it off, so a build that misreports the flag fails.

CI enables the flag on the 25.2 cells only: on 25.0 it breaks @Nested test resolution without reporting it. The demo README now says which registrations belong to an application, enums included.

Verified on CE 25.2.4: 2601 tests pass with the flag and 2601 without.


hgschmie:
Can we turn this into an actual PR? We have all the native stuff "bleeding edge" anyway.

Issue #2247, opened 2023-01-10
Make modules OSGi compatible


TomasTokaMrazek:
OSGi compatible JARs have Bundle-* specific headers inside MANIFEST.MF file. That is done by [bnd maven plugin](https://github.com/bndtools/bnd/blob/master/maven-plugins/bnd-maven-plugin/README.md) automatically.

From quick inspection of Jdbi codebase, there are few gotchas. 
* Any shaded packages (ANTLR) should be excluded from `Export-Package`.
* Usage of SPI (JdbiPlugin) in OSGi environment is slightly complicated. However how to deal with that is explained [here](https://bnd.bndtools.org/chapters/240-spi-annotations.html).

The OSGi framework is not the most popular solution, but it is still under active development, used by big names like WSO2 and Adobe AEM and vast majority of third party libraries are more or less compatible with OSGi.

hgschmie:
Hi @TomasTokaMrazek,

Thank you for opening an issue with Jdbi! 

I agree that OSGi compatibility would be great. I don't think anyone on the team right now is an expert on OSGi. So we would be relying on PRs (even a sketch on how this would integrate with our build system and maybe convert one module) to get this started.

We do support the java 9 module system.

TomasTokaMrazek:
I am no expert myself, but I've had to wrap few non-OSGi libraries, so we can use it in our environment. I'll try to create draft PR for it in a freetime then.

TomasTokaMrazek:
@hgschmie Sorry, took me some time but I finally have time (and need) to implement it. We are currently testing in DEV environment newest OSGi runtime with SLF4J 2.x in conjunction with few jdbi modules with OSGi headers I implemented. So far so good, so I opened up a PR.

PR #3010, opened 2026-08-10
Add OSGi bundle manifests to all published modules

TomasTokaMrazek:
This implements #2247: all published jdbi3 modules now ship OSGi bundle manifests, generated at build time by `bnd-maven-plugin` (`bnd-process` goal). The plugin version is managed in the build parent; each module carries its own inline `<bnd>` instructions, and `maven-jar-plugin` picks up the generated manifest on the `default-jar` execution only (`-tests` jars are unchanged).

### Design

- `Bundle-SymbolicName` = the JPMS `Automatic-Module-Name` (`${moduleName}`).
- `-exportcontents: org.jdbi.v3.*` — every package in each jar is exported, including `.internal` packages. Jdbi's internals are a de-facto inter-module API (e.g. sqlobject imports `org.jdbi.v3.core.internal`), so hiding them would make the downstream bundles unresolvable. Exports carry the bundle version, so inter-module imports get proper computed ranges (`[3.54,4)`).
- Imports mirror the poms: Maven `<optional>` dependencies that are referenced in bytecode are declared `resolution:=optional` (`com.pgvector` and `org.jdbi.v3.json` in postgres, `org.jdbi.v3.sqlobject` in freemarker/stringtemplate4/jpa, `jakarta.annotation`, `jdk.jfr`, errorprone annotations, the junit/flyway/driver packages in testing, …). Everything else keeps bnd's computed mandatory import with a version range.
- core additionally has:
  - `Import-Package: !org.antlr.v4.runtime*` — the antlr runtime is inlined/relocated by `inline-maven-plugin` *after* the manifest is generated, so it must not be imported (the relocated packages stay private, verified);
  - `DynamicImport-Package: *` — so `Jdbi.create(url)`/`DriverManager` can see driver classes registered by driver bundles (same approach as EclipseLink and Hibernate);
  - Aries SPI-Fly consumer headers (`osgi.serviceloader.processor`, both `resolution:=optional`) so the deprecated `installPlugins()` ServiceLoader path works under an SPI-Fly-equipped framework.
- Modules that ship `META-INF/services/org.jdbi.v3.core.spi.JdbiPlugin` use `-metainf-services: auto`, which advertises the plugins via `osgi.serviceloader` capabilities; SPI-Fly registers them as OSGi services.
- bnd is deliberately **not** gated behind `basepom.check.*`, so release builds keep the headers, and reproducible builds are preserved (no `Bnd-LastModified`; the existing `compare-reproducible` job passes).

### Verification

- `make install` passes on the full reactor.
- Runtime-tested on WSO2 Micro Integrator 4.6.0 (Equinox + Aries SPI-Fly). All module bundles resolve and start:

```
g! ss jdbi
"Framework is launched."

id	State       Bundle
175	ACTIVE      org.jdbi.v3.cache.noop_3.54.1.SNAPSHOT
176	ACTIVE      org.jdbi.v3.caffeine_3.54.1.SNAPSHOT
177	ACTIVE      org.jdbi.v3.commonstext_3.54.1.SNAPSHOT
178	ACTIVE      org.jdbi.v3.core_3.54.1.SNAPSHOT
179	ACTIVE      org.jdbi.v3.freemarker_3.54.1.SNAPSHOT
180	ACTIVE      org.jdbi.v3.generator_3.54.1.SNAPSHOT
181	ACTIVE      org.jdbi.v3.gson2_3.54.1.SNAPSHOT
182	ACTIVE      org.jdbi.v3.guava_3.54.1.SNAPSHOT
183	ACTIVE      org.jdbi.v3.guice_3.54.1.SNAPSHOT
184	ACTIVE      org.jdbi.v3.jackson2_3.54.1.SNAPSHOT
185	ACTIVE      org.jdbi.v3.jackson3_3.54.1.SNAPSHOT
186	ACTIVE      org.jdbi.v3.jodatime2_3.54.1.SNAPSHOT
187	ACTIVE      org.jdbi.v3.jpa_3.54.1.SNAPSHOT
188	ACTIVE      org.jdbi.v3.json_3.54.1.SNAPSHOT
189	ACTIVE      org.jdbi.v3.kotlin_3.54.1.SNAPSHOT
190	ACTIVE      org.jdbi.v3.moshi_3.54.1.SNAPSHOT
191	ACTIVE      org.jdbi.v3.mysql_3.54.1.SNAPSHOT
192	ACTIVE      org.jdbi.v3.opentelemetry_3.54.1.SNAPSHOT
193	ACTIVE      org.jdbi.v3.oracle12_3.54.1.SNAPSHOT
194	ACTIVE      org.jdbi.v3.postgis_3.54.1.SNAPSHOT
195	ACTIVE      org.jdbi.v3.postgres_3.54.1.SNAPSHOT
196	ACTIVE      org.jdbi.v3.spring_3.54.1.SNAPSHOT
197	ACTIVE      org.jdbi.v3.spring5_3.54.1.SNAPSHOT
198	ACTIVE      org.jdbi.v3.sqlite_3.54.1.SNAPSHOT
199	ACTIVE      org.jdbi.v3.sqlobject_3.54.1.SNAPSHOT
200	ACTIVE      org.jdbi.v3.sqlobject.kotlin_3.54.1.SNAPSHOT
201	ACTIVE      org.jdbi.v3.stringtemplate4_3.54.1.SNAPSHOT
202	ACTIVE      org.jdbi.v3.testcontainers_3.54.1.SNAPSHOT
203	ACTIVE      org.jdbi.v3.testing_3.54.1.SNAPSHOT
204	ACTIVE      org.jdbi.v3.vavr_3.54.1.SNAPSHOT
```

- End-to-end test of core/sqlobject/json/jackson2/postgres inside the running container: `@SqlUpdate` methods executed against a PostgreSQL database, including a `Map<String, String>` ↔ `hstore` mapping.
- ServiceLoader path tested deliberately: with all explicit `installPlugin(...)` calls removed, `installPlugins()` discovers and installs the plugins via SPI-Fly bytecode weaving — providers register correctly. The usual OSGi caveat applies: `installPlugins()` takes a one-shot snapshot, so bundle start ordering matters; explicit plugin registration (as the deprecation note already recommends) or consuming the `JdbiPlugin` OSGi services that SPI-Fly registers are the robust alternatives.

### Notes

- core needs slf4j 2.x at runtime (`org.slf4j;[2.0,3)`) — `Slf4JSqlLogger` uses slf4j 2 API (`Logger.atLevel`). This is the honest range for what the code requires.
- moshi and pgvector publish no OSGi metadata upstream, so those (optional/mandatory) imports are unversioned.
- guice 5.1.0's own manifest pins guava at `[30.1,31)`, which conflicts with the guava range jdbi builds against; running jdbi3-guice under OSGi works with Guice 7.x, whose range matches current guava. Not something jdbi's manifests can fix.
- bnd 7 emits `java.*` package imports (OSGi R7+ style); frameworks older than ~2018 would need `-noimportjava: true`, which I did not add.

Per AI-GUIDELINES.md: this PR was developed with AI assistance (Claude Code). I reviewed all changes, and built and runtime-verified the result myself. Please apply the `AI` label — I cannot set labels on this repository.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- add OSGi bundle manifests to all published modules (#2247)
+
 # 3.54.0
 
 - fix deadlock in configuration caching (#2980)
--- a/cache/caffeine-cache/pom.xml
+++ b/cache/caffeine-cache/pom.xml
@@ -68,6 +68,39 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-invoker-plugin</artifactId>
             </plugin>
+            <plugin>
+                <groupId>biz.aQute.bnd</groupId>
+                <artifactId>bnd-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>bnd-process</id>
+                        <goals>
+                            <goal>bnd-process</goal>
+                        </goals>
+                        <configuration>
+                            <bnd><![CDATA[
+                                -exportcontents: org.jdbi.v3.*
+                                -metainf-services: auto
+                                Bundle-SymbolicName: ${moduleName}
+                            ]]></bnd>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-jar</id>
+                        <configuration>
+                            <archive>
+                                <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+                            </archive>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 </project>
--- a/cache/noop-cache/pom.xml
+++ b/cache/noop-cache/pom.xml
@@ -64,6 +64,39 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-invoker-plugin</artifactId>
             </plugin>
+            <plugin>
+                <groupId>biz.aQute.bnd</groupId>
+                <artifactId>bnd-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>bnd-process</id>
+                        <goals>
+                            <goal>bnd-process</goal>
+                        </goals>
+                        <configuration>
+                            <bnd><![CDATA[
+                                -exportcontents: org.jdbi.v3.*
+                                -metainf-services: auto
+                                Bundle-SymbolicName: ${moduleName}
+                            ]]></bnd>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-jar</id>
+                        <configuration>
+                            <archive>
+                                <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+                            </archive>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 </project>
--- a/commons-text/pom.xml
+++ b/commons-text/pom.xml
@@ -75,4 +75,41 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>biz.aQute.bnd</groupId>
+                <artifactId>bnd-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>bnd-process</id>
+                        <goals>
+                            <goal>bnd-process</goal>
+                        </goals>
+                        <configuration>
+                            <bnd><![CDATA[
+                                -exportcontents: org.jdbi.v3.*
+                                Bundle-SymbolicName: ${moduleName}
+                            ]]></bnd>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-jar</id>
+                        <configuration>
+                            <archive>
+                                <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+                            </archive>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
 </project>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -139,6 +139,48 @@
                 </executions>
             </plugin>
             <plugin>
+                <groupId>biz.aQute.bnd</groupId>
+                <artifactId>bnd-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>bnd-process</id>
+                        <goals>
+                            <goal>bnd-process</goal>
+                        </goals>
+                        <configuration>
+                            <bnd><![CDATA[
+                                -exportcontents: org.jdbi.v3.*
+                                Bundle-SymbolicName: ${moduleName}
+                                DynamicImport-Package: *
+                                Import-Package:\
+                                    !org.antlr.v4.runtime*,\
+                                    com.google.errorprone.annotations*;resolution:=optional,\
+                                    jakarta.annotation;resolution:=optional,\
+                                    jdk.jfr;resolution:=optional,\
+                                    *
+                                Require-Capability:\
+                                    osgi.extender;filter:="(osgi.extender=osgi.serviceloader.processor)";resolution:=optional,\
+                                    osgi.serviceloader;filter:="(osgi.serviceloader=org.jdbi.v3.core.spi.JdbiPlugin)";cardinality:=multiple;resolution:=optional
+                            ]]></bnd>
+                        </configuration>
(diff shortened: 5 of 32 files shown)


hgschmie:
Hi,

First thank you for the PR!

Some comments:

The PR is pretty repetitive and it touches all of our pom files. We have gone to great lengths to minimize the size of the pom files by abstracting almost everything into the build pom. Most of our poms only contain boilerplate settings and dependencies. 

Making this many changes and introducing an new (and pretty much unknown and not from an authorized source such as the maven-bundle-plugin for osgi) plugin without context into every pom is not something that we would be doing lightly. In addition, these changes introduce import dependencies such as https://github.com/jdbi/jdbi/pull/3010/changes#diff-8d04401f1cc51365fe3e32f019cd720135ba920a1a7da7f19e9c9208478701fcR151-R164 which at least I can not make a lot of sense of it nor maintain it when it breaks. 

To consider such an invasive PR, it would need at the very least

- abstract down to minimal changes, common code moved to the build pom
- clearly document how the importing mechanism works, what needs to be done if we were to introduce new dependencies and how to fix errors
- being optional, e.g. only run when we build the release, otherwise being run by an explicit make command (similar to what we do with e.g. the API compatibility checker)
- have a set of tests that we can run automated (e.g. as part of our test suite or gh actions), not just manual testing

And at that point, it would still need to be debated whether it makes sense for us to add OSGi manifests. This is a technology that has been around for 25 years and no one has ever asked for adding it to Jdbi. And would adding those manifests make it harder for us to maybe take on the Java Module effort in the future. 

Adding support for OSGi comes at a cost for the project. Providing such a PR is simple in the days of AI generated code and even when reviewing it and adhering to our AI guidelines before submitting, it still will require longterm maintenance by our small team of core contributors. 

At this point, I am not ready to move forward with this. I am interested to hear comments from others. @stevenschlansker ?

hgschmie (changes requested):
[Longer comment submitted in the conversation]

stevenschlansker:
I do not really know anything about OSGi, I have not used it myself. I don't see harm in shipping the metadata, but agree that we should figure out a better factored approach for the configuration. Is it possible to configure the bnd plugin once in the parent, and only ship the configuration block separately in each module?

TomasTokaMrazek:
@hgschmie @stevenschlansker I'll try to answer everything with my best knowledge and intent.

- Yes, I was assisted by AI, but I manually tested all jars in OSGi environment until everything installed correctly. As I said, we use this in production and I used to wrap jdbi in OSGi JAR for years.
- OSGi is 25 years old, but the technology is still used heavily, for example Apache Karaf, Apache Felix, Eclipse Equinox, WSO2 Micro Integrator, and well, Eclipse IDE. One of its main advantages, that is still unresolved in today's world of Java, is classloading multiple versions of the same jars. This is most notably seen with SLF4J or Jackson, who had to change maven coordinates for Jackson 3 (they kept OSGi support). The OSGi environment doesn't care, we for example run two Jackson 2 versions side by side, one latest for our needs, one 2.17 for Azure SDK if I remember correctly.
- We use around 20 - 25 third party dependencies and every single one with the exception of Oracle drivers and Azure SDK uses bnd to add OSGi support. That includes core libraries like Netty, Jackson, ASM, Bouncy Castle, Apache Commons, and database stack of HikariCP, PgJdbc, Mongo, RabbitMQ, MariaDB. It's not mandatory by any means, it is however nice to have.
- For non-OSGi users, nothing will change, there is zero risk. It is really only about metadata in `MANIFEST.MF`.
- Scoping this for release only is viable through `jdbi-release` profile and maven property.
- I agree that `maven-jar-plugin` execution is repeating and could be moved to `build`. Majority of `bnd-maven-plugin` configurations is repeating too and could be also moved to `build` with local overrides, I'll prepare it, test it and update the PR accordingly.
- The `bnd` itself scans classes and automatically searches for packages to import. Unfortunately the plugin cannot know, what packages are optional and not needed at runtime, unless explicitly opted in. Generally when the dependency is optional in maven, that is the signal it should be optional in bnd, and that's how I modeled the instructions. I even think that all those optional third-party dependencies can be declared in `build`, because `bnd` will ignore them, if the scanning actually doesn't find anything related, but I have to verify that.
- There are two major plugins for OSGi manifest, one is `maven-bundle-plugin`, which is Apache Felix wrapper over `bnd`. The other is `bnd-maven-plugin`, which is in active development by the people, who also develop `bnd` tool itself. This is the one I chose to use.
- OSGi can easily live side-by-side with JPMS, Jackson again is the authority here. I used the same mechanism, you can see that in their [pom.xml](https://repo1.maven.org/maven2/com/fasterxml/jackson/jackson-base/2.21.2/jackson-base-2.21.2.pom). They use the other plugin, but the mechanism is equal.
- For the `core` module, the most problematic is `Require-Capability`, whose sole existence is to enable SPI through `installPlugins()`, which is deprecated. I wanted to be thorough, even if we explicitly in our code use `installPlugin(xxx)`. If you're ok that for OSGi environments this single deprecated feature will not work and won't autoresolve plugins, I can drop it.
- About testing, basically nobody does this. For Jackson and Guava for example, they don't test it in any way. One repo I know of, that has some kind of test, is Gson. They inspect the build jar and read MANIFEST.MF. I didn't introduce this kind of test, because in my opinion it depends on the OSGi knowledge itself and doesn't provide much value. When OSGi breaks in terms of wrong MANIFEST.MF, people generally raise an issue about it and it's fixed in the next release. I myself reported it few times already. And if there is a bug, those people can easily fix that on their side temporarily by rewriting `MANIFEST.MF`, they don't have to do the full circle with building own jars.
- Where the documentation should live? The plugin is really just a wrapper over `bnd` and that itself is documented at https://bndtools.org/

stevenschlansker:
Yes, I think if we can eliminate the repetition by factoring it out to the parent pom, the bnd plugin use is OK.
I agree, let's not support `installPlugins()` and then we can skip the capability.
I think a test would be good, here is an example of how Gson does it: https://github.com/google/gson/blob/main/gson/src/test/java/com/google/gson/integration/OSGiManifestIT.java
With that, I think we could merge this.

PR #3015, opened 2026-08-24
Review feedback and interface tightening on 2942

### LINKED ISSUE #2445: Multiple Constructor Mappers of same type

stevenschlansker:
Closes #2445. Builds on #2942 by @phinjensen; his commits are included unchanged.

A query that joins the same table twice cannot map both objects through RowView: getRow(Class) finds mappers by type only, and the mapper registered last always wins. This PR makes the registration prefix part of an opt-in lookup: a new PrefixedRowMapper interface (implemented by BeanMapper, ConstructorMapper, FieldMapper, KotlinMapper), RowMappers.findFor(Type, String), and RowView.getRow(..., String prefix) overloads. All new API is @Alpha.

Changes relative to #2942, from the review discussion:

1. Strict matching. A prefixed lookup matches only a PrefixedRowMapper with an equal prefix. When no mapper matches, getRow throws NoSuchMapperException. It does not fall back to an unprefixed mapper, because that would silently return data from the wrong columns.
2. The prefix argument must be non-null. Use the single-argument methods for lookup by type alone. The empty string matches mappers registered without a prefix.
3. The unprefixed hot path is unchanged. Unprefixed lookups keep their Type-keyed caches; prefixed lookups use separate caches keyed by (type, prefix). RowViewPrefixBenchmark measures a registry cache hit at 3.3 ns unprefixed and 5.4 ns prefixed, and reduceRows over 1000 rows at 132 µs vs 139 µs, within error.
4. Smaller surface: one interface and five methods instead of thirteen methods. The StatementContext overloads are removed.
5. RowView.getRow(Type) stays abstract and the prefix overload has a concrete default, so external RowView subclasses still compile and japicmp passes at MINOR (the branch bumps the version to 3.55.0-SNAPSHOT for that reason).
6. Core tests (TestRowViewPrefixedMappers), the sqlobject tests from #2942, and a user guide section.


core/src/main/java/org/jdbi/v3/core/mapper/RowMappers.java
```diff
@@ -181,6 +187,49 @@ public Optional<RowMapper<?>> findFor(Type type) {
...
+     * Use this lookup to tell apart multiple mappers for the same type that read differently
+     * prefixed column sets, for example when a query joins the same table twice. To look up a
+     * mapper by type alone, use {@link #findFor(Type)}.
+     *
+     * @param type the target type to map to
+     * @param prefix the column name prefix the mapper must declare, never null
+     * @return a RowMapper for the given type and prefix, or empty if no matching row mapper is registered.
+     */
+    @Alpha
+    @SuppressWarnings("java:S2789") // the cache stores empty Optionals for negative results, so null means "not cached yet"
```
hgschmie, line 210: what warning does that suppress?

hgschmie (approved):
Looks good to me. I would have preferred it separately from the update to 3.55, the number of files (50+) made it a bit intimidating.

MikeNeilson:
Thanks for taking the time to do this. You and @phinjensen. This definitely looks reasonable to me. The additional path and cache, since while self referencing tables are common certainly aren't the most common, definitely keeps timing for existing usage the same. Not sure if I'm surprised that the lookup difference was only 2.1 ns or that it was 2.1 ns, but that can certainly add up with heavy usage.

My only annoyance (and this is in jest)... I think I'm going to have to rename my local PrefixRowMapper class. Does more than just the prefix anyways.


Not that you need to wait for me but I'm planning tomorrow morning to build jdbi locally and modify one of my usages of self referencing tables to use it; I don't expect any issues but will definitely report any I find.

core/src/main/java/org/jdbi/v3/core/mapper/RowMappers.java
stevenschlansker: SonarQube complains about storing the Optional type, but it is intentional here.

PR #3023, opened 2026-09-06
Fix commit failure skipping afterRollback callbacks

### LINKED ISSUE #3021: Handle afterRollback callbacks do not execute during SerializableTransactionRunner restarts
## Reporting an issue to Jdbi

[x] Jdbi Version:<br>
3.54.0
[x] Jdbi module:<br>
core
[x] Java Version:<br>
26.0.2
[x] I have read and understood the [AI contribution guidelines](https://github.com/jdbi/jdbi/blob/master/AI-GUIDELINES.md)<br>
[x] I understand that by contributing code to Jdbi, it may be distributed to others under the [Apache Software License 2.0](https://github.com/jdbi/jdbi/blob/master/LICENSE). I am authorized to contribute this code.<br>

----
Please describe the issue below. If you have a piece of code to reproduce it, add it as a code block.

AI has not been used for this bug report, though AI has been used post-fact to gain additional confidence with the findings described.

First, we notice that there is a direct call to rollback() in the exception path of BoundLocalTransctionHandler::commit, which is provided below the commit():

```java
        @Override
        public void commit(Handle handle) {
            try {
                if (handlerState != State.OUTSIDE_TRANSACTION) {
                    handle.getConnection().commit();
                }
                handlerState = getInitialHandlerState();
                restoreAutoCommitState(handle);
            } catch (SQLException e) {
                try {
                    rollback(handle); // <== executes, but doesn't go through handle
                } catch (Exception rollbackException) {
                    e.addSuppressed(rollbackException);
                }

                throw new TransactionException("Failed to commit transaction", e);
            }
        }

        @Override
        public void rollback(Handle handle) {
            try {
                if (handlerState != State.OUTSIDE_TRANSACTION) {
                    handle.getConnection().rollback();
                }
            } catch (SQLException e) {
                throw new TransactionException("Failed to rollback transaction", e);
            } finally {
                handlerState = getInitialHandlerState(); // <== also resets state
                restoreAutoCommitState(handle);
            }
        }
```

When handle.getConnection().commit() fails under SERIALIZABLE, it throws an SQLException. The rollback is in the same class, and resets the database but bypasses the handle's rollback(). Thus, handle does not execute any of its rollback handlers.

The rollback(), in addition, also updates the transaction's state. This prevents BoundLocalTransactionHandler from invoking its own handle.rollback() later:

```java
        @Override
        public <R, X extends Exception> R inTransaction(Handle handle,
                                                        HandleCallback<R, X> callback) throws X {
            if (isInTransaction(handle)) {
                throw new IllegalStateException("Already in transaction");
            }
            final R returnValue;

            try {
                handle.begin();
                handlerState = State.IN_TRANSACTION;

                returnValue = callback.withHandle(handle);

                if (handlerState == State.IN_TRANSACTION) {
                    handle.commit();
                }

                return returnValue;

            } catch (Throwable e) {
                // something went wrong within the operation. If this
                // can still be rolled back, roll it back and handle any
                // additional exception we may encounter
                if (handlerState == State.IN_TRANSACTION) {
                    try {
                        handle.rollback(); // <== not executed
                        e.addSuppressed(rollback);
                    }
                }
                // now call mark the operation as failed.
                throw e;
            }
        }
```
It seems to me that the rollback(handle) called during commit exception could perhaps be deleted.

The impact in our own case is that we do serializable work that runs a slow task, yielding an object in memory which we wish to persist in database. Our generic storage system uses the ID field of the object to decide whether it expects upsert to update or insert a row in the underlying database. In this case, the object is always new, so ID field has null; this triggers expectation that upsert returns the generated ID, which we save into the object. However, if that transaction rolls back, then the ID is stale and must be nulled, which should be done by afterRollback hook, which didn't run, and so the process looks like it wants an update, except there was no row in database that changed, and we caught that.

OswaldOniSango:
fixes #3021

When `commit()` fails with a `SQLException` (for example SERIALIZABLE conflict), `BoundLocalTransactionHandler.commit()` does a manual rollback (`rollback(handle)`) instead of going through`handle.rollback()`.

That means:
- `afterRollback` hooks never fire (the callback drain only happens inside `Handle.rollback()`)
- internal state gets reset before `inTransaction()` gets a chance to attempt its own fallback rollback, so it doesn't run there either

This bit me with an upsert: the transaction fails, the object is left with an ID that never got cleared (because the `afterRollback` that clears it never ran), and the next attempt mistakes it for an update instead of an insert.

Fix: use `handle.rollback()` instead of the local method. Added a test (`testCommitFailureStillFiresAfterRollbackCallback`) that fails without the fix and passes with it.

### DIFF
--- a/core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
+++ b/core/src/main/java/org/jdbi/v3/core/transaction/LocalTransactionHandler.java
@@ -141,7 +141,7 @@ public class LocalTransactionHandler implements TransactionHandler {
                 restoreAutoCommitState(handle);
             } catch (SQLException e) {
                 try {
-                    rollback(handle);
+                    handle.rollback();
                 } catch (Exception rollbackException) {
                     e.addSuppressed(rollbackException);
                 }
--- a/core/src/test/java/org/jdbi/v3/core/transaction/TestLocalTransactionHandler.java
+++ b/core/src/test/java/org/jdbi/v3/core/transaction/TestLocalTransactionHandler.java
@@ -14,6 +14,8 @@
 package org.jdbi.v3.core.transaction;
 
 import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.Jdbi;
@@ -46,6 +48,23 @@ public class TestLocalTransactionHandler {
             assertThat(e.getSuppressed()).hasSize(1);
             assertThat(e.getSuppressed()[0]).isSameAs(inner);
         }
+    }
+
+    @Test
+    public void testCommitFailureStillFiresAfterRollbackCallback() throws Exception {
+        Mockito.when(c.getAutoCommit()).thenReturn(true);
+        Mockito.doThrow(new SQLException("commit failed")).when(c).commit();
+
+        AtomicBoolean rolledBack = new AtomicBoolean();
+
+        assertThatThrownBy(() ->
+            h.inTransaction(handle -> {
+                handle.afterRollback(() -> rolledBack.set(true));
+                return null;
+            }))
+            .isInstanceOf(TransactionException.class);
+
+        assertThat(rolledBack).isTrue();
     }
 
     @Test


stevenschlansker:
Merged as #3034

